PR #7, opened 2020-04-25
visualize FSA via GraphViz.

csukuangfj:
I know it's difficult, if not impossible, to visualize an FSA with lots of states and transitions.

Visualization can be helpful to debug later algorithms, such as epsilon removal, trim, determinization ... ...

-----

Visualization of the following fsa is given below.

```cpp
  std::vector<Arc> arcs = {
      {0, 1, 2}, {0, 2, 1}, {1, 2, 0}, {1, 3, 5}, {2, 3, 6},
  };
  std::vector<Range> leaving_arcs = {
      {0, 2}, {2, 4}, {4, 5}, {0, 0},  // the last state has no leaving arcs
  };
```

abc

It is obtained by

```bash
./k2/csrc/fsa_renderer_test 2>&1 >/dev/null | dot -Tpng > abc.png
```


danpovey:
Great!  Merging.

PR #8, opened 2020-04-25
Rework RmEpsilons

danpovey:


danpovey:
Guys, this will break the compilation, but I am going to merge it anyway.. the GetEnteringArcs thing needs to be reworked a bit.  (The way it was done left no mechanism to keep track of arc-indexes, which is necessary..)
Actually, if we stick to mostly acyclic FSAs the GetEnteringArcs thing may not turn out to be necessary.  We'll see.

danpovey:
The main point of this PR is to show pseudocode for RmEpsilonsPruned(), and update the function declaration.

PR #9, opened 2020-04-25
fix GetEnteringArcs.

csukuangfj:


k2/csrc/fsa_util.cc
```diff
@@ -11,34 +11,29 @@
...
   int32_t num_arcs = 0;
+  int32_t k = 0;
   for (const auto &arc : fsa.arcs) {
-    auto src_state = arc.src_state;
     auto dest_state = arc.dest_state;
-    auto label = arc.label;
-    vec[dest_state].emplace_back(label, src_state);
+    vec[dest_state].push_back(k);
     ++num_arcs;
+    ++k;
```
qindazhu, line 26: why not just use `num_arcs = fsa.arcs.size()` as `vector.size()` is always `O(1)`

csukuangfj: oh, yes, you are right. I will change it in the next commit.

csukuangfj: Fixed.

PR #13, opened 2020-04-25
implement Connect.

csukuangfj:
Only `arc_map` is missing.

For the following fsa,

trim

the algorithm prints:

```
state 5 is not accessible
state 3 is not coaccessible
state 4 is not coaccessible
state 5 is not coaccessible
```

The implementation is inspired by OpenFST. Only a single forward pass is enough.
No backward pass is performed.

----

State 5 is considered as not coaccessible because the current implementation assumes
that every path has to start with the initial state 0. @danpovey , if you think
this is not suitable, I can change the implementation.

----

It does NOT require a topo-sorted input fsa.

### DIFF
--- a/k2/csrc/fsa.h
+++ b/k2/csrc/fsa.h
@@ -18,12 +18,11 @@ using StateId = int32_t;
 using Weight = float;
 
 enum {
-  kFinalSymbol = -1,     // final-costs are represented as arcs with
-                         // kFinalSymbol as their label, to the final
-                         // state (see Fst::final_state).
-  kEpsilon = 0,          // Epsilon, which means "no symbol", is numbered zero,
-                         // like in OpenFst.
-  kInvalidStateId = -1,  // it MUST be less than zero, like in OpenFst.
+  kFinalSymbol = -1,  // final-costs are represented as arcs with
+                      // kFinalSymbol as their label, to the final
+                      // state (see Fst::final_state).
+  kEpsilon = 0,       // Epsilon, which means "no symbol", is numbered zero,
+                      // like in OpenFst.
 };
 
 struct Arc {
--- a/k2/csrc/fsa.h
+++ b/k2/csrc/fsa.h
@@ -18,11 +18,12 @@ using StateId = int32_t;
 using Weight = float;
 
 enum {
-  kFinalSymbol = -1,  // final-costs are represented as arcs with
-                      // kFinalSymbol as their label, to the final
-                      // state (see Fst::final_state).
-  kEpsilon = 0        // Epsilon, which means "no symbol", is numbered zero,
-                      // like in OpenFst.
+  kFinalSymbol = -1,     // final-costs are represented as arcs with
+                         // kFinalSymbol as their label, to the final
+                         // state (see Fst::final_state).
+  kEpsilon = 0,          // Epsilon, which means "no symbol", is numbered zero,
+                         // like in OpenFst.
+  kInvalidStateId = -1,  // it MUST be less than zero, like in OpenFst.
 };
 
 struct Arc {
--- a/k2/csrc/fsa_algo.cc
+++ b/k2/csrc/fsa_algo.cc
@@ -6,6 +6,7 @@
 
 #include "k2/csrc/fsa_algo.h"
 
+#include <map>
 #include <stack>
 
 namespace {
@@ -24,6 +25,8 @@ namespace k2 {
 // The implementation of this function is inspired by
 // http://www.openfst.org/doxygen/fst/html/connect_8h_source.html
 void ConnectCore(const Fsa &fsa, std::vector<int32_t> *state_map) {
+  // DCHECK_NOTNULL(state_map)
+  state_map->clear();
   if (IsEmpty(fsa)) return;
 
   auto num_states = fsa.NumStates();
@@ -80,13 +83,58 @@ void ConnectCore(const Fsa &fsa, std::vector<int32_t> *state_map) {
     }
   }
 
-  state_map->clear();
   state_map->reserve(num_states);
 
   for (int32_t i = 0; i != num_states; ++i) {
-    if (accessible[i] && coaccessible[i]) {
-      state_map->push_back(i);
+    if (accessible[i] && coaccessible[i]) state_map->push_back(i);
+  }
+}
+
+void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
+  // DCHECK_NOTNULL(b);
+  std::vector<int32_t> state_map;
+  ConnectCore(a, &state_map);
+  if (state_map.empty()) return;
+
+  b->arc_indexes.resize(state_map.size());
+  b->arcs.clear();
+  b->arcs.reserve(a.arcs.size());
+
+  if (arc_map) {
+    arc_map->clear();
+    arc_map->reserve(a.arcs.size());
+  }
+
+  std::map<int32_t, int32_t> valid_states;  // map from state_a to state_b
+  int32_t num_states_b = b->NumStates();
+  for (int32_t i = 0; i != num_states_b; ++i) {
+    auto state_a = state_map[i];
+    valid_states.insert({state_a, i});
+  }
+
+  int32_t begin = 0;
+  int32_t end = 0;
+  int32_t final_state_a = a.NumStates() - 1;
+  for (int32_t i = 0; i != num_states_b; ++i) {
+    auto state_a = state_map[i];
+    begin = a.arc_indexes[state_a];
+    if (state_a != final_state_a)
+      end = a.arc_indexes[state_a + 1];
+    else
+      end = begin;
+
+    b->arc_indexes[i] = static_cast<int32_t>(b->arcs.size());
+    for (; begin != end; ++begin) {
+      auto arc = a.arcs[begin];
+      auto dest_state = arc.dest_state;
+      if (!valid_states.count(dest_state))
+        continue;  // dest_state is unreachable
+      arc.src_state = i;
+      arc.dest_state = valid_states[dest_state];
+      b->arcs.push_back(arc);
     }
+
+    // TODO(fangjun): handle arc_map
   }
 }
 
--- a/k2/csrc/fsa_algo.cc
+++ b/k2/csrc/fsa_algo.cc
@@ -6,8 +6,6 @@
 
 #include "k2/csrc/fsa_algo.h"
 
-#include <iostream>
-#include <sstream>
 #include <stack>
 
 namespace {
@@ -82,22 +80,14 @@ void ConnectCore(const Fsa &fsa, std::vector<int32_t> *state_map) {
     }
   }
 
-  // TODO(fangjun): remove states that are not accessible or coaccessible;
-  // the following is for debug purpose only
-  int32_t i = 0;
-  std::ostringstream os;
-  for (const auto b : accessible) {
-    if (!b) os << "state " << i << " is not accessible\n";
-    ++i;
-  }
+  state_map->clear();
+  state_map->reserve(num_states);
 
-  i = 0;
-  for (const auto b : coaccessible) {
-    if (!b) os << "state " << i << " is not coaccessible\n";
-    ++i;
+  for (int32_t i = 0; i != num_states; ++i) {
+    if (accessible[i] && coaccessible[i]) {
+      state_map->push_back(i);
+    }
   }
-
-  std::cout << os.str() << "\n";
 }
 
 }  // namespace k2
--- a/k2/csrc/fsa_algo.cc
+++ b/k2/csrc/fsa_algo.cc
@@ -1,0 +1,103 @@
+// k2/csrc/fsa_algo.cc
+
+// Copyright (c)  2020  Fangjun Kuang (csukuangfj@gmail.com)
+
+// See ../../LICENSE for clarification regarding multiple authors
+
+#include "k2/csrc/fsa_algo.h"
+
+#include <iostream>
+#include <sstream>
+#include <stack>
+
+namespace {
+
+// depth first search state
+struct DfsState {
+  int32_t s;      // state number of the visiting node
+  int32_t begin;  // arc index of the visiting arc
+  int32_t end;    // end of the arc index of the visiting node
+};
+
+}  // namespace
+
+namespace k2 {
+
+// The implementation of this function is inspired by
+// http://www.openfst.org/doxygen/fst/html/connect_8h_source.html
+void ConnectCore(const Fsa &fsa, std::vector<int32_t> *state_map) {
+  if (IsEmpty(fsa)) return;
+
+  auto num_states = fsa.NumStates();
+  auto final_state = num_states - 1;
+
+  std::vector<bool> accessible(num_states, false);
+  std::vector<bool> coaccessible(num_states, false);
+  std::vector<bool> visited(num_states, false);
+
+  accessible.front() = true;
+  coaccessible.back() = true;
+
+  std::stack<DfsState> stack;
+  stack.push({0, fsa.arc_indexes[0], fsa.arc_indexes[1]});
+  visited[0] = true;
+
+  while (!stack.empty()) {
+    auto &top = stack.top();
+
+    if (top.begin == top.end) {
+      // we have finished visiting this state
+      auto s = top.s;  // get a copy since we will destroy it
+      stack.pop();
+      if (!stack.empty()) {
+        // if it has a parent, set the parent's coaccessible flag
+        if (coaccessible[s]) {
+          auto &parent = stack.top();
+          coaccessible[parent.s] = true;
+          ++parent.begin;  // process the next child
+        }
+      }
+      continue;
+    }
+
+    const auto &arc = fsa.arcs[top.begin];
+    auto next_state = arc.dest_state;
+    bool is_visited = visited[next_state];
+    if (!is_visited) {
+      // this is a new discovered state
+      visited[next_state] = true;
+      auto begin = fsa.arc_indexes[next_state];
+      if (next_state != final_state)
+        stack.push({next_state, begin, fsa.arc_indexes[next_state + 1]});
+      else
+        stack.push({next_state, begin, begin});
+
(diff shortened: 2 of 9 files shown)


k2/csrc/fsa_algo.cc
```diff
@@ -0,0 +1,114 @@
...
+namespace {
+
+struct DfsState {
+  int32_t s;
+  int32_t begin;
+  int32_t end;
+};
+
+enum class DfsStateColor : int8_t {
+  kWhite,  // a state which has not been discovered
```
danpovey, line 23: For me it might be clearer to just use bool and name it 'discovered' or something like that, but up to you.

k2/csrc/fsa_util.cc
```diff
@@ -15,14 +15,13 @@ void GetEnteringArcs(const Fsa &fsa, std::vector<int32_t> *arc_index,
                      std::vector<int32_t> *end_index) {
   // CHECK(CheckProperties(fsa, KTopSorted));
 
-  int32_t num_states = fsa.NumStates();
+  auto num_states = fsa.NumStates();
   std::vector<std::vector<int32_t>> vec(num_states);
-  int32_t num_arcs = 0;
+  auto num_arcs = static_cast<int32_t>(fsa.arcs.size());
```
danpovey, line 20: I think this variable isn't needed any more since it's only accessed once; could just use fsa.arcs.size().

qindazhu: I have changed this in previous PR.

danpovey: cool.  BTW there are conflicts now.  I'll look at this again when you remove WIP.

k2/csrc/fsa_algo.cc
```diff
@@ -0,0 +1,114 @@
...
+
+#include <iostream>
+#include <memory>
+#include <sstream>
+#include <stack>
+
+namespace {
+
+struct DfsState {
+  int32_t s;
```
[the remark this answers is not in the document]
csukuangfj: Comment has been added.

k2/csrc/fsa_algo.cc
csukuangfj: Fixed.

k2/csrc/fsa_util.cc
csukuangfj: conflicts resolved.

k2/csrc/fsa_algo.cc
```diff
@@ -0,0 +1,103 @@
...
+  int32_t end;    // end of the arc index of the visiting node
+};
+
+}  // namespace
+
+namespace k2 {
+
+// The implementation of this function is inspired by
+// http://www.openfst.org/doxygen/fst/html/connect_8h_source.html
+void ConnectCore(const Fsa &fsa, std::vector<int32_t> *state_map) {
```
csukuangfj, line 28: @danpovey 
It seems the signature has to be changed.

There is no argument for the **out** fsa.

k2/csrc/fsa_algo_test.cc
```diff
@@ -0,0 +1,39 @@
...
+
+namespace k2 {
+
+TEST(FsaAlgo, Connect) {
+  std::vector<Arc> arcs = {
+      {0, 1, 1}, {0, 2, 2}, {1, 3, 3}, {1, 6, 6},
+      {2, 4, 2}, {2, 6, 3}, {5, 0, 1},
+  };
+
+  std::vector<Range> leaving_arcs = {
```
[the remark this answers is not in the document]
csukuangfj: Fixed.

csukuangfj:
The connected FSA is shown below:

trim (1)

----

For ease of reference , the orignial FSA is given below:
trim

danpovey:
@csukuangfj this is still marked WIP which is why I'm not merging/reviewing..

k2/csrc/fsa_algo.cc
```diff
@@ -0,0 +1,141 @@
...
+  b->arc_indexes.resize(state_map.size());
+  b->arcs.clear();
+  b->arcs.reserve(a.arcs.size());
+
+  if (arc_map) {
+    arc_map->clear();
+    arc_map->reserve(a.arcs.size());
+  }
+
+  std::map<int32_t, int32_t> valid_states;  // map from state_a to state_b
```
qindazhu, line 108: use `unordered_map`?

csukuangfj: OK, will change it.

I am going to replace it with a `std::vector<int32_t>`. 
If a state is invalid, its corresponding value will be set to a negative number. Maybe `kInvalidStateId` will be introduced.

If lots of states are valid, `std::vector<int32_t>` may be more efficient:
- O(1) lookup, even faster than a hash map.
- O(N) memory space

csukuangfj:
@danpovey 
I will finish it tonight and will `@` you when it is ready.

k2/csrc/fsa_algo.cc
```diff
@@ -0,0 +1,141 @@
...
+
+  std::vector<bool> accessible(num_states, false);
+  std::vector<bool> coaccessible(num_states, false);
+  std::vector<bool> visited(num_states, false);
+
+  accessible.front() = true;
+  coaccessible.back() = true;
+
+  std::stack<DfsState> stack;
+  stack.push({0, fsa.arc_indexes[0], fsa.arc_indexes[1]});
```
qindazhu, line 43: I just think that maybe we can only store the `end` and do `--end` and check `end == 0` in the loop

k2/csrc/fsa_algo.cc
```diff
@@ -0,0 +1,141 @@
...
+
+void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
+  // DCHECK_NOTNULL(b);
+  std::vector<int32_t> state_map;
+  ConnectCore(a, &state_map);
+  if (state_map.empty()) return;
+
+  b->arc_indexes.resize(state_map.size());
+  b->arcs.clear();
+  b->arcs.reserve(a.arcs.size());
```
qindazhu, line 101: maybe there is no need to `clear`, just `reserve` or `resize`, since you will override all values

k2/csrc/fsa_algo.cc
```diff
@@ -0,0 +1,141 @@
...
+
+  if (arc_map) {
+    arc_map->clear();
```
qindazhu, line 104: delete?

k2/csrc/fsa_algo.cc
qindazhu: Sure. Actually I did the same thing in `RandomPath`, but I set it with `-1`. if you define `kInvalidStateId`, I'll update after merging your code.

k2/csrc/fsa_algo.cc
csukuangfj: The issue is that I do not know how many arcs b will contain at this point.

If I use reserve, then I can use `push_back` later.

k2/csrc/fsa_algo.cc
csukuangfj: I do not have the information about the size of arc_map at this point, so I cannot use resize.

k2/csrc/fsa_algo.cc
csukuangfj: Fixed.

csukuangfj:
@qindazhu @danpovey 
please have a review.

I also added glog into this pullrequest.

k2/csrc/fsa_algo.cc
csukuangfj: In your pullrequest, you can remove the two variables: num state vistited and num arcs visited.
Then you can also use `reserve` instead of `resize`. 

Two fewer variables will be used.

k2/csrc/fsa.h
```diff
@@ -18,11 +18,12 @@ using StateId = int32_t;
...
-                      // kFinalSymbol as their label, to the final
-                      // state (see Fst::final_state).
-  kEpsilon = 0        // Epsilon, which means "no symbol", is numbered zero,
-                      // like in OpenFst.
+  kFinalSymbol = -1,     // final-costs are represented as arcs with
+                         // kFinalSymbol as their label, to the final
+                         // state (see Fst::final_state).
+  kEpsilon = 0,          // Epsilon, which means "no symbol", is numbered zero,
+                         // like in OpenFst.
+  kInvalidStateId = -1,  // it MUST be less than zero, like in OpenFst.
```
qindazhu, line 26: maybe it's a bad idea to define two enums with the same value, this enables users to call `some_symbol == kInvalidStateId` successfully even if they should check `some_symbol == kFinalSymbol`

k2/csrc/fsa_algo.cc
qindazhu: In my pullrequest, actually if I removed `num_visited_states`, I need to get the size of  `state_map_b2a` again and again in the loop which is inefficient. Furthermore, If I remove `num_visited_arcs`, as there may be cycle and duplicated visited arcs, I'll need to iterate over `visited_arcs` first to get `num_visited_arcs`.

k2/csrc/fsa_algo_test.cc
```diff
@@ -0,0 +1,79 @@
...
+
+  {
+    Fsa fsa;
+    fsa.arc_indexes = std::move(arc_indexes);
+    fsa.arcs = std::move(arcs);
+
+    std::vector<int32_t> state_map(10);  // an arbitrary number
+    ConnectCore(fsa, &state_map);
+
+    EXPECT_EQ(state_map.size(), 4u);
```
qindazhu, line 39: Use `ASSERT_EQ` here as below tests (such as `EXPECT_EQ(state_map[3], ..)`) would not make any sense if the `size < 4`

k2/csrc/fsa_algo_test.cc
```diff
@@ -0,0 +1,79 @@
...
+    EXPECT_EQ(state_map[0], 0);
```
qindazhu, line 40: use `expect_that` from `gmock`, such as
' EXPECT_THAT(state_map, ::testing::ElementsAre(0, 1, 2, 6))'

k2/csrc/fsa_algo.cc
```diff
@@ -0,0 +1,149 @@
...
+
+#include <stack>
+
+#include "glog/logging.h"
+
+namespace {
+
+// depth first search state
+struct DfsState {
+  int32_t s;      // state number of the visiting node
```
qindazhu, line 17: `state` may be better.

k2/csrc/fsa_algo_test.cc
csukuangfj: Will fix it this night.

k2/csrc/fsa_algo_test.cc
csukuangfj: Will fix it this night.

k2/csrc/fsa_algo.cc
csukuangfj: Will fix it this night.

k2/csrc/fsa_algo.cc
```diff
@@ -0,0 +1,149 @@
...
+  coaccessible.back() = true;
+
+  std::stack<DfsState> stack;
+  stack.push({0, fsa.arc_indexes[0], fsa.arc_indexes[1]});
+  visited[0] = true;
+
+  while (!stack.empty()) {
+    auto &top = stack.top();
+
+    if (top.begin == top.end) {
```
qindazhu, line 51: `curr_state.arc_begin == curr_state.arc_end` may be more readable, just a suggestion, do what you prefer.

k2/csrc/fsa_algo.h
```diff
@@ -127,7 +128,7 @@ void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map = nullptr);
 void RmEpsilonsPruned(const WfsaWithFbWeights &a,
                       float beam,
                       Fsa *b,
-                      std::vector<std::vector> *arc_map);
+                      std::vector<std::vector<int32_t>> *arc_map);
```
qindazhu, line 131: add `/*=nullptr*/` ?

csukuangfj: I am going to resolve all the conflicts this night out of working hours.

k2/csrc/fsa_algo_test.cc
```diff
@@ -0,0 +1,79 @@
...
+namespace k2 {
+
+TEST(FsaAlgo, Connect) {
+  std::vector<Arc> arcs = {
+      {0, 1, 1}, {0, 2, 2}, {1, 3, 3}, {1, 6, 6},
+      {2, 4, 2}, {2, 6, 3}, {5, 0, 1},
+  };
+
+  std::vector<int32_t> arc_indexes = {0, 2, 4, 6, 6, 6, 7};
+
```
qindazhu, line 23: maybe another case that: get an empty fsa after connecting a non-empty one

k2/csrc/fsa_algo_test.cc
```diff
@@ -0,0 +1,79 @@
...
+    EXPECT_EQ(state_map[0], 0);
+    EXPECT_EQ(state_map[1], 1);
+    EXPECT_EQ(state_map[2], 2);
+    EXPECT_EQ(state_map[3], 6);
+
+    Fsa connected;
+    std::vector<int32_t> arc_map(10);  // an arbitrary number
+    Connect(fsa, &connected, &arc_map);
+
+    EXPECT_EQ(connected.NumStates(), 4u);  // state 3,4,5 from fsa are removed
```
qindazhu, line 49: `ASSERT_EQ `

k2/csrc/fsa_algo_test.cc
```diff
@@ -0,0 +1,79 @@
...
+    EXPECT_EQ(connected.arc_indexes[0], 0);
```
qindazhu, line 50: `EXPECT_THAT`

k2/csrc/fsa_algo_test.cc
```diff
@@ -0,0 +1,79 @@
...
+
+    EXPECT_EQ(connected.arcs[2].src_state, 1);
+    EXPECT_EQ(connected.arcs[2].dest_state, 3);
+    EXPECT_EQ(connected.arcs[2].label, 6);
+
+    EXPECT_EQ(connected.arcs[3].src_state, 2);
+    EXPECT_EQ(connected.arcs[3].dest_state, 3);
+    EXPECT_EQ(connected.arcs[3].label, 3);
+
+    EXPECT_EQ(arc_map.size(), 4u);
```
qindazhu, line 71: `ASSERT_EQ` and `EXPECT_THAT`

k2/csrc/fsa.h
danpovey: I expect some compilers won't let you declare conflicting enum's.  In any case, my own preference would just be to use a literal -1 for "invalid state id", and to test it with "if (states[i] < 0) { ... }" which should compile to less code than "if (states[i] == -1)", I'd expect.

qindazhu: actually I don't think so, as if we define `-2, -3..` in this enum in the future and users misue the value to initlialize `states[i] = -2` , check `states[i] < 0` will still pass..

BTW, I don't think `==-1` and `<0` will make a big difference in time consumed (I guess for most compilers, they all just call one line CPU instruction)

danpovey:
OK.  But I don't want kInvalidStateId unless it appears in interfaces - and
even then maybe not.
I just don't want this kind of coding style, like OpenFST, that had so much
interconnection and boilerplate.
Inside a function, using -1 for invalid state is readable and obvious- no
need for a definition.

qindazhu:
> Inside a function, using -1 for invalid state is readable and obvious- no
need for a definition
> unless it appears in interfaces

I'm OK  with this.

k2/csrc/fsa.h
csukuangfj: To my best knowledge, you **can** define multiple enum entries with the same value.
It is a bug if a compiler does not support it.

----

In kaldi, there is

```cpp
static const int32 kNoPdf = -1;
```
https://github.com/kaldi-asr/kaldi/blob/master/src/hmm/hmm-topology.h#L86

----

A named constant improves readability.

danpovey: At that time I was coding more like OpenFST.  But my decision on this is: no named constant.  Just use -1.

csukuangfj: Fine, I will change it.

csukuangfj: Removed and fixed.

k2/csrc/fsa_algo_test.cc
csukuangfj: Done.

k2/csrc/fsa_algo.cc
csukuangfj: Done.

k2/csrc/fsa_algo.cc
csukuangfj: Done.

k2/csrc/fsa_algo.h
csukuangfj: This is the function signature and it does not have a default value.

`/*=nullptr*/ ` is useful only in the implementation file when  its default value is a nullptr.

k2/csrc/fsa_algo_test.cc
csukuangfj: Done.

k2/csrc/fsa_algo_test.cc
csukuangfj: Done.

k2/csrc/fsa_algo_test.cc
csukuangfj: Done.

csukuangfj:
@qindazhu @danpovey 
Comments are resolved. Please review and merge.

k2/csrc/fsa_algo_test.cc
danpovey: We shouldn't need to specify the arc_indexes to the constructor of these functions because the begin-arc is part of the supplied arcs.  Would save coders time.  Fine for now though.

PR #14, opened 2020-04-26
remove Range in Fsa

qindazhu:


qindazhu:
@csukuangfj I have updated all code with `arc_indexes` except files related with `fsa_render`. You can remove `Range` and `leaving_arcs` after updating `fsa_render`.

csukuangfj:
+2

merging.

k2/csrc/properties.cc
```diff
@@ -33,18 +33,16 @@ bool IsValid(const Fsa &fsa) {
     } else {
       // every state contains at least one arc.
```
danpovey, line 34: This is an issue in previous code, but RE "only epsilon arcs enter the final state", this is not right, it should be kFinalSymbol (-1).  You may have to change other code.  This change is necessary, things will break later if you leave it as epsilon.

danpovey: RE "every state contains at least one arc": firstly, you should word it as "every state has at least one arc leaving it".  But actually this condition is not required for an FSA to be valid (it would just not be connected).

k2/csrc/properties.cc
```diff
@@ -33,18 +33,16 @@ bool IsValid(const Fsa &fsa) {
...
         return false;
       state = arc.src_state;
       num_arcs = 1;
     }
   }
   // check the last state
   if (final_state != state + 1) return false;
-  if ((fsa.leaving_arcs[final_state].begin - fsa.leaving_arcs[state].begin) !=
-      num_arcs)
+  if ((fsa.arc_indexes[final_state] - fsa.arc_indexes[state]) != num_arcs)
```
danpovey, line 45: ... this code is hard for me to understand and to verify that it's correct.  I think it would be much clearer if the iteration were done over fsa.arc_indexes.

k2/csrc/properties.cc
danpovey: .. also you should probably make it so that this code can't segfault even if the input is invalid.

PR #15, opened 2020-04-27
only kFinalSymobl arcs enter final state

qindazhu:


danpovey:
We can put something like assert(IsValid(input_fsa)) on other operations.  For operations that have other requirements, e.g. top-sortedness, can put assert(IsValid(..) && IsTopSorted(..)).

danpovey:
And, guys, please always be absolutely clear in your head and in the functions' documentation, what the pre-conditions and post-conditions of functions are.  There can also be conditional ones, e.g. "Input must be valid; if the input is top-sorted the output will be top-sorted".  These kinds of things can also be checked via asserts.

csukuangfj:
I will introduce `glog`  in my pending pullrequest to check preconditions.

PR #16, opened 2020-04-27
implement randomPath

qindazhu:


k2/csrc/fsa.h
```diff
@@ -11,6 +11,8 @@
 #include <utility>
 #include <vector>
 
+#include "k2/csrc/util.h"
```
csukuangfj, line 14: Is this header needed here?

k2/csrc/fsa.h
```diff
@@ -11,6 +11,8 @@
 #include <utility>
```
csukuangfj, line 11: I think utility can be removed.

k2/csrc/fsa.h
```diff
@@ -11,6 +11,8 @@
...
 #include <vector>
 
+#include "k2/csrc/util.h"
+
 namespace k2 {
 
 using Label = int32_t;
```
csukuangfj, line 18: should we place `Weight` in fsa.h? There is no weight in an acceptor.

k2/csrc/fsa_equivalent.cc
```diff
@@ -0,0 +1,82 @@
...
+#include <unordered_set>
+#include <vector>
+
+#include "k2/csrc/fsa.h"
+// #include "k2/csrc/fsa_algo.h"
+#include "k2/csrc/properties.h"
+
+namespace k2 {
+
+bool RandomPath(const Fsa &a, Fsa *b, std::vector<int32_t> *state_map) {
```
csukuangfj, line 21: It's better to use

```cpp
std::vector<int32_t> *state_map /*=nullptr*/
```

k2/csrc/fsa_equivalent.cc
```diff
@@ -0,0 +1,82 @@
...
+  if (!IsConnected(a)) {
+    Fsa connected_fsa;
+    // TODO(haowen): uncomment this
+    // Connect(a, &connected_fsa);
+    return RandomPath(connected_fsa, b, state_map);
+  }
+
+  int32_t num_states = a.NumStates();
+  std::vector<int32_t> state_map_b2a;
+  std::vector<int32_t> state_map_a2b(num_states, -1);
```
csukuangfj, line 32: I've added `kInvalidStateId`.

k2/csrc/fsa_equivalent.cc
```diff
@@ -0,0 +1,82 @@
...
+  // `visited_arcs[i]` stores `arcs` leaving from state `i`
+  std::vector<std::unordered_set<Arc, ArcHash>> visited_arcs;
```
csukuangfj, line 34:
```cpp
using ArcSet = std::unordered_set<Arc, ArcHash>;
std::vector<ArcSet> visited_arcs;
....

visited_arcs.push_bac(ArcSet{});
```

k2/csrc/fsa_equivalent.cc
```diff
@@ -0,0 +1,82 @@
...
+    for (const auto &arc : visited_arcs[i]) {
+      auto &b_arc = b->arcs[n];
+      b_arc.src_state = i;
+      b_arc.dest_state = state_map_a2b[arc.dest_state];
+      b_arc.label = arc.label;
+      ++n;
+    }
+  }
+  if (state_map != nullptr) {
+    state_map->swap(state_map_b2a);
```
csukuangfj, line 77: If `a` is not connected, then the state in `state_map` does not map to the state in `a`,
it is mapped to the connected fsa.

Should we just return `false` instead of run `Connect`?

k2/csrc/fsa_equivalent_test.cc
```diff
@@ -0,0 +1,61 @@
...
+    Fsa path;
+
+    {
+      bool status = RandomPath(fsa, &path);
+      EXPECT_TRUE(status);
+    }
+
+    {
+      for (size_t i = 0; i < 20; ++i) {
+        std::vector<int32_t> state_map;
```
csukuangfj, line 46: `state_map` can be moved out of the `for` loop.

k2/csrc/fsa_equivalent_test.cc
```diff
@@ -0,0 +1,61 @@
...
+    fsa.arcs = std::move(arcs);
+    Fsa path;
+
+    {
+      bool status = RandomPath(fsa, &path);
+      EXPECT_TRUE(status);
+    }
+
+    {
+      for (size_t i = 0; i < 20; ++i) {
```
csukuangfj, line 45: ```int32_t i = 0;```

or

```auto i = 0;```

File: k2/csrc/util.h [hunk 10]
```diff
@@ -0,0 +1,24 @@
...
+
+#include <functional>
+
+#include "k2/csrc/fsa.h"
+
+namespace k2 {
+
+// boost::hash_combine
+template <class T>
+inline void hash_combine(std::size_t *seed, const T &v) {
```
csukuangfj, line 18: To follow the google c++ style, the output paremter should be the last one.

So the siganature should be
```cpp
inline void hash_combine(const T &v, std::size_t *seed)
```

k2/csrc/fsa.h
csukuangfj: Maybe we can move the hash function to `fsa.h`.

k2/csrc/fsa_equivalent.cc
```diff
@@ -0,0 +1,82 @@
...
+  int32_t state = 0;
+  int32_t final_state = num_states - 1;
+  while (true) {
+    if (state_map_a2b[state] == -1) {
+      state_map_a2b[state] = num_visited_state;
+      state_map_b2a.push_back(state);
+      visited_arcs.push_back(std::unordered_set<Arc, ArcHash>());
+      ++num_visited_state;
+    }
+    if (state == final_state) break;
```
csukuangfj, line 51: It **might** not terminate and be an infinite loop here if there is a cycle in
the graph.

k2/csrc/fsa_equivalent_test.cc
```diff
@@ -0,0 +1,61 @@
...
+    EXPECT_FALSE(status);
+  }
+  // TODO(haowen): add tests for non-connected fsa
+}
+
+TEST(Properties, RandomPathSuccess) {
+  {
+    Fsa fsa;
+    std::vector<Arc> arcs = {
+        {0, 1, 1}, {0, 2, 2}, {1, 2, 3}, {2, 3, 4},
```
csukuangfj, line 31: To remove randomness, you can set up a case  such that every node has only one leaving arc,
so you can **check** the returned path.

k2/csrc/fsa_equivalent_test.cc
csukuangfj: This guy, Walter E. Brown (https://cppcon2019.sched.com/speaker/walter_e_brown)
suggests using
```cpp
for (auto i = 0; i != 20; ++i)
```

**!=** instead of **<**

---

The motivation is explained in this presentation
https://www.youtube.com/watch?v=OAmWHmwlMwI

At `37:57`, i.e., 37 minutes, 57 seconds.

k2/csrc/fsa_equivalent.cc
```diff
@@ -0,0 +1,82 @@
...
+    Fsa connected_fsa;
+    // TODO(haowen): uncomment this
+    // Connect(a, &connected_fsa);
+    return RandomPath(connected_fsa, b, state_map);
+  }
+
+  int32_t num_states = a.NumStates();
+  std::vector<int32_t> state_map_b2a;
+  std::vector<int32_t> state_map_a2b(num_states, -1);
+  // `visited_arcs[i]` stores `arcs` leaving from state `i`
```
csukuangfj, line 33: It's better to specify the state of which fsa.
In this case, it is the state of `b`.

k2/csrc/fsa.h
qindazhu: No,  I think it's better to put this in a new file, as `hash_combine` is not just for `ArcHash` and may be called by other struct, such as `std::pair`.

Note that there is a hash function named `ArcHash` in fsa.h
https://github.com/danpovey/k2/blob/1e362c1b3b603c57cde6bbbb08153f34ba99a88c/k2/csrc/fsa.h#L49-L55

k2/csrc/fsa.h
qindazhu: Sure

k2/csrc/fsa.h
qindazhu: I guess @danpovey worte this because there is another weighted fsa structure, 

https://github.com/danpovey/k2/blob/08145d03d6dac7fb442857227ab632158fbb4191/k2/csrc/weights.h#L61

k2/csrc/fsa_equivalent.cc
qindazhu: Sure

k2/csrc/fsa_equivalent.cc
qindazhu: As there is only two occurrences of  `std::unordered_set<Arc, ArcHash>`, I think using a name alias is not necessary.

k2/csrc/fsa_equivalent.cc
qindazhu: Yes, I have mentioned this yesterday, will make another PR after merging your code.

k2/csrc/fsa_equivalent.cc
qindazhu: Yes, you are right.

k2/csrc/fsa_equivalent.cc
qindazhu: Sure

k2/csrc/util.h
qindazhu: I think it's better here to put `seed` as the first parameter since: 

1. it it the input prarameter
2. when we read the code from left to right, we see first `seed` then `value` which is more readable.
3. consistent siganature with `boost::hash_combine`

k2/csrc/fsa_equivalent.cc
qindazhu: If the input fsa is connected, probabilistically the algorithm will always terminate in finite time.

k2/csrc/fsa_equivalent_test.cc
qindazhu: Sure

k2/csrc/fsa_equivalent_test.cc
qindazhu: change to `auto` and `!=` (but I still think `<` is better here as we don't really care about how many times the loop will run.)

k2/csrc/fsa_equivalent_test.cc
qindazhu: Sure, Add a case.

qindazhu:
Ready to merge.

k2/csrc/util.h
csukuangfj: @danpovey may have something to say about this.

k2/csrc/fsa.h
danpovey: Let's remove the Weight typedef.  I prefer to remove Label and StateId typedefs for now too, as well.

k2/csrc/fsa_equivalent.cc
danpovey: In this case, be sure to update the documentation of Connect() that it requires the input to be connected, and add the necessary asserts.  Always do this (if you weren't doing this already)... it's important to keep track of preconditions.

k2/csrc/util.h
danpovey: I think I'm OK to have seed first in this instance, specifically because it's following the boost::hash_combine tradition.
BTW, it's a bit annoying that I can't search for conversations in github because they are all resolved.  I dont know if there's a way to mark resolved without the text disappearing...

k2/csrc/fsa_equivalent_test.cc
danpovey: @csukuangfj  next time please summarize the reason in text, was hard to get to the video!
Sorry I don't buy this.  My problem is that it doesn't work when you increment by 2, and it doesn't work in cases where there was less than an empty range, e.g. if begin > end (in these cases you generally want to do nothing, not loop forever and segfault).

danpovey:
Great!   Thanks @csukuangfj  for your very detailed comments!  You guys can merge when you think it's ready.

csukuangfj:
+2

qindazhu:
@danpovey did you set a policy that one cannot merge his own PR? I cann not merge this by myself. Maybe you or @csukuangfj can help to merge it.

danpovey:
I didn't set that, might be a default.

k2/csrc/fsa_equivalent_test.cc
```diff
@@ -0,0 +1,86 @@
...
+#include <utility>
+#include <vector>
+
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+#include "k2/csrc/fsa.h"
+
+namespace k2 {
+
+TEST(Properties, RandomPathFail) {
```
csukuangfj, line 19: Forgot to mention that it should NOT be `Properties` here.

qindazhu: updated in this pr https://github.com/danpovey/k2/pull/17

PR #17, opened 2020-04-28
implement IsArcSorted

qindazhu:
Will implement `Intersect` based on this.

### DIFF
--- a/k2/csrc/fsa.h
+++ b/k2/csrc/fsa.h
@@ -58,7 +58,8 @@ struct ArcHash {
 
 struct ArcLabelCompare {
   bool operator()(const Arc &a, const Arc &b) const {
-    return a.label < b.label;
+    // compares `label` first, then `dest_state`
+    return std::tie(a.label, a.dest_state) < std::tie(b.label, b.dest_state);
   }
 };
 
--- a/k2/csrc/fsa.h
+++ b/k2/csrc/fsa.h
@@ -8,6 +8,7 @@
 #define K2_CSRC_FSA_H_
 
 #include <cstdint>
+#include <tuple>
 #include <vector>
 
 #include "k2/csrc/util.h"
--- a/k2/csrc/properties.cc
+++ b/k2/csrc/properties.cc
@@ -58,14 +58,14 @@ bool IsTopSorted(const Fsa &fsa) {
 bool IsArcSorted(const Fsa &fsa) {
   ArcLabelCompare comp;
   StateId final_state = fsa.NumStates() - 1;
-  auto begin = fsa.arcs.begin();
+  const auto begin = fsa.arcs.begin();
   const auto &arc_indexes = fsa.arc_indexes;
   // we will not check the final state as it has no arcs leaving it.
   for (StateId state = 0; state < final_state; ++state) {
     // as non-empty `fsa` contains at least two states,
     // we can always access `state + 1` validly.
-    if (!is_sorted(begin + arc_indexes[state], begin + arc_indexes[state + 1],
-                   comp))
+    if (!std::is_sorted(begin + arc_indexes[state],
+                        begin + arc_indexes[state + 1], comp))
       return false;
   }
   return true;
--- a/k2/csrc/properties.cc
+++ b/k2/csrc/properties.cc
@@ -55,6 +55,22 @@ bool IsTopSorted(const Fsa &fsa) {
   return true;
 }
 
+bool IsArcSorted(const Fsa &fsa) {
+  ArcLabelCompare comp;
+  StateId final_state = fsa.NumStates() - 1;
+  auto begin = fsa.arcs.begin();
+  const auto &arc_indexes = fsa.arc_indexes;
+  // we will not check the final state as it has no arcs leaving it.
+  for (StateId state = 0; state < final_state; ++state) {
+    // as non-empty `fsa` contains at least two states,
+    // we can always access `state + 1` validly.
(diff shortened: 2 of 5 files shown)


danpovey:
@qindazhu I realized that when we do arc-sorting, I need them to be doubly arc-sorted: by symbol, and then, in case of ties on the symbol, by destination-state.
Also I think we have to change kFinalSymbol to be a very large positive number, e.g. max possible int32, instead of -1.  Will make things like composition easier (epsilons will be first after arc sorting).

danpovey:
The doubly-arc-sorted thing is to make determinization easier (I am working on the algorithm).

qindazhu:
> @qindazhu I realized that when we do arc-sorting, I need them to be doubly arc-sorted: by symbol, and then, in case of ties on the symbol, by destination-state.
> Also I think we have to change kFinalSymbol to be a very large positive number, e.g. max possible int32, instead of -1. Will make things like composition easier (epsilons will be first after arc sorting).

OK, I'll update `ArcLabelCompare`.

qindazhu:
@danpovey , updated

k2/csrc/properties.cc
```diff
@@ -55,6 +55,22 @@ bool IsTopSorted(const Fsa &fsa) {
   return true;
 }
 
+bool IsArcSorted(const Fsa &fsa) {
+  ArcLabelCompare comp;
```
csukuangfj, line 59: Check that the fsa is not empty.

k2/csrc/properties.cc
```diff
@@ -55,6 +55,22 @@ bool IsTopSorted(const Fsa &fsa) {
...
+  StateId final_state = fsa.NumStates() - 1;
+  auto begin = fsa.arcs.begin();
```
csukuangfj, line 61:
```cpp
const auto begin = fsa.arcs.cbegin();
```

k2/csrc/properties.cc
```diff
@@ -55,6 +55,22 @@ bool IsTopSorted(const Fsa &fsa) {
...
+  const auto &arc_indexes = fsa.arc_indexes;
+  // we will not check the final state as it has no arcs leaving it.
+  for (StateId state = 0; state < final_state; ++state) {
+    // as non-empty `fsa` contains at least two states,
```
csukuangfj, line 65: Before saying this, you have to ensure that the passing fsa is not empty.

k2/csrc/properties.cc
```diff
@@ -55,6 +55,22 @@ bool IsTopSorted(const Fsa &fsa) {
...
+    // we can always access `state + 1` validly.
+    if (!is_sorted(begin + arc_indexes[state], begin + arc_indexes[state + 1],
```
csukuangfj, line 67:
```cpp
std::is_sorted
```

`std` is missing.

k2/csrc/properties.cc
qindazhu: it will work on empty fsa 
https://github.com/danpovey/k2/blob/aac136e0a90c54e0bd5a2400ecc1466b5115bbe2/k2/csrc/properties.cc#L64

also see tests
https://github.com/danpovey/k2/blob/aac136e0a90c54e0bd5a2400ecc1466b5115bbe2/k2/csrc/properties_test.cc#L189-L192

k2/csrc/properties.cc
qindazhu: It does not matter, as for empty fsa, `final_state` will be -1;

k2/csrc/properties.cc
qindazhu: fixed. It shocked me that `gcc` compiles it successfully.

k2/csrc/fsa.h
```diff
@@ -19,11 +20,11 @@ using StateId = int32_t;
...
 enum {
-  kFinalSymbol = -1,  // final-costs are represented as arcs with
-                      // kFinalSymbol as their label, to the final
-                      // state (see Fst::final_state).
-  kEpsilon = 0        // Epsilon, which means "no symbol", is numbered zero,
-                      // like in OpenFst.
+  kFinalSymbol = INT32_MAX,  // final-costs are represented as arcs with
+                             // kFinalSymbol as their label, to the final
+                             // state (see Fst::final_state).
+  kEpsilon = 0  // Epsilon, which means "no symbol", is numbered zero,
```
csukuangfj, line 26: It'd better place a comma `,` after 0.

qindazhu: is that from google style?

csukuangfj: no, it's optional.

csukuangfj:
there are conflicts now.

csukuangfj:
+2

merging

qindazhu:
Conflicts Resloved. Ready to merge.

Note that I will get below error with `glog`
```
 No rule to make target `third_party/glog/install/lib/libglog.a', needed by `k2/csrc/libfsa_algo.so'.  Stop.
```
The generated file on my machine is 
```
third_party/glog/install/lib64/libglogd.a
```

csukuangfj:
Oh, same with gtest.

I'll figure out a way to solve this problem. But for now, please change it locally by hand.

qindazhu:
> Oh, same with gtest.
> 
> I'll figure out a way to solve this problem. But for now, please change it locally by hand.

Note that the file name, it's <pre>libglog<b>d</b>.a</pre>, instead of `libglog.a`

PR #18, opened 2020-04-28
Implement topological sort.

csukuangfj:
For the following fsa

a

after `TopSort`, it produces the following fsa

b

with the state map (mapping state from the sorted fsa to the original fsa) 
```
0 -> 0
1 -> 4
2 -> 5
3 -> 2
4 -> 3
5 -> 1
6 -> 6
```

---

More unit test cases will be added later (possibly tomorrow).

### DIFF
--- a/k2/csrc/CMakeLists.txt
+++ b/k2/csrc/CMakeLists.txt
@@ -25,7 +25,6 @@ function(k2_add_fsa_test name)
       gtest
       gtest_main
   )
-
   add_test(NAME "Test.${name}"
     COMMAND
       $<TARGET_FILE:${name}>
--- a/k2/csrc/fsa_algo.cc
+++ b/k2/csrc/fsa_algo.cc
@@ -57,7 +57,7 @@ void ConnectCore(const Fsa &fsa, std::vector<int32_t> *state_map) {
       auto state = current_state.state;  // get a copy since we will destroy it
       stack.pop();
       if (!stack.empty()) {
-        // if it has a parent, set the parent's coaccessible flag
+        // if it has a parent, set the parent's co-accessible flag
         if (coaccessible[state]) {
           auto &parent = stack.top();
           coaccessible[parent.state] = true;
@@ -194,4 +194,110 @@ void ArcSort(const Fsa &a, Fsa *b,
   }
   if (arc_map != nullptr) arc_map->swap(indexes);
 }
+
+bool TopoSort(const Fsa &a, Fsa *b,
+              std::vector<int32_t> *state_map /*= nullptr*/) {
+  CHECK_NOTNULL(b);
+  b->arc_indexes.clear();
+  b->arcs.clear();
+
+  if (state_map) state_map->clear();
+
+  if (IsEmpty(a)) return false;
+  static constexpr int32_t kNotVisited = 0;  // a node that has not been visited
+  static constexpr int32_t kVisiting = 1;    // a node that is under visiting
+  static constexpr int32_t kVisited = 2;     // a node that has been visited
+
+  auto num_states = a.NumStates();
+  auto final_state = num_states - 1;
+  std::vector<int8_t> state_status(num_states, kNotVisited);
+
+  // map order to state.
+  // state 0 has the largest order, i.e., num_states - 1
+  // final_state has the least order, i.e., 0
+  std::vector<int32_t> order;
+
+  order.reserve(a.NumStates());
+
+  std::stack<DfsState> stack;
+  stack.push({0, a.arc_indexes[0], a.arc_indexes[1]});
+  state_status[0] = kVisiting;
+  bool is_acyclic = true;
+  while (is_acyclic && !stack.empty()) {
+    auto &current_state = stack.top();
+    if (current_state.arc_begin == current_state.arc_end) {
+      // we have finished visiting this state
+      state_status[current_state.state] = kVisited;
+      order.push_back(current_state.state);
+      stack.pop();
+      continue;
+    }
+    const auto &arc = a.arcs[current_state.arc_begin];
+    auto next_state = arc.dest_state;
+    auto status = state_status[next_state];
+    switch (status) {
+      case kNotVisited: {
+        // a new discovered node
+        state_status[next_state] = kVisiting;
+        auto arc_begin = a.arc_indexes[next_state];
+        if (next_state != final_state)
+          stack.push({next_state, arc_begin, a.arc_indexes[next_state + 1]});
+        else
+          stack.push({next_state, arc_begin, arc_begin});
+        ++current_state.arc_begin;
+        break;
+      }
+      case kVisiting:
+        // this is a back arc indicating a loop in the graph
+        is_acyclic = false;
+        break;
+      case kVisited:
+        // this is a forward cross arc, do nothing.
+        ++current_state.arc_begin;
+        break;
+      default:
+        LOG(FATAL) << "Unreachable code is executed!";
+        break;
+    }
+  }
+
+  if (!is_acyclic) return false;
+
+  std::vector<int32_t> state_a_to_b(num_states);
+  for (auto i = 0; i != num_states; ++i) {
+    state_a_to_b[order[num_states - 1 - i]] = i;
+  }
+
+  // start state maps to start state
+  CHECK_EQ(state_a_to_b.front(), 0);
+  // final state maps to final state
+  CHECK_EQ(state_a_to_b.back(), final_state);
+
+  b->arcs.reserve(a.arc_indexes.size());
+  b->arc_indexes.resize(num_states);
+
+  int32_t arc_begin;
+  int32_t arc_end;
+  for (auto state_b = 0; state_b != num_states; ++state_b) {
+    auto state_a = order[num_states - 1 - state_b];
+    arc_begin = a.arc_indexes[state_a];
+    if (state_a != final_state)
+      arc_end = a.arc_indexes[state_a + 1];
+    else
+      arc_end = arc_begin;
+
+    b->arc_indexes[state_b] = static_cast<int32_t>(b->arcs.size());
+    for (; arc_begin != arc_end; ++arc_begin) {
+      auto arc = a.arcs[arc_begin];
+      arc.src_state = state_b;
+      arc.dest_state = state_a_to_b[arc.dest_state];
+      b->arcs.push_back(arc);
+    }
+  }
+  if (state_map)
+    state_map->insert(state_map->end(), order.rbegin(), order.rend());
+
+  return true;
+}
+
 }  // namespace k2
--- a/k2/csrc/fsa_algo.cc
+++ b/k2/csrc/fsa_algo.cc
@@ -13,6 +13,7 @@
 #include <utility>
 
 #include "glog/logging.h"
+#include "k2/csrc/properties.h"
 
 namespace {
 
@@ -98,7 +99,7 @@ void ConnectCore(const Fsa &fsa, std::vector<int32_t> *state_map) {
 
 void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
   CHECK_NOTNULL(b);
-  if (arc_map) arc_map->clear();
+  if (arc_map != nullptr) arc_map->clear();
 
   std::vector<int32_t> state_b_to_a;
   ConnectCore(a, &state_b_to_a);
@@ -108,7 +109,7 @@ void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
   b->arcs.clear();
   b->arcs.reserve(a.arcs.size());
 
-  if (arc_map) {
+  if (arc_map != nullptr) {
     arc_map->clear();
     arc_map->reserve(a.arcs.size());
   }
@@ -144,9 +145,7 @@ void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
       arc.src_state = i;
       arc.dest_state = state_b;
       b->arcs.push_back(arc);
-      if (arc_map) {
-        arc_map->push_back(arc_begin);
-      }
+      if (arc_map != nullptr) arc_map->push_back(arc_begin);
     }
   }
 }
@@ -195,18 +194,20 @@ void ArcSort(const Fsa &a, Fsa *b,
   if (arc_map != nullptr) arc_map->swap(indexes);
 }
 
-bool TopoSort(const Fsa &a, Fsa *b,
-              std::vector<int32_t> *state_map /*= nullptr*/) {
+void TopSort(const Fsa &a, Fsa *b,
+             std::vector<int32_t> *state_map /*= nullptr*/) {
   CHECK_NOTNULL(b);
   b->arc_indexes.clear();
   b->arcs.clear();
 
-  if (state_map) state_map->clear();
+  if (state_map != nullptr) state_map->clear();
 
-  if (IsEmpty(a)) return false;
-  static constexpr int32_t kNotVisited = 0;  // a node that has not been visited
-  static constexpr int32_t kVisiting = 1;    // a node that is under visiting
-  static constexpr int32_t kVisited = 2;     // a node that has been visited
+  if (IsEmpty(a)) return;
+  if (!IsConnected(a)) return;
+
+  static constexpr int8_t kNotVisited = 0;  // a node that has not been visited
+  static constexpr int8_t kVisiting = 1;    // a node that is under visiting
+  static constexpr int8_t kVisited = 2;     // a node that has been visited
 
   auto num_states = a.NumStates();
   auto final_state = num_states - 1;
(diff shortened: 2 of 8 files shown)


danpovey:
Great.  Please call it TopSort though, not TopoSort.

k2/csrc/fsa_algo.cc
```diff
@@ -147,4 +147,109 @@ void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+    b->arc_indexes[state_b] = static_cast<int32_t>(b->arcs.size());
+    for (; arc_begin != arc_end; ++arc_begin) {
+      auto arc = a.arcs[arc_begin];
+      arc.src_state = state_b;
+      arc.dest_state = state_a_to_b[arc.dest_state];
+      b->arcs.push_back(arc);
+    }
+  }
+  if (state_map)
+    state_map->insert(state_map->end(), order.rbegin(), order.rend());
```
qindazhu, line 250:
```
std::reverse(order.begin(), order.end());
state_map->swap(order)
```
will reuse memory allocated by `order`.

k2/csrc/fsa_algo.cc
```diff
@@ -147,4 +147,109 @@ void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
 
+bool TopoSort(const Fsa &a, Fsa *b,
+              std::vector<int32_t> *state_map /*= nullptr*/) {
+  CHECK_NOTNULL(b);
+  b->arc_indexes.clear();
+  b->arcs.clear();
+
+  if (state_map) state_map->clear();
+
+  if (IsEmpty(a)) return false;
```
qindazhu, line 158: maybe `true` for empty fsa? just leave `b` and `state_map` empty

k2/csrc/fsa_algo.cc
```diff
@@ -147,4 +147,109 @@ void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+  static constexpr int32_t kNotVisited = 0;  // a node that has not been visited
```
qindazhu, line 159: `int8_t`?

k2/csrc/fsa_algo.h
```diff
@@ -207,6 +207,20 @@ void IntersectPruned2(const Fsa &a, const float *a_cost, const Fsa &b,
 void RandomPath(const Fsa &a, const float *a_cost, Fsa *b,
                 std::vector<int32_t> *state_map = nullptr);
 
+/**
+    Sort the input fsa topologically.
+
+    Since we can only sort an acyclic fsa. It returns false and an empty fsa
```
qindazhu, line 213: input `fsa` is required to be connected, right?

k2/csrc/fsa_algo.cc
```diff
@@ -147,4 +147,109 @@ void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+  auto num_states = a.NumStates();
+  auto final_state = num_states - 1;
+  std::vector<int8_t> state_status(num_states, kNotVisited);
+
+  // map order to state.
+  // state 0 has the largest order, i.e., num_states - 1
+  // final_state has the least order, i.e., 0
+  std::vector<int32_t> order;
+
+  order.reserve(a.NumStates());
```
qindazhu, line 172: use `num_states` as you have defined it before.

k2/csrc/fsa_algo.cc
```diff
@@ -147,4 +147,109 @@ void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+
+  auto num_states = a.NumStates();
+  auto final_state = num_states - 1;
+  std::vector<int8_t> state_status(num_states, kNotVisited);
+
+  // map order to state.
+  // state 0 has the largest order, i.e., num_states - 1
+  // final_state has the least order, i.e., 0
+  std::vector<int32_t> order;
+
```
qindazhu, line 171: also remove this blank line, as you `reserve` it just after defining it.

k2/csrc/fsa_algo.cc
```diff
@@ -147,4 +147,109 @@ void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+      default:
+        LOG(FATAL) << "Unreachable code is executed!";
+        break;
+    }
+  }
+
+  if (!is_acyclic) return false;
+
+  std::vector<int32_t> state_a_to_b(num_states);
+  for (auto i = 0; i != num_states; ++i) {
```
qindazhu, line 219: `std::reverse_copy`?

k2/csrc/fsa_algo.cc
```diff
@@ -147,4 +147,109 @@ void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
   }
 }
 
+bool TopoSort(const Fsa &a, Fsa *b,
+              std::vector<int32_t> *state_map /*= nullptr*/) {
+  CHECK_NOTNULL(b);
+  b->arc_indexes.clear();
+  b->arcs.clear();
+
+  if (state_map) state_map->clear();
```
qindazhu, line 156: Actually I prefer `state_map != nullptr` but I did't find the rule in Google Style, just an example in

https://google.github.io/styleguide/cppguide.html

csukuangfj: I assume you are referring to https://google.github.io/styleguide/cppguide.html#Run-Time_Type_Information__RTTI_

Tensorflow also uses
```cpp
if (x != nullptr)
```
Please see
- https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/tensor.cc#L177
- https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/tensor.cc#L213
- https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/tensor.cc#L251

-----

I will change it to `!= nullptr`

k2/csrc/fsa_algo.h
danpovey: Mm.  If possible it would be nice not to require that the input is connected.  We can have it so if the input was not connected the output won't necessarily be connected though.  Obviously make clear in the comment.

k2/csrc/fsa_algo.cc
csukuangfj: Tensorflow also uses

- https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/tensor.cc#L696
```cpp
  if (in_size == 0) {
```

----

Maybe we can also follow this style.

```cpp
if (x == 0)

if (b == false)
```

k2/csrc/fsa_algo.h
qindazhu: if  the input is not connected, then in our current implementation, those non-connected states may be not visited at all as we were starting from the start state, right? therefore, the program may crash here

https://github.com/danpovey/k2/blob/ad89b577f71dea5ead2b4a982227f57969ceb374/k2/csrc/fsa_algo.cc#L220

as `order[num_states - 1 - i]` may return meaningless value.

k2/csrc/fsa_algo.cc
qindazhu: Yes, I like all of those above except `if (bool_value == false)`, I prefer `if (!bool_value)` for boolean values, as `if` is checking true or false and boolean value is already a `true` or `false`.

Another example is that in STL code there is a lot of `empty()` function and people tend to call `if(v.empty())`

k2/csrc/fsa_algo.h
csukuangfj: I will change the implementation to support non-connected input fsa.

csukuangfj:
It happens that not all un-connected graphs can be sorted topologically.

For the following fsa:
trim

What should state 5 be numbered in a TopSorted FSA?

danpovey:
Hm, I see.
I guess you can require that it be connected then.
But can we make sure that Connect() is able to produce top-sorted output as
long as the input is free of cycles (which are not self-loops)?
This should be possible just by using the right queue topology, i.e. a fifo
queue, to do breadth first search.

danpovey:
If that's hard, it's OK, we can make them separate calls for now.

csukuangfj:
> But can we make sure that Connect() is able to produce top-sorted output as
long as the input is free of cycles (which are not self-loops)?

I am going to implement it in the current `ConnectCore`.

k2/csrc/fsa_algo.cc
csukuangfj: Done.

k2/csrc/fsa_algo.cc
csukuangfj: I've changed the function to return `void`.

k2/csrc/fsa_algo.cc
csukuangfj: Done.

k2/csrc/fsa_algo.h
csukuangfj: It checks explicitly that if the input fsa is not connected, then we return an empty fsa.

k2/csrc/fsa_algo.cc
csukuangfj: Done.

k2/csrc/fsa_algo.cc
csukuangfj: Done.

k2/csrc/CMakeLists.txt
```diff
@@ -1,95 +1,39 @@
...
-)
-
-# TODO(fangjun): write some helper functions to create targets.
+# sort the source files alphabetically
+add_library(fsa
+  fsa_algo.cc
+  fsa_equivalent.cc
+  fsa_renderer.cc
+  fsa_util.cc
+  properties.cc
```
csukuangfj, line 7: should we rename `properties.cc` to `fsa_properties.cc` ?

k2/csrc/fsa_algo.cc
csukuangfj: Done

k2/csrc/CMakeLists.txt
qindazhu: I'm OK with this. BTW, maybe we should move all FSA files to a `fsa` folder as we definitely will add other libraries （such as cuFsa?）

k2/csrc/fsa_algo.cc
```diff
@@ -194,4 +193,111 @@ void ArcSort(const Fsa &a, Fsa *b,
   }
   if (arc_map != nullptr) arc_map->swap(indexes);
 }
+
+void TopSort(const Fsa &a, Fsa *b,
```
qindazhu, line 197: I think make boolean as the return value may be better, as the caller will know what it should do next according to the return value:

- if the input fsa is empty, we return true, as sorted fsa on empty fsa is still empty.
- if the input fsa is non-connected or cyclic, we return false. Then the caller knows that the input is invalid, the caller may check the input fsa.

For now, as we always clear the output fsa at the beginning, the caller can not tell the difference when the function returns from different branches.

danpovey:
Lets leave the filenames as is for now.  Maybe fsa_algo could just be
algorithms.  We will already access via k2 prefix.

danpovey:
Maybe remove fsa prefix on filenames.

danpovey:
Have to talk with cuda experts re how to structure that part.

k2/csrc/fsa_algo.cc
csukuangfj: Fixed.

csukuangfj:
I've added support to build an FSA from a transition table, like OpenFST, simplifying
the task of creating FSAs in unittests.

k2/csrc/fsa_algo.cc
csukuangfj: fixed.

k2/csrc/CMakeLists.txt
csukuangfj: left as it is.

csukuangfj:
Unit tests are added. Please have a review and merge if necessary
@qindazhu @danpovey

k2/csrc/fsa_util.cc
```diff
@@ -6,9 +6,91 @@
...
+namespace {
+
+int32_t StringToInt(const std::string &s, bool *is_ok = nullptr) {
+  CHECK_EQ(s.empty(), false);
+
+  bool ok = false;
+  char *p = nullptr;
+  // std::strtol requires a `long` type
+  long n = std::strtol(s.c_str(), &p, 10);
+  if (*p == 0) ok = true;
```
qindazhu, line 32: `if (*p == '\0')`?

k2/csrc/fsa_util.cc
```diff
@@ -6,9 +6,91 @@
...
+
+  int32_t res = n;
```
qindazhu, line 34: `static_cast<int32_t>(n)`

k2/csrc/fsa_util.cc
csukuangfj: 0 is identical to `\0` with less typing.

k2/csrc/fsa_util.cc
csukuangfj: done.

k2/csrc/fsa_util.cc
```diff
@@ -6,9 +6,91 @@
...
+  size_t end = in.size();
+  while (true) {
+    auto pos = in.find_first_of(delim, start);
+    if (pos == in.npos) break;
+
+    auto sub = in.substr(start, pos - start);
+    start = pos + 1;
+
+    TrimString(&sub);
+    if (sub.empty() == false) out->emplace_back(std::move(sub));
```
qindazhu, line 95: really suggest `if(!sub.empty())` It's more readable

danpovey:
I agree on !sub.empty().

You guys can merge if you think itit's OK.

k2/csrc/fsa_util.cc
csukuangfj: done.

k2/csrc/fsa_util.cc
```diff
@@ -34,4 +116,53 @@ void GetEnteringArcs(const Fsa &fsa, std::vector<int32_t> *arc_index,
...
+
+    auto fields = StringVectorToIntVector(splits);
+    auto num_fields = fields.size();
+    if (num_fields == 3u) {
+      Arc arc;
+      arc.src_state = fields[0];
+      arc.dest_state = fields[1];
+      arc.label = fields[2];
+
+      if (arc.src_state >= vec.size()) vec.resize(arc.src_state + 1);
```
qindazhu, line 154: it's better to do `new_size = max(arc.arc_state, arc.dest_state)` and then `resize`, otherwise you may `resize` twice here.

csukuangfj:
please use **squash** and merge. there are too many commits in this pullrequest.

k2/csrc/fsa_util.cc
csukuangfj: Done.

k2/csrc/fsa_util.cc
qindazhu: check section `0 and nullptr/NULL` in [google style](https://google.github.io/styleguide/cppguide.html)

csukuangfj: thanks, will update it in the next commit.

k2/csrc/fsa_util.cc
```diff
@@ -6,9 +6,104 @@
...
+
+namespace {
+/** Convert a string to an integer.
+
+  @param [in]   s     The input string.
+  @param [out]  is    If non-null, it is set to true when the conversion
+                      is successful; false otherwise
+  @return The converted integer.
+ */
+int32_t StringToInt(const std::string &s, bool *is_ok = nullptr) {
```
qindazhu, line 25: I was just thinking that maybe we sould not put those helper functions in `fsa_utils` as all functions are not related with fsa except `stringToFsa`

k2/csrc/fsa_util.cc
csukuangfj: fixed in #23

k2/csrc/fsa_util.cc
csukuangfj: I agree.

When such helper functions reach a certain number, we can place them separately.

danpovey:
For now they are in a .cc file so it doesn't matter.
We don't have to be too focused on style issues at the moment.
I'm more concerned about the overall structure.

csukuangfj:
@danpovey 
> But can we make sure that Connect() is able to produce top-sorted output as
long as the input is free of cycles (which are not self-loops)?

I have a question about this.

For the following input fsa `a`
a

it is acyclic and topsorted.

I've changed the `ConnectCore` to produce a topsorted output fsa as long as the input
fsa is acyclic. It produces the following output for the above input fsa

b

----

Notice that the state 2 in the output fsa maps to the state 3 in the input fsa.

My question is that the topological sort of an acyclic fsa is **not unique**, does the
output fsa have to keep the same state number as the input fsa if the input fsa is topsorted,
i.e., does the state map has to be an identity map ?

pzelasko:
Shouldn't the topological sort respect the arcs ordering? E.g. if the FSA is label sorted then follow the same order in these cases (and if it's not sorted then whatever arc comes first in the underlying data structure)?

danpovey:
I dont want to impose any additional requirements on it.  Leave more
freedom for optimization.  We can sort the arcs separately.

PR #19, opened 2020-04-29
implement ArcSort and refactor IsConnect with ConnectCore

qindazhu:


qindazhu:
mm.. guys, can you kindly help to review this?

csukuangfj:
I will review this to-night.

danpovey:
Thanks!!!

k2/csrc/fsa_algo.cc
```diff
@@ -6,7 +6,9 @@
 
 #include "k2/csrc/fsa_algo.h"
 
+#include <algorithm>
```
csukuangfj, line 9: Your name in the copyright info is missing.

qindazhu: mm..OK, thanks.

k2/csrc/fsa_algo.cc
```diff
@@ -147,4 +149,48 @@ void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+  CHECK_NOTNULL(b);
+  b->arc_indexes = a.arc_indexes;
+  b->arcs.clear();
+  b->arcs.reserve(a.arcs.size());
+  if (arc_map != nullptr) arc_map->clear();
+
+  using ArcWithIndex = std::pair<Arc, int32_t>;
+  std::vector<int32_t> indexes(a.arcs.size());  // index mapping
+  int32_t n = 0;
+  std::generate(indexes.begin(), indexes.end(), [&n]() { return n++; });
```
csukuangfj, line 163: `std::iota` https://en.cppreference.com/w/cpp/algorithm/iota

No lambda function and no extra function invoke for `std::itoa`.

`int32_t n = 0; ` can be removed when using `std::itoa`.

-----

`++n` is preferred over `n++`.

k2/csrc/fsa_algo.cc
```diff
@@ -147,4 +149,48 @@ void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+  const auto arc_begin_iter = a.arcs.begin();
+  const auto index_begin_iter = indexes.begin();
+  // we will not process the final state as it has no arcs leaving it.
+  StateId final_state = a.NumStates() - 1;
+  for (StateId state = 0; state < final_state; ++state) {
+    int32_t begin = a.arc_indexes[state];
+    // as non-empty fsa `a` contains at least two states,
+    // we can always access `state + 1` validly.
+    int32_t end = a.arc_indexes[state + 1];
+    std::vector<ArcWithIndex> arc_range_to_sorted;
```
csukuangfj, line 173: `*_to_be_sorted`

k2/csrc/fsa_algo.cc
qindazhu: Sure

csukuangfj:
+2

k2/csrc/fsa_algo.cc
qindazhu: `int32_t n = -1` is required if using `++n` here, which is not so readable.

Anyway, changed to iota

qindazhu:
Thanks!

k2/csrc/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 add_library(properties properties.cc)
 target_include_directories(properties PUBLIC ${CMAKE_SOURCE_DIR})
 target_compile_features(properties PUBLIC cxx_std_11)
+target_link_libraries(properties PUBLIC fsa_algo)
```
csukuangfj, line 4: I am trying to add

```
target_link_libraries(fsa_algo PUBLIC properties)
```
since I want to use `IsConnected()` in `TopSort`;
but this results in circular dependencies:

```
CMake Error: The inter-target dependency graph contains the following strongly connected component (cycle):
  "properties" of type SHARED_LIBRARY
    depends on "fsa_algo" (weak)
  "fsa_algo" of type SHARED_LIBRARY
    depends on "properties" (weak)
At least one of these targets is not a STATIC_LIBRARY.  Cyclic dependencies are allowed only among static libraries.
```

-----

Should we remove all `add_library` and write a single `add_library` like the following

```
add_library(fsa
  fsa_equivalent.cc
  fsa_util.cc
  properties.cc
  ....
)
```

qindazhu: I'm OK with this (maybe except `fsa_render`?). Finally I guess we will have two `.so` files for common FSA operations, one for CPU and another for GPU?

PR #21, opened 2020-05-01
fix cmake configure errors.

### LINKED ISSUE #20: Compile issue
The code compiles fine on my MacBook (Darwin - 17.7.0 - x86_64), but it fails on Linux (Linux - 4.9.0-11-amd64 - x86_64), with the following error:

```
$ cmake ..
CMake Error at cmake/googletest.cmake:40 (target_include_directories):
  Cannot specify include directories for imported target "gtest".
Call Stack (most recent call first):
  cmake/googletest.cmake:50 (download_googltest)
  CMakeLists.txt:29 (include)
```

csukuangfj:
ExternalProject_Add is replaced with FetchContent. Since FetchContent
is available only in CMake 3.11+, we've placed a copy of it in k2.

This should fix #20 

@danpovey @qindazhu 
please have a review.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,8 +22,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        # TODO(fangjun): support "windows-latest"
-        os: [ubuntu-latest, macOS-latest]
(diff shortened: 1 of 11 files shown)


csukuangfj:
The problem is fixed, so merging.

PR #22, opened 2020-05-02
implement intersect

qindazhu:


k2/csrc/fsa.h
```diff
@@ -10,6 +10,7 @@
 #include <cstdint>
 #include <tuple>
 #include <vector>
+#include <utility>
```
csukuangfj, line 13: Have you installed `clang-format`?

It should reorder the headers for you automatically.

k2/csrc/fsa_algo.cc
```diff
@@ -150,6 +153,148 @@ void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
   }
 }
 
+inline int32_t InsertIntersectionState(
```
csukuangfj, line 156: It would be better if this function is placed inside the above unnamed namespace.

k2/csrc/fsa_algo_test.cc
```diff
@@ -19,10 +19,13 @@
 namespace k2 {
 
 TEST(FsaAlgo, Connect) {
-  std::vector<Arc> arcs = {
-      {0, 1, 1}, {0, 2, 2}, {1, 3, 3}, {1, 6, 6},
-      {2, 4, 2}, {2, 6, 3}, {5, 0, 1},
-  };
+  std::vector<Arc> arcs = {{0, 1, 1},
```
csukuangfj, line 22: `StringToFsa` can help you build the fsa.

k2/csrc/fsa.h
qindazhu: let me check.
```
Package              Version
-------------------- ----------
absl-py              0.9.0
argcomplete          1.9.4
astor                0.6.2
bleach               1.5.0
cachetools           4.0.0
certifi              2019.11.28
chardet              3.0.4
clang-format         9.0.0

```

csukuangfj:
could you visualize the result and post them here?

k2/csrc/fsa.h
qindazhu: seems it will not reorder in my machine...strange.

csukuangfj: Have you configured https://github.com/google/vim-codefmt ?

qindazhu: Yes, I have done it before. Anyway, changed the order manually for now.

k2/csrc/fsa_algo.cc
qindazhu: Sure

k2/csrc/fsa_algo_test.cc
qindazhu: Actually I dont think it's a good idea to use StringToFsa in testing, as it only save the line for defining `arc_indexes` but we cannot know what `arc_indexes` is by reading code.
However, I think `StringToFsa` will be used in the future for reading Fsa from files.

qindazhu:
Input FSA `a`
a

Input FSA `b`
b


Output FSA `c = a^b`.
c

csukuangfj:
Should the output be `trimmed` (connected) ?

qindazhu:
> Should the output be `trimmed` (connected) ?

I think it's better to leave this for the caller to do that.

csukuangfj:
OpenFST produces a connected fsa.

Please refer to http://www.openfst.org/twiki/bin/view/FST/IntersectDoc

qindazhu:
> OpenFST produces a connected fsa.
> 
> Please refer to http://www.openfst.org/twiki/bin/view/FST/IntersectDoc

OpenFst supports kinds of `filters`, so callers can decide what fsa they want. However, we don't support such `filters`, so maybe callers want do `RmEplision` then `Compose`, etc. 

Thus, I think it's better to let the caller do this instead of doing such operations (we make the decision for them) in one function call.

csukuangfj:
ok.

Another point: why is there a self-loop with label `epsilon` in the input fsa `a` and the output fsa?

qindazhu:
> ok.
> 
> Another point: why is there a self-loop with label `epsilon` in the input fsa `a` and the output fsa?

I guess what you are asking is why there is self-loop with label `epsilon` in the input fsa (as if there is one in the input, there may be such a state in the output).

Well, that's just because we support such input fsa, we only require that either `a` or `b` is epsilon-free, but not limit what labels can be on self-loop.

csukuangfj:
I think epsilon free means you can go to state `b` from `a`  with the epslion label (i.e., no input).
But `a` should be different from `b`. 

Does every node implicitly have a self-loop with label epsilon?

qindazhu:
We treat the epsilon free fsa has a self loop with epsilon at each state. Say if `a` has `epsilon` arc and `b` is epsilon-free. When we match `Arc(src_a, dest_a, epsilon)` in `a` with `Arc(src_b, dest_b, label)`, we always add a arc `(src_a, src_b) -> (dest_a, src_b)` with label epsilon (That means we add a self-loop in `src_b` with epsilon)

csukuangfj:
> Arc(src_a, dest_a, epsilon)

I think `dest_a != src_a`.

From wiki: https://en.wikipedia.org/wiki/Epsilon_transition
> An epsilon transition[1] (also epsilon move[1] or lambda transition) allows an automaton to ******change****** its state spontaneously, i.e. without consuming an input symbol. 

An epsilon transition -->changes<--- the state.

The input `fsa a` has a transition

```
3 3 0
```
which confuses me.

qindazhu:
As I said before, the caller can remove there arcs (even for input fsa) if they want or they should. That's not the responsiblity of `Intersect`.

k2/csrc/fsa_algo_test.cc
```diff
@@ -19,10 +19,13 @@
...
+                           {0, 2, 2},
+                           {1, 3, 3},
+                           {1, 6, 6},
+                           {2, 4, 2},
+                           {2, 6, 3},
+                           {5, 0, 1}, };
```
danpovey, line 28: Guys.... the `arc_indexes` arg of the constructor of Fsa is not needed because the constructor can work out what value the arc_indexes array should have.   There can be a version of the constructor where arc_indexes is not provided. This will save typing in tests.  I have said this several times before; if you don't understand what I mean, or you disagree, please say so!

qindazhu: Will add such constructor with another PR.

qindazhu:
If there no other issues, please help to merge.

k2/csrc/fsa_algo.cc
```diff
@@ -150,6 +166,135 @@ void Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+
+    auto curr_state_pair = qstates.front();
+    qstates.pop();
+    // as we have inserted `curr_state_pair` before.
+    int32_t curr_state_index = state_pair_map[curr_state_pair];
+
+    auto state_a = curr_state_pair.first;
+    ArcIterator a_arc_iter_begin = arc_a_begin + a.arc_indexes[state_a];
+    ArcIterator a_arc_iter_end =
+        (state_a != final_state_a) ? (arc_a_begin + a.arc_indexes[state_a + 1])
```
danpovey, line 215: I think we should make the arc_indexes array one larger, with a duplicate of the final number, to avoid these checks.  
Again, it won't require much work as long as you modify the constructor to not take the arc_indexes arg, and as long as you always call NumStates() to get the number of states, rather than using arc_indexes.size() (it will have to subtract one).

danpovey:
Regarding self-loop epsilon transitions: they may seem pointless, but in
the transducer case they can do something nontrivial so they should be
kept.  (This code can be used for transducers, you can use those lists of
source arcs to do that).

k2/csrc/fsa_algo.cc
qindazhu: OK, will add that state and check all of implemented algortithms.

PR #25, opened 2020-05-03
fix ConnectCore.

csukuangfj:
Return a topologically sorted fsa if the input fsa is acyclic.

The state map is not necessarily an identity map if the input fsa
is topo sorted, i.e., states may be renumbered.

----

Example:

Input fsa:
example-1-input

Output fsa:
example-1-output

### DIFF
--- a/k2/csrc/fsa_algo.cc
+++ b/k2/csrc/fsa_algo.cc
@@ -8,6 +8,7 @@
 #include "k2/csrc/fsa_algo.h"
 
 #include <algorithm>
+#include <limits>
 #include <numeric>
 #include <queue>
 #include <stack>
--- a/k2/csrc/fsa_algo.cc
+++ b/k2/csrc/fsa_algo.cc
@@ -9,8 +9,8 @@
 
 #include <algorithm>
 #include <numeric>
-#include <stack>
 #include <queue>
+#include <stack>
 #include <unordered_map>
 #include <utility>
 
@@ -20,6 +20,9 @@
 
 namespace {
 
+static constexpr int8_t kNotVisited = 0;  // a node that has not been visited
+static constexpr int8_t kVisiting = 1;    // a node that is under visiting
+static constexpr int8_t kVisited = 2;     // a node that has been visited
 // depth first search state
 struct DfsState {
   int32_t state;      // state number of the visiting node
@@ -57,14 +60,21 @@ void ConnectCore(const Fsa &fsa, std::vector<int32_t> *state_map) {
 
   std::vector<bool> accessible(num_states, false);
   std::vector<bool> coaccessible(num_states, false);
-  std::vector<bool> visited(num_states, false);
+  std::vector<int8_t> state_status(num_states, kNotVisited);
 
   accessible.front() = true;
   coaccessible.back() = true;
 
   std::stack<DfsState> stack;
   stack.push({0, fsa.arc_indexes[0], fsa.arc_indexes[1]});
-  visited[0] = true;
+  state_status[0] = kVisiting;
+
+  // map order to state.
+  // state 0 has the largest order, i.e., num_states - 1
+  // final_state has the least order, i.e., 0
+  std::vector<int32_t> order;
+  order.reserve(num_states);
+  bool is_acyclic = true;  // order and is_acyclic are for topological sort
 
   while (!stack.empty()) {
     auto &current_state = stack.top();
@@ -73,6 +83,11 @@ void ConnectCore(const Fsa &fsa, std::vector<int32_t> *state_map) {
       // we have finished visiting this state
       auto state = current_state.state;  // get a copy since we will destroy it
       stack.pop();
+      state_status[state] = kVisited;
+
+      if (is_acyclic)
+        order.push_back(state);  // save the state on if it is acyclic
+
       if (!stack.empty()) {
         // if it has a parent, set the parent's co-accessible flag
         if (coaccessible[state]) {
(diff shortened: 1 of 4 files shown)


csukuangfj:
Note that for the following fsa:
special-case

The input fsa is **not acyclic**. Although the cycle does not exist in the output fsa, the output
fsa is not topsorted. The user has to invoke `TopSort` to sort the output fsa.

The output of the above input fsa is given below
special-out

csukuangfj:
ready to merge.

@danpovey @qindazhu 
please have a review

k2/csrc/fsa_algo.cc
```diff
@@ -86,30 +101,49 @@ void ConnectCore(const Fsa &fsa, std::vector<int32_t> *state_map) {
...
+        break;
     }
   }
 
   state_map->reserve(num_states);
+  if (!is_acyclic) {
+    for (auto i = 0; i != num_states; ++i) {
+      if (accessible[i] && coaccessible[i]) state_map->push_back(i);
+    }
+    return;
```
danpovey, line 138: I think there should be a return code of some kind here.  E.g. return false if there were cycles found, maybe?

danpovey:
I don't undertsand your comment starting  "Note that for the following fsa..".

Those two FSAs in the picture are not equivalent.  I'm not sure what you are speaking of implementing here (TopSort or Connect()) but it's not right in either case.

csukuangfj:
> Note that for the following fsa:

What I want to explain is that if the input fsa contains a cycle that is removed in
the output fsa, the output fsa is no going to be topsorted.

The output fsa is not topsorted if the input fsa contains a cycle, no matter where the cycle is,
trimmed or not.

csukuangfj:
> Those two FSAs in the picture are not equivalent.

Sorry, I made a mistake.

danpovey:
OK, well make sure the return status indicates if there was a problem.

In general we can require that the input will be acyclic.  How it behaves if there are cycles doesn't matter much, it can even die with an assertion- the important thing is to inform the user of the error.

k2/csrc/fsa_algo.cc
csukuangfj: It returns true when
- the input fsa is empty
- the output fsa is acyclic
- the outupt is cyclic but contains only self-loops

The output fsa is topsorted for the above three cases.

It returns false otherwise.

csukuangfj:
The original implementation has a bug and cannot handle a graph containing strongly connected components. It has been fixed in the latest commit.

The following are some example input/output:

## Example 1
the input contains a cycle but the cycle is removed, so the output fsa is topsorted.

ex1-input

output
ex1-output

## Example 2
The input contains a cycle and the cycle remains in the output fsa, so the output fsa is NOT topsorted.
ex2-input

output

ex2-output

## Example 3
The input contains cycles but all the cycles are self-loops, so the output fsa is topsorted.

ex3-input

output

ex3-output

csukuangfj:
@danpovey 
ready to merge. Please review.

k2/csrc/fsa_algo.cc
```diff
@@ -44,27 +48,62 @@ inline int32_t InsertIntersectionState(
...
+// This function uses "Tarjan's strongly connected components algorithm"
+// (see
+// https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm)
+// to find co-accessible states in a single pass of the graph.
+//
+// The notations "lowlink", "dfnumber", and "onstack" are from the book
+// "The design and analysis of computer algorithms", on page 192,
+// Fig. 5.15 "Procedure to compute LOWLINK", written by John E. Hopcroft.
+//
 // http://www.openfst.org/doxygen/fst/html/connect_8h_source.html
```
danpovey, line 60: I am a little concerned by the complexity of this.  If it were me I would have implemented it in about 10 lines of code; that would be less bug-prone and easier to transfer it to GPU.  I'll merge, but I think it's something we should revisit later.

k2/csrc/fsa_algo.h
```diff
@@ -19,39 +19,52 @@ namespace k2 {
...
+                      two terminologies).
 
      @param [out] arc_map   If non-NULL, this function will
-                     output a map from the arc-index in `b` to
-                     the corresponding arc-index in `a`.
+                            output a map from the arc-index in `b` to
+                            the corresponding arc-index in `a`.
+
+     @return true if the output fsa is topsorted, or is empty.
+             false otherwise.
```
danpovey, line 54: I'd say: 

   Returns true on success (i.e. the output is topsorted).  The only failure condition is when the input had cycles that were not epsilon loops.  Caution: true return status does not imply that the returned FSA is nonempty.

k2/csrc/fsa_algo.cc
csukuangfj: The current approach is difficult for parallelization.

I can change the implementation if your pseudo code is available.

k2/csrc/fsa_algo.h
csukuangfj: Will change it.

danpovey:
The implementation would just be:

  std::vector<char> accessible(num_states, 0);  // really bool but
vector<bool> is slow.
  std::vector<int> queue_forward;
  queue_forward.push_back(0);
  while (!queue_forward.empty()) {
      pop element
      for each successor:
          if (!accessible[successor]) {
            accessible[successor] = true;
            queue_forward.push_back(successor);
          }
 }
 // and then the same backward.  And keep each state if it's both
accessible and coaccessible.
 // OK, it's much more than 10 lines, but still a lot simpler than your
version.

csukuangfj:
I agree that your implemenation is
- much simpler
- less error-prone
- easy to debug

**But**
- it requires two passes of the graph: first forward pass for `accessible`, second backward pass for `co-accessible`
- The backward pass is not so simple

  (1) You have to get the Entering Arcs of each node. If you compute them at once, it needs
lot of memory space to save them; If you compute them on demand, you have to visit the graph multiple times.

   (2) It's not easy to know whether there is a cycle in the graph

- Difficult to get a topsorted output fsa.

csukuangfj:
I can switch to your implementation if  what`Connect()`  does it to connect the graph, and
`topsort` is not required.

k2/csrc/fsa_algo.cc
csukuangfj: The current implementation visits every node and every edge only once, so the
time complexity is O(V + E).

We need several `vectors` to save the node state, so the space complexity is 
`some constant * V`

danpovey:
Yes, but still v. hard to parallelize.  Waiti, I thought of an algo that
can be parallelized... writing it..

k2/csrc/fsa_algo.h
csukuangfj: Fixed.

danpovey:
OK, I'll merge, we don't need to overthink this at this point, we can
change it later.
My algorithm idea turned out not to be workable.

PR #26, opened 2020-05-04
test if fsa is acyclic

mahsa7823:


k2/csrc/properties.cc
```diff
@@ -78,6 +79,41 @@ bool HasSelfLoops(const Fsa &fsa) {
   return false;
 }
 
+bool CheckCycles(StateId s, std::vector<bool> visited, std::vector<bool> back_arc,
```
csukuangfj, line 82: It would be more efficient if you pass  const references as function  arguments.

```cpp
const std::vector<bool>&
```

k2/csrc/properties.cc
```diff
@@ -78,6 +79,41 @@ bool HasSelfLoops(const Fsa &fsa) {
...
+  return false;
+}
+
+// Detect cycles using DFS traversal
+bool IsAcyclic(const Fsa &fsa) {
+  StateId num_states = fsa.NumStates();
+  std::vector<bool> visited(num_states, false);
+  std::vector<bool> back_arc(num_states, false);
+
+  std::unordered_set<StateId> *adj = new std::unordered_set<StateId>[num_states];
```
csukuangfj, line 105: `adj` is not freed and memory is leaked here.

danpovey:
Best solution for that is probably just to declare adj as a non-pointer
variable, i.e.

std::unordered_set<StateId> adj(num_states);

k2/csrc/properties.cc
mahsa7823: done. I passed by reference, and as const when possible.

mahsa7823:
> Best solution for that is probably just to declare adj as a non-pointer variable, i.e. std::unordered_set<StateId> adj(num_states);

Right. I removed the pointer.

k2/csrc/properties.cc
mahsa7823: I defined 'adj' as non-pointer and removed the 'new' operation.

k2/csrc/properties.cc
```diff
@@ -78,6 +79,43 @@ bool HasSelfLoops(const Fsa &fsa) {
...
+  StateId num_states = fsa.NumStates();
+  const std::vector<Arc> arcs = fsa.arcs;
+  std::vector<bool> visited(num_states, false);
+  std::vector<bool> back_arc(num_states, false);
+
+  for (StateId i = 0; i < num_states; i++)
+    if (CheckCycles(i, visited, back_arc, arcs))
+      return true;
+
+    return false;
```
csukuangfj, line 115: Better to add braces `{}` for the `for` loop.

Indentation does not work here.

k2/csrc/properties.cc
mahsa7823: The Google style checker gives me the following error despite I'm using `std::vector<bool> &visited`. Do I have to use pointer (*) instead of reference (&) in the function arguments?

```
Is this a non-const reference? If so, make const or use a pointer: std::vector<bool> &visited  [runtime/references]
```

k2/csrc/properties_test.cc
```diff
@@ -356,5 +357,41 @@ TEST(Properties, IsConnected) {
     EXPECT_TRUE(is_connected);
   }
 }
+TEST(FsaAlgo, IsAcyclic) {
+  // empty fsa is acyclic
+  {
+    Fsa fsa;
+    bool is_acyclic = IsAcyclic(fsa);
+    EXPECT_FALSE(is_acyclic);
```
csukuangfj, line 365: the comment says
```cpp
// empty fsa is acyclic
 ```

but you use `EXPECT_FALSE`. Should it be `EXPECT_TRUE` ?

File: k2/csrc/properties_test.cc [hunk 5]
```diff
@@ -356,5 +357,41 @@ TEST(Properties, IsConnected) {
...
+  }
+
+  // an acyclic fsa example
```
csukuangfj, line 368: It is an acyclic example, then why is it `EXPECT_FALSE` at line 379?

k2/csrc/properties.cc
csukuangfj: Yes. For output arguments, please use `pointer`.

k2/csrc/properties_test.cc
mahsa7823: I addressed this issue in [this commit](https://github.com/danpovey/k2/pull/26/commits/45763678844206c61f81918aab31dce7f46815de).

k2/csrc/properties.cc
mahsa7823: done.

k2/csrc/properties.cc
```diff
@@ -78,6 +79,51 @@ bool HasSelfLoops(const Fsa &fsa) {
...
+  for (StateId i = 0; i < num_states; i++)
+    if (CheckCycles(i, &visited, &back_arc, arcs))
+      return true;
+
+  return false;
+}
+
+bool IsAcyclic(const Fsa &fsa) {
+  if (IsEmpty(fsa)) return true;
+  if (IsCyclic(fsa))
```
qindazhu, line 121: just write `return !IsCyclic()`

k2/csrc/properties.cc
```diff
@@ -78,6 +79,51 @@ bool HasSelfLoops(const Fsa &fsa) {
...
+  }
+  (*back_arc)[s] = false;
+  return false;
+}
+
+// Detect cycles using DFS traversal
+bool IsCyclic(const Fsa &fsa) {
+  if (IsEmpty(fsa)) return false;
+  StateId num_states = fsa.NumStates();
+  const std::vector<Arc> arcs = fsa.arcs;
```
qindazhu, line 108: should be a reference

k2/csrc/properties.cc
```diff
@@ -78,6 +79,51 @@ bool HasSelfLoops(const Fsa &fsa) {
...
+  std::vector<bool> back_arc(num_states, false);
+
+  for (StateId i = 0; i < num_states; i++)
+    if (CheckCycles(i, &visited, &back_arc, arcs))
+      return true;
+
+  return false;
+}
+
+bool IsAcyclic(const Fsa &fsa) {
```
qindazhu, line 119: I suggest that we only implement `IsAcyclic`, then users can call `!IsAcyclic()` to check if it is cyclic.

k2/csrc/properties.h
```diff
@@ -17,6 +17,8 @@ enum Properties {
   kTopSorted,            // topologically sorted
   kTopSortedAndAcyclic,  // topologically sorted and no self-loops (which
                          // implies acyclic)
+  kAcyclic,              // acyclic
+  kCyclic,              // cyclic
```
qindazhu, line 21: Again, suggest we only implement `IsAcyclic`, so delete `kCyclic`

k2/csrc/properties.cc
```diff
@@ -78,6 +79,51 @@ bool HasSelfLoops(const Fsa &fsa) {
   return false;
 }
 
+bool CheckCycles(StateId s, std::vector<bool> *visited,
+                 std::vector<bool> *back_arc, const std::vector<Arc> &arcs) {
+  if ((*visited)[s] == false) {
```
qindazhu, line 84: Seems the implentation  will iterate over the arcs multiple times which is really inefficient. 
BTW, I strongly suggest we should try to avoid recursion

danpovey:
.. yeah, recursion can use up stack space and might cause crash if too deep.

k2/csrc/properties.cc
csukuangfj:
```cpp
const auto& arcs = fsa.arcs;
```

k2/csrc/properties.cc
mahsa7823: A directed graph is acyclic iff it has no strongly connected subgraph with more than one state. Since we already have an implementation of Tarjan's scc algorithm, that should be enough for detecting IsAcyclic.
In fact, it seems that the [ConnectCore function](https://github.com/danpovey/k2/blob/152ced3082914ffd1dfaeba71d2e5e11949678a2/k2/csrc/fsa_algo.cc#L62) does the job of detecting IsAcyclic; it returns true if the fsa is acyclic and false if it is cyclic. Do we still need to implement a separate function IsAcyclic?

danpovey:
If that works already it should be OK.
This will only be activated in debug mode, I think.

k2/csrc/properties_test.cc
```diff
@@ -356,5 +357,41 @@ TEST(Properties, IsConnected) {
...
+    bool is_acyclic = IsAcyclic(fsa);
+    EXPECT_TRUE(is_acyclic);
+  }
+
+  // an acyclic fsa example
+  {
+    std::vector<Arc> arcs = {
+      {0, 1, 2}, {0, 2, 1}, {1, 2, 0}, {1, 3, 5}, {2, 3, 6},
+    };
+    std::vector<int32_t> arc_indexes = {0, 2, 4, 5};
```
qindazhu, line 373: We have add a constructor only with `arcs`

https://github.com/danpovey/k2/blob/master/k2/csrc/fsa.h#L93

so we can write now `Fsa fsa(std::move(arcs), 3)`

qindazhu:
I'm (sort of) OK with calling `ConnectCore` in `IsAcyclic` (note that I strongly suggest we only implement `IsAcyclic` and let users call `!IsAcyclic()` to check if it is cyclic). My concern is that it seems now `ConnectCore` doing so much things (ssc, order, co-accessible,etc.), some of them may be not required by `IsAcyclic`.

BTW, for `IsAcyclic`, if we implement it with `SCC` approach, we can even exit the program earlier when we detect a cycle (in current `ConnectCore`, we need to wait the outer `while` exit to check)

csukuangfj:
> My concern is that it seems now ConnectCore doing so much things

This is because there are **so much** requirements for `ConnectCore`.

You do not need the whole `scc` algorithm to detect a cycle.

The `TopSort` will do
https://github.com/danpovey/k2/blob/master/k2/csrc/fsa_algo.cc#L468

k2/csrc/properties.cc
mahsa7823: will do.

mahsa7823:
> The `TopSort` will do
> https://github.com/danpovey/k2/blob/master/k2/csrc/fsa_algo.cc#L468

I'll use the relevant parts of the TopSort in the IsAcyclic function.

mahsa7823:
In my new commit, I cut parts of TopSort and used for IsAcyclic. Now, TopSort calls IsAcyclic for determining is_acyclic. The same thing can be probably done with ConnectCore (i.e. ConnectCore to call IsAcyclic).

k2/csrc/properties.cc
```diff
@@ -10,12 +11,23 @@
...
 
+static constexpr int8_t kNotVisited = 0;  // a node that has not been visited
+static constexpr int8_t kVisiting = 1;    // a node that is under visiting
+static constexpr int8_t kVisited = 2;     // a node that has been visited
+// depth first search state
+struct DfsState {
+  int32_t state;      // state number of the visiting node
+  int32_t arc_begin;  // arc index of the visiting arc
+  int32_t arc_end;    // end of the arc index of the visiting node
+};
```
qindazhu, line 29: If these struct and constexpr is useful in different files, I'm OK with it in namespace `k2`, but a better place is in a `header` file, such as `utils.h`.

As `fsa_algo.cc` defines same thing, we definitely should avoid such duplicate code.

k2/csrc/properties.cc
qindazhu: function `CheckCycles` is not needed anymore, right? I guess you can remove it.

k2/csrc/properties.cc
```diff
@@ -78,6 +90,78 @@ bool HasSelfLoops(const Fsa &fsa) {
...
+  stack.push({0, fsa.arc_indexes[0], fsa.arc_indexes[1]});
+  state_status[0] = kVisiting;
+  bool is_acyclic = true;
+  while (is_acyclic && !stack.empty()) {
+    auto &current_state = stack.top();
+    if (current_state.arc_begin == current_state.arc_end) {
+      // we have finished visiting this state
+      state_status[current_state.state] = kVisited;
+      if (order != nullptr)
+        (*order).push_back(current_state.state);
```
qindazhu, line 132: just `order->push_back`

k2/csrc/properties.cc
```diff
@@ -78,6 +90,78 @@ bool HasSelfLoops(const Fsa &fsa) {
...
+    }
+    const auto &arc = fsa.arcs[current_state.arc_begin];
+    auto next_state = arc.dest_state;
+    auto status = state_status[next_state];
+    switch (status) {
+      case kNotVisited: {
+        // a new discovered node
+        state_status[next_state] = kVisiting;
+        auto arc_begin = fsa.arc_indexes[next_state];
+        stack.push({next_state, arc_begin, fsa.arc_indexes[next_state + 1]});
```
qindazhu, line 144: change `arc_begin` here to `fsa.arc_indexes[next_state]`, it's confusing with below line `++current_state.arc_begin`

k2/csrc/properties.h
```diff
@@ -51,6 +52,11 @@ bool IsArcSorted(const Fsa &fsa);
  */
 bool HasSelfLoops(const Fsa &fsa);
 
+/*
+  Returns true if `fsa` is acyclic
+ */
```
qindazhu, line 57: it's better if can comment here to say what the `order` is

k2/csrc/properties_test.cc
```diff
@@ -216,5 +217,35 @@ TEST(Properties, IsConnected) {
...
```
qindazhu, line 233: `arc_indexes` is not needed anymore

k2/csrc/properties_test.cc
```diff
@@ -216,5 +217,35 @@ TEST(Properties, IsConnected) {
...
+    bool is_acyclic = IsAcyclic(fsa);
+    EXPECT_TRUE(is_acyclic);
+  }
+
+  // a cyclic fsa example
+  {
+    std::vector<Arc> arcs = {
+      {0, 1, 2}, {0, 4, 0}, {0, 2, 0}, {1, 2, 1}, {1, 3, 0}, {2, 1, 0},
+    };
+    std::vector<int32_t> arc_indexes = {0, 3, 5, 6, 6};
```
qindazhu, line 244: remove `arc_indexes`

k2/csrc/properties.cc
```diff
@@ -78,6 +90,78 @@ bool HasSelfLoops(const Fsa &fsa) {
...
+      else if ((*back_arc)[*i])
+        return true;
+    }
+  }
+  (*back_arc)[s] = false;
+  return false;
+}
+
+// Detect cycles using DFS traversal
+bool IsAcyclic(const Fsa &fsa, std::vector<int32_t> *order /*= nullptr*/) {
```
qindazhu, line 116: BTW, I just found any approach of `DFS` starting with the start state holds an assumption that the fsa is connected. It's OK in `ConnectCore` as we would only keep connected states in `ConnectCore`. However, this is not true in other situations such as `IsAcyclic`. Maybe implementing this function with `Topological Sorting`(a queue to maintain states with `out-degree == 0`) is straightforward and simpler than `scc`.

Guys, what do you think? I was thinking we may need to avoid to write so much `DFS` or `SCC` code (as they are nearly same structure), it will hard to maintain such code.  Maybe we should establish `openfst`-like structure for such structure(and only for such structrue), i.e. define a DFS structrue and provide different `next/done/visiting/visited` function for different operations.

And, seems sometime it will be very simple if we could maintain a `GetEnterArcs`.

danpovey: I don't think we care about cycles in parts of the FSA that are not connected.  We could just document that they won't be found.Also an "IsAcyclicAndConnected()" function might be useful because in practice people will want both of those conditions at the same time.

danpovey: Regarding a special structure for DFS: I doubt it will be needed except in these two specific instances.  I'm fine to leave that for later or never; I don't think it's a big priority.

mahsa7823:
Other than that the style check is cancelled on PR, this PR should be ready to merge.

qindazhu:
> I don't think we care about cycles in parts of the FSA that are not accessible (i.e. from the start state).  We could just document that they won't be found.Also an "IsAcyclicAndConnected()" function might be useful because in practice people will want both of those conditions at the same time.

As Dan said here, it's better if could add such comments in `IsAcyclic` and `IsTopSorted`. Maybe add a function `IsTopSortedAndConnected`?

mahsa7823:
> > I don't think we care about cycles in parts of the FSA that are not accessible (i.e. from the start state).  We could just document that they won't be found.Also an "IsAcyclicAndConnected()" function might be useful because in practice people will want both of those conditions at the same time.
> 
> As Dan said here, it's better if could add such comments in `IsAcyclic` and `IsTopSorted`. Maybe add a function `IsTopSortedAndConnected`?

Ok, I'll add that function.

k2/csrc/properties.cc
```diff
@@ -10,12 +11,23 @@
 #include <algorithm>
 #include <unordered_set>
 #include <vector>
+#include <stack>
```
csukuangfj, line 14: head files should be sorted in aphabetic order.

csukuangfj:
+1

qindazhu:
Merging, thanks!

PR #33, opened 2020-05-08
Unpruned declarations of RmEpsilon etc.

danpovey:


danpovey:
@qindazhu  please check this and merge if possible...
Esp. w.r.t. declaration of RmEpsilons()

qindazhu:
Sure, wait a minute

k2/csrc/fsa_algo.h
```diff
@@ -144,9 +146,50 @@ bool Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map = nullptr);
...
  */
-void RmEpsilonsPruned(const WfsaWithFbWeights &a, float beam, Fsa *b,
+void RmEpsilonsPrunedMax(const WfsaWithFbWeights &a, float beam, Fsa *b,
+                         std::vector<std::vector<int32_t>> *arc_map);
+
+/*
+  Version of RmEpsilonsPrunedMax that doesn't support pruning; see its
+  documentation.
+ */
+void RmEpsilonsMax(const Fsa &a, float *a_weights, Fsa *b,
```
qindazhu, line 156: The interfaces if OK for now. I guess we may need to define `semiring operation` (max/log-sum) to avoid duplicate code when we implement these alogrithms, what do you think?

danpovey: RE semirings: it's not quite as simple as that, as the algorithms needed to compute the derivative info can't, I think, be super easily templated.  The structures they require are a bit different.  As I continue implementing determinization I'll see if there are commonalities though.

qindazhu: Got it, thanks!

k2/csrc/determinize.cc
```diff
@@ -7,44 +7,136 @@
...
+                      // that owns this DetStateElement.
+
+  std::shared_ptr<PathElement> path;
+                      // The path from the start state to here (actually we will
+                      // only follow back `seq_len` links.  Will be nullptr if
+                      // seq_len == 0 and this belongs to the initial determinized
+                      // state.
+
+  DetStateElement &&Advance(float arc_weight, int32_t arc_index, int32_t arc_symbol) {
+    return DetStateElement(weight + arc_weight,
```
qindazhu, line 107: just a question, we always do `+` here, right? Suppose it it log-sum?

k2/csrc/determinize.cc
```diff
@@ -7,44 +7,136 @@
...
 };
 
+
+
+class Determinizer {
+ public:
+ private:
+
+  using DetStatePriorityQueue = priority_queue<shared_ptr<DetState>,
+                                               vector<shared_ptr<DetState> >,
```
qindazhu, line 133: Aha, I noticed that you are used to add a blank between `> >` when using STL, actually this is not required anymore in modern C++ compiler (it has been a bug).

k2/csrc/determinize.cc
danpovey: It's always + because this is along a path.  It's when we merge paths that it's max vs. log-sum.

k2/csrc/determinize.cc
danpovey: OK

k2/csrc/determinize.cc
danpovey: you can fix this later, BTW, let's not block merge on this.

k2/csrc/determinize.cc
```diff
@@ -125,18 +258,169 @@ struct DetState {
...
+  // lists of states in their paths.  This will be true becase of how we
+  // construct them (it requires on the IsArcSorted() property, whereby arcs
+  // leaving each state in the FSA are sorted first on label and then on
+  // dest_state.
+  if (seq_len == 0) {
+    CHECK(elements.size() == 1);
+    CHECK(elements.front().weight == 0.0);
+  }
+
+  std::vector<int32> prev_seq;
```
qindazhu, line 400: `int32_t`?

k2/csrc/determinize.cc
```diff
@@ -125,18 +258,169 @@ struct DetState {
...
+  for (auto iter = elements.begin(); iter != elements.end(); ++iter) {
+    auto path = iter->path;
+    std::vector<int32> cur_seq;
+    for (int32_t i = 0; i < seq_len; i++) {
+      cur_seq.push_back(input_fsa.arcs[path->arc_index].prev_state);
```
qindazhu, line 405: I got stuck. why `Fsa.arcs` has a property of `preve_state` as it's not `LogSumTracebackLink`

danpovey: should be src_state.   This is all an early draft.

qindazhu:
Will merge as we will refine them later.

PR #34, opened 2020-05-09
implement ComputeForward/BackwardMaxWeights

qindazhu:
- Implement `ComputeForward(Backward)MaxWeights`
- Add weights to `FsaRender`
- Changes in `determinize.cc` are just fixing some style issues.

### DIFF
--- a/k2/csrc/fsa_renderer.cc
+++ b/k2/csrc/fsa_renderer.cc
@@ -6,6 +6,7 @@
 
 #include "k2/csrc/fsa_renderer.h"
 
+#include <iomanip>
 #include <sstream>
 #include <string>
 
@@ -31,8 +32,8 @@ std::string GenerateEpilogue() { return "}"; }
 using k2::Arc;
 using k2::Fsa;
 
-
-std::string ProcessState(const Fsa &fsa, int32_t state) {
+std::string ProcessState(const Fsa &fsa, int32_t state,
+                         const float *arc_weights = nullptr) {
   std::ostringstream os;
   os << "  " << state << " [label = \"" << state
      << "\", shape = circle, style = bold, fontsize = 14]"
@@ -46,8 +47,10 @@ std::string ProcessState(const Fsa &fsa, int32_t state) {
     int32_t src = arc.src_state;
     int32_t dest = arc.dest_state;
     int32_t label = arc.label;
-    os << "          " << src << " -> " << dest << " [label = \"" << label
-       << "\", fontsize = 14];"
+    os << "          " << src << " -> " << dest << " [label = \"" << label;
+    if (arc_weights != nullptr)
+      os << "/" << std::fixed << std::setprecision(1) << arc_weights[begin];
+    os << "\", fontsize = 14];"
        << "\n";
   }
 
@@ -58,7 +61,8 @@ std::string ProcessState(const Fsa &fsa, int32_t state) {
 
 namespace k2 {
 
-FsaRenderer::FsaRenderer(const Fsa &fsa) : fsa_(fsa) {}
+FsaRenderer::FsaRenderer(const Fsa &fsa, const float *arc_weights /*=nullptr*/)
+    : fsa_(fsa), arc_weights_(arc_weights) {}
 
 std::string FsaRenderer::Render() const {
   int32_t num_states = fsa_.NumStates();
@@ -68,7 +72,7 @@ std::string FsaRenderer::Render() const {
   os << GeneratePrologue();
 
   for (int32_t i = 0; i != num_states - 1; ++i) {
-    os << ProcessState(fsa_, i);
+    os << ProcessState(fsa_, i, arc_weights_);
   }
 
   // now for the final state
--- a/k2/csrc/fsa_renderer.h
+++ b/k2/csrc/fsa_renderer.h
@@ -16,13 +16,14 @@ namespace k2 {
 // Get a GraphViz representation of an fsa.
 class FsaRenderer {
  public:
-  explicit FsaRenderer(const Fsa &fsa);
+  explicit FsaRenderer(const Fsa &fsa, const float *arc_weights = nullptr);
 
   // Return a GraphViz representation of the fsa
   std::string Render() const;
 
  private:
   const Fsa &fsa_;
+  const float *arc_weights_;
 };
 
 }  // namespace k2
--- a/k2/csrc/util.h
+++ b/k2/csrc/util.h
@@ -7,6 +7,8 @@
 #ifndef K2_CSRC_UTIL_H_
 #define K2_CSRC_UTIL_H_
 
+#include <cfloat>
+#include <cmath>
 #include <functional>
 #include <utility>
 
@@ -31,8 +33,8 @@ struct PairHash {
   }
 };
 
-static const double kMinLogDiffDouble = Log(DBL_EPSILON);  // negative!
-static const float kMinLogDiffFloat = Log(FLT_EPSILON);  // negative!
+static const double kMinLogDiffDouble = log(DBL_EPSILON);  // negative!
+static const float kMinLogDiffFloat = log(FLT_EPSILON);    // negative!
 
 // returns log(exp(x) + exp(y)).
 inline double LogAdd(double x, double y) {
@@ -48,13 +50,12 @@ inline double LogAdd(double x, double y) {
 
   if (diff >= kMinLogDiffDouble) {
     double res;
-    res = x + Log1p(Exp(diff));
+    res = x + log1p(exp(diff));
     return res;
   } else {
     return x;  // return the larger one.
   }
 }
-
 
 // returns log(exp(x) + exp(y)).
 inline float LogAdd(float x, float y) {
@@ -70,14 +71,12 @@ inline float LogAdd(float x, float y) {
 
   if (diff >= kMinLogDiffFloat) {
     float res;
-    res = x + Log1p(Exp(diff));
+    res = x + log1p(exp(diff));
     return res;
   } else {
     return x;  // return the larger one.
   }
 }
-
-
 
 }  // namespace k2
 #endif  // K2_CSRC_UTIL_H_
--- a/k2/csrc/util.h
+++ b/k2/csrc/util.h
@@ -34,7 +34,7 @@ struct PairHash {
 };
 
 static const double kMinLogDiffDouble = log(DBL_EPSILON);  // negative!
-static const float kMinLogDiffFloat = log(FLT_EPSILON);    // negative!
+static const float kMinLogDiffFloat = logf(FLT_EPSILON);   // negative!
 
 // returns log(exp(x) + exp(y)).
 inline double LogAdd(double x, double y) {
@@ -71,7 +71,7 @@ inline float LogAdd(float x, float y) {
 
   if (diff >= kMinLogDiffFloat) {
     float res;
-    res = x + log1p(exp(diff));
+    res = x + log1pf(expf(diff));
     return res;
   } else {
     return x;  // return the larger one.
--- a/k2/csrc/weights.cc
+++ b/k2/csrc/weights.cc
@@ -1,0 +1,78 @@
+// k2/csrc/weights.cc
+
+// Copyright (c)  2020  Xiaomi Corporation (author: Haowen Qiu)
+
+// See ../../LICENSE for clarification regarding multiple authors
+
+#include "k2/csrc/weights.h"
+
+#include <algorithm>
+#include <queue>
+#include <vector>
+
+#include "glog/logging.h"
+#include "k2/csrc/fsa.h"
+#include "k2/csrc/properties.h"
+
+namespace {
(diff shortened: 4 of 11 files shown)


qindazhu:
`ComputeForwardMaxWeights`:
https://github.com/danpovey/k2/blob/dbba25b59b1da8e8a671507ea737fbb98df9cbf3/k2/csrc/weights_test.cc#L53-L54

`ComputeBackwardMaxWeights`:
https://github.com/danpovey/k2/blob/dbba25b59b1da8e8a671507ea737fbb98df9cbf3/k2/csrc/weights_test.cc#L65-L66

qindazhu:
Cancelled style check as it took so long time to install `cppcheck`..

k2/csrc/weights.cc
```diff
@@ -0,0 +1,79 @@
...
+  int32_t num_states = fsa.NumStates();
+  std::fill_n(state_weights, num_states, kNegativeInfinity);
+
+  const auto &arcs = fsa.arcs;
+  state_weights[0] = 0;
+  for (std::size_t i = 0; i != arcs.size(); ++i) {
+    const auto &arc = arcs[i];
+    auto src_weight = state_weights[arc.src_state];
+    auto &dest_weight = state_weights[arc.dest_state];
+    if (src_weight > kNegativeInfinity)
```
danpovey, line 43: I don't think this if stateent is needed.

danpovey: .. also this code seems to require topologically sorted input.  Double check that it's documented; and I think we should have a check that only runs in debug mode, that it's top sorted and acyclic.  E.g. within the loop, `CHECK(arc.dest_state > arc.src_state);`

qindazhu: > I don't think this if stateent is needed.

It is needed here as we may update `dest_weight` to a value bigger than `kNegativeInfinity` (considering there's no standard `-inifity` in C++, I define it with `std::num_limits<float>::lowest`).

qindazhu: > .. also this code seems to require topologically sorted input. Double check that it's documented; and I think we should have a check that only runs in debug mode, that it's top sorted and acyclic. E.g. within the loop, `CHECK(arc.dest_state > arc.src_state);`

OK. Will Update. Should I remove the check of `IsValid` as well?

danpovey: No don't remove the check of IsValid.
Regarding infinity, just use -std::numeric_limits<float>::infinity() wherever we need that.  It works fine, it's been used in Kaldi a lot.  And no need for the if.

qindazhu: ok, seems compilers support `-std::numeric_limits::infinity()` well.

qindazhu:
Updated, Ready to merge.

k2/csrc/fsa_renderer.h
```diff
@@ -16,13 +16,14 @@ namespace k2 {
 // Get a GraphViz representation of an fsa.
 class FsaRenderer {
  public:
-  explicit FsaRenderer(const Fsa &fsa);
+  explicit FsaRenderer(const Fsa &fsa, float *arc_weights = nullptr);
```
csukuangfj, line 19: `const float *`

`FsaRenderer` is not supposed to change `arc_wegiths`, so it's better to add `const`.

k2/csrc/fsa_renderer.cc
```diff
@@ -68,7 +72,7 @@ std::string FsaRenderer::Render() const {
   os << GeneratePrologue();
 
   for (int32_t i = 0; i != num_states - 1; ++i) {
-    os << ProcessState(fsa_, i);
+    os << ProcessState(fsa_, i, arc_weights_);
   }
 
   // now for the final state
```
csukuangfj, line 78: sometimes the weight for the final state is not zero, so it should also handle this case.

k2/csrc/weights.cc
```diff
@@ -0,0 +1,81 @@
...
+  CHECK_NOTNULL(state_weights);
+
+  int32_t num_states = fsa.NumStates();
+  std::fill_n(state_weights, num_states, kNegativeInfinity);
+
+  const auto &arcs = fsa.arcs;
+  state_weights[0] = 0;
+  for (std::size_t i = 0; i != arcs.size(); ++i) {
+    const auto &arc = arcs[i];
+#ifndef NDEBUG
```
csukuangfj, line 39: please use `DCHECK_GE`, no need to guard it with `NDEBUG`.

Refer to https://github.com/google/glog/blob/3ba8976592274bc1f907c402ce22558011d6fc5e/src/glog/logging.h.in#L438

```cpp
#if defined(NDEBUG) && !defined(DCHECK_ALWAYS_ON)
#define DCHECK_IS_ON() 0
#else
#define DCHECK_IS_ON() 1
#endif
```

and

https://github.com/google/glog/blob/master/src/glog/logging.h.in#L1004

k2/csrc/weights.cc
```diff
@@ -0,0 +1,81 @@
...
+  if (IsEmpty(fsa)) return;
+  CheckInput(fsa, arc_weights);
+  CHECK_NOTNULL(state_weights);
+
+  int32_t num_states = fsa.NumStates();
+  std::fill_n(state_weights, num_states, kNegativeInfinity);
+
+  const auto &arcs = fsa.arcs;
+  state_weights[0] = 0;
+  for (std::size_t i = 0; i != arcs.size(); ++i) {
```
csukuangfj, line 37: Does the input fsa need to be TopSorted?

k2/csrc/weights.h
```diff
@@ -29,53 +34,52 @@ namespace k2 {
...
                 They will be 0 for the start-state (if fsa is
                 nonempty), and for later states will be the
                 largest (most positive) weight from the start-state
-                to that state along any path, or +infinity if no such
+                to that state along any path, or `kNegativeInfinity` if no such
                 path exists.
  */
-void ComputeForwardMaxWeights(const Fsa &fsa, float *state_weights);
+void ComputeForwardMaxWeights(const Fsa &fsa, const float *arc_weights,
+                              float *state_weights);
```
csukuangfj, line 49: Why is the argument `state_weights` of type `float`, but the members are of type `double`?

k2/csrc/weights.h
```diff
@@ -29,53 +34,52 @@ namespace k2 {
...
-  const double *backward_state_weights;
 
   // Records whether we use max or log-sum.
   FbWeightType weight_type;
 
   /*
     Constructor.
-       @param [in] fsa  Pointer to an FSA; must satisfy
-            IsValid(*fsa) and IsTopSorted(*fsa).
+       @param [in] fsa  Reference to an FSA; must satisfy
```
csukuangfj, line 81: > must satisfy IsValid(fsa) and IsTopSorted(fsa).

But `IsTopSorted(fsa)` is not enforced in the constructor.

k2/csrc/weights_test.cc
```diff
@@ -0,0 +1,70 @@
...
+    std::vector<float> weights = {1, 1, 2, 3, 4, 5, 2, 3, 4, 3, 5, 6};
+    std::copy_n(weights.begin(), num_arcs, arc_weights);
+  }
+
+  ~WeightsTest() {
+    delete fsa;
+    delete[] arc_weights;
+  }
+
+  Fsa *fsa;
```
csukuangfj, line 41: These member variables do not conform to the naming style.

They should be ended with `_`.

```cpp
Fsa *fsa_;
int32_t num_states_;
float *arc_weights_;
```

k2/csrc/weights_test.cc
```diff
@@ -0,0 +1,70 @@
...
+  int32_t num_states;
+  float *arc_weights;
+};
+
+TEST_F(WeightsTest, ComputeForwardMaxWeights) {
+  {
+    float *state_weights = new float[num_states];
```
csukuangfj, line 48:
```cpp
std::vector<float> state_weights(num_states);
```
avoids explicit memory allocation and free.

csukuangfj:
`cppcheck` can be disabled by commenting out the following code

https://github.com/danpovey/k2/blob/9bb1b26e9b3630a62f45f245ae96918779511f68/.github/workflows/style_check.yml#L47-L56

and 
https://github.com/danpovey/k2/blob/9bb1b26e9b3630a62f45f245ae96918779511f68/.github/workflows/style_check.yml#L71-L74

k2/csrc/fsa_renderer.cc
danpovey: @csukuangfj  can you be more explicit about what you mean here?  Or perhaps you misunderstand..
The final state doesn't have a weight, there is no way to express it.  Arcs to the final state have weights on them, which can be interpreted as final weights.

danpovey: .. in fact, if we were to convert to OpenFST models we would change the format; we'd remove the -1 labels and instead treat those arcs as final-probs; and we would not have the final-state.

k2/csrc/weights.cc
danpovey: The top-sorting is checked by: `CHECK_GE(arc.dest_state, arc.src_state)`

k2/csrc/fsa_renderer.cc
csukuangfj: you are right.

k2/csrc/fsa_renderer.h
qindazhu: Sure

k2/csrc/weights.cc
qindazhu: Thanks

k2/csrc/weights.h
qindazhu: `ComputeForwardMaxWeights` is not a member of `WfsaWithFbWeights`.

k2/csrc/weights.h
qindazhu: We will check this in debug mode and in the arc-loops as in `ComputeForwardMaxWeights`.

Added a TODO for now.

k2/csrc/weights_test.cc
qindazhu: Sure

k2/csrc/weights_test.cc
qindazhu: Sure

qindazhu:
> cppcheck can be disabled by commenting out the following code
mm..OK. BTW, I'm fine with removing `cppcheck` if you think `clang-tidy` is good enough.

qindazhu:
Ready to merge. Please ignore issues from `determinize.cc` for now

danpovey:
Cool, I'll wait for @csukuangfj  to merge.

danpovey:
BTW, I am considering whether at some point it might make sense to relax the constraint that final_state be numbered last (except for top-sorted FSAs).  It might turn out that removing that constraint will simplify code.  Just something to keep in mind as you code stuff.   We could put this into class Fsa for now, maybe, to make any future change less disruptive.
```
  int32_t FinalState() const {
    // It's not valid to call this if the FSA is empty.                                                                                                                            
    CHECK(!arc_indexes.empty());
    return arc_indexes.size() - 2;
  }
```

qindazhu:
Added! As wee have gotten final_state with `int32_t final_state = fsa.NumState() - 1` before, we may need to update those code one by one. ( but for now I believe numbering final state as the last one if convenient)

k2/csrc/util.h
```diff
@@ -31,8 +33,8 @@ struct PairHash {
   }
 };
 
-static const double kMinLogDiffDouble = Log(DBL_EPSILON);  // negative!
-static const float kMinLogDiffFloat = Log(FLT_EPSILON);  // negative!
+static const double kMinLogDiffDouble = log(DBL_EPSILON);  // negative!
+static const float kMinLogDiffFloat = log(FLT_EPSILON);    // negative!
```
csukuangfj, line 37: There is a function that is specific to `float`

```cpp
float logf(float x);
```

k2/csrc/util.h
```diff
@@ -70,14 +71,12 @@ inline float LogAdd(float x, float y) {
 
   if (diff >= kMinLogDiffFloat) {
     float res;
-    res = x + Log1p(Exp(diff));
+    res = x + log1p(exp(diff));
```
csukuangfj, line 74: `log1pf` and `expf` ?

k2/csrc/weights.h
csukuangfj: What I wanted to express is that both the arguments of this function and the data members of  `WfsaWithFbWeights` represent **state** weight, but the function arguments are `float`
but the data members are **double**.

Why not to change the arguments to type `double` ?

k2/csrc/util.h
qindazhu: Sure

k2/csrc/util.h
qindazhu: Done

csukuangfj:
+2 merging.

PR #36, opened 2020-05-11
implement Wfsa

qindazhu:


k2/csrc/weights.cc
```diff
@@ -68,11 +69,83 @@ WfsaWithFbWeights::WfsaWithFbWeights(const Fsa &fsa, const float *arc_weights,
...
+    DCHECK_GE(arc.dest_state, arc.src_state);
+    auto src_weight = forward_state_weights[arc.src_state];
+    auto &dest_weight = forward_state_weights[arc.dest_state];
+
+    // will be `kNegativeInfinity` for non-accessible state
+    double r = rd[arc.src_state] + arc_weights[i];
+    // will be `dest_weight` for non-accessible state
+    double d = AddWeights(dest_weight, r);
+    if (!ApproxEqual(dest_weight, d, kShortestDelta)) {
+      dest_weight = d;
```
danpovey, line 99: ? I don't see an iteration here.
For now we only need this for the acyclic, top-sorted case; there is no need for the delta stuff in that case.

k2/csrc/weights.cc
```diff
@@ -68,11 +69,83 @@ WfsaWithFbWeights::WfsaWithFbWeights(const Fsa &fsa, const float *arc_weights,
...
+      src_weight = d;
+      // do not overwrite original `rd` value of `dest_state`
+      // as we may use it in the next iteration
+      if (arc.src_state != arc.dest_state)
+        rd[arc.src_state] = AddWeights(rd[arc.src_state], r);
+    }
+  }
+}
+
+inline double WfsaWithFbWeights::AddWeights(double weight1,
```
danpovey, line 136: I don't like the way this is repeatedly called.  I'd rather have the switch or if statement be outside the loop.
The loop can be substantially simplified, anyway; all that delta stuff is not needed.

k2/csrc/weights.cc
qindazhu: Oh, ok, will delete

k2/csrc/weights.cc
qindazhu: This is exactly a question that I wanted to ask: one way do do this is make `Wfsa` class be a templetate class on `kMax/LogSum-Weight`, is that OK? (Another way is make a virtual function, but I think that approach holds overhead of virtual call)

danpovey:
No, this code is very simple so doesn't need template stuff.  Just use an
if statement, outside the loop.

k2/csrc/util.h
```diff
@@ -7,15 +7,25 @@
...
+#include <limits>
 #include <utility>
 
 #include "k2/csrc/fsa.h"
 
 namespace k2 {
 
+#define EXPECT_DOUBLE_ARRAY_APPROX_EQ(expected, actual, abs_error)             \
+  EXPECT_EQ(expected.size(), actual.size()) << "Different Array Size.";        \
+  for (std::size_t i = 0; i < std::min(expected.size(), actual.size()); ++i) { \
```
csukuangfj, line 23: Why is there an `std::min`?
`EXPECT_EQ` has already been used to check that they have the same size.

BTW: `EXPECT_EQ` is from `gtest`, so this may not be the best place to place
the macro `EXPECT_DOUBLE_ARRAY_APPROX_EQ`.

k2/csrc/util.h
```diff
@@ -7,15 +7,25 @@
...
+    EXPECT_TRUE(ApproxEqual(expected[i], actual[i], abs_error))                \
```
csukuangfj, line 24: There is a macro `EXPECT_NEAR` which does the same thing as this line, but much simpler.

Please refer to https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#floating-point-macros

qindazhu:
@danpovey , I changed it to a lambda function
https://github.com/danpovey/k2/blob/72e5bf81c6927501d7e9c8d2d36449f6a6bbc582/k2/csrc/weights.cc#L114
https://github.com/danpovey/k2/blob/72e5bf81c6927501d7e9c8d2d36449f6a6bbc582/k2/csrc/weights.cc#L126
https://github.com/danpovey/k2/blob/72e5bf81c6927501d7e9c8d2d36449f6a6bbc582/k2/csrc/weights.cc#L135

Please kindly help to check again.

danpovey:
No, this will lead to a pointer lookup or the like, in the middle of the
loop.  Please do it how I asked.
The loop can be simplified.  The rd stuff is not necessary.  We can assume
it's acyclic and top-sorted; check that.

k2/csrc/weights.cc
```diff
@@ -68,11 +69,81 @@ WfsaWithFbWeights::WfsaWithFbWeights(const Fsa &fsa, const float *arc_weights,
...
+  ComputeForwardWeights();
+  ComputeBackardWeights();
+}
+
+// Mohri, M. 2002. Semiring framework and algorithms for shortest-distance
+// problems, Journal of Automata, Languages and Combinatorics 7(3): 321-350,
+// 2002.
+void WfsaWithFbWeights::ComputeForwardWeights() {
+  auto num_states = fsa.NumStates();
+  forward_state_weights = std::unique_ptr<double[]>(new double[num_states]);
```
csukuangfj, line 81: I prefer to use
```
std::vector<double> forward_state_weights;
```

and
```
forward_state_weights.resize(num_states);
```

k2/csrc/weights.cc
```diff
@@ -68,11 +69,81 @@ WfsaWithFbWeights::WfsaWithFbWeights(const Fsa &fsa, const float *arc_weights,
...
+WfsaWithFbWeights::AddWeightsFun WfsaWithFbWeights::AddWeights() const {
+  switch (weight_type) {
+    case kMaxWeight:
+      return [](double w1, double w2) { return std::max(w1, w2); };
+      break;
+    case kLogSumWeight:
+      return [](double w1, double w2) { return LogAdd(w1, w2); };
+      break;
+    default:
+      LOG(FATAL) << "Unreachable code is executed!";
```
csukuangfj, line 144: You have to return something here to remove compiler warnings.

k2/csrc/weights.cc
qindazhu: I'm sort of fine with this, but there is an interface issue, users will get a vector instead of a raw pointer, so for our other functions, we may need to make them accept vector (instead of pointer) of weights.

k2/csrc/weights.cc
qindazhu: OK

k2/csrc/util.h
qindazhu: it's sad that macro does not work for `infinity`, I have tried before...

k2/csrc/util.h
qindazhu: as `EXPECT_EQ` will not terminate the program, maybe I should use `ASSERT_EQ`. 

`EXPECT_TRUE` is also from gtest, seems we need to write our own check for this?

csukuangfj: I find that this macro is used only in `weights_test.cc`, should we place it in that file?

qindazhu:
> No, this will lead to a pointer lookup or the like, in the middle of the loop. Please do it how I asked. The loop can be simplified. The rd stuff is not necessary. We can assume it's acyclic and top-sorted; check that.

@danpovey I really don't get your idea here. There are two problems here, right,

- Remove relaxation variables `rd`: but it is needed here a copy of orignal `r` when we updating `d[dest_state]` as we may update itsself if there is a self-loop.
- `AddWeights`: we have to call this function to merge two paths' weights in the loop, right? how can I call it before the loop.

k2/csrc/weights.cc
csukuangfj: `*_state_weights` are private data members, can we return `*_state_weights.data()`
instead of `*_state_weights.get()`?

Users can still get a pointer.

csukuangfj:
> how can I call it before the loop.

One possible solution might be

```
if (weight_type == kMaxWeight) {
  for (....) { ... }
} else if (weight_type == kLogSumWeight) {
 for (....) { ... }
}
```

But this results in duplicate code, and a macro might be introduced:

```cpp
#define FOR_LOOP(add_func) \
  for (...) { ... add_func(...)  ... }
```

where `add_func` can be `std::max` or `LogAdd`. The advantages are that
- it avoids branches inside the loop
- it avoids indirect function calling via a pointer, something inside `std::function`.

k2/csrc/util.h
qindazhu: It will be definitely used in other algortihms like `Rm/Intersect-Prue` in `fsa_algo.cc`.

I'm open with it, what do you think? write our own CHECK for this?

csukuangfj:
> Remove relaxation variables rd: but it is needed here a copy of orignal r when we updating d[dest_state] as we may update itsself if there is a self-loop.

If it is acyclic, can we assume that it has no self-loops?

k2/csrc/weights.h
```diff
@@ -18,7 +19,11 @@
 
 namespace k2 {
 
-constexpr float kNegativeInfinity = -std::numeric_limits<float>::infinity();
+constexpr float kFloatNegativeInfinity =
+    -std::numeric_limits<float>::infinity();
+constexpr float kDoubleNegativeInfinity =
```
csukuangfj, line 24: `constexpr double`

k2/csrc/weights.cc
qindazhu: but that will be: we use vector as a pointer only for the reason that we want its ability for memory managment..that is what `unique_ptr` doing..

k2/csrc/weights.h
```diff
@@ -18,7 +19,11 @@
...
+    -std::numeric_limits<double>::infinity();
+constexpr double kShortestDelta = 1e-6;
```
csukuangfj, line 26: Is this variable unused?

danpovey:
Yes, acyclic implies no self-loops.
Have if-statement outside the loop.

k2/csrc/weights.h
qindazhu: I deleted code that do `ApproxEqual` in `Wfsa`, will delete it

k2/csrc/weights.h
qindazhu: Sure

k2/csrc/weights.cc
csukuangfj: Another advantage of using `std::vector` here:

you can use `resize()` inside `ComputeForwardWeights`, which **may** reuse the memory
allocated last time.

`std::unique_ptr` here always results in memory allocation each time `ComputeForwardWeights` is called; I guess it will be called multiple times
and that's why Dan worries about its efficiency.

csukuangfj: `std::vector` has its own memory allocator, which has better support for cache.

qindazhu:
> > Remove relaxation variables rd: but it is needed here a copy of orignal r when we updating d[dest_state] as we may update itsself if there is a self-loop.
> 
> If it is acyclic, can we assume that it has no self-loops?

we are just assuming `IsTopSorted` here, see the interface.

> > how can I call it before the loop.
> 
> One possible solution might be
> 
> ```
> if (weight_type == kMaxWeight) {
>   for (....) { ... }
> } else if (weight_type == kLogSumWeight) {
>  for (....) { ... }
> }
> ```
> 
> But this results in duplicate code, and a macro might be introduced:
> 
> ```c++
> #define FOR_LOOP(add_func) \
>   for (...) { ... add_func(...)  ... }
> ```
> 
> where `add_func` can be `std::max` or `LogAdd`. The advantages are that
> 
> * it avoids branches inside the loop
> * it avoids indirect function calling via a pointer, something inside `std::function`.

Then I prefer just write lamda function before the loop, writing Macro for so much code will make the code ugly

k2/csrc/weights.cc
```diff
@@ -68,11 +69,67 @@ WfsaWithFbWeights::WfsaWithFbWeights(const Fsa &fsa, const float *arc_weights,
...
+    case kMaxWeight:
+      return [](double w1, double w2) { return std::max(w1, w2); };
+      break;
+    case kLogSumWeight:
+      return [](double w1, double w2) { return LogAdd(w1, w2); };
+      break;
+    default:
+      // just return a value to compress compiler warnings,
+      // the program will be terminated after `LOG(FATAL)`
+      return [](double w1, double w2) { return LogAdd(w1, w2); };
```
csukuangfj, line 129: It should be placed after LOG(FATAL), otherwise, LOG(FATAL) is never reached.

qindazhu: Oh, sorry, .

k2/csrc/weights.cc
qindazhu: mm.. the input fsa is const, so we will definitly create `state_weight` only one time in the constructor.. there's no effciency issue before the loop.

k2/csrc/weights.cc
```diff
@@ -68,11 +69,67 @@ WfsaWithFbWeights::WfsaWithFbWeights(const Fsa &fsa, const float *arc_weights,
...
+  const auto &arcs = fsa.arcs;
+  forward_state_weights[0] = 0;
+  for (std::size_t i = 0; i != arcs.size(); ++i) {
+    const auto &arc = arcs[i];
+    DCHECK_GE(arc.dest_state, arc.src_state);
+    auto src_weight = forward_state_weights[arc.src_state];
+    auto &dest_weight = forward_state_weights[arc.dest_state];
+
+    double r = src_weight + arc_weights[i];
+    dest_weight = add_weights_func(dest_weight, r);
```
danpovey, line 94: No:  How many times do I have to say it, put a if statement outside the for loop!
This will require a pointer to some kind of function object, which can be quite slow (e.g. may defeat prefetch).

k2/csrc/weights.cc
danpovey: > Another advantage of using `std::vector` here:
> 
> you can use `resize()` inside `ComputeForwardWeights`, which **may** reuse the memory
> allocated last time.
> 
> `std::unique_ptr` here always results in memory allocation each time `ComputeForwardWeights` is called; I guess it will be called multiple times
> and that's why Dan worries about its efficiency.

ComputeForwardWeights() won't be called more than once for any given object.  It can only be called from the constructor.  From a memory allocation point of view I don't think std::vector and unique_ptr really differ, and I don't mind.

k2/csrc/weights.cc
danpovey: If you really dont want to duplicate the code, you can template Compute*Weights() on the ForwardBackwardType enum, and put the if-statement inside the for loop:
  if (i's max) { do something } else { do something else; }
Then the if-statement will be compiled out.

qindazhu: mm..I am currently writing the lambda (and try to avoid duplicate lambda)...this version is not the final version for `switch` issue, I just pushed code for relaxation issue..

danpovey: .. then, of course, you'd have to have an if-statement where you invoke Compute*Weights(), so you can use the right one.    It's a bit ugly.  I would probably prefer what I have been asking for all along, which is an if-statement outside the for loop with 2 versions.  That's easier for newbies to understand.

danpovey:
I'll wait to see what it is, but you can waste a lot of time trying to be
clever.  It would have taken about 10 seconds to duplicate that code like I
asked and it would be easy to read and debug, and we wouldn't have had to
discuss it so long.

k2/csrc/weights.cc
danpovey: The way you are currently doing it with lambdas/function-objects is a definite no.  That requires at the very least a function-object lookup each time round.  It defeats optimization and lots of other things.

danpovey: I think you shoudl accept that sometimes I'll make decisions you won't like.  If you want to argue about something, why not wait till it's something where the stakes are higher?  This is just a waste of time.

danpovey:
> > how can I call it before the loop.
> 
> One possible solution might be
> 
> ```
> if (weight_type == kMaxWeight) {
>   for (....) { ... }
> } else if (weight_type == kLogSumWeight) {
>  for (....) { ... }
> }
> ```
> 
> But this results in duplicate code, and a macro might be introduced:
> 
> ```c++
> #define FOR_LOOP(add_func) \
>   for (...) { ... add_func(...)  ... }
> ```
> 
> where `add_func` can be `std::max` or `LogAdd`. The advantages are that
> 
> * it avoids branches inside the loop
> * it avoids indirect function calling via a pointer, something inside `std::function`.

BTW, this is the solution I was talking about (the first one, not the one with the #define).
There's no need to go to excessive lengths to avoid duplicate code.  SOmetimes duplicate
code is the easiest solution.

qindazhu:
OK, I'll implement this version then.

danpovey:
Thanks!!

csukuangfj:
+2

danpovey:
Great, thanks!  And sorry for getting annoyed.  I thought you had
understood what I was saying and were just being obstainate.  You guys can
merge when you think it's ready.

qindazhu:
Merging

PR #37, opened 2020-05-11
[WIP] Determinization

danpovey:
This is still a work in progress, but closer to being finished.
Putting up a PR so you guys can see the code, it might give you some implementation ideas for things like epsilon removal.

### DIFF
--- a/k2/csrc/determinize.cc
+++ b/k2/csrc/determinize.cc
@@ -1,0 +1,230 @@
+// k2/csrc/fsa_algo.cc
+
+// Copyright (c)  2020  Xiaomi Corporation (authors: Daniel Povey dpove@gmail.com, Haowen Qiu qindazhu@gmail.com)
+
+// See ../../LICENSE for clarification regarding multiple authors
+
+#include "k2/csrc/fsa_algo.h"
+
+#include <utility>
+#include <vector>
+
+namespace k2 {
+
+
+struct DetStateElement {
+  // Element of the doubly linked list whose start/end are
+  // members 'head' and 'tail' of DetState.
+  // We can trace back the `parent` links, which will take
+  // us backward along a path in the original FSA.
+  DetStateElement *parent = nullptr;
+  int32_t arc_index;  // Index of most recent arc in path to the dest-state.
+                      // This data-structure represents a path through the FSA,
+                      // with this arc being the most recent arc on that path.
+  int32_t symbol;     // Symbol on the arc numbered `arc_index` of the input FSA
+                      // (copied here for convenience).
+
+  double weight;      // Weight from reference state to this state, along
+                      // the path taken by following the 'parent' links
+                      // (the path would have `seq_len` arcs in it).
+                      // Note: by "this state" we mean the destination-state of
+                      // the arc at `arc_index`.
+
+  // `prev` and `next` form the doubly linked list of DetStateElement
+  DetStateElement *prev = nullptr;
+  DetStateElement *next = nullptr;
+
+  // This comparator function compares the weights, but is careful in case of
+  // ties to ensure deterministic behavior.
+  bool operator < (const DetStateElement &other) const {
+    if (weight < other.weight) return true;
+    else if (weight > other.weight) return false;
+
+
+  }
+
+};
+
+
+
+
+
+/*
+  Conceptually a determinized state in weighted FSA determinization would normally
+  be a weighted subset of states in the input FSA, with the weights normalized
+  somehow (e.g. subtracting the sum of the weights).
+
+  Two determinized states are equal if the states and weights are the same.  To
+  ensure differentiability, our assumption is that in general no two arcs in the
+  input FSA have identical weights.  We argue that two determinized states can
+  always be represented as a base-state and a symbol sequence.  Imagine that we
+  follow arcs with that symbol sequence from the base-state, and then in case we
+  reach the same states in the different ways we always select the best path
+  from the base-state.  That process gives us a set of states and weights.  We
+  argue that this representation is unique.  (If not, it won't matter actually;
+  it will just give us an output that's less minimal than it could be).
+
+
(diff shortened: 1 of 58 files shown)


qindazhu:
OK, thanks. I'll implement `RmEpsilons*` as I have done with `Wfsa`.

danpovey:
OK, this code is basically finished, but I haven't tried to compile it.
Does anyone (@qindazhu ?) want to try to compile it and debug it?
I don't want to be the only one who has some idea how it works; and
if there are aspects that are not well documened we can find them this way.

qindazhu:
OK, will do

k2/csrc/determinize.cc
```diff
@@ -5,423 +5,706 @@
...
+     @param [in] arc_weight   Weight on the input arc
+   */
+  MaxTracebackState(int32_t state_id,
+                    const shared_ptr<MaxTracebackState> &src,
+                    int32_t incoming_arc_index,
+                    int32_t arc_weight):
+      state_id(state_id),
+      arc_id(incoming_arc_index),
+      prev_state(src),
+      forward_prob(element->forward_prob + arc_weight) { }
```
qindazhu, line 188: `src->forward_prob`, right?

k2/csrc/determinize.cc
```diff
@@ -5,423 +5,706 @@
...
-  ensure differentiability, our assumption is that in general no two arcs in the
-  input FSA have identical weights.  We argue that two determinized states can
-  always be represented as a base-state and a symbol sequence.  Imagine that we
-  follow arcs with that symbol sequence from the base-state, and then in case we
-  reach the same states in the different ways we always select the best path
-  from the base-state.  That process gives us a set of states and weights.  We
-  argue that this representation is unique.  (If not, it won't matter actually;
-  it will just give us an output that's less minimal than it could be).
+  shared_ptr<LogSumTracebackState> prev_state;
+                      // `prev_state` is the state that this points back to.
```
qindazhu, line 219: `prev_state(src)` instead of `src(src)`, right?
https://github.com/danpovey/k2/blob/a79f527e32d6fe064f704e0987c263bbf7894e16/k2/csrc/determinize.cc#L237-L240

k2/csrc/determinize.cc
```diff
@@ -5,423 +5,706 @@
...
+
+template <class TracebackState>  // TracebackState == MaxTracebackState or LogSumTracebackState
+class DetState {
+ public:
+  using DerivType = typename TracebackState::DerivType;
+  // DerivType == int32_t for MaxTracbackState, or
+  // pair<int32_t, float> for LogSumTracebackState.
+
+  // Constructor for the initial state of the determinized FSA
+  DetState(): seq_len(0), output_state(-1), normalized(true) {
```
qindazhu, line 495: what is `output_state`, it seems it's never used in `DetState`?

k2/csrc/determinize.cc
```diff
@@ -459,15 +742,229 @@ class DetStateMap {
...
+    vector<Arc> *arcs_out,
+    vector<float> *arc_weights_out,
+    vector<vector<typename TracebackState::DerivType> > *derivs_per_arc,
+    DetStateMap *state_map) {
+  float arc_weight;
+  std::vector<DerivType> deriv_info;
+  state->Normalize(wfsa_in, &arc_weight, &deriv_info);
+  int32_t next_state_id;
+  bool is_new_state = state_map->GetOutputState(state);
+  arcs_out->push_back({this->state_id, next_state_id, label});
```
qindazhu, line 765: what does `this` point to? what's `next_state_id`, seems we can remove this function as there is a function `Normalize` in `DetState`?

k2/csrc/determinize.cc
```diff
@@ -459,15 +742,229 @@ class DetStateMap {
...
+  for (auto iter = label_to_state.begin();
+       iter != label_to_state.end(); ++iter) {
+    DetState<TracebackState> *det_state = iter->second;
+
+    float arc_weight;
+    std::vector<DerivType> deriv_info;
+    det_state->Normalize(wfsa_in, &arc_weight, &deriv_info);
+    if (det_state->forward_backward_prob >= prune_cutoff) {
+      bool is_new_state = state_map->GetOutputState(state);
+      arcs_out->push_back({this->state_id, next_state_id, label});
```
qindazhu, line 823: `GetOutputState(det_state)`, right?
what's the `new_state_id` and `label`

k2/csrc/determinize.cc
```diff
@@ -459,15 +742,229 @@ class DetStateMap {
...
+  using DS = DetState<TracebackState>;
+
+  shared_ptr<DS> start_state = std::make_shared<DS>();
+
+  std::vector<Arc> arcs_out;
+  arc_weights_out->clear();
+  arc_derivs_out->clear();
+
+  bool ans = map.GetOutputState(start_state.get());
+  CHECK(ans && ans->state_id == 0);
```
qindazhu, line 916: can I remove `ans->state_id`?

k2/csrc/determinize.cc
danpovey: mm.. maybe this function should not exist but should be inline code, if it's called only once now.  this->state_id was the state id of the parent state, which is no longer in scope (was in ProcessArcs()).  next_state_id should be state->state_id.

k2/csrc/determinize.cc
danpovey: yes, GetOutputState(det_state).  next_state_id == state->state_id.  You can get `label` as `iter->first`.

k2/csrc/determinize.cc
danpovey: actually, possibly this function is never called and can be deleted.

csukuangfj:
Can we close this one since it has already been merged in #41 ?

PR #39, opened 2020-05-13
Generate randomized FSA.

### LINKED ISSUE #38: Randomized testing
It would be nice to have utilities for randomized testing.
I'm thinking of a mechanism of generating random FSAs.
We have something in Kaldi's fstext/ directory.  There could be an option for when it needs to be acyclic.

csukuangfj:
Closes #38

### DIFF
--- a/k2/csrc/fsa_util.cc
+++ b/k2/csrc/fsa_util.cc
@@ -9,10 +9,13 @@
 #include <algorithm>
 #include <cctype>
 #include <cstdlib>
+#include <random>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
 #include "glog/logging.h"
+#include "k2/csrc/fsa_algo.h"
 #include "k2/csrc/properties.h"
 
 namespace {
@@ -101,6 +104,15 @@ void SplitStringToVector(const std::string &in, const char *delim,
     if (!sub.empty()) out->emplace_back(std::move(sub));
   }
 }
+
+struct PairHash {
+  size_t operator()(const std::pair<int32_t, int32_t> &pair) const {
+    std::size_t result = 0;
+    k2::hash_combine(&result, pair.first);
+    k2::hash_combine(&result, pair.second);
+    return result;
+  }
+};
 
 }  // namespace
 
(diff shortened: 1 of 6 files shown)


csukuangfj:
@danpovey @qindazhu 
please have a review. Ready to merge.

csukuangfj:
Some generated rand FSAs are shown below:

## 1
a

## 2
b

## 3
c

csukuangfj:
Some random cyclic FSAs are

d
e

danpovey:
That was fast!!  Cool.  You can merge.
See if any of the tests can make use of this somehow.

k2/csrc/fsa_util.cc
```diff
@@ -102,6 +105,15 @@ void SplitStringToVector(const std::string &in, const char *delim,
   }
 }
 
+struct PairHash {
```
qindazhu, line 108: there is a template `PairHash` in `until.h`, just using that.

k2/csrc/fsa_util.cc
```diff
@@ -201,4 +213,83 @@ std::string FsaToString(const Fsa &fsa) {
...
+      if (seen.count(std::make_pair(src_state, dest_state))) continue;
+
+      seen.insert(std::make_pair(src_state, dest_state));
+
+      if (dest_state == num_states - 1)
+        label = kFinalSymbol;
+      else
+        label = Rand(0, static_cast<int32_t>(opts.num_syms - 1));
+
+      state_to_arcs[src_state].push_back({src_state, dest_state, label});
```
qindazhu, line 275: `emplace_back`

k2/csrc/fsa_util.cc
```diff
@@ -201,4 +213,83 @@ std::string FsaToString(const Fsa &fsa) {
   return os.str();
 }
 
+int32_t Rand(int32_t low /* = std::numeric_limits<int32_t>::min() */,
```
qindazhu, line 216: Maybe we should add some comments on `Rand` to say that it is not suggested to call it many times in one function, as it will create `gen` and `dis` again and again. Anyway, it's OK we call it in `GenerateRandFsa` as it's just a testing function

k2/csrc/fsa_util.cc
danpovey: mm, in this case I just want to take the 1st element as it contains the same information as the 2nd, just differently hashed.  It's faster.

k2/csrc/fsa_util.cc
danpovey: OK

csukuangfj:
Will resolve all the comments during the weekend.

k2/csrc/fsa_util.cc
csukuangfj: Done.

k2/csrc/fsa_util.cc
csukuangfj: Fixed.

k2/csrc/fsa_util.cc
csukuangfj: A class `RandInt` for generating random numbers is implemented.

danpovey:
Thanks!

PR #40, opened 2020-05-14
 implement RmEpsilonPruneMax

qindazhu:
Input
ori

Output

eps_max

### DIFF
--- a/k2/csrc/fsa_algo.cc
+++ b/k2/csrc/fsa_algo.cc
@@ -8,11 +8,13 @@
 #include "k2/csrc/fsa_algo.h"
 
 #include <algorithm>
+#include <functional>
 #include <limits>
 #include <numeric>
 #include <queue>
 #include <stack>
 #include <unordered_map>
+#include <unordered_set>
 #include <utility>
 
 #include "glog/logging.h"
@@ -295,53 +297,11 @@ void RmEpsilonsPrunedMax(const WfsaWithFbWeights &a, float beam, Fsa *b,
   // identify all states that should be kept
   std::vector<char> non_eps_in(num_states_a, 0);
   non_eps_in[0] = 1;
-  using WeightsPair = std::pair<double, std::vector<int32_t>>;
-  // (label, dest_state) -> `sum` of weights along all paths from current state
-  // to `dest_state` with label == `label`(or plus numbers of epsilon),
-  // `vector<int32_t>` in `WeightsPair` records all arc-indexes in `a` that
-  // contributes the current arc in `b`.
-  using ArcMap =
-      std::unordered_map<std::pair<int32_t, int32_t>, WeightsPair, PairHash>;
-  std::vector<ArcMap> arcs_b(num_states_a);
-  for (int32_t i = static_cast<int32_t>(arcs_a.size()) - 1; i >= 0; --i) {
-    const auto &arc = arcs_a[i];
-    const auto src_state = arc.src_state;
-    const auto dest_state = arc.dest_state;
-    const auto label = arc.label;
-    DCHECK_GE(dest_state, src_state);
-
-    double arc_weight = arc_weights_a[i];
-    if (label != kEpsilon) {
-      non_eps_in[dest_state] = 1;
-      WeightsPair weights_pair =
-          std::make_pair(arc_weight, std::vector<int32_t>{i});
-      auto insert_result = arcs_b[src_state].emplace(
-          std::make_pair(label, dest_state), weights_pair);
-      if (!insert_result.second) {
-        auto &old_weights_pair = insert_result.first->second;
-        // compare `arc_weights`
-        if (weights_pair.first > old_weights_pair.first) {
-          std::swap(old_weights_pair, weights_pair);
-        }
-      }
-    } else {
-      // remove epsilon arcs
-      for (const auto &item : arcs_b[dest_state]) {
-        auto weights_pair = item.second;  // copy intended
-        // `times` the arc weights along path
-        weights_pair.first += arc_weight;
-        weights_pair.second.push_back(i);
-        auto insert_result =
-            arcs_b[src_state].emplace(item.first, weights_pair);
-        if (!insert_result.second) {
-          auto &old_weights_pair = insert_result.first->second;
-          // compare `arc_weights`
-          if (weights_pair.first > old_weights_pair.first) {
-            std::swap(old_weights_pair, weights_pair);
-          }
-        }
-      }
-    }
+  for (const auto &arc : arcs_a) {
+    // We suppose the input fsa `a` is top-sorted, but only check this in DEBUG
+    // time.
+    DCHECK_GE(arc.dest_state, arc.src_state);
+    if (arc.label != kEpsilon) non_eps_in[arc.dest_state] = 1;
   }
 
   // remap state id
@@ -350,28 +310,74 @@ void RmEpsilonsPrunedMax(const WfsaWithFbWeights &a, float beam, Fsa *b,
   for (int32_t i = 0; i != num_states_a; ++i) {
     if (non_eps_in[i] == 1) state_map_a2b[i] = num_states_b++;
   }
+  b->arc_indexes.reserve(num_states_b + 1);
+  int32_t arc_num_b = 0;
 
-  // prune and output `b`
   const double *forward_state_weights = a.ForwardStateWeights();
   const double *backward_state_weights = a.BackwardStateWeights();
   const double best_weight = forward_state_weights[final_state] - beam;
-  b->arc_indexes.reserve(num_states_b + 1);
-  int32_t arc_num_b = 0;
-  for (int32_t s = 0; s < num_states_a; ++s) {
-    if (non_eps_in[s] == 1) {
-      b->arc_indexes.push_back(arc_num_b);
-      for (const auto &arcs : arcs_b[s]) {
-        int32_t dest_state = arcs.first.second;
-        double weight = arcs.second.first;
-        if (forward_state_weights[s] + weight +
-                backward_state_weights[dest_state] >
-            best_weight) {
-          b->arcs.emplace_back(state_map_a2b[s], state_map_a2b[dest_state],
-                               arcs.first.first);
-          auto curr_arc_deriv = std::move(arcs.second.second);
-          std::reverse(curr_arc_deriv.begin(), curr_arc_deriv.end());
-          arc_derivs->emplace_back(curr_arc_deriv);
-          ++arc_num_b;
+  for (int32_t i = 0; i != num_states_a; ++i) {
+    if (non_eps_in[i] != 1) continue;
+    b->arc_indexes.push_back(arc_num_b);
+    int32_t curr_state_b = state_map_a2b[i];
+    // as the input FSA is top-sorted, we use a heap here so we can process
+    // states when they already have the best cost they are going to get
+    std::priority_queue<int32_t, std::vector<int32_t>, std::greater<int32_t>> q;
+    // stores states that have been queued
+    std::unordered_set<int32_t> qstates;
+    // state -> local_forward_state_weights of this state
+    std::unordered_map<int32_t, double> local_forward_weights;
+    // state -> (src_state, arc_index) entering this state which contributes to
+    // `local_forward_weights` of this state.
+    std::unordered_map<int32_t, std::pair<int32_t, int32_t>>
+        local_backward_arcs;
+    local_forward_weights.emplace(i, forward_state_weights[i]);
+    // `-1` means we have traced back to current state `i`
+    local_backward_arcs.emplace(i, std::make_pair(i, -1));
+    q.push(i);
+    qstates.insert(i);
+    while (!q.empty()) {
+      int32_t state = q.top();
+      q.pop();
+      int32_t arc_end = fsa.arc_indexes[state + 1];
+      for (int32_t arc_index = fsa.arc_indexes[state]; arc_index != arc_end;
+           ++arc_index) {
+        int32_t next_state = arcs_a[arc_index].dest_state;
+        int32_t label = arcs_a[arc_index].label;
+        double next_weight =
+            local_forward_weights[state] + arc_weights_a[arc_index];
+        if (next_weight + backward_state_weights[next_state] >= best_weight) {
+          if (label == kEpsilon) {
+            auto result =
+                local_forward_weights.emplace(next_state, next_weight);
+            if (result.second) {
+              local_backward_arcs[next_state] =
+                  std::make_pair(state, arc_index);
+            } else {
+              if (next_weight > result.first->second) {
+                result.first->second = next_weight;
+                local_backward_arcs[next_state] =
+                    std::make_pair(state, arc_index);
+              }
+            }
+            if (qstates.find(next_state) == qstates.end()) {
+              q.push(next_state);
+              qstates.insert(next_state);
+            }
+          } else {
+            b->arcs.emplace_back(curr_state_b, state_map_a2b[next_state],
+                                 label);
+            std::vector<int32_t> curr_arc_deriv;
+            std::pair<int32_t, int32_t> curr_backward_arc{state, arc_index};
+            auto *backward_arc = &curr_backward_arc;
+            while (backward_arc->second != -1) {
+              curr_arc_deriv.push_back(backward_arc->second);
+              backward_arc = &(local_backward_arcs[backward_arc->first]);
+            }
+            std::reverse(curr_arc_deriv.begin(), curr_arc_deriv.end());
+            arc_derivs->emplace_back(std::move(curr_arc_deriv));
+            ++arc_num_b;
+          }
         }
       }
     }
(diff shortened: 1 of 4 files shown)


k2/csrc/fsa_algo.h
```diff
@@ -120,35 +122,40 @@ bool Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map = nullptr);
...
                     add an arc to the output FSA, and create the appropriate
                     arc_map entry by following backpointers (hopefully you can
-                    figure out the details).  Note: the output FSA's weights can be
-                    computed later on, by calling code, using the info in arc_map.
+                    figure out the details).  Note: the output FSA's weights can
+   be computed later on, by calling code, using the info in arc_map.
  */
 void RmEpsilonsPrunedMax(const WfsaWithFbWeights &a, float beam, Fsa *b,
-                         std::vector<std::vector<int32_t>> *arc_map);
+                         std::vector<float> *arc_weights_b);
```
qindazhu, line 158: Can I change the interface to this? Is `arc_map` needed by following steps?

k2/csrc/fsa_algo.cc
```diff
@@ -274,6 +274,85 @@ bool Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+  }
+  const double *forward_state_weights = a.ForwardStateWeights();
+  const double *backward_state_weights = a.BackwardStateWeights();
+  const double best_weight = forward_state_weights[final_state] - beam;
+  b->arc_indexes = fsa.arc_indexes;
+  auto &arc_indexes_b = b->arc_indexes;
+  int32_t arc_num_b = 0;
+  for (int32_t s = 0; s < b->NumStates(); ++s) {
+    arc_indexes_b[s] = arc_num_b;
+    if (non_eps_in[s] == 1) {
```
qindazhu, line 339: should I remap states id here? i.e. remove all epsilon states, or just leave it to users to call `Connect`

k2/csrc/fsa_algo.cc
```diff
@@ -274,6 +274,85 @@ bool Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+      for (const auto &arc_map : arcs_b[s]) {
+        int32_t dest_state = arc_map.first.second;
+        double weight = arc_map.second;
+        if (forward_state_weights[s] + weight +
+                backward_state_weights[dest_state] >
+            best_weight) {
+          b->arcs.emplace_back(Arc{s, dest_state, arc_map.first.first});
+          arc_weights_b->push_back(static_cast<float>(weight));
```
qindazhu, line 347: can I `static_cast` type (double->float) here? otherwise we may change the `arc_weights_b` to `double`

k2/csrc/fsa_algo.h
danpovey: No!   The whole point of this framework is to add differentiability.
You'll see why once we do the pytorch integration.

k2/csrc/fsa_algo.cc
danpovey: I think it would be better to remap yourself.

k2/csrc/fsa_algo.cc
danpovey: I don't want this interface... probably we won't be outputting the weights at all.

danpovey: ...they can be computed outside, by utility functions.

k2/csrc/fsa_algo.h
danpovey: Read notes/python.txt to get some feel for how this will interact with PyTorch, that may make it more clear why the interface needs to be like this.

qindazhu: OK, then I'll revert it back to the interface, and add the arc_maps computation

k2/csrc/fsa_algo.cc
qindazhu: OK

k2/csrc/fsa_algo.cc
qindazhu: OK

qindazhu:
Will go back to this PR after compile `determinize.cc`. Just go over to validate the correctness of the whole algorthim, will add `arc_map` later.

k2/csrc/fsa_algo.cc
```diff
@@ -274,6 +274,85 @@ bool Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+  for (int32_t s = 0; s < b->NumStates(); ++s) {
+    arc_indexes_b[s] = arc_num_b;
+    if (non_eps_in[s] == 1) {
+      for (const auto &arc_map : arcs_b[s]) {
+        int32_t dest_state = arc_map.first.second;
+        double weight = arc_map.second;
+        if (forward_state_weights[s] + weight +
+                backward_state_weights[dest_state] >
+            best_weight) {
+          b->arcs.emplace_back(Arc{s, dest_state, arc_map.first.first});
```
csukuangfj, line 346:
```cpp
b->arcs.emplace_back(s, dest_state, arc_map.first.first);
```

No need to write `Arc{` and `}`.  You have to rebase your change to avoid compiler errors.

k2/csrc/fsa_algo.cc
```diff
@@ -274,6 +274,85 @@ bool Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
 
+void RmEpsilonsPrunedMax(const WfsaWithFbWeights &a, float beam, Fsa *b,
+                         std::vector<float> *arc_weights_b) {
+  CHECK_GT(beam, 0);
+  CHECK_NOTNULL(b);
+  CHECK_NOTNULL(arc_weights_b);
+  b->arc_indexes.clear();
+  b->arcs.clear();
+  arc_weights_b->clear();
+
```
csukuangfj, line 285: whe have to check that

```cpp
CHECK_EQ(a.weight_type, kMaxWeight);
```

k2/csrc/fsa_algo.cc
```diff
@@ -274,6 +274,85 @@ bool Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+      }
+    }
+  }
+  const double *forward_state_weights = a.ForwardStateWeights();
+  const double *backward_state_weights = a.BackwardStateWeights();
+  const double best_weight = forward_state_weights[final_state] - beam;
+  b->arc_indexes = fsa.arc_indexes;
+  auto &arc_indexes_b = b->arc_indexes;
+  int32_t arc_num_b = 0;
+  for (int32_t s = 0; s < b->NumStates(); ++s) {
```
csukuangfj, line 337: It'd be better if `b->NumStates()` is moved outside of the `for` loop.

k2/csrc/fsa_algo.cc
qindazhu: Sure

qindazhu:
Remapped `state_id` in output `b`, added `arc_derivs`

qindazhu:
Sorry, forgot to remove `WIP` in the title. Ready to merge if there's no other issues.

k2/csrc/fsa_algo.cc
```diff
@@ -275,6 +275,111 @@ bool Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+  int32_t num_states_a = fsa.NumStates();
+  int32_t final_state = fsa.FinalState();
+  const auto &arcs_a = fsa.arcs;
+  const float *arc_weights_a = a.arc_weights;
+
+  // identify all states that should be kept
+  std::vector<char> non_eps_in(num_states_a, 0);
+  non_eps_in[0] = 1;
+  using WeightsPair = std::pair<double, std::vector<int32_t>>;
+  // (label, dest_state) -> `sum` of weights along all paths from current state
```
danpovey, line 299: clarify that sum here == max.  probably just say max would be clearer.

danpovey: You don't need to sum over a specific label here.  We are just removing epsilons, not determinizing.

k2/csrc/fsa_algo.cc
```diff
@@ -275,6 +275,111 @@ bool Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+  std::vector<ArcMap> arcs_b(num_states_a);
+  for (int32_t i = static_cast<int32_t>(arcs_a.size()) - 1; i >= 0; --i) {
+    const auto &arc = arcs_a[i];
+    const auto src_state = arc.src_state;
+    const auto dest_state = arc.dest_state;
+    const auto label = arc.label;
+    DCHECK_GE(dest_state, src_state);
+
+    double arc_weight = arc_weights_a[i];
+    if (label != kEpsilon) {
```
danpovey, line 314: This algorithm would take quadratic time in the length of epsilon paths.  Better to use a traceback-based algorithm, like what I sketched in the header.

k2/csrc/fsa_algo.cc
```diff
@@ -275,6 +275,111 @@ bool Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+  // contributes the current arc in `b`.
+  using ArcMap =
+      std::unordered_map<std::pair<int32_t, int32_t>, WeightsPair, PairHash>;
+  std::vector<ArcMap> arcs_b(num_states_a);
+  for (int32_t i = static_cast<int32_t>(arcs_a.size()) - 1; i >= 0; --i) {
+    const auto &arc = arcs_a[i];
+    const auto src_state = arc.src_state;
+    const auto dest_state = arc.dest_state;
+    const auto label = arc.label;
+    DCHECK_GE(dest_state, src_state);
```
danpovey, line 311: You are relying on topological sorting.. comment so near the top of this function and point the reader to this line where you check it.

k2/csrc/fsa_algo.cc
qindazhu: OK, will change sum to max.

RE `sum over label`, if there is two epsilon arcs from `a->b`, seems we need to do `sum(max)` here, right?

danpovey: yes, for epsilons, but doing that for non-epsilon labels is unnecessary.

k2/csrc/fsa_algo.cc
qindazhu: OK, will add in `.cc` file (actually you have commented in the header file to require TopSorted)
https://github.com/danpovey/k2/blob/master/k2/csrc/fsa_algo.h#L84-L87

k2/csrc/fsa_algo.cc
danpovey: .. calling epsilons a label in this context would be a little strange, given that epsilon means "no label here".

k2/csrc/fsa_algo.cc
qindazhu: if there is path `a->b->c->d` and `e->b->c->d` with all labels are epsilon, the `traceback-based` algorithm will iterate over `b->c->d` twice, right? And the current implementation will only iterate over the path one time (Of course with a cost of copying non-epsilon output arcs for those states that has epsilon arcs in).

Fo example,  if there is copy arc in the current implementation (suppose a->b with label epsilon (no label), and b->c with label `l`, then there is a  copy arc  `a->c` with label `l`), then there's also a trace path from a->b->c (and may multiple such iterations as there maybe `e->a` with epsilon and `f->a` with epsilon) in traceback-based algorithm, so seems the current implenmentation will  take less time?

Did I express clearly? Correct me If I'm wrong.

danpovey: The problem is that your algorithm can in general be quadratic in the size of the output FSA, because for a sequence of epsilons, you have to append those vector<int> things.  Mine is linear in the size of the output FSA, assuming the input is connected.

danpovey: The problem is that your algorithm can in general be quadratic in the size of the output FSA, because for a sequence of epsilons, you have to append those vector<int> things.  Mine is linear in the size of the output FSA, assuming the input is connected.

danpovey: .. each time you trace e->a or f->a or the like, you are taking time O(length of that sequence a->b->c->d) as you are manipulating those vector<int> things.

qindazhu: Oh, ok, will revert back to the traceback-based algorithm.

qindazhu:
@danpovey  implemented with traceback-based algorithm, please have a review.

danpovey:
OK, later.

k2/csrc/fsa_algo.h
```diff
@@ -91,71 +91,13 @@ bool Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map = nullptr);
...
-   + backward_state_weights[arc_dest_state] < best_path_weight - beam: if arc
-   label is epsilon: if next_weight < local_forward_state_weight[next_state]:
-                           local_forward_state_weight[next_state] = next_weight
-                           local_backpointers[next_state] = ji
-                else:
-                    add an arc to the output FSA, and create the appropriate
-                    arc_map entry by following backpointers (hopefully you can
-                    figure out the details).  Note: the output FSA's weights can
-   be computed later on, by calling code, using the info in arc_map.
+    @param [out] arc_derivs  Indexed by arc in `b`, this is the sequence of
```
csukuangfj, line 94: Line 93
```cpp
will be in the same order that they were in in `a`.
```

---> in in `a` <-----

There is an extra `in`.

k2/csrc/fsa_algo.h
```diff
@@ -91,71 +91,13 @@ bool Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map = nullptr);
...
+                      arcs in `a` that this arc in `b` corresponds to; the
+                      weight of the arc in b will equal the sum of those input
+                      arcs' weights
  */
 void RmEpsilonsPrunedMax(const WfsaWithFbWeights &a, float beam, Fsa *b,
-                         std::vector<std::vector<int32_t>> *arc_map);
+                         std::vector<std::vector<int32_t>> *arc_derivs);
```
csukuangfj, line 100: Can we keep `arc_map`?
I am thinking of arc derivatives once I see `arc_derivs`, though it has nothing
to do with derivatives.

k2/csrc/fsa_algo_test.cc
```diff
@@ -283,6 +283,54 @@ TEST(FsaAlgo, Connect) {
...
+  int32_t num_states_;
+  float *arc_weights_;
+};
+
+TEST_F(RmEpsilonTest, RmEpsilonsPrunedMax) {
+  Fsa b;
+  std::vector<std::vector<int32_t>> arc_derivs_b;
+  RmEpsilonsPrunedMax(*max_wfsa_, 8, &b, &arc_derivs_b);
+
+  IsEpsilonFree(b);
```
csukuangfj, line 325:
```cpp
EXPECT_TRUE(IsEpsilonFree(b));
```

csukuangfj:
Looks good to me.

danpovey:
Sorry I may have to look at this tomorrow.

k2/csrc/fsa_algo_test.cc
qindazhu: Ah, thanks

k2/csrc/fsa_algo.h
qindazhu: `arc_derivs` may be better, as  `arc_derivs` here actually is `vector<vector<pair<int32_t, int32_t>` where `pair.second` is always `1`. Another reason is that `map` usually means `1 to 1 `mapping in convention, especially in the perspective of function/algorithm, but `arc_derivs` here usually is a `one to many` mapping. Dan aslo uses this name in many other algorithms, such as determinize.

danpovey: mm, I prefer arc_derivs too.  it does actually represent derivatives in a sense, as in it says how the weights on the output arcs vary with those of the input arcs.

k2/csrc/fsa_algo.cc
```diff
@@ -275,6 +277,115 @@ bool Connect(const Fsa &a, Fsa *b, std::vector<int32_t> *arc_map /*=nullptr*/) {
...
+    if (non_eps_in[i] != 1) continue;
+    b->arc_indexes.push_back(arc_num_b);
+    int32_t curr_state_b = state_map_a2b[i];
+    // as the input FSA is top-sorted, we use a heap here so we can process
+    // states when they already have the best cost they are going to get
+    std::priority_queue<int32_t, std::vector<int32_t>, std::greater<int32_t>> q;
+    // stores states that have been queued
+    std::unordered_set<int32_t> qstates;
+    // state -> local_forward_state_weights of this state
+    std::unordered_map<int32_t, double> local_forward_weights;
```
danpovey, line 329: I wonder whether using std::map<int32_t, double> and not using the priority_queue (just the map, as it's ordered) might be faster in most cases?  For chains of arcs, which is the common case most likely, the map would usually only have 1 or 2 elements in it.

qindazhu: Great insight. thanks! Have changed to map, please check the last commit. You can merge it if there is no other issues.

danpovey:
Cool!  Merging.

PR #42, opened 2020-05-16
fix determinize issues

qindazhu:
Fixed some issues of `Determinize`.  The alogrithm now can ouput FSA correctly.

Input:
ori_max

Output
new_max (2)

**Noted that** the `weights_out` and `derivs_out` are still **not correct**, I'll debug into it and make **another PR** to fix them.

### DIFF
--- a/k2/csrc/determinize.cc
+++ b/k2/csrc/determinize.cc
@@ -115,6 +115,7 @@ void TraceBack(std::unordered_set<MaxTracebackState *> *cur_states,
     // `deriv_out` is just a list of arc indexes in the input FSA
     // that this output arc depends on (it's their sum).
     (*deriv_out)[i] = state->arc_id;
+    state = state->prev_state.get();
   }
   double prev_forward_prob = state->forward_prob;
   *weight_out = cur_forward_prob - prev_forward_prob;
--- a/k2/csrc/determinize.h
+++ b/k2/csrc/determinize.h
@@ -20,6 +20,7 @@
 #include <vector>
 
 #include "k2/csrc/fsa.h"
+#include "k2/csrc/fsa_algo.h"
 #include "k2/csrc/util.h"
 #include "k2/csrc/weights.h"
 
@@ -371,20 +372,20 @@ class DetState;
 
 template <class TracebackState>
 struct DetStateCompare {
-  bool operator()(const std::unique_ptr<DetState<TracebackState>> &a,
-                  const std::unique_ptr<DetState<TracebackState>> &b) {
+  bool operator()(const std::shared_ptr<DetState<TracebackState>> &a,
+                  const std::shared_ptr<DetState<TracebackState>> &b) {
     return a->forward_backward_prob < b->forward_backward_prob;
   }
 };
 // Priority queue template arguments:
-//   item queued = unique_ptr<DetState> (using pointer equality as comparison)
-//   container type = vector<unique_ptr<DetState> >
+//   item queued = shared_ptr<DetState> (using pointer equality as comparison)
+//   container type = vector<shared_ptr<DetState> >
 //   less-than operator = DetStateCompare (which compares the
 //   forward_backward_prob).
 template <class TracebackState>
 using DetStatePriorityQueue =
-    std::priority_queue<std::unique_ptr<DetState<TracebackState>>,
-                        std::vector<std::unique_ptr<DetState<TracebackState>>>,
+    std::priority_queue<std::shared_ptr<DetState<TracebackState>>,
+                        std::vector<std::shared_ptr<DetState<TracebackState>>>,
                         DetStateCompare<TracebackState>>;
 
 template <class TracebackState>
@@ -448,7 +449,7 @@ class DetState {
                          const std::shared_ptr<TracebackState> &src,
                          int32_t incoming_arc_index, int32_t arc_weight) {
     auto ret = elements.insert({state_id, nullptr});
-    if (!ret.second) {  // No such state existed in `elements`
+    if (ret.second) {  // No such state existed in `elements`
       ret.first->second = std::make_shared<TracebackState>(
           state_id, src, incoming_arc_index, arc_weight);
     } else {  // A state with this staste_id existed in `elements`.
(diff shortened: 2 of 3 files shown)


k2/csrc/fsa_algo_test.cc
```diff
@@ -502,6 +502,53 @@ TEST(FsaAlgo, TopSort) {
...
+  Fsa *fsa_;
+  int32_t num_states_;
+  float *arc_weights_;
+};
+
+TEST_F(DeterminizeTest, DeterminizePrunedMax) {
+  Fsa b;
+  std::vector<float> b_arc_weights;
+  std::vector<std::vector<int32_t>> arc_derivs;
+  DeterminizePrunedMax(*max_wfsa_, 10, 100, &b, &b_arc_weights, &arc_derivs);
```
qindazhu, line 542: Just leave `EXPECT` statements empty for now, will add after fix issues of `weights`. please ignore this for now.

k2/csrc/determinize.cc
```diff
@@ -28,6 +28,7 @@ LogSumTracebackLink::LogSumTracebackLink(
 
 int32_t GetMostRecentCommonAncestor(
     std::unordered_set<LogSumTracebackState *> *cur_states) {
+  if (cur_states->size() == 1) return 1;
```
danpovey, line 31: My intention was that it return 0 in this case; even the documentation says that.  What happened if it returned zero?

k2/csrc/determinize.h
```diff
@@ -371,20 +372,20 @@ class DetState;
...
+//   item queued = shared_ptr<DetState> (using pointer equality as comparison)
+//   container type = vector<shared_ptr<DetState> >
 //   less-than operator = DetStateCompare (which compares the
 //   forward_backward_prob).
 template <class TracebackState>
 using DetStatePriorityQueue =
-    std::priority_queue<std::unique_ptr<DetState<TracebackState>>,
-                        std::vector<std::unique_ptr<DetState<TracebackState>>>,
+    std::priority_queue<std::shared_ptr<DetState<TracebackState>>,
+                        std::vector<std::shared_ptr<DetState<TracebackState>>>,
```
danpovey, line 388: What was the problem with unique_ptr here?  I made it unique_ptr as it's more lightweight than shared_ptr (and this is only ever owned in one place, I believed.)

k2/csrc/determinize.cc
qindazhu: if it's `0`, there `label` will not be used in `DetStateToCompact`
https://github.com/danpovey/k2/blob/75c4cd1b1fdd2a20009997c269b905b438575cdb/k2/csrc/determinize.h#L745-L746

Different states with same `seq_len` will mapped to a same `det_state`.

k2/csrc/determinize.h
qindazhu: `priority_queue.top()` returns a const reference so we cannnot call `std::move()` on its returned value to construct any `unqiue_ptr` or `shared_ptr`.

qindazhu: see here 
https://github.com/danpovey/k2/blob/75c4cd1b1fdd2a20009997c269b905b438575cdb/k2/csrc/determinize.h#L818

danpovey:
It's supposed to be that way.

However, I see a problem there.  In fact, when seq_len is 0 it's OK but
otherwise it is not.

The state that it should be taking into account there is not the final
state (what's called input_state in the code), the state that we get to
after seeing the sequence, but the base_state which is the state we're in
before we see that sequence of symbols.  So we should add the
elem->state_id *after* the loop.

k2/csrc/determinize.h
danpovey: I see; shared_ptr is fine.

qindazhu:
Oh, yes, it should be zero as I have changed the `state_id` in `DetStateToCompact` to `input_state_id` (it is originally zero for any unnormalized states).

RE add `elem->state_id`, Can you show some details with code? I really didn't get your idea.

Revert it back to `zero`.

k2/csrc/determinize.h
```diff
@@ -766,8 +756,10 @@ class DetStateMap {
                          const Fsa &fsa, std::pair<uint64_t, uint64_t> *vec) {
     assert(d.normalized);
 
-    uint64_t a = d.state_id + 17489 * d.seq_len,
-             b = d.state_id * 103979 + d.seq_len;
+    int32_t input_state_id = d.elements.begin()->first;
+
+    uint64_t a = input_state_id + 17489 * d.seq_len,
```
danpovey, line 761: Remove input_state_id here, just have 17489 * d.seq_len and so on... and I'll say below..

danpovey, line 761: Remove input_state_id here, just have 17489 * d.seq_len and so on... and I'll say below..

danpovey: well, around line 776, insert: 
```
    a += elem->state_id;   // This is `base_state`: the state from which we 
                                          // start (and accept the specified symbol sequence).
```
(we don't need it in b, actually).

danpovey: actually, no, to reduce the chance of collisions, do instead:
       a = elem->state_id + 14051 * a

danpovey: .. I mean collisions on a, which is used for hashing.  collisions should still be vanishingly rare if we take a and b together.

qindazhu: oh, add the `base_state`, got it, thanks!

qindazhu: Updated

k2/csrc/determinize.h
```diff
@@ -781,6 +768,7 @@ class DetStateMap {
       b = symbol + 102983 * b;
       elem = elem->prev_elements[0].prev_state;
     }
+    a += elem->state_id;
```
danpovey, line 771: Two things: (1) after suggesting this I suggested a different version involving multiplying by a prime.. please use that one.  And (2) in my original suggested code there was a comment; please include that!

k2/csrc/fsa_algo_test.cc
```diff
@@ -502,6 +502,53 @@ TEST(FsaAlgo, TopSort) {
...
+}
+
+TEST_F(DeterminizeTest, DeterminizePrunedLogSum) {
+  Fsa b;
+  std::vector<float> b_arc_weights;
+  std::vector<std::vector<std::pair<int32_t, float>>> arc_derivs;
+  DeterminizePrunedLogSum(*log_wfsa_, 10, 100, &b, &b_arc_weights, &arc_derivs);
+}
```
danpovey, line 550: I'd like, to have this code automatically:
 (1) check that the result is deterministic,
 (2) check that the result is equivalent to the original (in the appropriate semiring)

and if possible check that the arc_derivs make sense somehow, although that is more complex and can be left for now.

danpovey: I can merge this PR when you fix the other issue, though;  you can work on this later.

k2/csrc/determinize.h
qindazhu: Oh, sorry for the comments (have dinner just now..)

RE prime: seems there's notification sync issue on my chrome, will add, thanks

k2/csrc/fsa_algo_test.cc
qindazhu: will add all `EXPECT` statements after fix issues of `weights` (As I said before,  the `weights_out` now are not correct)

PR #43, opened 2020-05-18
fix some issues of determinize and add some basic tests

qindazhu:


qindazhu:
Will see the test failure tomorrow, seems related with the implmentation of STL in different compilers.

danpovey:
Great, merging.
Run this in valgrind to check for memory errors (these might explain the test failure).

qindazhu:
The original test failure is because of 
https://github.com/qindazhu/k2/blob/ea97ccde899b956968863fcde3870625ae8936f8/k2/csrc/fsa_algo_test.cc#L552-L554

And there's memory leak in `label_to_state`, I'll fix it together with another PR (just add `else delete det_state` after below code)
https://github.com/qindazhu/k2/blob/ea97ccde899b956968863fcde3870625ae8936f8/k2/csrc/determinize.h#L610-L611

danpovey:
Ah, OK.  The right way to do that test is to test equivalence and test that
the output is deterministic, and not rely on "gold output".

qindazhu:
Yes, that's why I comment a TODO with `we need to check the equivalence automatically`, will replace the TODO with tests after implementing `RandEquivalent` for WFsa

PR #44, opened 2020-05-25
Implement RandEquivalent and ShorestDistance for weighted FSA

qindazhu:
- Implemented `ComputeForwardLogSumWeights` and  `ComputeBackwardLogSumWeights`
- Implemented `ShortestDistance`
- Implemented `RandEquivalent` for WFSA.

### DIFF
--- a/k2/csrc/fsa_equivalent.cc
+++ b/k2/csrc/fsa_equivalent.cc
@@ -45,111 +45,12 @@ void SetDifference(const std::unordered_set<int32_t> &a,
     if (a.find(v) == a.end()) c->insert(v);
   }
 }
-}  // namespace
 
-namespace k2 {
-
-bool IsRandEquivalent(const Fsa &a, const Fsa &b, std::size_t npath /*=100*/) {
-  // We will do `intersect` later which requires either `a` or `b` is
-  // epsilon-free, considering they should hold same set of arc labels, so both
-  // of them should be epsilon-free.
-  // TODO(haowen): call `RmEpsilon` here instead of checking.
-  if (!IsEpsilonFree(a) || !IsEpsilonFree(b)) return false;
-
-  Fsa connected_a, connected_b, valid_a, valid_b;
-  Connect(a, &connected_a);
-  Connect(b, &connected_b);
-  ArcSort(connected_a, &valid_a);  // required by `intersect`
-  ArcSort(connected_b, &valid_b);
-  if (IsEmpty(valid_a) && IsEmpty(valid_b)) return true;
-  if (IsEmpty(valid_a) || IsEmpty(valid_b)) return false;
-
-  // Check that arc labels are compatible.
-  std::unordered_set<int32_t> labels_a, labels_b;
-  for (const auto &arc : valid_a.arcs) labels_a.insert(arc.label);
-  for (const auto &arc : valid_b.arcs) labels_b.insert(arc.label);
-  if (labels_a != labels_b) return false;
-
-  Fsa c, connected_c, valid_c;
-  if (!Intersect(valid_a, valid_b, &c)) return false;
-  Connect(c, &connected_c);
-  ArcSort(connected_c, &valid_c);
-  if (IsEmpty(valid_c)) return false;
-
-  std::random_device rd;
-  std::mt19937 gen(rd());
-  std::bernoulli_distribution coin(0.5);
-  for (auto i = 0; i != npath; ++i) {
-    const auto &fsa = coin(gen) ? valid_a : valid_b;
-    Fsa path, valid_path;
-    RandomPath(fsa, &path);  // path is already connected
-    ArcSort(path, &valid_path);
-    Fsa cpath, valid_cpath;
-    Intersect(valid_path, valid_c, &cpath);
-    Connect(cpath, &valid_cpath);
-    if (IsEmpty(valid_cpath)) return false;
-  }
-
-  return true;
-}
-
-template <FbWeightType Type>
-bool IsRandEquivalent(const Fsa &a, const float *a_weights, const Fsa &b,
-                      const float *b_weights, std::size_t npath /*= 100*/) {
-  Fsa connected_a, connected_b, valid_a, valid_b;
-  std::vector<int32_t> connected_a_arc_map, connected_b_arc_map,
-      valid_a_arc_map, valid_b_arc_map;
-  Connect(a, &connected_a, &connected_a_arc_map);
-  Connect(b, &connected_b, &connected_b_arc_map);
-  ArcSort(connected_a, &valid_a, &valid_a_arc_map);  // required by `intersect`
-  ArcSort(connected_b, &valid_b, &valid_b_arc_map);
-  if (IsEmpty(valid_a) && IsEmpty(valid_b)) return true;
-  if (IsEmpty(valid_a) || IsEmpty(valid_b)) return false;
-
-  // Get arc weights
-  std::vector<float> valid_a_weights(valid_a.arcs.size());
-  std::vector<float> valid_b_weights(valid_b.arcs.size());
-  GetArcWeights(a_weights, connected_a_arc_map, valid_a_arc_map,
-                &valid_a_weights);
-  GetArcWeights(b_weights, connected_b_arc_map, valid_b_arc_map,
-                &valid_b_weights);
-
-  // Check that arc labels are compatible.
-  std::unordered_set<int32_t> labels_a, labels_b, labels_difference;
-  for (const auto &arc : valid_a.arcs) labels_a.insert(arc.label);
-  for (const auto &arc : valid_b.arcs) labels_b.insert(arc.label);
-  SetDifference(labels_a, labels_b, &labels_difference);
-  if (labels_difference.size() >= 2 ||
-      (labels_difference.size() == 1 &&
-       (*(labels_difference.begin()) != kEpsilon)))
-    return false;
-
-  std::random_device rd;
-  std::mt19937 gen(rd());
-  std::bernoulli_distribution coin(0.5);
-  for (auto i = 0; i != npath; ++i) {
-    const auto &fsa = coin(gen) ? valid_a : valid_b;
-    Fsa path, valid_path;
-    RandomPathWithoutEpsilonArc(fsa, &path);  // path is already connected
-    ArcSort(path, &valid_path);
-
-    Fsa a_compose_path, b_compose_path;
-    std::vector<float> a_compose_weights, b_compose_weights;
-    Intersect(valid_a, valid_a_weights.data(), path, &a_compose_path,
-              &a_compose_weights);
-    Intersect(valid_b, valid_b_weights.data(), path, &b_compose_path,
-              &b_compose_weights);
-    double sum_a =
-        ShortestDistance<Type>(a_compose_path, a_compose_weights.data());
-    double sum_b =
-        ShortestDistance<Type>(b_compose_path, b_compose_weights.data());
-    if (!DoubleApproxEqual(sum_a, sum_b)) return false;
-  }
-  return true;
-}
-
-bool RandomPath(const Fsa &a, Fsa *b, bool no_epsilon_arc,
-                std::vector<int32_t> *state_map /*=nullptr*/) {
+bool RandomPathHelper(const k2::Fsa &a, k2::Fsa *b, bool no_epsilon_arc,
+                      std::vector<int32_t> *state_map = nullptr) {
+  using k2::Arc;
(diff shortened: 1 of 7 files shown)


k2/csrc/fsa_equivalent.cc
```diff
@@ -62,8 +90,63 @@ bool IsRandEquivalent(const Fsa &a, const Fsa &b, std::size_t npath /*=100*/) {
...
+    RandomPathWithoutEpsilonArc(fsa, &path);  // path is already connected
+    ArcSort(path, &valid_path);
+
+    Fsa a_compose_path, b_compose_path;
+    std::vector<float> a_compose_weights, b_compose_weights;
+    Intersect(valid_a, valid_a_weights.data(), path, &a_compose_path,
+              &a_compose_weights);
+    Intersect(valid_b, valid_b_weights.data(), path, &b_compose_path,
+              &b_compose_weights);
+    double sum_a = ShortestDistance<FbWeightType>(a_compose_path,
```
qindazhu, line 139: @danpovey, Seems we cannot suppose `a_compose_path` is top-sorted and need to implement the version ShortestDistance with relax things, right?

danpovey: I can't find those commits on github so can't se the whole thing, but if that's an issue, it would be possible to top-sort those things separately, no?  I'm open to having a version of ShortestDistance that doesn't require top-sorting though, or to make the whole thing not require top-sorting if it would be as fast.   (But I do slightly worry about whether that would be as easy to do on GPU).

qindazhu: This is a new PR for checking equivalence for weighted FSA which will be used to check equivalence of `determinize/RmEpsilon/Interset..`. There's now no any implementation of `ShortestDistance` for any version in our repo.

In the above scenarios (determinze/RmEpsilon/Interset..), if both the input and output of these algorithms are top-sorted FSAs, then we can just only implement a version of `ShortestDistance` for top-sorted FSA, if not, we need to implement a version for non-sorted FSA. (if two FSAs are top-sorted, their intersection will also be top-sorted according to our implemeation). 

BTW, the current implementation of `RandEquivalent` does not require top-sorting in all steps except in `ShortestDistance`, that's why I wonder if we should provide a general version, or just TopSorted before calling `ShorestDistance` (may fail if there is cycle and then we will fail to the equivalence)

qindazhu:
> If we ever need these RandomPath functions at a user-level, it will be necessary to provide an `arc_map` instead of a `state_map` as the output, because that's required for differentiability. arc_map provides more information than state_map.

Sure, For now seems it will only be used in `RandEquivalent`, we can add arc_map if we find it will be called at user-level.

k2/csrc/fsa_equivalent.cc
danpovey: How about adding a boolean flag that's true if the input is top-sorted.

Eventually we should add the notion of properties.  But there are some other changes that I want to make first, and I want to wait a bit on that.

qindazhu: I'm OK with add a boolean flag.

And what's your changes? As this PR is not finished (including tests), I'm working on the PR of `RmEpsilon` now. I can do this PR first if you think it has a higher priority.

danpovey:
You decide what to do first.
I want to add a different, more general interface for FSAs that are inputs
to operations, so that we can use linearized FSAs (living inside
tensors) as inputs to things.  And add a format for a vector of FSAs that
can live inside a PyTorch tensor.

k2/csrc/fsa_equivalent.h
```diff
@@ -22,13 +23,58 @@ namespace k2 {
  */
 bool IsRandEquivalent(const Fsa &a, const Fsa &b, std::size_t npath = 100);
 
+/*
+  Returns true if the Fsa `a` is stochastically equivalent to `b` by randomly
+  generating `npath` paths from one of them and then checking if each path exist
+  in the other one and the sum of weights along that path are the same.
+ */
+template <class FbWeightType>
```
csukuangfj, line 31: It should be

```cpp
template<FbWeightType Type>
```

And you can invoke it with `IsRandEquivalent<kMaxWeight>(....)`
or `IsRandEquivalent<kLogSumWeight>(....)`


The same applies to the subsequent `ShortestDistance `.

k2/csrc/fsa_equivalent.h
```diff
@@ -22,13 +23,58 @@ namespace k2 {
  */
 bool IsRandEquivalent(const Fsa &a, const Fsa &b, std::size_t npath = 100);
 
+/*
+  Returns true if the Fsa `a` is stochastically equivalent to `b` by randomly
+  generating `npath` paths from one of them and then checking if each path exist
```
csukuangfj, line 28: if each path exist**s**. A typo here.

k2/csrc/fsa_equivalent.cc
```diff
@@ -62,8 +90,63 @@ bool IsRandEquivalent(const Fsa &a, const Fsa &b, std::size_t npath /*=100*/) {
   return true;
 }
 
-bool RandomPath(const Fsa &a, Fsa *b,
-                std::vector<int32_t> *state_map /*=nullptr*/) {
+template <class FbWeightType>
+bool IsRandEquivalent(const Fsa &a, const float *a_weights, const Fsa &b,
```
csukuangfj, line 94: Should we place it in the header file? 

Since it is a template and it is not instantiated in this `.cc` file, we may have compiler errors
later if it is used in another `.cc` file.

k2/csrc/fsa_equivalent.cc
```diff
@@ -62,8 +90,63 @@ bool IsRandEquivalent(const Fsa &a, const Fsa &b, std::size_t npath /*=100*/) {
...
+                                                  a_compose_weights.data());
+    double sum_b = ShortestDistance<FbWeightType>(b_compose_path,
+                                                  b_compose_weights.data());
+    if (!DoubleApproxEqual(sum_a, sum_b)) return false;
+  }
+  return true;
+}
+
+bool RandomPath(const Fsa &a, Fsa *b, bool no_epsilon_arc,
+                std::vector<int32_t> *state_map) {
```
csukuangfj, line 149:
```cpp
std::vector<int32_t> *state_map /* = nullptr */
```

k2/csrc/fsa_equivalent.cc
```diff
@@ -62,8 +90,63 @@ bool IsRandEquivalent(const Fsa &a, const Fsa &b, std::size_t npath /*=100*/) {
...
+    double sum_a = ShortestDistance<FbWeightType>(a_compose_path,
+                                                  a_compose_weights.data());
+    double sum_b = ShortestDistance<FbWeightType>(b_compose_path,
+                                                  b_compose_weights.data());
+    if (!DoubleApproxEqual(sum_a, sum_b)) return false;
+  }
+  return true;
+}
+
+bool RandomPath(const Fsa &a, Fsa *b, bool no_epsilon_arc,
```
csukuangfj, line 148: This overload is not declared in the header file `fsa_equivalent.h`.

csukuangfj: It would be great if this function is placed in the above unnamed namespace
if it is used only inside this `.cc` file.

k2/csrc/fsa_equivalent.cc
```diff
@@ -123,4 +214,128 @@ bool RandomPath(const Fsa &a, Fsa *b,
...
+  if (!IsArcSorted(a) || !IsArcSorted(b)) return false;
+  if (!IsEpsilonFree(b)) return false;
+
+  int32_t final_state_a = a.NumStates() - 1;
+  int32_t final_state_b = b.NumStates() - 1;
+  const auto arc_a_begin = a.arcs.begin();
+  const auto arc_b_begin = b.arcs.begin();
+  using ArcIterator = std::vector<Arc>::const_iterator;
+
+  const int32_t final_state_c = -1;  // just as a placeholder
```
csukuangfj, line 249: The name of a `const` variable has the style `kConst`

k2/csrc/fsa_equivalent.cc
csukuangfj: Or we can do explicitly instantiation in this `.cc` file and place an `extern template`
declaration in the header file.

k2/csrc/fsa_equivalent.h
qindazhu: Ah, thanks, that's what I wanted to do. Will change

k2/csrc/fsa_equivalent.cc
qindazhu: Yes, I plan do `explicit template instantiation` here as there are only two different types for FbWeightType

k2/csrc/fsa_equivalent.cc
danpovey: BTW, if something is used only in a .cc file I'd rather it be `static` to clarify that.

danpovey:
Thanks for the detailed comments, @csukuangfj !

danpovey:
Thanks for the detailed comments, @csukuangfj !

k2/csrc/fsa_equivalent.cc
qindazhu: Sure, will move to unnamed namespace and add `static`

k2/csrc/fsa_equivalent.cc
qindazhu: Sure

k2/csrc/fsa_equivalent.cc
```diff
@@ -7,14 +7,119 @@
...
 
 #include "k2/csrc/fsa.h"
 #include "k2/csrc/fsa_algo.h"
 #include "k2/csrc/properties.h"
+#include "k2/csrc/util.h"
+#include "k2/csrc/weights.h"
+
+namespace {
+// out_weights[i] = weights[arc_map1[arc_map2[i]]]
+static void GetArcWeights(const float *weights,
```
csukuangfj, line 25: > When definitions in a .cc file do not need to be referenced outside that file, place them in an unnamed namespace or declare them static

Please refer to https://google.github.io/styleguide/cppguide.html#Unnamed_Namespaces_and_Static_Variables

So there is no need to add `static` here since they are already in
an anonymous namespace.

qindazhu: Yes, you're right, both of them give internal linkage. Will remove, thanks!

qindazhu:
Ready to merge if there is no other issue.

k2/csrc/fsa_equivalent.cc
```diff
@@ -7,14 +7,118 @@
...
 
 #include "k2/csrc/fsa.h"
 #include "k2/csrc/fsa_algo.h"
 #include "k2/csrc/properties.h"
+#include "k2/csrc/util.h"
+#include "k2/csrc/weights.h"
+
+namespace {
+// out_weights[i] = weights[arc_map1[arc_map2[i]]]
+void GetArcWeights(const float *weights, const std::vector<int32_t> &arc_map1,
```
danpovey, line 25: can we make things static if they are local to the file?
I know the anon namespace guarantees this, I just feel like it would be clearer.

k2/csrc/fsa_equivalent.cc
```diff
@@ -62,64 +166,197 @@ bool IsRandEquivalent(const Fsa &a, const Fsa &b, std::size_t npath /*=100*/) {
...
-  int32_t state = 0;
-  int32_t final_state = num_states - 1;
-  while (true) {
-    if (state_map_a2b[state] == -1) {
-      state_map_a2b[state] = num_visited_state;
-      state_map_b2a.push_back(state);
-      visited_arcs.emplace_back(std::unordered_set<Arc, ArcHash>());
-      ++num_visited_state;
+  if (IsEmpty(a) || IsEmpty(b)) return true;
+  if (!IsArcSorted(a) || !IsArcSorted(b)) return false;
```
danpovey, line 262: I think we should be crashing rather than returning false if these conditions are not met.  Could be CHECK statements.

k2/csrc/fsa_equivalent.cc
qindazhu: mm..OK, I just added it and remove just before..

csukuangfj: How about replacing unnamed namespace with static ?

k2/csrc/fsa_equivalent.cc
qindazhu: Oh, yes, you're right, Thanks. I also changed the return type from `bool` to `void`. You can merge it if everything is fine.

PR #45, opened 2020-05-25
Add interface for ConstFsa

danpovey:


k2/csrc/fsa.h
```diff
@@ -122,6 +126,139 @@ struct Fsa {
...
+
+             [pad as needed for memory-alignment purposes then...]
+
+             - arcs[tot_arcs]
+  */
+  CfsaVec(size_t size, void *data);
+
+  int32_t NumFsas() { return num_fsas; }
+
+  Cfsa &&operator [] (int32_t f) const;
```
csukuangfj, line 247: Why does it return an `rvalue reference`?

k2/csrc/fsa.h
```diff
@@ -122,6 +126,139 @@ struct Fsa {
...
+             - arc_indexes[tot_states + 1]   This gives the indexes into the `arcs`
+                             array of where we can find the first of each state's arcs.
+
+             [pad as needed for memory-alignment purposes then...]
+
+             - arcs[tot_arcs]
+  */
+  CfsaVec(size_t size, void *data);
+
+  int32_t NumFsas() { return num_fsas; }
```
csukuangfj, line 245: `int32_t NumFsas() **const**`

k2/csrc/fsa.h
```diff
@@ -122,6 +126,139 @@ struct Fsa {
...
+
+  Cfsa &&operator [] (int32_t f) const;
+ private:
+  // TODO: fix names by adding underscores?
+  // Note: we will already have checked `size` to make sure
+  int32_t num_fsas;
+
+  // The raw underlying data
+  int32_t *data;
```
csukuangfj, line 254: Should we disable its copy constructor and assignment operator?

k2/csrc/fsa.h
```diff
@@ -122,6 +126,139 @@ struct Fsa {
...
+  Arc *arcs;   // Note: arcs[BeginArcIndex()] through arcs[EndArcIndex() - 1]
+               // are valid.
+
+  // Constructor from Fsa
+  Cfsa(const Fsa &fsa);
+
+  Cfsa &operator = (const Cfsa &cfsa) = default;
+  Cfsa(const Cfsa &cfsa) = default;
+
+  Cfsa(int32_t size, int32_t *data);
```
csukuangfj, line 168: Is the format of `data` the same as that of `CfsaVec`?

k2/csrc/fsa.h
```diff
@@ -122,6 +126,139 @@ struct Fsa {
...
+                             where the `state_offsets` array is, in int32_t
+                             (4-byte) elements.
+             - arc_indexes_start   The offset from the start of `data` of
+                             where the `arc_indexes` array is, in int32_t
+                             (4-byte) elements.
+             - arcs_start    The offset from the start of `data` of where
+                             the first Arc is, in sizeof(Arc) multiples, i.e.
+                             Arc *arcs = ((Arc*)data) + arcs_start
+
+            [possibly some padding here]
```
csukuangfj, line 225: Why would there be some padding here? 

Since all the above fields are `int32_t`, so if the first field is aligned, this array should
also be aligned in memory.

csukuangfj: Are the paddings for adding extra fields later?

k2/csrc/fsa.h
danpovey: I thought that was when you wanted to `move` the data?  So there aren't unnecessary constructors?  I may misunderstand.

k2/csrc/fsa.h
danpovey: will delete this constructor

k2/csrc/fsa.h
danpovey: On GPUs, I believe data alignment can sometimes be important.  And it might be helpful even on CPU to make sure these things match up with cache lines.

csukuangfj:
@danpovey 
Have you finished the interface? I would like to write the implementation.

danpovey:
Fantastic, and great timing!!!
Yes I think I have finished.
Let me push some other changes about DenseFsa, too.

danpovey:
BTW, I'm starting to wish we had some kind of Python interface.  You'll see
in ConstFsa how it's supposed to interact with Python.. we'll mostly be
dealing with FSAs that are actually stored as Tensors, and have functions
defined in C++ that allow us to invoke FSA algorithms in parallel over
these things defined as Tensors (i.e. a single memory region storing
multiple FSAs, linearized).

We can figure out later where to have the Python/C++ boundary though.  For
now we could wrap lots of stuff and whenever something becomes a
bottleneck, put it to C++.

csukuangfj:
@danpovey 
As for the Python interface, should we depend on `torch::Tensor` in c++ directly?

danpovey:
I'd rather not depend on torch::Tensor in our actual C++ code itself.
My plan was to arrange things so our C++ code can tell us how much memory
we need to allocate, then we allocate
a Tensor with the right number of elements and our C++ code writes to it.
We could probably use the buffer protocol to
get the shape and data-pointer at the C++-wrapping-to-python level.

danpovey:
.. look at GetCfsaVecSize() for an example.

csukuangfj:
thanks, I see.

csukuangfj:
There are conflicts now. Could you please resolve them and then merge?
I'll add implementation tomorrow.

csukuangfj:
Merging now. If there are any problems, they can be fixed later.

PR #46, opened 2020-05-27
add beam in RandEquivalent

qindazhu:


k2/csrc/fsa_equivalent.cc
```diff
@@ -199,6 +202,12 @@ bool IsRandEquivalent(const Fsa &a, const float *a_weights, const Fsa &b,
        (*(labels_difference.begin())) != kEpsilon))
     return false;
 
+  double best_distance = 0.0;
```
danpovey, line 205: I don't think you need all these if-statements.  Just let best_distance become -infinity in case beam is infinity.

qindazhu: just want to avoid to compute `ShortestDistance` over the whole FSA if users call this function without beam,

qindazhu: I can remove the second if in the loop, for the first one, what do you think?

k2/csrc/fsa_equivalent.cc
```diff
@@ -199,6 +202,12 @@ bool IsRandEquivalent(const Fsa &a, const float *a_weights, const Fsa &b,
        (*(labels_difference.begin())) != kEpsilon))
     return false;
 
+  double best_distance = kDoubleNegativeInfinity;
```
danpovey, line 205: I think distance_cutoff would be a name than best_distance.  Also these are not distances, they are loglikes (or negated cost), so actually loglike_cutoff would be better.

qindazhu: OK

k2/csrc/fsa_equivalent.h
```diff
@@ -32,6 +32,10 @@ bool IsRandEquivalent(const Fsa &a, const Fsa &b, std::size_t npath = 100);
   @param [in]  a_weights  Arc weights of `a`
   @param [in]  b          The other FSA to be checked the equivalence
   @param [in]  b_weights  Arc weights of `b`
+  @param [in]  beam       beam > 0 that affects pruning in `a`; the algorithm
+                          will keep paths that are within `beam` of the best
+                          path in `a` to check the equivalence. Just keep
+                          `kFloatInfinity` if you don't want pruning.
```
danpovey, line 38: so I assume that paths that are not within the beam don't have to be equivalent?  because thats not what the code currently does.

k2/csrc/fsa_equivalent.cc
```diff
@@ -222,6 +231,7 @@ bool IsRandEquivalent(const Fsa &a, const float *a_weights, const Fsa &b,
     CHECK(top_sorted);
     double sum_a =
         ShortestDistance<Type>(a_compose_path, a_compose_weights.data());
+    if (sum_a < best_distance) sum_a = kDoubleNegativeInfinity;
```
danpovey, line 234: Here, shouldn't you be doing `continue` instead of sum_a = kDoubleNegativeInfinity ?  Else, I don't see that this function does what it's supposed to do.

k2/csrc/fsa_equivalent.cc
```diff
@@ -199,6 +202,12 @@ bool IsRandEquivalent(const Fsa &a, const float *a_weights, const Fsa &b,
        (*(labels_difference.begin())) != kEpsilon))
     return false;
 
+  double best_distance = kDoubleNegativeInfinity;
+  if (beam != kFloatInfinity) {
+    best_distance =
+        ShortestDistance<Type>(valid_a, valid_a_weights.data()) - beam;
+  }
```
danpovey, line 209: I prefer if you set it to -infinity in the `else` statement of this, and just not initialize it when you declare it.  Would be clearer.

k2/csrc/fsa_equivalent.h
qindazhu: Yes, as we only keep those paths within `beam`, like we do `Prune(a, beam)` first in the algorithm.

k2/csrc/fsa_equivalent.h
```diff
@@ -32,6 +32,10 @@ bool IsRandEquivalent(const Fsa &a, const Fsa &b, std::size_t npath = 100);
...
   @param [in]  top_sorted If both `a` and `b` are topological sorted or not.
```
danpovey, line 39: Comment here is not very clear.  I'd say:

`The user may set this to true if both `a` and `b` are topologically sorted; this makes this
  function faster.  Otherwise it must be set to false.`

k2/csrc/fsa_equivalent.cc
qindazhu: what if `sum_b >  kDoubleNegativeInfinity`, then we should return false. As `a` has no such a path after pruning, but b has one.

k2/csrc/fsa_equivalent.h
qindazhu: OK, noted that now we don't implement the version for non-sorted fsa, so I will keep `CHECK(top_sorted)` in the cc file.

k2/csrc/fsa_equivalent.cc
qindazhu: Ok

danpovey:
Haowen, the logic here is not right because if you do some kind of pruned algorithm on a that will discard arcs that are not within `beam` of the best path, that does not imply that no paths exist that are within `beam` of the best path.  Think about a lattice with many confusable parts, one after another.

qindazhu:
yes, you are right, thanks! Changed title to wip and will work on it later

k2/csrc/properties.h
```diff
@@ -32,8 +32,7 @@ enum Properties {
   `fsa` is valid if:
   1. it is empty, if not, it contains at least two states.
   2. only kFinalSymbol arcs enter the final state.
-  3. every state contains at least one arc except the final state.
```
qindazhu, line 35: seems this is not a reasonable requirement (case we dont do `Connect` after some algorthims).

k2/csrc/fsa_equivalent.cc
```diff
@@ -169,8 +169,11 @@ bool IsRandEquivalent(const Fsa &a, const Fsa &b, std::size_t npath /*=100*/) {
 
 template <FbWeightType Type>
 bool IsRandEquivalent(const Fsa &a, const float *a_weights, const Fsa &b,
-                      const float *b_weights, bool top_sorted /*=true*/,
-                      std::size_t npath /*= 100*/) {
+                      const float *b_weights, float beam /*=kFloatInfinity*/,
```
qindazhu, line 172: will implement the version for RmEpsilonPrunedLogSum in another PR(after implement RmEpsilonPrunedLogSum)

k2/csrc/fsa_equivalent.cc
```diff
@@ -199,10 +202,25 @@ bool IsRandEquivalent(const Fsa &a, const float *a_weights, const Fsa &b,
...
+    loglike_cutoff_a =
+        ShortestDistance<Type>(valid_a, valid_a_weights.data()) - beam;
+    loglike_cutoff_b =
+        ShortestDistance<Type>(valid_b, valid_b_weights.data()) - beam;
+    if (Type == kMaxWeight &&
+        !DoubleApproxEqual(loglike_cutoff_a, loglike_cutoff_b))
+      return false;
+  } else {
+    loglike_cutoff_b = kDoubleNegativeInfinity;
+    loglike_cutoff_b = kDoubleNegativeInfinity;
```
danpovey, line 216: one should be a

k2/csrc/fsa_equivalent_test.cc
```diff
@@ -149,6 +149,19 @@ TEST(FsaEquivalent, IsWfsaRandEquivalent) {
                                                   c_weights.data());
     EXPECT_FALSE(status);
   }
+
+  // check equivalence with beam
+  {
+    bool status = IsRandEquivalent<kMaxWeight>(a, a_weights.data(), b,
+                                               b_weights.data(), 4);
```
danpovey, line 156: 4.0 ?  If this is the beam.  And why this value?

k2/csrc/fsa_equivalent.cc
qindazhu: oh,sorry. will changed

k2/csrc/fsa_equivalent.cc
danpovey: cool

k2/csrc/fsa_equivalent_test.cc
qindazhu: ok, will chaned to 4.0. It just a value I get for the testing FSA (will prune one path in the input fsa with this value

k2/csrc/properties.h
danpovey: You're right, that can be removed, it was an early draft.

danpovey: i will remove

k2/csrc/fsa_equivalent.cc
```diff
@@ -220,22 +238,43 @@ bool IsRandEquivalent(const Fsa &a, const float *a_weights, const Fsa &b,
...
+      if (Type == kLogSumWeight) {
+        loglike_path_a =
+            ShortestDistance<Type>(a_compose_path, a_compose_weights.data());
+        loglike_path_b =
+            ShortestDistance<Type>(b_compose_path, b_compose_weights.data());
+      } else {
+        loglike_path_a = max_loglike_path_a;
+        loglike_path_b = max_loglike_path_b;
+      }
+      if (!DoubleApproxEqual(loglike_path_a, loglike_path_b)) return false;
```
danpovey, line 261: we might want the option to pass the delta into this function.

k2/csrc/fsa_equivalent.cc
```diff
@@ -220,22 +238,43 @@ bool IsRandEquivalent(const Fsa &a, const float *a_weights, const Fsa &b,
     // find out that we don't need that version, we will remove flag
     // `top_sorted` and add requirements as comments in the header file.
     CHECK(top_sorted);
-    double sum_a =
-        ShortestDistance<Type>(a_compose_path, a_compose_weights.data());
-    double sum_b =
-        ShortestDistance<Type>(b_compose_path, b_compose_weights.data());
-    if (!DoubleApproxEqual(sum_a, sum_b)) return false;
+    double max_loglike_path_a =
```
danpovey, line 241: It's the `ShortestDistance<Type>` loglike that I would have thought you would be using here.. at least, for testing the determinization algorithm's equivalence.  Different algorithms may require slightly different tests though.
Mostly we'll be using the kMaxWeight option anyway.

qindazhu: why do we need kLogSum here for determinization? as we are trying to know if there is at least one path with beam for symbol sequence `path` (I thought LogSum tells that there is such a path but actually not)

k2/csrc/fsa_equivalent.cc
qindazhu: ok, will add one.

k2/csrc/fsa_equivalent.cc
danpovey: I mean, if we are testing equivalence for purposes of testing the correctness of determinization that has been done with log semiring, what we care about here is whether the log-sum versions of the shortest-path is better than the beam.

PR #48, opened 2020-05-30
Implement RandomEquivalent for RmEpsilonPrunedLogSum

qindazhu:


k2/csrc/fsa_equivalent.h
```diff
@@ -60,6 +60,49 @@ bool IsRandEquivalent(const Fsa &a, const float *a_weights, const Fsa &b,
                       float delta = 1e-6, bool top_sorted = true,
                       std::size_t npath = 100);
 
+/*
```
qindazhu, line 63: Put this function in file `fsa_equivalent.h` as:
- It may be easy for user to check these similar `RandEquivalent` functions together.
- We may need to refactor those functions to extract some common sub-functions (such as ::GetArcWeights that should not be in k2 namespace), it's easy to call those sub-functions if we put these `RandEquivalent` functions in one file.

danpovey: cool.

danpovey:
I don't have time for a thorough review of this, we'll clean up any issues if we encounter bugs on testing.  You guys can merge when you're comfortable; note sur whether Fangjun wants to review.

k2/csrc/fsa_algo.cc
```diff
@@ -52,6 +52,34 @@ inline int32_t InsertIntersectionState(
   return result.first->second;
 }
 
+/**
+   A TraceBack() function used in RmEpsilonPrunedLogSum.  It finds derivative
```
csukuangfj, line 56: The function name is `RmEpsilon **s** PrunedLogSum`.

csukuangfj:
+2

k2/csrc/fsa_algo.cc
qindazhu: Oh, yes, updated, thanks!

qindazhu:
Ready to merge now.

PR #49, opened 2020-05-31
Construct a CfsaVec from torch::Tensor via DLPack

csukuangfj:
~~Still working in progress.~~ Any comments are welcome.

TODOS:
- [x] add more tests
- [x] add python wrapper for FsaVec
- [x] add Python wrapper for cfsa and support `torch::Tensor`

---

Our current plan is to share memory with `torch::Tensor` with
DLPack (https://github.com/dmlc/dlpack)


danpovey:
Fantastic!  Haven't looked at it in detail but what I saw looks very plausible.  Thanks!

k2/csrc/fsa.cc
```diff
@@ -0,0 +1,250 @@
...
+  num_fsas_ = header->num_fsas;
+}
+
+Cfsa CfsaVec::operator[](int32_t f) const {
+  DCHECK_GE(f, 0);
+  DCHECK_LT(f, num_fsas_);
+
+  Cfsa cfsa;
+
+  const auto header = reinterpret_cast<const CfsaVecHeader *>(data_);
```
qindazhu, line 79: I guess it would be efficient if we cache `header` in the constructor as we may cast `data` to get `header` in `operator[](f)` again and again.

k2/csrc/fsa.cc
```diff
@@ -0,0 +1,250 @@
...
+    num_states += cfsas[0].NumStates();
+  }
+
+  offset += sizeof(int32_t) * (cfsas.size() + 1);
+  offset = AlignTo(offset, kAlignment);
+  header->arc_indexes_start = offset / sizeof(int32_t);
+
+  auto arc_indexes_array =
+      reinterpret_cast<int32_t *>(data) + header->arc_indexes_start;
+  offset += sizeof(int32_t) * (num_states + cfsas.size());
```
qindazhu, line 221: it would be greate if we can add comments as you do in `GetCfsaVecSize(const Cfsa &cfsa)` to clarify the purpose when updating `offset` each time, as well the `+csfsa.size()` trick (as we add a duplicate final state)

k2/csrc/fsa.cc
```diff
@@ -0,0 +1,250 @@
...
+
+  if (!cfsas.empty()) {
+    if (cfsas.back().NumStates() != 0) {
+      state_offsets_array[i] =
+          state_offsets_array[i - 1] + cfsas.back().NumStates() + 1;
+    } else {
+      state_offsets_array[i] = state_offsets_array[i - 1];
+    }
+
+    num_states += cfsas[0].NumStates();
```
qindazhu, line 212: it's strange to add `cfsas[0].NumStates()` here,  maybe we can do this before the for-loop just above.

k2/csrc/fsa.cc
```diff
@@ -0,0 +1,250 @@
...
+      state_offsets_array[i] =
+          state_offsets_array[i - 1] + cfsas[i - 1].NumStates() + 1;
+    } else {
+      state_offsets_array[i] = state_offsets_array[i - 1];
+    }
+
+    num_states += cfsas[i].NumStates();
+  }
+
+  if (!cfsas.empty()) {
```
qindazhu, line 204: It would be clearer we can put some default value in `data` and return in advance if `cfsas` is empty.

k2/csrc/fsa.cc
```diff
@@ -0,0 +1,250 @@
...
+    if (cfsas.back().NumStates() != 0) {
```
qindazhu, line 205: Then we can merge below inlines into the for-loop with `i  != cfsas.size()`. (Of course add a check when updating `num_states`)

k2/csrc/fsa.h
```diff
@@ -50,13 +50,20 @@ struct Arc {
...
     // compares `label` first, then `dest_state`
     return std::tie(label, dest_state) <
            std::tie(other.label, other.dest_state);
   }
 };
 
+std::ostream &operator<<(std::ostream &os, const Arc &arc);
+
+template <class T>
+void hash_combine(std::size_t *seed, const T &v);
```
qindazhu, line 65: why do we need to declare `hash_combine` here again? I thought I put one inline version in `util.h`

k2/csrc/fsa.h
```diff
@@ -159,18 +166,50 @@ struct Cfsa {
...
+  explicit Cfsa(const Fsa &fsa);
 
   Cfsa &operator = (const Cfsa &cfsa) = default;
   Cfsa(const Cfsa &cfsa) = default;
 
   int32_t NumStates() const { return num_states; }
-  int32_t FinalState() const { return num_states - 1; }
+  int32_t NumArcs()  const { return end_arc - begin_arc; }
+  int32_t FinalState() const {
+    DCHECK_GE(num_states, 2) << "It's an error to invoke this method for "
```
qindazhu, line 180: I guess using `CHECK_GE` instead of `DCHECK` is better here, as users usually would not call `FinalState` too many times on one cfsa, there is no efficiency issue here (but release check will make sure the program can stop early when there's error)

k2/csrc/fsa_test.cc
```diff
@@ -0,0 +1,316 @@
...
+      1 3 3
+      3 10 -1
+      10
+  )";
+  auto fsa2 = StringToFsa(s2);  // 4 arcs, 11 states
+  Cfsa cfsa2(*fsa2);
+
+  std::vector<Cfsa> cfsa_vec = {cfsa1, cfsa2};
+
+  size_t bytes = GetCfsaVecSize(cfsa_vec);
```
qindazhu, line 125: `std::size_t` as `size_t` is in std namespace

qindazhu: you may need to check other usage of `size_t` as well.

k2/csrc/fsa_test.cc
```diff
@@ -0,0 +1,316 @@
...
+  // 4-byte padding             (252) -> to be multiple of sizeof(Arc)
+  // 108-byte arcs_array        (360)
+  EXPECT_EQ(bytes, 360u);
+}
+
+TEST(CfsaVec, Empty) {
+  Cfsa cfsa;
+  std::vector<Cfsa> cfsas;
+  size_t bytes = GetCfsaVecSize(cfsas);
+  std::unique_ptr<void, decltype(&MemFree)> data(MemAlign(bytes, 64), &MemFree);
```
qindazhu, line 142: why do we need align here?  I just thought we have no requirement on the end of data (to be aligned).

k2/csrc/fsa_test.cc
```diff
@@ -0,0 +1,316 @@
...
+
+  CreateCfsaVec(cfsas, data.get(), bytes);
+
+  CfsaVec cfsa_vec(bytes / sizeof(int32_t), data.get());
+  EXPECT_EQ(cfsa_vec.NumFsas(), 1);
+}
+
+TEST(CfsaVec, OneNoneEmptyCfsa) {
```
qindazhu, line 162: `One**Non**EmptyCfsa`?

k2/csrc/fsa_test.cc
```diff
@@ -0,0 +1,316 @@
...
+    size_t bytes = GetCfsaVecSize(cfsas);
+    std::unique_ptr<void, decltype(&MemFree)> data(MemAlign(bytes, 64),
+                                                   &MemFree);
+
+    CreateCfsaVec(cfsas, data.get(), bytes);
+
+    CfsaVec cfsa_vec(bytes / sizeof(int32_t), data.get());
+    EXPECT_EQ(cfsa_vec.NumFsas(), 2);
+
+    Cfsa f = cfsa_vec[0];
```
qindazhu, line 251: using const reference instead of copy assignment? (even if the copy is cheap for Cfsa)

k2/csrc/util.cc
```diff
@@ -0,0 +1,38 @@
...
+
+#include "k2/csrc/util.h"
+
+#include <stdlib.h>
+
+#include "glog/logging.h"
+
+namespace k2 {
+
+void *MemAlign(size_t nbytes, size_t alignment) {
```
qindazhu, line 15: I prefer name `MemAlign(ed)Malloc` as posix naming style is just like: we only do (padding) alignment..

k2/csrc/util.h
```diff
@@ -101,5 +102,8 @@ inline bool DoubleApproxEqual(double a, double b, double delta = 1e-6) {
   return a <= b + delta && b <= a + delta;
 }
 
+void *MemAlign(size_t nbytes, size_t alignment);
```
qindazhu, line 105: std::size_t

k2/csrc/fsa.cc
qindazhu, line 79: I guess it would be efficient if we cache `header` (and `arcs_array`) in the constructor as we actually do this in `operator[](f)` again and again if user access cfsa in `CfsaVec` one by one, even if `reinterpret_cast` has no runtime cost for pointers (but we excute those statements again and again)

k2/csrc/fsa_test.cc
danpovey: There is no real requirement for alignment.  It's just a nice-to-have as in some cases it might improve efficiency.  For test code it's probably not necessary, but also doesn't hurt.  We can assume that when we allocate tensors from things like PyTorch, it will be memory aligned to whatever boundaries the hardware cares about.

k2/csrc/fsa_test.cc
danpovey: cfsa_vec operator [] doesn't return a reference (last I checked, anyway).  Hopefully compiler will inline.

k2/csrc/fsa_test.cc
qindazhu: OK, then I suggest we add what you said as comments in the test code here if we want to keep the alignment code (so the readers will not misunderstand when reading the test code here)

danpovey:
sure..  hopefully one of you guys can do it.

k2/csrc/fsa_test.cc
csukuangfj: Not sure what will happen if the returned pointer is not aligned to `alignof(CfsaVecHeader)`
and `alignof(Arc)` when `MemAlign` is not used.

k2/csrc/fsa.h
csukuangfj: Consider what will happen while compiling `util.cc`:

(1) It includes `k2/csrc/util.h`, which defines a header guard to avoid including itself multiple times
(2) `k2/csrc/util.h` includes `k2/csrc/fsa.h`
(3) `k2/csrc/fsa.h` includes `k2/csrc/util.h`, which is an no-op because of
the header guard of `k2/csrc/util.h`, which means `hash_combine` is not visible
in `k2/csrc/fsa.h`

csukuangfj: There are circular dependencies between `k2/csrc/util.h` and `k2/csrc/fsa.h`,
which indicates a bad design.

qindazhu: It may be a history issue to include `fsa.h` in `util.h`. Actually we can remove it from `util.h` as it seems there's no dependency on FSA in `util.h`

csukuangfj:
I will resolve all comments tomorrow night.

danpovey:
Thanks!!!

k2/csrc/fsa.cc
csukuangfj: fixed.

k2/csrc/fsa.cc
csukuangfj: `cfsas[0]` is not defined if `cfsas` is empty. That's the reason it is placed here.

k2/csrc/fsa.cc
csukuangfj: `data` is allocated outside of `k2`, e.g., from `torch::Tensor`, which is assumed
to contain garbage content once allocated. Not sure what default values
should be placed into it.

I think it does not cost much if the input vector is empty.

k2/csrc/fsa.cc
csukuangfj: It is placed here because I want to reduce the number of branches inside the `for` loop.

k2/csrc/fsa.h
csukuangfj: fixed.

k2/csrc/fsa.h
csukuangfj: fixed.

k2/csrc/fsa_test.cc
csukuangfj: fixed.

k2/csrc/fsa_test.cc
csukuangfj: fixed

k2/csrc/fsa_test.cc
csukuangfj: mark as resolved.

k2/csrc/util.cc
csukuangfj: Done.

k2/csrc/util.h
csukuangfj: fixed.

k2/csrc/fsa.cc
csukuangfj: I guess the optimizer will optimize it out.

Change it to a data field will occupy extra memory (sizeof a pointer).

csukuangfj:
@qindazhu @danpovey 
ready to merge, please have a review.

Now we can construct a `CfsaVec` from a `torch::Tensor`; the memory
is shared with `torch::Tensor`.

Please refer to
https://github.com/danpovey/k2/blob/a9ef9534ebb96f28e4e2f3de3faf31b47c4a3656/k2/python/tests/fsa_test.py#L174-L184 for example usage.

csukuangfj:
I've just disabled Windows in CI since I have not figured out how to install PyTorch in Windows.

qindazhu:
Great!! I guess we can merge it now and fix other small issues/TODOs in the future.

k2/csrc/fsa.cc
csukuangfj: The doc of this function says the input vector`cfsas` MUST not be empty.

I will fix the implementation later.

danpovey:
Fantastic!  You guys can merge if you think it's ready.

csukuangfj:
merging.

PR #50, opened 2020-06-01
implement MaxAuxLabels1(2)

qindazhu:
Changes in `dense_fsa.h` and `fsa.h` are only for passing style check.

### DIFF
--- a/k2/csrc/aux_labels.cc
+++ b/k2/csrc/aux_labels.cc
@@ -20,19 +20,19 @@ void MapAuxLabels1(const AuxLabels &labels_in,
   auto &start_pos = labels_out->start_pos;
   auto &labels = labels_out->labels;
   start_pos.clear();
+  start_pos.reserve(arc_map.size() + 1);
   labels.clear();
 
   int32_t num_labels = 0;
+  auto labels_in_iter_begin = labels_in.labels.begin();
   for (const auto &arc_index : arc_map) {
     start_pos.push_back(num_labels);
     int32_t pos_start = labels_in.start_pos[arc_index];
     int32_t pos_end = labels_in.start_pos[arc_index + 1];
-    for (int32_t pos = pos_start; pos != pos_end; ++pos) {
-      int32_t label = labels_in.labels[pos];
-      DCHECK_NE(label, kEpsilon);
-      labels.push_back(label);
-      ++num_labels;
-    }
+    // TODO(haowen): should we check labels contains no Epsilon?
+    labels.insert(labels.end(), labels_in_iter_begin + pos_start,
+                  labels_in_iter_begin + pos_end);
+    num_labels += pos_end - pos_start;
   }
   start_pos.push_back(num_labels);
 }
@@ -44,20 +44,19 @@ void MapAuxLabels2(const AuxLabels &labels_in,
   auto &start_pos = labels_out->start_pos;
   auto &labels = labels_out->labels;
   start_pos.clear();
+  start_pos.reserve(arc_map.size() + 1);
   labels.clear();
 
   int32_t num_labels = 0;
+  auto labels_in_iter_begin = labels_in.labels.begin();
   for (const auto &arc_indexes : arc_map) {
     start_pos.push_back(num_labels);
     for (const auto &arc_index : arc_indexes) {
       int32_t pos_start = labels_in.start_pos[arc_index];
       int32_t pos_end = labels_in.start_pos[arc_index + 1];
-      for (int32_t pos = pos_start; pos != pos_end; ++pos) {
-        int32_t label = labels_in.labels[pos];
-        DCHECK_NE(label, kEpsilon);
-        labels.push_back(label);
-        ++num_labels;
-      }
+      labels.insert(labels.end(), labels_in_iter_begin + pos_start,
+                    labels_in_iter_begin + pos_end);
+      num_labels += pos_end - pos_start;
     }
   }
   start_pos.push_back(num_labels);
--- a/k2/csrc/aux_labels.cc
+++ b/k2/csrc/aux_labels.cc
@@ -1,0 +1,66 @@
+// k2/csrc/aux_labels.cc
+
+// Copyright (c)  2020  Xiaomi Corporation (author: Haowen Qiu)
+
+// See ../../LICENSE for clarification regarding multiple authors
+
+#include "k2/csrc/aux_labels.h"
+
+#include <numeric>
+#include <vector>
+
+#include "glog/logging.h"
+#include "k2/csrc/fsa.h"
+
+namespace k2 {
+
+void MapAuxLabels1(const AuxLabels &labels_in,
+                   const std::vector<int32_t> &arc_map, AuxLabels *labels_out) {
+  CHECK_NOTNULL(labels_out);
+  auto &start_pos = labels_out->start_pos;
+  auto &labels = labels_out->labels;
+  start_pos.clear();
+  labels.clear();
+
+  int32_t num_labels = 0;
+  for (const auto &arc_index : arc_map) {
+    start_pos.push_back(num_labels);
+    int32_t pos_start = labels_in.start_pos[arc_index];
+    int32_t pos_end = labels_in.start_pos[arc_index + 1];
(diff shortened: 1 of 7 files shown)


danpovey:
Great!  Sorry dont have time for detailed review...

qindazhu:
OK, let's wait for @csukuangfj 's review.

k2/csrc/aux_labels.cc
```diff
@@ -0,0 +1,66 @@
...
+namespace k2 {
+
+void MapAuxLabels1(const AuxLabels &labels_in,
+                   const std::vector<int32_t> &arc_map, AuxLabels *labels_out) {
+  CHECK_NOTNULL(labels_out);
+  auto &start_pos = labels_out->start_pos;
+  auto &labels = labels_out->labels;
+  start_pos.clear();
+  labels.clear();
+
```
csukuangfj, line 24:
```cpp
start_pos.reserve(arc_map.size() + 1);
labels.reserve(labels_in.labels.size());
```

Inside the `for` loop, it would be better to use

```cpp
labels.insert(labels.end(), labels_in.start_pos.begin() + arc_index,
                            labels_in.start_pos.begin() + arc_index + 1);
```

We can assume that none of the labels are epsilons.

k2/csrc/aux_labels.cc
```diff
@@ -0,0 +1,66 @@
...
+  int32_t num_labels = 0;
+  for (const auto &arc_index : arc_map) {
+    start_pos.push_back(num_labels);
+    int32_t pos_start = labels_in.start_pos[arc_index];
+    int32_t pos_end = labels_in.start_pos[arc_index + 1];
```
csukuangfj, line 29: `num_labels += pos_end - pos_start;`

Change `num_labels` outside of the `for` loop.

k2/csrc/aux_labels.cc
```diff
@@ -0,0 +1,66 @@
...
+}
+
+void MapAuxLabels2(const AuxLabels &labels_in,
+                   const std::vector<std::vector<int32_t>> &arc_map,
+                   AuxLabels *labels_out) {
+  CHECK_NOTNULL(labels_out);
+  auto &start_pos = labels_out->start_pos;
+  auto &labels = labels_out->labels;
+  start_pos.clear();
+  labels.clear();
```
csukuangfj, line 47: Please pre-allocate space for `start_pos` and `labels`.

k2/csrc/aux_labels.cc
```diff
@@ -0,0 +1,66 @@
...
+
+  int32_t num_labels = 0;
+  for (const auto &arc_indexes : arc_map) {
+    start_pos.push_back(num_labels);
+    for (const auto &arc_index : arc_indexes) {
+      int32_t pos_start = labels_in.start_pos[arc_index];
+      int32_t pos_end = labels_in.start_pos[arc_index + 1];
+      for (int32_t pos = pos_start; pos != pos_end; ++pos) {
+        int32_t label = labels_in.labels[pos];
```
csukuangfj, line 56: The `insert` method of `std::vector` might be more efficient.

k2/csrc/aux_labels.cc
csukuangfj: We can invoke `labels.resize(num_labels)` after the `for` loop in case
some of the arcs are removed from the output fsa.

k2/csrc/aux_labels_test.cc
```diff
@@ -0,0 +1,86 @@
...
+    aux_labels_out.labels = {4, 5};
+    MapAuxLabels1(aux_labels_in_, arc_map, &aux_labels_out);
+
+    EXPECT_TRUE(aux_labels_out.labels.empty());
+    EXPECT_EQ(aux_labels_out.start_pos.size(), 1);
+    EXPECT_EQ(aux_labels_out.start_pos[0], 0);
+  }
+
+  {
+    std::vector<int32_t> arc_map = {2, 0, 3};
```
csukuangfj, line 46: Another case might be that no arcs are removed from the output fsa.
For example, the `TopSort` does not reduce the number of arcs.

k2/csrc/aux_labels_test.cc
```diff
@@ -0,0 +1,86 @@
...
+    AuxLabels aux_labels_out;
+    MapAuxLabels1(aux_labels_in_, arc_map, &aux_labels_out);
+
+    EXPECT_EQ(aux_labels_out.start_pos.size(), 4);
```
csukuangfj, line 50: `ASSERT_EQ(aux_labels_out.start_pos.size(), 4);`

Same for the following `aux_labels_out`

k2/csrc/dense_fsa.h
```diff
@@ -53,30 +52,31 @@ struct DenseFsa {
    */
   int32_t num_states() { return T + 2; }
```
csukuangfj, line 53: The method names of this `struct` do not follow Google C++ style.

Some methods should have a `const` modifier, such as `int32_t NumStates() const`.

k2/csrc/fsa.h
```diff
@@ -219,11 +213,11 @@ class CfsaVec {
 
   int32_t NumFsas() const { return num_fsas_; }
 
-  Cfsa operator[] (int32_t f) const;
+  Cfsa operator[](int32_t f) const;
 
  private:
-  CfsaVec &operator = (const CfsaVec &);  // Disable
-  CfsaVec(const CfsaVec&);  // Disable
+  CfsaVec &operator=(const CfsaVec &);  // Disable
```
csukuangfj, line 219: The modern style with `=delete` is more readable.

```cpp
public:
  CfsaVec &operator=(const CfsaVec &) = delete;

  CfsaVec(const CfsaVec &) = delete;
```

`public` is to suppress the warnings from clang-tidy.


----

I've changed this in my pullrequest.

qindazhu: OK, thanks (I'll reslove it here)!

k2/csrc/dense_fsa.h
qindazhu: Sure

k2/csrc/aux_labels.cc
qindazhu: Sure, thanks!

k2/csrc/aux_labels.cc
qindazhu: We cannot actually pre-allocate here if we have not iterated over `arc_map`, as we don't know how many arcs in it.

csukuangfj: The maximum number of arcs is `labels_in.labels.size()`.
We can resize it at the end if extra memory is allocated.

k2/csrc/aux_labels.cc
qindazhu: will remove `DCHECK_NE(label, kEpsilon)` and use insert, I prefer to only reserve for start_pos here (we may always allocate too much memory (even if we resize at last) for most of FSA algorithms).

k2/csrc/aux_labels_test.cc
qindazhu: Added

danpovey:
RE assumptions about epsilons etc.: make sure to document any assumptions.
There may be certain situations where we will want to treat epsilons as
normal symbols.
Important thing now is to make sure everything is clearly documented.

qindazhu:
It is documented in .h file, @csukuangfj and I were talking about if we should check labels!=kEpsilon in the implementation (DCHECK).

danpovey:
I may have changed my mind.. my feeling is if we can treat epsilon the same
as other symbols, and document so, it would be OK.
I have realized there are scenarios where epsilon is best treated as a
normal symbol.

qindazhu:
OK, I will remove the requirement (labels != kEpsilon) in `.h` file.

qindazhu:
Ready to merge.

csukuangfj:
+2

k2/csrc/aux_labels.cc
qindazhu: actually it's not (think about for log-sum version for most of algortihms, there may be many duplicate arcs in arc_derivs_out)

csukuangfj:
Merging.

PR #52, opened 2020-06-04
implement invertFst

qindazhu:


k2/csrc/aux_labels.cc
```diff
@@ -61,4 +136,93 @@ void MapAuxLabels2(const AuxLabels &labels_in,
...
+
+  int32_t num_non_eps_ilabel_processed = 0;
+  start_pos.push_back(0);
+  for (auto i = 0; i != fsa_in.arcs.size(); ++i) {
+    const auto &arc = fsa_in.arcs[i];
+    int32_t pos_start = labels_in.start_pos[i];
+    int32_t pos_end = labels_in.start_pos[i + 1];
+    int32_t src_state = arc.src_state;
+    int32_t dest_state = arc.dest_state;
+    if (dest_state == final_state_in) {
```
qindazhu, line 183: Seems we need to create a function `IsValid` for `AuxLabels`. We need to check if `start_pos` and `labels` in this state are consistent and any arc entering final state should have an olabel with kFinalSymbol

danpovey: We'll have to decide whether the kFinalSymbol should be swapped, or if we're OK with the aux label there being 0 (and just not swap that one).  I'm not sure right now.  Lean towards 0.

k2/csrc/aux_labels.cc
```diff
@@ -6,14 +6,89 @@
...
+  CHECK_EQ(expanded_states[0], 0);
+  CHECK_EQ(s_map[0], 0);
+  CHECK_EQ(s_ids[0], 0);
+  auto num_states_in = expanded_states.size();
+  // process from the second state
+  int32_t num_states_out = 0;
+  for (auto i = 1; i != num_states_in; ++i) {
+    s_ids[i] = num_states_out;
+    // `+1` as we did not count state `i` itself in `expanded_states`
+    num_states_out += expanded_states[i] + 1;
```
csukuangfj, line 77:
```
states[arc.dest_state] += std::max(0, pos_end - pos_start - 1);
```

If `pos_end - pos_start - 1` is `-1`, then 
```
num_states_out += expanded_states[i] + 1;
```
an extra state is added to `num_states_out`.

Maybe it should be
```
num_states_out += expanded_states[i] == 1 ? 1 : expanded_states + 1;
```

k2/csrc/aux_labels.cc
qindazhu: If olable is `-1 = kFinalSymbol`, we can just treat `-1` as normal symbol when inverting (remove line 183-185 above), while we always need to check if `dest_state == final_state_in` if olabel of such arc is `0` as we need to add `-1` as the ilabel of the output arc.

k2/csrc/aux_labels.cc
```diff
@@ -6,14 +6,89 @@
...
+
+/*
+  Map the state in the input FSA to state in the output inverted FSA.
+
+  @param [in] expanded_states  Output of function `CountExpandedStatesNum`
+                               which gives how many extra states we need
+                               to create for each state in the input FSA.
+  @param [out] state_map       Map state `i` in the input FSA to state
+                               `state_map[i]` in the output FSA.
+  @param [in] state_ids        state_id[i] == state_map[i-1], it will be
```
danpovey, line 54: This documentation is not clear.  If `state_id[i]` were equal `state_map[i-1]`, why is there a need for this?

k2/csrc/aux_labels.cc
```diff
@@ -6,14 +6,89 @@
...
+    int32_t pos_start = labels_in.start_pos[i];
+    int32_t pos_end = labels_in.start_pos[i + 1];
+    states[arc.dest_state] += std::max(0, pos_end - pos_start - 1);
+  }
+}
+
+/*
+  Map the state in the input FSA to state in the output inverted FSA.
+
+  @param [in] expanded_states  Output of function `CountExpandedStatesNum`
```
danpovey, line 49: The name num_extra_states would be clearer than `expanded_states`.
Also I think CountExtraStates() would be  better for that other function.

k2/csrc/aux_labels.h
```diff
@@ -46,6 +46,9 @@ struct AuxLabels {
   std::vector<int32_t> labels;
 };
 
+// swap AuxLabels
+void Swap(AuxLabels *labels1, AuxLabels *labels2);
```
danpovey, line 50: clarify that's a shallow swap

k2/csrc/aux_labels.h
```diff
@@ -46,6 +46,9 @@ struct AuxLabels {
   std::vector<int32_t> labels;
 };
```
danpovey, line 47: Note: when we integrate the Array stuff, we would have `using AuxLabels = Array2<int32_t, int32_t>` or something like that.
Does one of you have time to take what I sketched out, take into account you guys' comments (you found some small issues) and start switching stuff over?

danpovey: ... just using Array for AuxLabels, and not for the other stuff, may be a good way to start that is not too overwhelming.

k2/csrc/aux_labels.cc
```diff
@@ -61,4 +136,93 @@ void MapAuxLabels2(const AuxLabels &labels_in,
   start_pos.push_back(num_labels);
 }
 
+void InvertFst(const Fsa &fsa_in, const AuxLabels &labels_in, Fsa *fsa_out,
+               AuxLabels *aux_labels_out) {
```
danpovey, line 140: Obviously for output AuxLabels we'll have to have a different data structure as Array2 and the like do not own their data.  What I am thinking is that these kinds of algorithms could be wrapped as a class.  It would have a constructor which would do the work, a function called size() that returns the number of bytes (or elements?) required, and a function called `output(void *data ? int32_t *data?, size_t data_size /*for checking*/, Fs *fsa_out, AuxLabels *aux_labels_out)`.
At least that's an initial proposal.  LMK your thoughts.
Eventually we'll be using Tensors for memory, so we need to think about this now.

k2/csrc/aux_labels.cc
qindazhu: `state_map` will be not changed after the function call, while the `state_id[i]` will be increaed by one each time when we number extra states created for state `i`, to get different state index for those extra states.
 Will add comments.

k2/csrc/aux_labels.cc
qindazhu: Sure

qindazhu:
Ready to merge, add I will replace `AuxLabel` with `Array2` later based on this.

k2/csrc/aux_labels.cc
danpovey: If it will be changed then you should mark it [in,out].  Also, when you say `state_id[i] == state_map[i-1],`, is that true at entry or at exit?

danpovey: .. if you talk about this being changed *after* this function call (i.e. outside the function, by calling code), then that probably doesn't belong in this comment.  If you mean being changed *by* this function call, you need to be very clear about the values at entry and exit.

k2/csrc/aux_labels.cc
```diff
@@ -6,14 +6,95 @@
...
+}
+
+/*
+  Map the state in the input FSA to state in the output inverted FSA.
+
+  @param [in] expanded_states  Output of function `CountExpandedStatesNum`
+                               which gives how many extra states we need
+                               to create for each state in the input FSA.
+  @param [out] state_map       Map state `i` in the input FSA to state
+                               `state_map[i]` in the output FSA.
```
danpovey, line 53: .. if each state in the input FSA corresponds to a chain of several states in the output FSA, you need to clarify which member of the chain `state_map` will be set to.  (and are those elements of the chain consecutively numbered?)

k2/csrc/aux_labels.cc
qindazhu: what about we wrapper each Array type with a template class, e.g.
```
class Array2Storage: Array2{
   private:
         Ptr* data_storage; // allocate memory here for `data` in `Array2`
}
```
and define `AuxLabel = Array2<int32_t, int32_t> *` which is a pointer, then we can pass `&Array2` or `&Array2Storage`

k2/csrc/aux_labels.cc
qindazhu: Updated

k2/csrc/aux_labels.cc
qindazhu: I will set `state_id[0]` to 0, to it will just be `[out]`.

k2/csrc/aux_labels.cc
```diff
@@ -6,14 +6,99 @@
...
+
+/*
+  Map the state in the input FSA to state in the output inverted FSA.
+
+  @param [in] num_extra_states Output of function `CountExtraStates`
+                               which gives how many extra states we need
+                               to create for each state in the input FSA.
+  @param [out] state_map       Map state `i` in the input FSA to state
+                               `state_map[i]` in the output FSA.
+                               Generally, if there are n extra states we
```
danpovey, line 54: I think this extra text just confuses the issue.  Less is more.

qindazhu: OK, thanks!

qindazhu:
merging, as I will rework `invert` based on Dan's pull request about `Array`. just leave your comments and we can fix any issues later.

PR #55, opened 2020-06-09
support StridedPtr in Array2 and add some tests

qindazhu:


k2/csrc/array.h
```diff
@@ -52,70 +56,150 @@ struct Array2Size {
...
-  // `size2` is the nunber of elements in the array, equal to
+  // `size2` is the number of elements in the array, equal to
   // o->indexes[o->size] - o->indexes[0] (if the Array2 object o is
   // initialized).
   I size2;
 };
 
-template <typename I, typename Ptr>
+template <typename Ptr, typename I = int32_t>
 struct Array2 {
```
qindazhu, line 66: @csukuangfj I don't define any constructor or `Init` method in `Array2`, you can merge yours if you think it will be convenient in Python code.

k2/csrc/array.h
```diff
@@ -52,70 +56,150 @@ struct Array2Size {
...
-
-        An Array2 object is initialized if its `size` member is set and its
-        `indexes` and `data` pointer allocated, and the values of its `indexes`
-        array are set for indexes[0] and indexes[size].
+  /*
+        An Array2 object is non-empty if its `size` member is set and its
+        `indexes` and `data` pointer allocated,
   */
+  bool Empty() const {
+    return size == 0 || indexes == nullptr || data == nullptr;
```
danpovey, line 90: The 'initialized' definition was actually needed, and an Array2 object can be initialized if it has size == 0 and indexes== pointer to e.g. [0], and data == nullptr.  It would just be empty in the sense of no elements.  The point of `initialized` is that it is set up.
The reason for having a definition is that some functions may require the Array2 object to be initialized (and some not).
Maybe just keep that comment?
Empty() would just return {size == 0}.

k2/csrc/array.h
```diff
@@ -52,70 +56,150 @@ struct Array2Size {
...
+
+  void FillData(const std::vector<ValueType> &data) {
+    CHECK_EQ(data.size(), array2_size_.size2);
+    for (auto i = 0; i != array2_size_.size2; ++i) array2_.data[i] = data[i];
+  }
+
+  Array2<Ptr, I> &GetArray2() { return array2_; }
+
+ private:
+  Array2<Ptr, I> array2_;
```
danpovey, line 189: I'd just call these array_ and size_ ?

k2/csrc/array.h
qindazhu: OK, will get back the comment and update Empty()

k2/csrc/array.h
qindazhu: Sure (removed size_ as we now have `size2` in Array2)

danpovey:
Great! OK to merge, for me.

csukuangfj:
+2



Sent from myMail for iOS


Tuesday, 9 June 2020, 22:30 +0800 from notifications@github.com  <notifications@github.com>:

qindazhu:
Merging.

k2/csrc/array.h
```diff
@@ -1,45 +1,49 @@
...
-  IndexT end;
-
-  PtrT data;
+  T *data;   // it is NOT owned here
+  I stride;  // in number of elements, NOT number of bytes
+  T &operator[](I i) { return data[i * stride]; }
+  const T &operator[](I i) const { return data[i * stride]; }
+  StridedPtr(T *data = nullptr, I stride = 0)  // NOLINT
+      : data(data), stride(stride) {}
+  StridedPtr(const StridedPtr &other)
```
csukuangfj, line 36: Since `StridedPtr` is trivially constructible and copyable, I think
the default copy constructor and assignment operator provided
by the compiler are good enough and there is no need to implement
our own.

In the constructor `StridedPtr(T *data = nullptr, I stride = 0) `, we should check that

```
DCHECK_GT(stride, 1);
```

qindazhu: If we have defined a use-defined constructor, then the compiler will not generate other constructors.  Will add `DCHECK_GT`

k2/csrc/array.h
```diff
@@ -49,73 +53,154 @@ struct Array1 {
...
-  const IndexT *indexes;  // indexes[0,1,...size] should be defined; note, this
-                          // means the array must be of at least size+1.  We
-                          // require that indexes[i] <= indexes[i+1], but it is
-                          // not required that indexes[0] == 0, it may be
-                          // greater than 0.
-
+  using ValueType = typename std::iterator_traits<Ptr>::value_type;
+
+  IndexT size1;
+  IndexT *indexes;  // indexes[0,1,...size1] should be defined; note, this
```
csukuangfj, line 73: It'd be better if the definition order of `indexes` and `size2` is swapped.

On 64-bit machines where `sizeof(indexes) == 8`, there will be 4-byte padding
after `size1` for the current implementation.

k2/csrc/array.h
qindazhu: RE `DCHECK_GT`, as we will declare `PtrT data` without parameter in Array2, or we may write `data = nullptr`, we actually cannot require `stride > 1`.

csukuangfj: > If we have defined a use-defined constructor, then the compiler will not generate other constructors.

Could you provide some sources for the above comment? I cannot find one. All I've foud is that
> The copy constructor is auto-generated if there is no user-declared move constructor or move assignment operator (because there are no move constructors or move assignment operators in C++03, this simplifies to "always" in C++03) (§12.8/8)

----

I just wrote a test program:

```cpp
#include <iostream>

struct Foo {
  int a;
  int *p;
  Foo(int a = 0, int *p = nullptr) : a(a), p(p) {}
};

int main() {
  int b;
  Foo f(10, &b);
  Foo d;
  d = f;
  std::cout << d.a << "\n";
}
```

The assignment operator of `Foo` is still available even if we have a user defined constructor.

csukuangfj: > RE DCHECK_GT, as we will declare PtrT data without parameter in Array2, or we may write data = nullptr, we actually cannot require stride > 1.

Instead of writing
```cpp
DCHECK_GT(stride, 1);
```

can we use

```cpp
DCHECK_NE(stride, 1);

```

k2/csrc/array.h
```diff
@@ -49,73 +53,154 @@ struct Array1 {
...
+};
+
+// Allocate memory for Array2, this class is just for test purpose.
+// It will be usually called after those `GetSizes` functions (e.g.
+// FstInverter.GetSizes) to allocate memory in test code.
+template <typename Ptr, typename I>
+struct Array2Storage {
+  using ValueType = typename Array2<Ptr, I>::ValueType;
+  explicit Array2Storage(const Array2Size<I> &array2_size, int32_t stride)
+      : indexes_storage_(new I[array2_size.size1 + 1]),
```
csukuangfj, line 170: There are memory leaks here.

It should be

```cpp
  std::unique_ptr<I[]> indexes_storage_;
  std::unique_ptr<ValueType[]> data_storage_;
```

k2/csrc/array.h
```diff
@@ -49,73 +53,154 @@ struct Array1 {
...
+  }
+};
+
+// Allocate memory for Array2, this class is just for test purpose.
+// It will be usually called after those `GetSizes` functions (e.g.
+// FstInverter.GetSizes) to allocate memory in test code.
+template <typename Ptr, typename I>
+struct Array2Storage {
+  using ValueType = typename Array2<Ptr, I>::ValueType;
+  explicit Array2Storage(const Array2Size<I> &array2_size, int32_t stride)
```
csukuangfj, line 169: I guess it should be `I stride` instead of `int32_t stride`.

k2/csrc/array.h
```diff
@@ -49,73 +53,154 @@ struct Array1 {
...
+        data_storage_(new ValueType[array2_size.size2 * stride]) {
+    array_.size1 = array2_size.size1;
+    array_.size2 = array2_size.size2;
+    array_.indexes = indexes_storage_.get();
+    array_.data = DataPtrCreator<Ptr, I>::Create(data_storage_, stride);
+  }
+
+  void FillIndexes(const std::vector<I> &indexes) {
+    CHECK_EQ(indexes.size(), array_.size1 + 1);
+    for (auto i = 0; i != indexes.size(); ++i) array_.indexes[i] = indexes[i];
```
csukuangfj, line 180: I would suggest using `std::copy`.

k2/csrc/array.h
qindazhu: Oh yes, you're right, I just checked https://en.cppreference.com/w/cpp/language/copy_constructor 
https://en.cppreference.com/w/cpp/language/move_constructor
https://en.cppreference.com/w/cpp/language/copy_assignment
Updated, thanks

qindazhu:
@csukuangfj fix all issues here https://github.com/danpovey/k2/pull/61, please have a review

PR #59, opened 2020-06-15
replace Fsa with Array2 in Aux_labels

qindazhu:
Note that as we have defined `(const) begin` and `(const) end` for `Array2`, we can now use range-based for-loop, as I do in `IsTopSorted`
```
 for (const auto &arc : fsa) {
   if (arc.dest_state < arc.src_state) return false;
 }
```

### DIFF
--- a/k2/csrc/array.h
+++ b/k2/csrc/array.h
@@ -139,9 +139,13 @@ struct Array2 {
   */
   bool Empty() const { return size1 == 0; }
 
-  PtrT begin() const { return data + indexes[0]; }
+  PtrT begin() { return size2 != 0 ? data + indexes[0] : nullptr; }
+  const PtrT begin() const { return size2 != 0 ? data + indexes[0] : nullptr; }
 
-  PtrT end() const { return data + indexes[size1]; }
+  PtrT end() { return size2 != 0 ? data + indexes[size1] : nullptr; }
+  const PtrT end() const {
+    return size2 != 0 ? data + indexes[size1] : nullptr;
+  }
 
   // just to replace `Swap` functions for Fsa and AuxLabels for now,
   // may delete it if we finally find that we don't need to call it.
--- a/k2/csrc/array.h
+++ b/k2/csrc/array.h
@@ -101,12 +101,7 @@ struct Array2 {
   using PtrT = Ptr;
   using ValueType = typename std::iterator_traits<Ptr>::value_type;
 
-  // We need default constructor now just because we would check
-  // `indexes == nullptr` in NumStates and FinalStates in class Fsa.
-  // We may change it to `Array2() == default` if we don not need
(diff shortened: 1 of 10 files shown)


k2/csrc/array.h
```diff
@@ -139,9 +139,13 @@ struct Array2 {
   */
   bool Empty() const { return size1 == 0; }
 
-  PtrT begin() const { return data + indexes[0]; }
+  PtrT begin() { return size2 != 0 ? data + indexes[0] : nullptr; }
```
danpovey, line 142: It might be worthwhile to let `indexes` point to a zero (e.g. could use the address of `size` itself) in the case where size == 0.  Would avoid if-statements, which can be slow.

danpovey:
.. otherwise OK.

k2/csrc/array.h
qindazhu: Updated, thanks!

k2/csrc/array.h
```diff
@@ -106,7 +106,7 @@ struct Array2 {
   // We may change it to `Array2() == default` if we don not need
   // it anymore (after we replace Fsa with Array2).
   // Array2() = default;
-  Array2() : size1(0), indexes(nullptr) {}
+  Array2() : size1(0), indexes(nullptr), size2(0), data(nullptr) {}
```
danpovey, line 109: I would have thought you wuld be initializing indexes with &size1 or something like that.
Also the comment refers to checking `indexes == nullptr`... I think that might not be necessary now?

qindazhu: `indexes = nullptr` is required in constructor of FSA (otherwise it may break current code of FSA).

I can add a constructor in FSA to set it to `nullptr` to test if it's OK (and leave indexes as &size1 in array2).

danpovey: You could remove that check and change it to check that size1 == 0 (if necessary).
I don't know why that (indexes1 == nullptr) would be a requirement anyway.

qindazhu: mm. It's because we keep both `indexes` and `arc_indexes` in Fsa for now, I check `indexes == nullptr` in those FSA-related algorithms (`Numstates`, `FinalStates`, properties, etc.) to support both version for now (to replace FSA with Array2 one by one in those fsa_algo algorithms)

qindazhu: so we can't just check `size1 == 0` as  it will be 0 in both cases

qindazhu:
@danpovey I set `indexes = &size1` in `Array2` and reset it to `nullptr` in the constructor of Fsa (it's just temporary code for distinguishing whether the Fsa object is constructed using Array2 or origianl `arc_indexes` and `arcs`, will remove it finally).

danpovey:
OK... OK to merge for me.

qindazhu:
Merging.

PR #60, opened 2020-06-19
replace FSA with Array2 in WFSA

qindazhu:
replaced FSA with Array2 in WFSA and updated some WFSA-related code.


danpovey:
Looks fine from quick scan!

qindazhu:
Merging

PR #61, opened 2020-06-22
fix some issues of Array2

qindazhu:


k2/csrc/array.h
```diff
@@ -112,21 +105,20 @@ struct Array2 {
   }
 
   IndexT size1;
+  IndexT size2;     // the number of elements in the array,  equal to
```
csukuangfj, line 108: The initialization order of data members in the initialization list of the constructors should match the definition order.

csukuangfj:
+2

k2/csrc/array.h
qindazhu: Actually I prefer to keep the original order (size1, indexes, size2, data) here as it's more readable (indexes is related to size1, and data is related to data).

csukuangfj: Different people have different opinions about what is more readable.

The current approach is more memory-efficient.

qindazhu: Will merge and leave it with `unresolved`. We can update it later if other guys (@danpovey) think we should change the order in the initialization list

danpovey:
I agree with haowen, more memory efficient.

csukuangfj:
> I agree with haowen, more memory efficient.

I am confused.

csukuangfj:
> The initialization order of data members in the initialization list of the constructors should match the definition order.

It is a best practice to let the initialization order match the definition order,  though the current implementation is fine.

danpovey:
Oh sorry.. Make it match to avoid warnings is what i would do.

qindazhu:
There's no any warning in current code. 
Anyway, I'll change the initialization order in another PR as I don't want to take more time to argue about what is more readable here.

PR #64, opened 2020-07-01
replace Fsa with Array2 in RandEquivalent/Connect/Intersect/TopSort/ArcSort/..

qindazhu:
Replace `FSA` with `Array2` in `RandEquivalent/Connect/Intersect/TopSort/ArcSort/RandPath/GenerateRandFsa` and fix two bugs in `Intersect` and `RandEquivalent`, sorry for the big PR as those code are interdependent.


k2/csrc/arcsort.h
```diff
@@ -0,0 +1,57 @@
...
+  /*
+    Finish the operation and output the arc-sorted FSA to `fsa_out` and
+    arc mapping information to `arc_map` (if provided).
+    @param [out] fsa_out   The output FSA; Will be arc-sorted;
+                           Must be initialized; search for 'initialized
+                           definition' in class Array2 in array.h for meaning.
+    @param [out] arc_map   If non-NULL, will output a map from the arc-index
+                           in `fsa_out` to the corresponding arc-index in
+                           `fsa_in`.
+                           If non-NULL, Must be allocated with size num-arcs
```
danpovey, line 46: For `num-arcs of fsa_out`, I prefer if you make this more explicit by writing something usable as code, e.g. fsa_out->size or whatever it happens to be.

qindazhu: OK, will change those files.

k2/csrc/arcsort.cc
```diff
@@ -0,0 +1,67 @@
...
+  fsa_size->size2 = fsa_in_.size2;
+}
+
+void ArcSorter::GetOutput(Fsa *fsa_out, int32_t *arc_map /*= nullptr*/) {
+  CHECK_NOTNULL(fsa_out);
+  CHECK_EQ(fsa_out->size1, fsa_in_.size1);
+  CHECK_EQ(fsa_out->size2, fsa_in_.size2);
+
+  using ArcWithIndex = std::pair<Arc, int32_t>;
+  std::vector<int32_t> indexes(fsa_in_.size2);  // arc index mapping
```
danpovey, line 30: Need to be clearer here, it's a mapping from what to what? 
E.g. if indexes[i] == j, is index i an arc-index into fsa_out at entry, and index j an arc-index into fsa_out at exit?
Or vice versa?

k2/csrc/arcsort.cc
```diff
@@ -0,0 +1,67 @@
...
+  std::iota(indexes.begin(), indexes.end(), 0);
+  const auto arc_begin_iter = fsa_in_.data;
+  const auto index_begin_iter = indexes.begin();
+  int32_t num_states = fsa_in_.NumStates();
+  int32_t num_arcs = 0;
+  for (int32_t state = 0; state != num_states; ++state) {
+    fsa_out->indexes[state] = num_arcs;
+    int32_t begin = fsa_in_.indexes[state];
+    int32_t end = fsa_in_.indexes[state + 1];
+    std::vector<ArcWithIndex> arc_range_to_be_sorted;
```
danpovey, line 40: This will lead to multiple reallocations (malloc()) inside the loop; would be more efficient to declare the vector outside the loop.

k2/csrc/arcsort.cc
```diff
@@ -0,0 +1,67 @@
...
+    arc_range_to_be_sorted.reserve(end - begin);
+    std::transform(arc_begin_iter + begin, arc_begin_iter + end,
```
danpovey, line 42: I don't think it's actually necessary to store the Arcs in this vector, though.  What I would do is simply sort a vector of integers, but use a comparator which had as a member the begin() of the original (input) arcs, and treats the integer args as indexes into that.

danpovey: ... of course you'd still need a temporary place to put the arcs while copying, but it would require just 2 copies per arc, instead of as many as the sort requires.

danpovey: BTW, I'm not convinced that I would have made arc-sorting a non-in-place operation.  I don't see a scenario where you'd want to sort the arcs but still keep the original.

qindazhu: That's the original interface you wrote. So could we change it to an in-place interface now as it seems there's no such scenario (non-in-place)?

danpovey:
Mm, but I was thinking of other operations.. I feel like arc sorting
doesn't require that.
It can be a simple function call.

qindazhu:
> Mm, but I was thinking of other operations.. I feel like arc sorting doesn't require that. It can be a simple function call.


You mean a function like `ArcSort(const Fsa &fsa_in, Fsa *fsa_out, arc_map)` but requires that `fsa_out` must be initialized at entry? still non-in-place?  or a inplace version `ArcSort(Fsa *fsa)` (we will never need the arc_map in this version, right?)?

k2/csrc/arcsort.cc
qindazhu: by saying `original interface`, I mean the non-in-place version for `ArcSort`

danpovey:
Mm.   It's OK to leave that code I guess, but I think there should be an
in-place version. In-place is what I was talking about.

qindazhu:
> In-place is what I was talking about.
Sorry, I'm confused by the thread, can you write the interface of the in-place version? and I would like to implement it in another PR.

danpovey:
Oh wait, I'm sorry.. I forgot about the arc_map.
We'd likely need the arc map even for in-place.
In-place version would be like:
```
 ArcSort(Fsa *fsa, whatever *arc_map);
```
note, the size of arc_map is known in advance, it's the number of arcs in fsa.

k2/csrc/connect.h
```diff
@@ -0,0 +1,111 @@
...
+        Caution: true return status does not imply that the returned FSA is
+        nonempty.
+   */
+  bool GetOutput(Fsa *fsa_out, int32_t *arc_map = nullptr);
+
+ private:
+  const Fsa &fsa_in_;
+
+  bool is_acyclic_;  // if the input FSA is acyclic
+  // indicates that there's no any accessible & co-accessible state
```
danpovey, line 78: this comment is unclear.  I'd write: // if true, there is no state in the input FSA that is both accessible and co-accessible.
(Assuming that's what you mean!  Or do you mean there exists at least one state that is either not accessible or not co-accessible?)

qindazhu: Your first version, updated, thanks!

qindazhu:
> Oh wait, I'm sorry.. I forgot about the arc_map.
> We'd likely need the arc map even for in-place.
> In-place version would be like:
> 
> ```
>  ArcSort(Fsa *fsa, whatever *arc_map);
> ```
> 
> note, the size of arc_map is known in advance, it's the number of arcs in fsa.

Added, thanks.

danpovey:
OK thanks.. probably OK to merge unless Fangun plans to review.

csukuangfj:
+2

I will continue with the pybind11 wrapper.

qindazhu:
Merging, thanks!

PR #68, opened 2020-07-18
add pybindings for Array2

qindazhu:


danpovey:
Thanks a lot!  I have a had a quick look to try to stay familiar with what you are doing.
Right now I am focusing on mapping out the GPU code.

k2/python/csrc/array.cc
```diff
@@ -0,0 +1,141 @@
...
+    StridedPtr<ValueType>), the requirement of `cap_data` is nearly same with
+    case 2 above except cap_data.strides[0] will be greater than
+    num-primitive-values in `ValueType`.
+
+*/
+template <typename ValueType, bool IsPrimitive, typename I>
+class DLPackArray2;
+
+template <typename ValueType, typename I>
+class DLPackArray2<ValueType *, true, I> : public Array2<ValueType *, I> {
```
csukuangfj, line 51: The compiler should be able to figure out whether ValueType is primitive or not.
You do not need to pass true or false by yourself.

k2/python/csrc/array.cc
```diff
@@ -0,0 +1,141 @@
...
+ public:
+  DLPackArray2(py::capsule cap_indexes, py::capsule cap_data)
+      : indexes_tensor_(new Tensor(cap_indexes)),
+        data_tensor_(new Tensor(cap_data)) {
+    CHECK_EQ(indexes_tensor_->NumDim(), 1);
+    CHECK_GE(indexes_tensor_->Shape(0), 1);  // must have one element at least
+    CHECK_EQ(indexes_tensor_->Stride(0), 1);
+
+    CHECK_EQ(data_tensor_->NumDim(), 1);
+    CHECK_GE(data_tensor_->Shape(0), 0);  // num-elements
```
csukuangfj, line 61: CHECK_GT ?

k2/python/csrc/array.cc
```diff
@@ -0,0 +1,141 @@
...
+};
+
+// Note: we can specialized for `StridedPtr` later if we need it.
+
+}  // namespace k2
+
+template <typename Ptr, bool IsPrimitive, typename I = int32_t>
+void PybindArray2Tpl(py::module &m, const char *name) {
+  using PyClass = k2::DLPackArray2<Ptr, IsPrimitive, I>;
+  using Parent = k2::Array2<Ptr, I>;
```
csukuangfj, line 111: I would suggest defining `Parent` inside  `DLPackArray2` and using

```cpp
using Parent = typename PyClass::Parent;
```

k2/python/csrc/array.cc
```diff
@@ -0,0 +1,141 @@
...
+          "__iter__",
+          [](const PyClass &self) {
+            return py::make_iterator(self.begin(), self.end());
+          },
+          py::keep_alive<0, 1>())
+      .def_readonly("size1", &PyClass::size1)
+      .def_readonly("size2", &PyClass::size2)
+      .def("indexes", [](const PyClass &self, I i) { return self.indexes[i]; })
+      .def("data", [](const PyClass &self, I i) { return self.data[i]; });
+  // TODO(haowen): expose `indexes` and `data` as a array
```
csukuangfj, line 126: an array

k2/python/csrc/array.cc
```diff
@@ -0,0 +1,141 @@
...
+  // instead of a function call?
+}
+
+void PybindArray(py::module &m) {
+  // Note: all following wrappers whose name start with `_` are only used by
```
csukuangfj, line 131: all the following,
whose name starts

k2/python/tests/array_test.py
```diff
@@ -0,0 +1,63 @@
...
+#
+
+import unittest
+
+import torch
+
+import k2
+
+
+class TestFsa(unittest.TestCase):
```
csukuangfj, line 19: class TestArray

k2/python/tests/array_test.py
```diff
@@ -0,0 +1,63 @@
...
+        for i, v in enumerate(array):
+            self.assertEqual(i, v)
+
+        self.assertEqual(indexes.numel(), array.size1 + 1)
+        self.assertEqual(data.numel(), array.size2)
+
+        # the underlying memory is shared between k2 and torch;
+        # so change one will change another
+        data[0] = 100
+        self.assertEqual(array.data(0), 100)
```
csukuangfj, line 40: actually, you are unable to use
```python
array.data(0) = 1000
```
to change the `data` in torch in the current implementation.

k2/python/csrc/array.cc
```diff
@@ -0,0 +1,141 @@
...
+    CHECK_GE(indexes_tensor_->Shape(0), 1);  // must have one element at least
+    CHECK_EQ(indexes_tensor_->Stride(0), 1);
+
+    CHECK_EQ(data_tensor_->NumDim(), 1);
+    CHECK_GE(data_tensor_->Shape(0), 0);  // num-elements
+    CHECK_EQ(data_tensor_->Stride(0), 1);
+
+    int32_t size1 = indexes_tensor_->Shape(0) - 1;
+    int32_t size2 = data_tensor_->Shape(0);
+    this->Init(size1, size2, indexes_tensor_->Data<I>(),
```
csukuangfj, line 66: I think we need to check that `tensor->dtype()` is the same as ValueType.
For instance, if ValueType is `int32_t`, we should prevent the user from passing
a tensor of dtype == torch.float32.

csukuangfj:
@qindazhu  There are 10 branches in k2 right now. Could you delete those that are already merged?

k2/python/csrc/array.cc
```diff
@@ -0,0 +1,141 @@
...
+
+#include "k2/csrc/array.h"
+#include "k2/python/csrc/tensor.h"
+
+namespace k2 {
+/*
+   DLPackArray2 initializes Array2 with `cap_indexes` and `cap_data` which are
+   DLManagedTensors.
+
+   `cap_indexes` is usually a one dimension contiguous array, i.e.,
```
csukuangfj, line 20: a one dimensional

k2/python/csrc/array.cc
qindazhu: How? there is a function `std::is_fundamental`, I can pass it as the value of `IsPrimitive`, but that also requires the template parameter.

k2/python/csrc/array.cc
qindazhu: we can have empty Array2

k2/python/csrc/array.cc
qindazhu: what's the syntax you are writing? We can do this in C++?

k2/python/tests/array_test.py
qindazhu: Yes, I know that. That's why I add a `TODO` to support `[]`. BTW, I even doubt whether or not we need to write data in Python, as we will create Array2 with `tensors` or `Array2Size`  and will write data to it in C++

k2/python/csrc/array.cc
csukuangfj: One possible method would be

```cpp
template<typename T> struct SupportedPrimitiveType : public std::false_type {};

template<> struct SupportedPrimitiveType<int32_t> : public std::true_type {};
template<> struct SupportedPrimitiveType<float> : public std::true_type {};

// add other supported types here
```

Then you can use

```cpp
template <typename ValueType, typename I>
class DLPackArray2;

// for primitive types
template<typename ValueType, typename I>
class DLPackArray2<typename std::enable_if<SupportedPrimitiveType<ValueType>::value, ValueType*>::type, I> {...};


// for non-primitive types

template<typename ValueType, typename I>
class DLPackArray2<typename std::enable_if<!SupportedPrimitiveType<ValueType>::value, ValueType*>::type, I> {...};

```

k2/python/csrc/array.cc
qindazhu: We could not do this here completely as we cannot check this for non-primitive types (that will be complex). So I plan to do this in Python Class (such as `IntArray`, `Fsa`, etc.) as we will only expose those Python Classes to users.

k2/python/csrc/array.cc
csukuangfj:
```cpp
template <typename ValueType, typename I>
class DLPackArray2<ValueType *, false, I> : public Array2<ValueType *, I> {
  public:
     using Parent = Array2<ValueType *, I>;
};

template <typename Ptr, bool IsPrimitive, typename I = int32_t>
void PybindArray2Tpl(py::module &m, const char *name) {
  using PyClass = k2::DLPackArray2<Ptr, IsPrimitive, I>;
  using Parent = typename PyClass::Parent;
};
```

k2/python/csrc/array.cc
qindazhu: mm..that's the same way with `std::is_fundamental`, you even do not need to write those `SupportedPrimitiveType` structs.

I prefer the simple one, as users will only expose a few types in `k2` (that's why I did not use `is_fundamental` here.)

k2/python/csrc/array.cc
qindazhu: mm,  I prefer the original one as we only use it one time. Will resolve this issue

qindazhu:
> @qindazhu There are 10 branches in k2 right now. Could you delete those that are already merged?

those branches are all created by @danpovey , I think it's better to let he decide it.

k2/python/tests/array_test.py
qindazhu: @danpovey maybe you can comment on this: do we need to provide `setter` function for `Fsa/Array2` (`indexes` and `data`) in Python code? I just thought it does not make sense (and maybe dangerous) if user can change data in Python directly instead of calling function in C++.

danpovey:
Sure, delete merged ones.

qindazhu:
deleted, only `cuda_draft` left.

k2/python/tests/array_test.py
danpovey: Mm.  We'll need some method to create these things, but perhaps you have created one already?
I can see instances where we might need to set the data, at least.

qindazhu:
Merged. Leave the two comments unsolved as I may work on them in the next PR.

PR #69, opened 2020-07-22
rewrite pybind interface for Array2 and Fsa

qindazhu:
Guys, please have a review. This PR is relatively **important**, if the whole structure/style is fine, I will follow it to wrap other algorithms to python (most of them will be straightforward).

Note as in Python, we cannot pass primitive type (int, float, etc.) to function as reference/pointer, I get `Array1` back to support those `arc_map` (see example in `arc_sort`)

### DIFF
--- a/k2/python/csrc/array.cc
+++ b/k2/python/csrc/array.cc
@@ -1,5 +1,3 @@
-// k2/python/csrc/array.cc
-
 // Copyright (c)  2020  Xiaomi Corporation (author: Haowen Qiu)
 
 // See ../../../LICENSE for clarification regarding multiple authors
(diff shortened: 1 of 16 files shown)


csukuangfj:
You can leave the interfaces as they are

To wrap them to python, you only need to add an indirection.

Take
```cpp
void ArcSorter::GetOutput(Fsa *fsa_out, int32_t *arc_map /*= nullptr*/) const
```
as an example, you can use
```cpp
pyclass.def("get_output", [](const ArcSorter& self, Fsa* fsa_out, std::vector<int32_t>* arc_map = nullptr) {
        self.GetOutput(fsa_out, arc_map ? arc_map->data() : nullptr);
}, py::arg("fsa_out"), py::arg("arc_map") = nullptr
);
```
I assume that you have made `std::vector<int32_t>` an opaque type and `arc_map` is saved in `std::vector<int32_t>`.

qindazhu:
I make `Array1` share memory with `torch.tensor` as we do in `Array2`, so does `vector<int32_t>` here share memory with `tensor`? how do you do that with above code?

csukuangfj:
> I make Array1 share memory with torch.tensor as we do in Array2, so does vector<int32_t> here share memory with tensor? how do you do that with above code?

You can still leave the interfaces as they are. Just replace `std::vector<int32_t>` above with `Array1`.

csukuangfj:
Alternatively, you can replace `std::vector<int32_t>` with a `py::capsule` and construct an `Array1` from the `py::capsule` inside the lambda function.

qindazhu:
Oh, you're right, as we could just change the interface in Python code. 

> You can still leave the interfaces as they are. Just replace std::vector<int32_t> above with Array1.

I would like to choose this one as constructing `Array1` outside of the function is clearer and consistent with what we do for Fsa.

csukuangfj:
> I would like to choose this one as constructing Array1 outside of the function is clearer and consistent with what we do for Fsa.

Anyway, there is no need to change `k2/csrc/*.cc` files.

qindazhu:
Updated, thanks @csukuangfj for the suggestions.

k2/python/csrc/array.cc
```diff
@@ -100,11 +128,32 @@ class DLPackArray2<ValueType *, false, I> : public Array2<ValueType *, I> {
...
+template <typename Ptr, typename I = int32_t>
+void PybindArray1Tpl(py::module &m, const char *name) {
+  using PyClass = k2::DLPackArray1<Ptr, I>;
+  using Parent = k2::Array1<Ptr, I>;
+  py::class_<PyClass, Parent>(m, name)
+      .def(py::init<py::capsule>(), py::arg("data"))
+      .def("empty", &PyClass::Empty)
+      .def(
+          "get_base",
+          [](PyClass &self) { return reinterpret_cast<Parent *>(&self); },
```
csukuangfj, line 144:
```cpp
[](PyClass &self) -> Parent* {return self;}
```
and the explicit cast can be removed.

qindazhu: I guess you mean `return &self`. Updated.

csukuangfj:
+1

danpovey:
Thanks for the review Fangjun!
It's OK with me.  I'm not doing a detailed review as I want to focus on
figuring out some next-steps stuff.

qindazhu:
Merging..

PR #73, opened 2020-07-25
implement GetContext with parameter pack

qindazhu:
Implemented https://github.com/danpovey/k2/issues/72

### DIFF
--- a/k2/csrc/cuda/context.h
+++ b/k2/csrc/cuda/context.h
@@ -65,6 +65,18 @@ class Context {
 
 typedef std::shared_ptr<Context> ContextPtr;
 
+template <typename T> ContextPtr GetContext(const T& t) {
+    // suppose T has member method `Context`
+    return t.Context();
+}
+
+template <typename First, typename... Rest> ContextPtr GetContext(const First& first, const Rest&... rest) {
+    ContextPtr ans1 = GetContext(first),  ans2 = GetContext(rest...);
+    assert(*ans1 == *ans2 && "Contexts mismatch");
+    // std::cout << sizeof...(Rest)<< std::endl;
+    return ans1;
+}
+
 /*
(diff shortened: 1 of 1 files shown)


k2/csrc/cuda/context.h
```diff
@@ -65,6 +65,18 @@ class Context {
...
+    return t.Context();
+}
+
+template <typename First, typename... Rest> ContextPtr GetContext(const First& first, const Rest&... rest) {
+    ContextPtr ans1 = GetContext(first),  ans2 = GetContext(rest...);
+    assert(*ans1 == *ans2 && "Contexts mismatch");
+    // std::cout << sizeof...(Rest)<< std::endl;
+    return ans1;
+}
+
```
qindazhu, line 79: we can use it with:
```
A a;
B b;
C c;
GetContext<A,B,C,A>(a,b,c,a);
GetContext<A,B>(a,b);
GetContext<A>(a);
```

danpovey: Thanks!  I'm assuming in the normal case the template args won't be used.  (?)  That's the idea.  LMK if that's not the case.

danpovey: .. I mean, they should be automatically inferred.

qindazhu: Do you mean for the case that there is only one template argument? I think we can just specialize the template function for those classes. Is that fine? As far as I know, in C++, we cannot convert a template function to a normal function (i.e. type inferred in the above case).

danpovey:
No, I'm saying I'm hoping the template prams can be automatically inferred
so they dont have to be typed out.

qindazhu:
Oh, you can, sorry for the examples. Actually you can write code below with the above implementation:
```
GetContext(a,b,c,b);
GetContext(a,b);
GetContext(a);
```

PR #75, opened 2020-07-27
Update the build system to support Cuda.

csukuangfj:
Now Cub is supported and the cuda branch is paritially compilable.

I've added a  naive implementation for context and a test for ExclusivePrefixSum.

A google [colab notebook][1] is also added in case someone wants to try it out but does not have a GPU.

----

Note that CMake 3.8 is required. C++ 14 is used to remove the warnings reported by Cub.

@danpovey @qindazhu 
please have a review.


k2/csrc/cuda/utils_inl.h
```diff
@@ -1,23 +1,48 @@
...
+
+// Copyright (c)  2020  Xiaomi Corporation (authors: Daniel Povey)
+//                      Fangjun Kuang (csukuangfj@gmail.com)
+
+// See ../../LICENSE for clarification regarding multiple authors
 
 /* Don't include this file directly; it is included by utils.h.
    It contains implementation code. */
 
+#include <assert.h>
```
qindazhu, line 11: #include <cassert>

.github/workflows/build.yml
```diff
@@ -10,9 +10,11 @@ on:
   push:
     branches:
       - master
+      - cuda_draft
```
qindazhu, line 13: maybe we should create a new branch (e.g. `cuda`), as Dan may push to `cuda_draft` directly.

k2.ipynb
```diff
@@ -0,0 +1,858 @@
+{
```
qindazhu, line 1: do we need to include this file? seems a lot of logs in the file? I noticed that you have add the corresponding link of `Google Colab` in `README.md`

k2/csrc/cuda/context.h
```diff
@@ -1,10 +1,29 @@
...
 
-enum DeviceType {
+// Copyright (c)  2020  Xiaomi Corporation (authors: Daniel Povey)
+
+// See ../../LICENSE for clarification regarding multiple authors
+
+#ifndef K2_CSRC_CUDA_CONTEXT_H_
+#define K2_CSRC_CUDA_CONTEXT_H_
+
+#include <assert.h>
```
qindazhu, line 10: `<cassert>`?

k2/csrc/cuda/context.cu
```diff
@@ -0,0 +1,66 @@
+// k2/csrc/cuda/context.cu
+
+// Copyright (c)  2020  Fangjun Kuang (csukuangfj@gmail.com)
+
+// See ../../LICENSE for clarification regarding multiple authors
+
+// WARNING(fangjun): this is a naive implementation to test the build system
+#include "k2/csrc/cuda/context.h"
+
+#include <stdlib.h>
```
qindazhu, line 10: `<cstdlib>`?

qindazhu:
Fantastic work! Only a few comments, +2!

csukuangfj:
thanks, I will resolve the comments tonight.

.github/workflows/build.yml
csukuangfj: I've disabled the GitHub workflow for the CUDA branch since it requires CUDA and won't compile, which always fails.

k2/csrc/cuda/utils_inl.h
csukuangfj: fixed

.github/workflows/build.yml
csukuangfj: Once this pullrequest is merged, we can create a new branch from cuda_draft and leave cuda_draft for Dan.

k2.ipynb
csukuangfj: I've deleted it and changed the link in README.md

k2/csrc/cuda/context.h
csukuangfj: done

k2/csrc/cuda/context.cu
csukuangfj: fixed.

danpovey:
Great, thanks!

danpovey:
@csukuangfj why did you upgrade the build to c++14 from c++11, and what file did you edit to do that?
It's causing me problems building it, as the version of gcc here doesn't support that.  I don't like to use any newer
version of c++ than we need to.

csukuangfj:
To use c++11, you can change 
https://github.com/danpovey/k2/blob/a662f9bfea895ad5eef09abcb8a02b6657d7c1f8/CMakeLists.txt#L42

to

set(CMAKE_CXX_STANDARD 11 CACHE STRING "The C++ version to be used.") 

---

The reason to switch to c++14 is that the following code from CUB: https://github.com/NVlabs/cub/blob/master/cub/util_cpp_dialect.cuh

```cpp
#if !defined(CUB_IGNORE_DEPRECATED_CPP_DIALECT) && CUB_CPP_DIALECT < 2014 && \
    (CUB_CPP_DIALECT != 2011 || !defined(CUB_IGNORE_DEPRECATED_CPP_11))
  CUB_COMPILER_DEPRECATION(C++14, pass -std=c++14 to your compiler);
#endif
```

issues warnings during the compilation if you are not using c++14.

csukuangfj:
I can suppress the CUB warnings by deleting `CUB_COMPILER_DEPRECATION(C++14, pass -std=c++14 to your compiler);`
if you like.

danpovey:
OK thanks... yeah suppress the warning for now.  I guess it's more
efficient with c++14, but I'd like to make sure our code works with c++11,
so we can support a wider range of compilers.

csukuangfj:
I will make a change this night.

You can use

```
cd build
cmake -DCMAKE_CXX_STANDARD=11 ..
```
to switch to c++11 without changing the CMakeLists.txt

danpovey:
thanks

PR #76, opened 2020-07-30
Wrap AuxLabels/Wfsa/RandEquivalent to Python

qindazhu:
Note that

-  I have changed the type of `arc_map` in `AuxLabelsMapper` from `vector` to `Array1` and `Array2` so that we can feed output from other algorithms (e.g. `ArcSort(fsa_out, arc_map)`) into `AuxLabelsMapper` directly in Python code, which will be simpler for users (Of course we may change the interface later when Cuda interface is stable). 

- I have changed to allocate memory outside (e.g. in Python) for `Wfsa.forward/backward_states_weight`.

- Support double in `Tensor.cc`.


k2/csrc/aux_labels.cc
```diff
@@ -90,9 +90,10 @@ namespace k2 {
 
 void AuxLabels1Mapper::GetSizes(Array2Size<int32_t> *aux_size) {
```
csukuangfj, line 91: void AuxLabels1Mapper::GetSizes(Array2Size<int32_t> *aux_size) **const**

A `Get` function should be `const`.

k2/csrc/aux_labels_test.cc
```diff
@@ -33,7 +34,7 @@ class AuxLablesTest : public ::testing::Test {
 TEST_F(AuxLablesTest, AuxLabels1Mapper) {
   {
     // empty arc_map
-    std::vector<int32_t> arc_map;
+    Array1<int32_t *> arc_map;
```
csukuangfj, line 37: Should we define
```cpp
using IntArray1 = Array1<int32_t *>;
```

danpovey: I'd rather use the template directly, but are you sure it should be int32_t*  ?  Seems to me it should be int32_t.

danpovey: I'd prefer to use the template rather than a typedef.
Incidentally, all of this code is going to have to be changed to use the templates defined in the cuda/ directory, like Array1 (it's different), Ragged2, Ragged3.

qindazhu: It's `int32_t*`, as we define `Array` to accept `Ptr` (raw pointer or StridedPtr) as its template argument. 
RE Ragged2, yes, of couse we will update those bindings later after cuda interface is stable.

k2/csrc/aux_labels.cc
qindazhu: Fixed, thanks! Note that `GetSizes` of `Connect/Determinize/RmEpsilons/...` is not `const` as we need to do most of work in it and save some varibales into the fields of class object.

k2/csrc/aux_labels_test.cc
qindazhu: Resolved, will not change.

danpovey:
Thanks.. OK to merge I guess.

PR #77, opened 2020-07-30
fix CUB warnings when using c++11.

csukuangfj:
@qindazhu @danpovey 
please have a review.

### DIFF
--- a/cmake/cub.cmake
+++ b/cmake/cub.cmake
@@ -24,7 +24,7 @@ function(download_cub)
   message(STATUS "cub is downloaded to ${cub_SOURCE_DIR}")
   add_library(cub INTERFACE)
   target_include_directories(cub INTERFACE ${cub_SOURCE_DIR})
-
+  target_compile_definitions(cub INTERFACE CUB_IGNORE_DEPRECATED_CPP_DIALECT)
 endfunction()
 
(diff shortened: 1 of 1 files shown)


qindazhu:
+1, maybe add some comments to say why we do this?

danpovey:
Guys, actually I dont want to suppress these warnings because it's there for a good reason and will remind us and others that c++11 is not ideal.

csukuangfj:
@danpovey 
I guess you are using GCC 4.9 or 4.8. Are there any reasons not to upgrade gcc to a version >= 5.0 ?

danpovey:
No reason, except that it's a hassle to install.  How can you tell I'm
using that version, and why does it matter?

csukuangfj:
C++ 14 is supported for GCC >= 5.0

GCC 4.8 supports  only C++11, so I think you are using   4.8 <= GCC < 5.0

csukuangfj:
It does not matter since CUB compiles for C++11.

It does matter if you want to avoid the warnings, but you are unable to since your compiler supports only c++11.
> It's causing me problems building it, as the version of gcc here doesn't support that

danpovey:
sorry gonna close this, we can worry uch later about things like suppressing warnings.

PR #79, opened 2020-07-31
wrap determinize/rmepsilon to Python

qindazhu:


k2/csrc/array.h
```diff
@@ -189,6 +189,46 @@ struct Array2 {
  */
 };
 
+// Specialized Array2 for std::pair*, we represents something like
+// vector<vector<std::pair<U,V>> as an Array2 object with two data members
+// `U* data1` and `V* data2`. Given any `indexes[i]`,
+// `data1[indexes[i]]` and `data2[indexes[i]]` can be viewed as a pair
+// in the original vector of vector<std::pair<U,V>>.
+template <typename U, typename V, typename I>
+struct Array2<std::pair<U, V> *, I> {
```
qindazhu, line 198: Note I specialize `Array2` for `std::pair` with two data members (`data1` and `data2`), so that there's no need to cast type when creating pybindings for `std::pair` (e.g. `LogSumTracebackState::DerivType`)

danpovey: Sorry, but I don't think I like this design.  The problem is that this Array2 looks like a regular Array2 but doesn't behave like one; it doesn't have the same interface or the same members.  It's templated on the pair, but how do you access the pair type?  I don't see any data-member that points to it.

I think it would be better to have the underlying data be a float array (say), but reinterpret the memory as int where needed.

qindazhu: OK, let's say we have `arc_derivs = tensor.FloatTensor([num_arcs_derivs, 2])` on Python side, and `Array2<std::pair<int32_t, float>*` on C++ side. Suppose we set `arc_id = 1` on C++ side, then we'll get value like `1.4013e-45`  on Python side as a float. As most of algorithms accept `int*` or `Array1<int32*>` as `arc_map`, we need to reinterpret the python's float memory as int on C++ side  every time when we interact with code in C++, is that right? However, as we will wrap those algorithms to Python, so on Python side, those algorithms will accpet `IntTensor`, if we want to pass the first column of `arc_derivs` (i.e. `arc_derivs[: 0]` to get arc-ids) above to those Python algorithms, we need to convert `tensor.FloatTensor` as `tensor.IntTensor`,right?   I just worry about the accessibility for users and the scalability for other frameworks.

RE  the interface, if we want it behave like `Array2`, i.e. access `data1` and `data2` as pair type, maybe we can make `data1` and `data2` as private, and define some getter and setter like `void set_data(std::pair<int, float>, int32_t index)`? Well, still not like `Array2` but make it easy to interact with other algorithms in Python code.

Actually, I propose that we create a new struct of Array2 (maybe a new name instead of `Array2`) as the specialized version of `std::pair`, i.e. we have `IndexT indexes` in this version `Array2<some_tuple_type/some_struct_type, I>`, but will map all fields in type `some_tuple_type/some_struct_type` to mutiple `data` (`data1, data2, data3...`) in this `Array2`. So there's no need to reinterpret or cast in any framework.

danpovey: In general we're going to have these kinds of problems, e.g. in the cuda stuff I am putting a float field in Arc so it has mixed int and float fields.  But I think the best way to handle it is to have it of a fixed arbitrary type on the Python side (e.g. float here) and cast as needed.  These will largely be used in side k2 algorithms as temporaries, user code won't, generally speaking, need to access these kinds of things.  So I don't think it really matters.  Anyway we can wrap them in some kind of class on the python side that knows how to deal with them, if needed.  So just do it the way I said, for now.  I think you are over-thinking this.

qindazhu: ok

k2/python/tests/determinize_test.py
```diff
@@ -0,0 +1,103 @@
...
+        self.assertEqual(fsa_out.size2, 9)
+        self.assertEqual(arc_derivs.size1, 9)
+        self.assertEqual(arc_derivs.size2, 15)
+        self.assertTrue(
+            k2.is_rand_equivalent_logsum_weight(self.fsa, self.weights,
+                                                fsa_out, arc_weights_out,
+                                                beam))
+        # cast float to int
+        arc_ids = k2.StridedIntArray1.from_float_tensor(arc_derivs.data[:, 0])
+        self.assertEqual(arc_ids.get_data(1), 9)
```
qindazhu, line 99: @danpovey here I reinterpret `tensor.FloatArray` (i.e. `arc_derivs[:,0]`) as `StridedIntArray1`, so we can pass this `(Strided)IntArray1` object to C++ code that accepts `IntArray1` or `int*`. We can do same thing in the future when create bindings for new struct `Arc` in `cuda_draft`

k2/python/tests/rmepsilon_test.py
```diff
@@ -0,0 +1,106 @@
...
+        self.assertEqual(fsa_out.size1, 6)
+        self.assertEqual(fsa_out.size2, 11)
+        self.assertEqual(arc_derivs.size1, 11)
+        self.assertEqual(arc_derivs.size2, 20)
+        self.assertTrue(
+            k2.is_rand_equivalent_after_rmeps_pruned_logsum(
+                self.fsa, self.weights, fsa_out, arc_weights_out, beam))
+        # cast float to int
+        arc_ids = k2.StridedIntArray1.from_float_tensor(arc_derivs.data[:, 0])
+        self.assertEqual(arc_ids.get_data(1), 1)
```
qindazhu, line 102: @danpovey here I reinterpret `tensor.FloatArray` (i.e. `arc_derivs[:,0]`) as `StridedIntArray1`, so we can pass this `(Strided)IntArray1` object to C++ code that accepts `IntArray1` or `int*`. We can do same thing in the future when create bindings for new struct `Arc` in `cuda_draft`

qindazhu:
Error from above build checks
```
Collecting torch==1.5.0
  Downloading torch-1.5.0-cp35-cp35m-manylinux1_x86_64.whl (752.0 MB)
ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    torch==1.5.0 from https://files.pythonhosted.org/packages/29/2a/7ef31c8057c6679ad0019974dc9f6d0ad9dbd71d45eaa60f2e33f1b2baf8/torch-1.5.0-cp35-cp35m-manylinux1_x86_64.whl#sha256=931b79aed9aba50bf314214be6efaaf7972ea9539a3d63f82622bc5860a1fd81:
        Expected sha256 931b79aed9aba50bf314214be6efaaf7972ea9539a3d63f82622bc5860a1fd81
             Got        e2c233410691aa0d0b1d36aba416f025e1534cf51cae1e5d8920e50467264e66
```

@csukuangfj  Any insights? should we add flag ` --no-cache-dir` to our github workflows when installing python dependencies?

danpovey:
is the error reproducible?

qindazhu:
> is the error reproducible?

Yes, it is. But as I said above, it does not occur every time. Next time when it fails I will paste the build  link (I don't know how to get the build history...)

danpovey:
maybe just some kind of corruption or firewell-related, may not be easy to
fix in general

k2/python/tests/rmepsilon_test.py
qindazhu: @danpovey  have you reviewed this PR? I updated it last Saturday.

csukuangfj:
@qindazhu 

maybe we can download and install the whl package from the PyTorch website directly via

```bash
pip install torch==1.5.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
```

qindazhu:
> @qindazhu
> 
> maybe we can download and install the whl package from the PyTorch website directly via
> 
> ```shell
> pip install torch==1.5.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
> ```

OK, thanks, I'll update it in `build.yml` to have a try.

.github/workflows/build.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: Install Python dependencies
         run: |
           python3 -m pip install --upgrade pip
-          python3 -m pip install torch==1.5.0
+          python3 -m pip install torch==1.5.0 -f https://download.pytorch.org/whl/torch_stable.html
```
qindazhu, line 40: @csukuangfj , I changed it to use `torch=1.5.0`, as name `1.5.0+cpu` for MacOS is not on the website of `https://download.pytorch.org/whl/torch_stable.html` (instead it's `1.5.0-cp35/36/37/38` on MacOS). Seems it works well, I'll run the job more times to check. Thanks!

qindazhu:
Still get the same error (I run three times, get the  error in one time)  https://github.com/danpovey/k2/pull/79/checks?check_run_id=938904235

danpovey:
Looks good from a quick look, thanks very much!!

qindazhu:
thanks, merging. We can fix any issue of the interface when creating bindings for cuda.

PR #83, opened 2020-08-12
cmake: set "Debug" as default, and add "-O0" for debugging

songmeixu:
[cmake set debug](https://app.gitkraken.com/glo/board/5f1984580a23d90011f14154/card/5f33c6b9ab0eea0011ab76da)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,9 +23,20 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin")
 
 if(NOT CMAKE_BUILD_TYPE)
   message(STATUS "No CMAKE_BUILD_TYPE given, default to Release")
-  set(CMAKE_BUILD_TYPE Release CACHE STRING
+  set(CMAKE_BUILD_TYPE Debug CACHE STRING
     "Set the build type. Available values are: Debug Release RelWithDebInfo MinSizeRel"
     FORCE)
+endif()
+
+if(CMAKE_BUILD_TYPE STREQUAL Debug)
+  message(STATUS "Build in Debug mode")
+  # Add -O0 to remove optimizations for debugging.
+  set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0")
+  set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -O0")
+elseif(CMAKE_BUILD_TYPE STREQUAL Release)
+  message(STATUS "Build in Release mode")
+else()
+  message(FATAL_ERROR "Unknown build type!! (only accept: Debug, Release)")
 endif()
 
 if(WIN32 AND BUILD_SHARED_LIBS)


CMakeLists.txt
```diff
@@ -23,11 +23,22 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin")
...
-  set(CMAKE_BUILD_TYPE Release CACHE STRING
+  set(CMAKE_BUILD_TYPE Debug CACHE STRING
     "Set the build type. Available values are: Debug Release RelWithDebInfo MinSizeRel"
     FORCE)
 endif()
 
+if(CMAKE_BUILD_TYPE STREQUAL Debug)
+  message(STATUS "Build in Debug mode")
+  # Add -O0 to remove optimizations for debugging.
+  set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0")
```
danpovey, line 34: thanks... @csukuangfj does this look OK?

csukuangfj: It looks fine to me.

---

I would suggest saving `CMAKE_CXX_FLAGS_DEBUG` to cmake cache, i.e.,

```cmake
set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0" CACHE STRING "" FORCE)
```

To make it more robust, we can convert `CMAKE_BUILD_TYPE` to lowercase or uppercase and compare it against `debug` or `DEBUG`.

songmeixu:
Guys, I learn about `CMAKE_BUILD_TYPE` and make some changes.

keys:
- CMAKE_BUILD_TYPE and CMAKE_CXX_FLAGS_DEBUG both are build-in variables that have "cached" property already (No need to declare cache, unless we want to change the allowable values).
- CMAKE_CXX_FLAGS_DEBUG: no need to put it within `CMAKE_BUILD_TYPE == Debug`, as the flags will append to CMAKE_CXX_FLAGS if "Debug".
- CMake is case sensitive. And the build types are defined as one string from "None Debug Release RelWithDebInfo MinSizeRel". Thus, no need to take care of the upper case. And checking CMAKE_BUILD_TYPE is a bad pattern as to its value could be effectless in the situation of multi-configuration generators. The practical two ways for setting compiler based on debug/.. flags are:
	```cmake
	set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} --my-debug-flags")
	# or
	target_compile_definitions(MyTarget PUBLIC "$<$<CONFIG:Debug>:MYDEBUG_MACRO>")
	```

csukuangfj:
Are you able to find `-O0` in the `CMakeCache.txt` for `CMAKE_CXX_FLAGS_DEBUG` ?

CMakeLists.txt
```diff
@@ -13,6 +13,23 @@ cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
 
 project(k2)
 
+# ----------------- Supported build types for K2 project -----------------
+set(allowableBuildTypes Debug Release RelWithDebInfo MinSizeRel)
```
csukuangfj, line 17: what is your_style toName variables?

PR #91, opened 2020-08-24
Cuda draft2

danpovey:


danpovey:
@qindazhu  I am hoping you can help fix compilation problems?
I am making the RaggedShape and Ragged classes generic (independent of the number of axes).
ragged_shape.h will be removed but for now I keep it, for reference; the generic code will go in ragged.h.

I think the top-level interfaces are very close to finished, and the next step would be to convert the already-implemented stuff (CPU algorithms) to use this.  These will be implementations that only work for CPU contexts.
For algorithms that operate on a single Fsa, if helpful we can let the input be FsaVec but pass in the index to operate on.  The interfaces will be much simpler and more functional, i.e. they'll generally return an object (that they allocate themselves).

qindazhu:
OK, will do.

danpovey:
Great, thanks!!

PR #92, opened 2020-08-24
fix compilation issues

qindazhu:
@danpovey, note that I comment `MaxPerSublist` for now as it's the only one interface in `ops.h`  that uses `ragged.h` , I will try to implement some simple algorithms about `raggedArray` later to compile and test `ragged.h`

### DIFF
--- a/k2/csrc/cuda/array.h
+++ b/k2/csrc/cuda/array.h
@@ -20,9 +20,6 @@ class Array1 {
  public:
   int32_t Dim() const { return size_; }  // dimension of only axis (axis 0)
 
-  T *Data();  // Returns pointer to 1st elem.  Could be a GPU or CPU pointer,
-  // depending on the context.
-
   // Returns pointer to 1st elem.  Could be a GPU or CPU pointer,
   // depending on the context.
   T *Data() {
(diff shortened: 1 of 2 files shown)


k2/csrc/cuda/array.h
```diff
@@ -95,24 +92,24 @@ class Array1 {
...
+    ContextPtr c = Context();
+    assert(c->IsCompatible(indexes.GetContext()));
     int32_t ret_dim = indexes.Dim();
     Array1<T> ans(c, ret_dim);
     const T *this_data = Data();
     T *ans_data = ans.Data();
     int32_t *indexes_data = indexes.Data();
-    auto lambda_copy_elems = [=] __host__ __device__ (int32_t i) -> void {
-       ans_data[i] = this_data[indexes_data[i]];
+    auto lambda_copy_elems = [=] __host__ __device__(int32_t i) -> void {
```
csukuangfj, line 105: It would be better if you specify explicitly what to be captured in the lambda.

qindazhu: Sure, we will definitely do this when implementing other methods in `array.h`, the code from @danpovey now in `array.h` is just to delcare the interface

danpovey:
Fangjun: most of the lambdas are going to have a lot of arguments, and the
idea of listing everything doesn't seem to scale adequately.
That's why I have decided to not list the captures.

k2/csrc/cuda/array.h
csukuangfj: The signature of this function says it returns `Array1`, but this function returns nothing.

danpovey: ah, it should return `ans`

PR #93, opened 2020-08-25
Cuda draft2

danpovey:


danpovey:
@qindazhu could you please fix any new compilation problems from this?

qindazhu:
There is no new compilation issues for those commits. It's OK to merge. (Note currently we only compile op_test and util_test, so there are some other files that not included in the compilation process, we can fix issues from those files later )

PR #94, opened 2020-08-26
[WIP] CUDA error checking/debugging

songmeixu:
add debug.cuh and related changes.


songmeixu:
Guys, it's WIP, but review/advice are welcome. 

Todo:
- write gtest/usages for this
- replace current CUDA error checks (glog, not functional)

k2/csrc/cuda/debug.cuh
```diff
@@ -0,0 +1,169 @@
...
+// In debug mode and include this header,
+// define K2_MAKE_ERROR_CHECK to turn on error checking
+#if (!defined(NDEBUG) && !defined(K2_MAKE_ERROR_CHECK))
+  #define K2_MAKE_ERROR_CHECK
+#endif
+
+/**
+ * \brief A static assertion
+ * \param exp the compile-time boolean expression that must be true
+ * \param msg an error message if exp is false
```
qindazhu, line 25: The comment format is not consistent with our conventions, we use `@param` in other files and no `\brief`

k2/csrc/cuda/debug.cuh
```diff
@@ -0,0 +1,169 @@
...
+  #if (K2_PTX_ARCH == 0)
+    fprintf(stderr, "CUDA error ID=%d, NAME=%s, [%s, %d]: %s\n",
+            error, cudaGetErrorName(error),
+            filename, line,
+            cudaGetErrorString(error));
+    fflush(stderr);
+  #elif (K2_PTX_ARCH >= 200)
+    printf("CUDA error ID=%d, NAME=%s, [block (%d,%d,%d) thread (%d,%d,%d), %s, %d]: %s\n",
+           error, cudaGetErrorName(error),
+           blockIdx.z, blockIdx.y, blockIdx.x, threadIdx.z, threadIdx.y, threadIdx.x,
```
qindazhu, line 62: it's strange that the order is `z, y, x` instead of `x, y, z`. and can we add kernel name here?

qindazhu: BTW, where this function is supposed to be called? It seems it will be called with return value (`cudaError_t error`) from some cuda api, if it's not a kernel, how can we get `threadIdx` and `blockIdx` here?

k2/csrc/cuda/debug_test.cu
```diff
@@ -0,0 +1,52 @@
...
+
+TEST(DebugTest, K2CudaKernelSafeCall) {
+  K2_CUDA_KERNEL_SAFE_CALL(cudaDeviceSynchronize());
+}
+
+TEST(DebugTest, K2Dlog) {
+  int *unallocated_array = static_cast<int *>(malloc(30 * sizeof(int)));
+  // device call
+  {
+    FillContents<<<2, 3>>>(5, unallocated_array);
```
qindazhu, line 43: Can this test run successfully? It's weird that it can access host memory in kernel.

k2/csrc/cuda/debug.cuh
```diff
@@ -0,0 +1,169 @@
...
+      }
+    #ifndef __CUDA_ARCH__
+      #define K2_DLOG(format, ...) va_printf(format,__VA_ARGS__)
+    #else
+      #define K2_DLOG(format, ...) va_printf("[block (%d,%d,%d), thread (%d,%d,%d)]: " format, __VA_ARGS__)
+    #endif
+  #endif
+#endif
+
+}  // end namespace k2
```
qindazhu, line 167: according to our conventions, there's no `end`, just `} // namespace k2`

cmake/gtest_utils.cmake
```diff
@@ -0,0 +1,76 @@
+enable_testing()
+
+function(k2_add_gtest)
+  cmake_parse_arguments(
```
danpovey, line 4: This cmake code is very confusing, for me and I'm sure the majority of readers .    Can you please add some kind of comment that's aimed at someone who doesn't really understand cmake?  That explains what implications it has for usage, in terms of what flags you can pass or what behaviors it causes?

k2/csrc/cuda/debug.cuh
```diff
@@ -0,0 +1,169 @@
+// k2/csrc/cuda/errors.h
+
```
danpovey, line 2: The convention is that we're just calling files .h even if they may contain cuda code, because due to the structure of the library almost every file can potentially instantiate cuda code even if it's not explicitly present.  I know this may seem odd, but for me it's convenient as my emacs recognizes as C++, and I think it will make the files seem less scary to people.

k2/csrc/cuda/debug.cuh
danpovey: The \brief is OK with me as it's likely more correct.  I am using a kind of retarded doxygen that I don't attempt to compile.

k2/csrc/cuda/debug.cuh
danpovey: Can you please give an example of how this would be called? 
I want to see what kind of code it's intended for.
I see that there are examples in CUDA API's of device functions that return cudaError_t.  But they're probably not common... if this is not intended to be used frequently, please clarify.

k2/csrc/cuda/debug.cuh
```diff
@@ -0,0 +1,169 @@
...
+      exit(EXIT_FAILURE);
+    }
+  }
+#endif
+  return error;
+}
+
+/**
+ * \brief Macro for checking cuda error.
+ * If not cudaSuccess and K2_MAKE_ERROR_CHECK, print and exit.
```
danpovey, line 76: Be more precise: "if K2_MAKE_ERROR_CHECK is defined and e is not cudaSuccess"... and clarify that it does nothing otherwise.

cmake/gtest_utils.cmake
```diff
@@ -0,0 +1,76 @@
...
+    target_link_libraries(${tgt} PRIVATE gtest_main)
+  else()
+    target_link_libraries(${tgt} PRIVATE gmock_main)
+  endif()
+
+  add_test(NAME ${name} COMMAND $<TARGET_FILE:${tgt}>)
+endfunction()
+
+# directly give the suite name source files,
+function(k2_add_gtests test_set_name)
```
csukuangfj, line 54: Is the following template
https://github.com/danpovey/k2/blob/2a6ba24b4a7845bab757fbf7044b0d6253bcb2d8/k2/csrc/CMakeLists.txt#L29-L41
not good enough that you want to make your own?

I find this cmake script hard to read and understand.

k2/csrc/cuda/debug.cuh
```diff
@@ -0,0 +1,169 @@
...
+    #elif (K2_PTX_ARCH >= 200)
+      #define K2_DLOG(format, ...)                                            \
+        printf("[block (%d,%d,%d), thread (%d,%d,%d)]: " format, blockIdx.z,  \
+               blockIdx.y, blockIdx.x, threadIdx.z, threadIdx.y, threadIdx.x, \
+               __VA_ARGS__)
+#endif
+  #else
+    // A hack to implement the variadic printf as clang for other c-compilers,
+    // and sielence the warning
+    #pragma clang diagnostic ignored "-Wc++11-extensions"
```
csukuangfj, line 149: What happens if a user does not use clang?

What are the purposes of these two pragmas?

If they are really needed it, please wrap it with pragma push and pop.

k2/csrc/cuda/debug.cuh
```diff
@@ -0,0 +1,169 @@
...
+/**
+ * \brief Log macro for printf statements.
+ * `printf` is supported by both host and device.
+ * This Log is for debugging, the error msg is printed to the stderr.
+ *
+ * Refer to:
+ * https://github.com/NVlabs/cub/blob/c3cceac115c072fb63df1836ff46d8c60d9eb304/cub/util_debug.cuh#L109
+ */
+#if !defined(K2_DLOG)
+  #if !(defined(__clang__) && defined(__CUDA__))
```
csukuangfj, line 137: The indentation looks weird.

k2/csrc/cuda/debug.cuh
```diff
@@ -0,0 +1,169 @@
...
+ * If K2_MAKE_ERROR_CHECK is defined and error is not cudaSuccess,
+ * the corresponding error message is printed to:
+ *   - host: stderr
+ *   - device: stdout in device
+ * along with the supplied source context.
+ * It's used to make other macros convenient.
+ *
+ * \return The CUDA error.
+ */
+__host__ __device__ __forceinline__ cudaError_t K2_CUDA_DEBUG(
```
csukuangfj, line 44: The function `K2_CUDA_DEBUG` is already inside the namespace `k2`, so there is
no need to put a `k2` in it.

`K2_CUDA_DEBUG` should be a macro name, not a function name.

k2/csrc/cuda/debug.cuh
csukuangfj: I would recommend you installing `clang-format` to format your code automatically on save.

danpovey:
Mm, I'm OK with leaving the name as K2_CUDA_DEBUG just to follow the
pattern of other things like that, which are macros.

k2/csrc/cuda/debug.cuh
```diff
@@ -0,0 +1,169 @@
+// k2/csrc/cuda/errors.h
+
+// Copyright (c) 2020, Xiaomi Corporation ( authors: Meixu Song )
+
+// See LICENSE for clarification regarding multiple authors
+
+#ifndef K2_CSRC_CUDA_ERRORS_H_
```
csukuangfj, line 7: Please run the cpplint script in the `scripts` folder before committing.
It can help you spot the incorrect header guard.

k2/csrc/cuda/debug.cuh
qindazhu: > K2_CUDA_DEBUG should be a macro name, not a function name

I agree with @csukuangfj  here , it's better if we can change `K2_CUDA_DEBUG` to follow name convention for functions (as it's a function instead of macro), e.g. `CudaDebug`, as you have defined macros below (e.g. `K2_CUDA_CHECK_ERROR`) to provide `__FILE__` and `__LINE__`. I suppose function `K2_CUDA_DEBUG` will not be called directly by users.
 
BTW, I'm in favour of macros with `K2_` prefix.

danpovey:
OK.

k2/csrc/cuda/debug.cuh
csukuangfj:
```cpp
void CudaDebug(....) {....}

#define K2_CUDA_DEBUG(...) CudaDebug(...)
```

This is what I meant.

It's fine to have `K2` in the macro, but I don't think it is needed in a function which is inside the namespace k2.

k2/csrc/cuda/debug.cuh
songmeixu: Thx! I turn to learn about doxygen and will change to some suggesting conventions in the coming update.

k2/csrc/cuda/debug.cuh
songmeixu: > it's strange that the order is `z, y, x` instead of `x, y, z`. and can we add kernel name here?

As the CUDA thread/block indexes are column-major distributed. I.e. they are virtual concepts for developers. For the indexing of them, if it's print like `(A, B, C)`, then it implies `[a][b][c]` as a way to index them. Thus, according to the column-major, `a=z, b=y, c=x`. However, I can't found a convention to log them. So, anyway is ok to me.

songmeixu: > Can you please give an example of how this would be called?
> I want to see what kind of code it's intended for.
> I see that there are examples in CUDA API's of device functions that return cudaError_t. But they're probably not common... if this is not intended to be used frequently, please clarify.

OK. I am writing gtests/examples.

k2/csrc/cuda/debug.cuh
songmeixu: > The function `K2_CUDA_DEBUG` is already inside the namespace `k2`, so there is
> no need to put a `k2` in it.

I am not very certain. But if it gets called out the `k2` namespace.. as it's an inline function.

cmake/gtest_utils.cmake
songmeixu: I agree it's not quite readable. My motivation was to make this function reusable (and could keep gtest-compile out of `make` process, e.g. only compile and run if `make test`.). As simply put the above codes into a CMake module, it still requires the sources and dep-libs. If you will, please help make this function reusable?

k2/csrc/cuda/debug.cuh
qindazhu: > As the CUDA thread/block indexes are column-major distributed

IIRC，the index is row-major...anyway, I think (x,y,z) is more straightforward to most people, this is also used by most of cuda examples

qindazhu: > The K2_PTX_ARCH is used to solve this. During compiling, as to the host part, it's 0; as to the kernel part, it's not.

Sorry, I don't understand it clearly, Can you explain it with some code? I mean, How can you know it's host code or device code and annotate the code with macro `K2_PTX_ARCH` ? I did't find the related code in the PR.

k2/csrc/cuda/debug.cuh
songmeixu: Done.

k2/csrc/cuda/debug.cuh
songmeixu: Done.

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,249 @@
...
+      (__VA_ARGS__);                                  \
+      K2_CUDA_API_SAFE_CALL(cudaDeviceSynchronize()); \
+      K2_CUDA_CHECK_ERROR(cudaGetLastError());        \
+    } while (0)
+#endif
+
+/**
+ * @def K2_DLOG
+ *
+ * @brief Log macro for printf statements.
```
danpovey, line 195: Please clarify how it is supposed to be used with an example!

k2/csrc/cuda/debug.cuh
songmeixu: 😂 Please give more details. As to me, the auto-format of my IDE had aligned the macros but it made it less readable i think.

k2/csrc/cuda/debug.cuh
songmeixu: OK. I change it manually this time. And will find a way to make it take effect for my IDE.

k2/csrc/cuda/debug_test.cu
```diff
@@ -0,0 +1,104 @@
...
+
+  /* free the CPU memory */
+  free(roots);
+}
+
+TEST(DebugTest, K2DLog) {
+  int *unallocated_array = static_cast<int *>(malloc(30 * sizeof(int)));
+  // device call K2_DLOG
+  {
+    helloCUDA<<<1, 5>>>(1.2345f);
```
danpovey, line 99: Inconsistent capitalization here

k2/csrc/cuda/debug.h
songmeixu: Updated. Please see the gtest about it. Though I'm afraid it's not as good as expected, I need to know the exactly expected usages so I can make changes accordingly.

k2/csrc/cuda/debug_test.cu
songmeixu: Fixed.

k2/csrc/cuda/debug.cuh
songmeixu: > > The K2_PTX_ARCH is used to solve this. During compiling, as to the host part, it's 0; as to the kernel part, it's not.
> 
> Sorry, I don't understand it clearly, Can you explain it with some code? I mean, How can you know it's host code or device code and annotate the code with macro `K2_PTX_ARCH` ? I did't find the related code in the PR.

It's defined in `cuda/arch.h` with comments.

k2/csrc/cuda/debug.h
danpovey: @songmeixu please make sure there is an example in the documentation!  People won't know to look in the test file.

k2/csrc/cuda/debug_test.cu
```diff
@@ -0,0 +1,104 @@
...
+    /* print out 100 evenly spaced square roots just to see that it worked */
+    unsigned int i;
+    for (i = 0; i < N; i += (N / 100)) {
+      K2_DLOG("sqrt(%d) = %lf\n", i, roots[i]);
+    }
+  }
+
+  /* free the CPU memory */
+  free(roots);
+}
```
qindazhu, line 93: seems ther's memory leak for GPU memory, i.e. gpu_roots

k2/csrc/cuda/debug_test.cu
```diff
@@ -0,0 +1,104 @@
...
+
+  /* free the memory */
+  K2_CUDA_API_SAFE_CALL(cudaFree(gpu_roots));
+
+  // host call K2_DLOG
+  {
+    /* print out 100 evenly spaced square roots just to see that it worked */
+    unsigned int i;
+    for (i = 0; i < N; i += (N / 100)) {
+      K2_DLOG("sqrt(%d) = %lf\n", i, roots[i]);
```
qindazhu, line 87: maybe something just like `assert(i, roots[i])` is better than print values here

k2/csrc/cuda/debug_test.cu
```diff
@@ -0,0 +1,104 @@
...
+  free(roots);
+}
+
+TEST(DebugTest, K2DLog) {
+  int *unallocated_array = static_cast<int *>(malloc(30 * sizeof(int)));
+  // device call K2_DLOG
+  {
+    HelloCUDA<<<1, 5>>>(1.2345f);
+    FillContents<<<2, 3>>>(5, unallocated_array);
+  }
```
qindazhu, line 101: free(unallocated_array)

csukuangfj: Why not to use `std::array` or `std::vector`? They manage the memory for you.

danpovey: I guess this uses unified memory... is this on by default, and does it impose a constraint on compute capability?
I had no plans to make use of unified memory, actually.  but OK for testing.

qindazhu: Actually it's not unified memory, unified memory should be allocated with `cudaMallocManage`. BTW, I think we should not use unified memory even in test code, as for our code, all memory should be allocated by using `Context`.

danpovey:
So how does the CUDA code successfully write to the CPU memory?

qindazhu:
> So how does the CUDA code successfully write to the CPU memory?

As I comment above, I think the current code can not be run successfully as kernel tries to access host memory directly, I guess this is only WIP test code. Maybe @songmeixu is the best one to answer this.

k2/csrc/cuda/debug_test.cu
csukuangfj: What I meant is to use `std::array` or `std::vector` to replace `malloc`. If it needs to be accessed by GPU,
we can use `cudaMalloc` and `cudaMemcpy` to make it available to GPU.

k2/csrc/cuda/debug.h
```diff
@@ -176,7 +176,7 @@ __host__ __device__ __forceinline__ cudaError_t K2_CUDA_DEBUG(
  *
  * @remark
  *  macro `__VA_ARGS__` is used to pass the kernel<<<...>>> as one argument,
- *  otherwise the compiler rise a error "passed 2 arguments, but takes just 1".
+ *  otherwise the compiler rises a error "passed 2 arguments, but takes just 1".
```
csukuangfj, line 179: raises an error

k2/csrc/cuda/debug_test.cu
songmeixu: Fixed.

k2/csrc/cuda/debug.cuh
songmeixu: Done.

danpovey:
Thanks!  Will look tomorrow...

cmake/gtest_utils.cmake
songmeixu: Hmm, maybe I can change another way to reuse the gtest function. But for now, as to add a CMake module doesn't break things, I would like to put it here for convenience.

k2/csrc/cuda/debug.cuh
songmeixu: I made a mistake with the comments. This `pragma` is for clang, and you are right, a push and pop guard is added.

k2/csrc/cuda/debug.cuh
songmeixu: OK, I think I know what you mean. An indent error here is fixed now.

k2/csrc/cuda/debug.cuh
songmeixu: > > K2_CUDA_DEBUG should be a macro name, not a function name
> 
> I agree with @csukuangfj here , it's better if we can change `K2_CUDA_DEBUG` to follow name convention for functions (as it's a function instead of macro), e.g. `CudaDebug`, as you have defined macros below (e.g. `K2_CUDA_CHECK_ERROR`) to provide `__FILE__` and `__LINE__`. I suppose function `K2_CUDA_DEBUG` will not be called directly by users.
> 
> BTW, I'm in favour of macros with `K2_` prefix.

The inline function name changed from "K2_CUDA_DEBUG" to "_K2DebugLog", "_" indicates it's an inner func.

k2/csrc/cuda/debug.h
songmeixu: Now, the usage example is put @code tag as doxygen suggests.

k2/csrc/cuda/debug_test.cu
songmeixu: gtest `EXPECT_EQ` is now used to check this.

k2/csrc/cuda/debug_test.cu
songmeixu: > free(unallocated_array)

Fixed.

songmeixu: > Why not to use `std::array` or `std::vector`? They manage the memory for you.

Do you mean use thrust::vector for device? Anyway, this code is a unittest. I like thrust, but it hides kernel so that it's not good at testing.

songmeixu: > What I meant is to use `std::array` or `std::vector` to replace `malloc`. If it needs to be accessed by GPU,
> we can use `cudaMalloc` and `cudaMemcpy` to make it available to GPU.

I see. As the device code like C-style data argument, and it's not an important thing.. So, I add the missing `free` and keep C-style here.

songmeixu:
> So how does the CUDA code successfully write to the CPU memory?

It's an error and gets fixed. SORRY guys would be more on-time on-call for issues.

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,383 @@
...
+
+#include <stdio.h>
+#include <assert.h>
+
+#include "k2/csrc/cuda/arch.h"
+
+namespace k2 {
+
+/**
+ * @brief Define a macro `K2_MAKE_ERROR_CHECK` that turn on bellow macros
```
danpovey, line 33: `the macros below`.

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,383 @@
...
+ *  as cuda kernel launches are asynchronous also,
+ *  the same macro for kernel could be used. See
+ *  `K2_CUDA_KERNEL_SAFE_CALL`.
+ *
+ * @code{.cpp}
+ * K2_CUDA_API_SAFE_CALL(CUDA_RUNTIME_API());
+ * @endcode
+ */
+#ifndef K2_CUDA_API_SAFE_CALL
+  #ifdef K2_MAKE_ERROR_CHECK
```
danpovey, line 228: I don't understand why `K2_MAKE_ERROR_CHECK` is named this way.  What does `MAKE` signify?  
This only seems to get turned on if NDEBUG is not defined... how is it different from NDEBUG?

k2/csrc/cuda/arch.h
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @brief
+ *  K2_PTX_ARCH get the target PTX version through the active compiler.
+ *
+ * @details
+ *  - for host, it's 0.
+ *  - for device, it's `__CUDA_ARCH__`, which indicates the
+ *                compute compatibility, should >= 200.
+ */
+#ifndef K2_PTX_ARCH
```
danpovey, line 22: You have macro guards around each definition of a macro.  Can't you just use an include guard around the entire header?  Wouldn't that be simpler?

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,383 @@
...
+ * K2_CUDA_KERNEL_SAFE_CALL(kernel_func<<<...>>>());
+ * @endcode
+ */
+#ifndef K2_CUDA_KERNEL_SAFE_CALL
+  #ifdef K2_MAKE_ERROR_CHECK
+    #define K2_CUDA_KERNEL_SAFE_CALL(...)               \
+      do {                                              \
+        K2_CUDA_CHECK_ERROR(cudaGetLastError());        \
+        (__VA_ARGS__);                                  \
+        K2_CUDA_API_SAFE_CALL(cudaDeviceSynchronize()); \
```
danpovey, line 269: Even if NDEBUG is defined I think we should still be checking for the cuda error code
(I think it's OK to do this either before or after the kernel invocation, but not both).

k2/csrc/cuda/debug.h
songmeixu: If NDEBUG, and `K2_MAKE_ERROR_CHECK` is defined manually before include "debug.h". The macros still take effect (except assert that under NDEBUG control). The macro name is bad and could be changed.

k2/csrc/cuda/arch.h
songmeixu: The include guard is already there. The intension is to avoid conflict with following redefine the same name macro outer the header. Though it rarely happens, sometimes could, e.g. conflict with `K2_PARANOID_ASSERT` from kaldi-error.h, `LOG()` from glog.

k2/csrc/cuda/debug.h
songmeixu: After `GetLastError()`, the error flag is set to cudaSuccess. Thus, check before a kernel is to rule out before errors. As the kernel is async, it should not be in production codes. As to what you mean, the `K2_MAKE_ERROR_CHECK` is used in this situation, define which manually in any situation. These error checks are turned on. And by default, its effect is the same as `NDEBUG`, I mean it only activate in DEBUG mode.

k2/csrc/cuda/debug.h
songmeixu: Done.

k2/csrc/cuda/debug.h
songmeixu: Done

k2/csrc/cuda/arch.h
danpovey: Macros that start with K2_ don't need include guards as they're specific to this project.

k2/csrc/cuda/debug.h
danpovey: OK, can you please simplify this by removing K2_MAKE_ERROR_CHECK and just relying on NDEBUG?  Less confusing.
But I want you to check the cuda error codes from kernels once (say, after the kernel invocation, although it doesn't matter) even if NDEBUG is defined, because we need to catch errors at *some* point, even if late.

songmeixu: Maybe I didn't express clearly, but what I meant was that `K2_MAKE_ERROR_CHECK` is just what satisfies what you're saying. `K2_MAKE_ERROR_CHECK` regardless of whether `NDEBUG` is defined or not, if `K2_MAKE_ERROR_CHECK` is defined manually, then it will definitely do error checking.

songmeixu: Check only once after kernel is ok. Changes are made.

k2/csrc/cuda/arch.h
```diff
@@ -0,0 +1,32 @@
+// k2/csrc/cuda/arch.h
+
+// Copyright (c) 2020 Xiaomi Corporation ( authors: Meixu Song )
+
+// See ../../LICENSE for clarification regarding multiple authors
+#ifndef K2_CSRC_CUDA_ARCH_H_
+#define K2_CSRC_CUDA_ARCH_H_
+
+namespace k2 {
```
csukuangfj, line 9: Does it make sense to use a namespace here?

k2/csrc/cuda/arch.h
```diff
@@ -0,0 +1,32 @@
...
+ * @details
+ *  - for host, it's 0.
+ *  - for device, it's `__CUDA_ARCH__`, which indicates the
+ *                compute compatibility, should >= 200.
+ */
+#ifndef K2_PTX_ARCH
+  #ifndef __CUDA_ARCH__
+    #define K2_PTX_ARCH 0
+  #else
+    #define K2_PTX_ARCH __CUDA_ARCH__
```
csukuangfj, line 26: Why not to use `__CUDA_ARCH__` directly in k2?

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,382 @@
...
+ * @param[in] msg an error message if exp is false
+ *
+ * @note `static_assert` is supported by both of host and device.
+ *
+ * @code{.cpp}
+ * K2_STATIC_ASSERT(DEFINED_SHAPE % DEFINED_X == 0);
+ * @endcode
+ */
+#ifndef K2_STATIC_ASSERT
+#define K2_STATIC_ASSERT(exp, msg) static_assert(exp, msg)
```
csukuangfj, line 55: `static_assert` is part of C++11 standard. What is the reason to add another indirection here?

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,382 @@
...
+ *  `assert` can be disabled at compile time by defining the `NDEBUG`
+ *  preprocessor macro before including assert.h.
+ *  Thus, `K2_MAKE_ERROR_CHECK` is ignored here.
+ *
+ *  @code{.cpp}
+ *  K2_CHECK_EQ(1, 1);
+ *  @endcode
+ */
+#ifndef K2_CHECK_EQ
+#define K2_CHECK_EQ(a, b) assert( a == b )
```
csukuangfj, line 109: Please **always** add parenthesis to your macro arguments.

```cpp
assert((a) == (b))
```

k2/csrc/cuda/debug.h
danpovey: agreed, I guess we can just use static_assert.

k2/csrc/cuda/debug.h
csukuangfj: So we do not need something like `K2_CHECK_EQ(a, b) << "some error message"` ?

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,382 @@
...
+ * @def K2_CUDA_CHECK_ERROR(cudaError)
+ *
+ * @brief Macro for checking cuda error.
+ *
+ * @details
+ *  If `K2_MAKE_ERROR_CHECK` is defined and e is not cudaSuccess,
+ *  print the error message and exit with the error enum value.
+ *  Otherwise, it does nothing.
+ *
+ * @param[in] e an enum type indicating CUDA errors.
```
danpovey, line 178: use the type please, e.g. cudaError_t.  There are many such enum's in CUDA and they are not all the same.

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,382 @@
...
+#ifdef K2_MAKE_ERROR_CHECK
+  (void)filename;
+  (void)line;
+  if (cudaSuccess != error) {
+  #if (K2_PTX_ARCH == 0)
+    fprintf(stderr, "CUDA error ID=%d, NAME=%s, [%s, %d]: %s\n",
+            error, cudaGetErrorName(error),
+            filename, line,
+            cudaGetErrorString(error));
+    fflush(stderr);
```
csukuangfj, line 153: I don't think you need `fflush` here. `stderr` is unbuffered by default.

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,382 @@
...
+ *
+ * @todo
+ *  Should the error code return by this macro?
+ *
+ * @code{.cpp}
+ * K2_CUDA_CHECK_ERROR(cudaGetLastError());
+ * @endcode
+ */
+#ifndef K2_CUDA_CHECK_ERROR
+  #ifdef K2_MAKE_ERROR_CHECK
```
csukuangfj, line 194: no need to indent for nested marcos.

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,382 @@
...
+    #define K2_CUDA_CHECK_ERROR(e)                                    \
```
csukuangfj, line 195: What happens if someone uses

```cpp

if (xxx)  K2_CUDA_CHECK_ERROR(blabla)
else {some statements here}.
```

----

Please wrap your `if` inside `do {...} while(0)`

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,382 @@
...
+      if (::k2::_K2CudaDebug((cudaError_t)(e), __FILE__, __LINE__)) { \
+        exit(e);                                                      \
```
csukuangfj, line 197: I would suggest `abort()`.

----

By the way, `__func__` represents the function name. It can be included in the output message.

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,382 @@
...
+      #define K2_DLOG(format, ...) printf(format,__VA_ARGS__)
+    #elif (K2_PTX_ARCH >= 200)
+      #define K2_DLOG(format, ...)                                            \
+        printf("[block (%d,%d,%d), thread (%d,%d,%d)]: " format, blockIdx.x,  \
+               blockIdx.y, blockIdx.z, threadIdx.x, threadIdx.y, threadIdx.z, \
+               __VA_ARGS__)
+    #endif
+  #else
+    /**
+     * A hack to implement the variadic printf for clang,
```
csukuangfj, line 308: Why not to use `vprintf` or `vfprintf`? They are from the standard C library
and does not need special compiler support.

k2/csrc/cuda/debug_test.cu
```diff
@@ -0,0 +1,190 @@
...
+
+/* number of threads per block (max 1,024) */
+#define THREADS 200
+
+/* N is now the number of blocks times the number of threads
+   in this case it is 10 million */
+#define N (BLOCKS * THREADS)
+
+/* the kernel fills an array up with square roots */
+__global__ void square_roots(double* array) {
```
csukuangfj, line 39: Please follow the code style and rename your function name.

csukuangfj: Please pay attention to where to put the star. It is `double *array`, not `double* array`.

k2/csrc/cuda/debug_test.cu
```diff
@@ -0,0 +1,190 @@
...
+  array[idx] = sqrt((double) idx);
+}
+
+__global__ void HelloCUDA(float f) {
+  if (threadIdx.x == 0)
+    K2_DLOG("Hello thread %d, f=%f\n", threadIdx.x, f);
+}
+
+// A vector add kernel definition
+__global__ void VecAdd(float* A, float* B, float* C) {
```
csukuangfj, line 59: Please add `const` to function arguments that you do not intend to modify inside the function.

k2/csrc/cuda/debug_test.cu
```diff
@@ -0,0 +1,190 @@
...
+TEST(DebugTest, K2CudaCheckError) {
+  ASSERT_DEATH(K2_CUDA_CHECK_ERROR(cudaErrorMemoryAllocation),
+               "cudaErrorMemoryAllocation");
+}
+
+TEST(DebugTest, K2CudaSafeCall) {
+
+  /* store the square roots of 0 to (N-1) on the CPU
+   * stored on the heap since it's too big for the stack for large values of N */
+  double* roots = (double*) malloc(N * sizeof(double));
```
csukuangfj, line 89: 1. Please use `reinterpret_cast`.
2. This is c++ code and there is no need to follow the c style. Once again,
I would suggest you using either `new double[N]` or `std::vector<double> roots(N)`;

k2/csrc/cuda/debug_test.cu
```diff
@@ -0,0 +1,190 @@
...
+  /* copy the data back */
+  K2_CUDA_API_SAFE_CALL(cudaMemcpy(roots, gpu_roots, N * sizeof(double),
+                        cudaMemcpyDeviceToHost));
+
+  /* free the memory */
+  K2_CUDA_API_SAFE_CALL(cudaFree(gpu_roots));
+
+  // host call K2_DLOG
+  {
+    /* print out 100 evenly spaced square roots just to see that it worked */
```
csukuangfj, line 108: What does the comment mean?

 Where is 100 in the code?

k2/csrc/cuda/debug_test.cu
```diff
@@ -0,0 +1,190 @@
...
+    unsigned int i = 100000;
+    testing::internal::CaptureStdout();
+    K2_DLOG("sqrt(%d) = %lf\n", i, roots[i]);
+    std::string output = testing::internal::GetCapturedStdout();
+    EXPECT_EQ(output, "sqrt(100000) = 316.227766\n");
```
csukuangfj, line 113: Did you compute `sqrt(100000)` by hand? why not to let the computer compute
and check the result for you?

k2/csrc/cuda/debug_test.cu
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  /* free the CPU memory */
+  free(roots);
+}
+
+TEST(DebugTest, K2DLog) {
+  int *h_A = static_cast<int *>(malloc(6 * sizeof(int)));
```
csukuangfj, line 121: what is `h_A` for? Where is it used?

cmake/gtest_utils.cmake
```diff
@@ -0,0 +1,75 @@
+enable_testing()
+
+function(k2_add_gtest)
+  cmake_parse_arguments(
+      ARGS                                 # prefix of output variables
+      "NO_MAIN;NO_GMOCK;EXCLUDE_FROM_ALL"  # list of names of the boolean arguments (only defined ones will be true)
+      ""                                   # list of names of mono-valued arguments
+      "SOURCES;DEPENDS"                    # list of names of multi-valued arguments (output variables are lists)
```
csukuangfj, line 8: I think `gtest`, `gmock`, `gtest_main` and `gmock_main` all can be passed via
`DEPENDS` and there is no need to use separate options for `NO_MAIN`, `NO_GMOCK`,
which can simplify this function significantly.

cmake/gtest_utils.cmake
```diff
@@ -0,0 +1,75 @@
...
+
+  if(ARGS_EXCLUDE_FROM_ALL)
+    set(EXCLUDE_FROM_ALL "EXCLUDE_FROM_ALL")
+  endif()
+
+  string(REGEX REPLACE "^.*\\." "" tgt ${name})
+  add_executable(${tgt} ${sources} ${EXCLUDE_FROM_ALL})
+
+  if(ARGS_NO_GMOCK)
+    target_link_libraries(${tgt} PUBLIC ${dependencies} gtest)
```
csukuangfj, line 35: why is it `PUBLIC` for `gtest` and `gmock`, but `private` for the subsequent `gtest_main` and `gmock_main`?

k2/csrc/cuda/CMakeLists.txt
```diff
@@ -1,28 +1,15 @@
-add_library(context context.cu)
-target_include_directories(context PUBLIC ${CMAKE_SOURCE_DIR})
-target_link_libraries(context PUBLIC cub)
-target_link_libraries(context PUBLIC glog)
+include_directories(${CMAKE_SOURCE_DIR})
```
csukuangfj, line 1: No need to add this statement. It can be removed.

k2/csrc/cuda/debug_test.cu
songmeixu: Actually, the checking string was got from stdout. The gtest can pass verifies this.

k2/csrc/cuda/CMakeLists.txt
```diff
@@ -1,28 +1,15 @@
...
-foreach(name IN LISTS cuda_tests)
-  k2_add_cuda_test(${name})
-endforeach()
+add_library(k2_cuda ${CUDA_SOURCE})
+target_include_directories(k2_cuda PUBLIC ${CMAKE_SOURCE_DIR})
+target_link_libraries(k2_cuda PUBLIC cub)
+target_link_libraries(k2_cuda PUBLIC glog)
+
+k2_add_gtests(Test.Cuda
+    SOURCES debug_test.cu utils_test.cu #ops_test.cu
```
csukuangfj, line 13: please put the source filenames in a variable, like the above `CUDA_SOURCE`.

k2/csrc/cuda/debug.cuh
csukuangfj: All functions in the namespace k2, except this one, do not contain `k2`, so what is the reason for adding k2 to this function?

k2/csrc/cuda/debug_test.cu
songmeixu: Fixed. Last gtest example changed, it's the old comment.

k2/csrc/cuda/debug_test.cu
songmeixu: Yeah, it's useless. 
Fixed.

k2/csrc/cuda/debug_test.cu
csukuangfj: I would suggest

```cpp
EXPECT_NEAR(roots[100000], sqrt(100000), 1e-5);
```

or put the `EXPECT_NEAR` inside a for loop to check all elements.

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,382 @@
...
+ * @code{.cpp}
+ * _K2CudaDebug(cudaGetLastError(), __FILE__, __LINE__);
+ * @endcode
+ */
+__host__ __device__ __forceinline__ cudaError_t _K2CudaDebug(
+    cudaError_t error,
+    const char *filename,
+    int line) {
+#ifdef K2_MAKE_ERROR_CHECK
+  (void)filename;
```
csukuangfj, line 145: Why do you put these two lines here?

k2/csrc/cuda/debug.h
csukuangfj: > What does MAKE signify?

I guess it means `enable`.

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,382 @@
...
+      inline __host__ __device__
+      void va_printf(char const* format, Args const&... args) {
+    #ifdef __CUDA_ARCH__
+        printf(format, blockIdx.x, blockIdx.y, blockIdx.z,
+               threadIdx.x, threadIdx.y, threadIdx.z, args...);
+    #else
+        printf(format, args...);
+    #endif
+      }
+    #ifndef __CUDA_ARCH__
```
csukuangfj, line 324: As you use `__CUDA_ARCH__` here directly, I really don't understand why you define
`K2_PTX_ARCH`. Could you please explain?

k2/csrc/cuda/debug_test.cu
csukuangfj: Not sure whether `style` is an important thing.

cmake/gtest_utils.cmake
songmeixu: Done.

cmake/gtest_utils.cmake
songmeixu: The PRIVATE keep the most depends on the target. It's suggested to compile executables.

songmeixu: _main would lead to an executable.

k2/csrc/cuda/CMakeLists.txt
songmeixu: Done.

k2/csrc/cuda/CMakeLists.txt
songmeixu: In this module, cmake_parse_arguments is used, which accepts these into a set. So we don't need to do it out of func.

k2/csrc/cuda/debug.h
danpovey: Agreed, I think it's better to use __CUDA_ARCH__ instead- simpler and  more standard.

k2/csrc/cuda/debug.h
songmeixu: They should suppress the warning: variables not used. But I made a mistake here. These two lines should be out of the above macro. 
Fixed

k2/csrc/cuda/debug.h
songmeixu: The __CUDA_ARCH_ is not 0 for the host side, it's undefined.

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,382 @@
...
+ *  https://github.com/NVlabs/cub/blob/ \
+ *  c3cceac115c072fb63df1836ff46d8c60d9eb304/cub/util_debug.cuh#L109
+ *
+ * @code{.cpp}
+ * K2_DLOG("Value is %d, string is %s ..", i, str);
+ * @endcode
+ */
+#if !defined(K2_DLOG)
+  #if !(defined(__clang__) && defined(__CUDA__))
+    #if (K2_PTX_ARCH == 0)
```
danpovey, line 298: Here can just do #ifndef __CUDA_ARCH__

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,382 @@
...
+      #define K2_DLOG(format, ...) printf(format,__VA_ARGS__)
+    #elif (K2_PTX_ARCH >= 200)
```
danpovey, line 300: and here: 
```
 #if (defined(__CUDA_ARCH__ ) && (__CUDA_ARCH__ >= 200)
```
(the c preprocessor si OK with this).  so no need for K2_PTX_ARCH.  Keep it simple!

k2/csrc/cuda/debug.h
danpovey: see my other comment, can do if defined(x) && x > 4

k2/csrc/cuda/arch.h
danpovey: can remove arch.h...

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,382 @@
...
+namespace k2 {
+
+/**
+ * @brief Define a macro `K2_MAKE_ERROR_CHECK` that turn on the macros bellow
+ *        to make errors get checked, if NDEBUG is undefined.
+ *
+ * @details If in NDEBUG mode, all macros here would do nothing.
+ */
+#if (!defined(NDEBUG) && !defined(K2_MAKE_ERROR_CHECK))
+#define K2_MAKE_ERROR_CHECK
```
danpovey, line 39: ..remove this..

k2/csrc/cuda/debug.h
songmeixu: Done

songmeixu:
According to the above discussion, I get a core principle, that to keep codes simple or minimal for usage. So I made some changes as above.

k2/csrc/cuda/debug.h
songmeixu: Done.

k2/csrc/cuda/debug.h
songmeixu: Done.

k2/csrc/cuda/debug.h
songmeixu: Done.

k2/csrc/cuda/debug_test.cu
songmeixu: This test is to check it do print to stdout. So the value is not important. To meet what you say, need to parse the stdout, it wastes my time.

k2/csrc/cuda/debug_test.cu
songmeixu: Hmm, I don't want to keep change this kind of what's your favorite thing. And it's not cpp code as its filename indicate, and the target error comes from device.

k2/csrc/cuda/debug_test.cu
songmeixu: This function and its argument init are from <<CUDA C Programming Guide>> Chapter 2. I would like to untouch it.

k2/csrc/cuda/debug_test.cu
songmeixu: Done.

k2/csrc/cuda/debug.h
danpovey: @songmeixu I'd rather have it print out the error message, including the values of a and b,  and line numbers, if it fails.  This will make debugging much easier.

danpovey: .. also of course there will be other similar macros like K2_CHECK, K2_CHECK_GE, K2_CHECK_LT and so on.

k2/csrc/cuda/debug.h
csukuangfj: They are used in `printf` or `fprintf`; why will the compiler complain?

csukuangfj:
Just want to make sure that any C++ code checked into K2 is of high quality.

If you have read the comment, you can find that a large part of them are about **code style**. Not sure whether you have
read the google c++ code style guide or have read the K2 source code. If you read your own code carefully, you can find
that even your own code has different styles, e.g., function names,  function arguments, placement of `const` and `*` for variables, indentation of macros, etc. It is totally fine that you have your own style. But for an existing code base, I think
the code becomes less readable if it contains different styles.

> I agree it's not quite readable.
>  as to add a CMake module doesn't break things, I would like to put it here for convenience.

To get things done is one thing, readability is also very important. You'll make the later maintainer cry.

I've read the code from <https://github.com/XiaoMi/mace> and I guess people from Xiaomi also care about code readability
and consistency.

csukuangfj:
I would be grateful if someone can take his/her time to review my code.

k2/csrc/cuda/debug_test.cu
```diff
@@ -56,7 +56,7 @@ __global__ void HelloCUDA(float f) {
 }
 
 // A vector add kernel definition
-__global__ void VecAdd(float* A, float* B, float* C) {
+__global__ void VecAdd(float *A, float *B, float *C) {
```
csukuangfj, line 59: I find that you have deleted the comment about adding `const` for the two input arguments `A` and `B`. 

The signature of this function tells us `A` and `B` will be changed inside the function,
but actually they are not. So it is really confusing for such a signature without `const`.

I would give a -1 vote.

k2/csrc/cuda/debug.h
danpovey: space after comma (I'm trying to follow Google C++ style generally.

k2/csrc/cuda/debug.h
danpovey: .. I meant, mention the type cudaError_t when you document the variable

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,313 @@
...
+ *  This code is refered from:
+ *  https://github.com/NVlabs/cub/blob/ \
+ *  c3cceac115c072fb63df1836ff46d8c60d9eb304/cub/util_debug.cuh#L109
+ *
+ * @code{.cpp}
+ * K2_DLOG("Value is %d, string is %s ..", i, str);
+ * @endcode
+ */
+#ifndef __CUDA_ARCH__
+  #define K2_DLOG(format, ...) printf(format,__VA_ARGS__)
```
danpovey, line 259: space after comma

k2/csrc/cuda/debug_test.cu
danpovey: Agreed, should use const where possible.
Also please add the extra checking macros like K2_CHECK_GE and so on.

k2/csrc/cuda/debug.cuh
songmeixu: > All functions in the namespace k2, except this one, do not contain `k2`, so what is the reason for adding k2 to this function?

I know what you mean, but you may not try to get my point. For example:
```cpp
#include <iostream>

namespace k2 {
  inline void Print() {
    std::cout << "Hello World" << std::endl;
  }

#define PP Print()
}

int main () {
  PP;
  return 0;
}
```

compile and get error:
```bash
note: 'k2::Print' declared here
  inline void Print() {
              ^
1 error generated. 
```

k2/csrc/cuda/debug_test.cu
songmeixu: 😂 That what I want to say. The thing is you wanna me change to the other style... and you know it's not an important thing.

k2/csrc/cuda/debug.h
songmeixu: Well, as it was that situation, but during our latest comments, I removed the K2_MAKE_ERROR_CHECK as Dan required.

k2/csrc/cuda/debug.h
songmeixu: done

k2/csrc/cuda/debug.h
songmeixu: Done.

songmeixu:
I do read the K2 code and google style is always my favorite. I made style check as pre-commit/push requirement for one branch about K2 called cppyy in my fork. And I know code review is valuable. I appreciate it. And as many company code repos, the serious code cannot get pass w/o +2 vote after automatic style checking. 

---

However, you must know my point: **I don't feel the same standard is work here for everyone every commits**. I don't find the master branch is all in right and consistent style, isn't it? And in the case that C-styles or Cpp-style, why I need to change to the other way? And even the static checker wouldn't remind the const thing for c-style. To be perfect, won't `const &` instead of `const *` be preferred for cpp-style? Thus, to be perfect, a lot of things need to change. I really appreciate the review work. I'm not a new-bee. In the case that I have explained I can't take advantage of automatic tools right now (b/c my IDE and be short of time). I really don't want to manually make it PERFECT in the current situation. As Dan knows, compare to the style thing, some implementation definitely with faults (e.g. the printf in _K2CudaDebug would delay cuda, but we need it for development), but to be simple and on-time, workable/minimal coding is running here as I observed. 

---

However, I am definitely the guy pursues the clean code (I dare say I know the most tools that help), but with the least man effort and strong requirement from tools to get the goal, instead of all by man write and check. As I planed, I should already be working on cppyy and made it help us at this weekend. But I took much time to be a perfect coder.

---

I apology for my words. But I insist on tools instead of man should take these work. SORRY to be rude. You would know I am usually a kind man, just sometimes I know there is a better way then I cannot step the road underfoot.

k2/csrc/cuda/debug_test.cu
songmeixu: I didn't delete things or IDK what you mean. The comment of issue or code, as to the issue it's still above.. As to the code, I post the page of doc:
image

songmeixu:
Sorry to be some rube. @csukuangfj. Found I was kind of unsatisfying with myself's progress. Feel not right now, would hug you next meet. Let's look ahead. Please keep your way, you're right. Thx!!

csukuangfj:
Not everything from NVIDIA is that right.

I think you have the ability to change the demo code to your needs instead
of just copying and pasting.

k2/csrc/cuda/CMakeLists.txt
csukuangfj: Some time later, this line will contain tens of cuda source filenames, which harms readability.

k2/csrc/cuda/debug.cuh
csukuangfj: Could you use

```cpp
#define PP k2::Print()
```

The compiler gives a hint that you can use `k2::Print` in the macro.

k2/csrc/cuda/debug_test.cu
csukuangfj: this is about code style; it can help you reduce the number of programming errors, e.g., memory leaks.

Anyway, you can leave it as is; my comments are just advice.

---

By the way, I find that a large part of your code (even the comment) is just copying and pasting from someplace.
That may explain why your code contains different styles. Please leave a notice indicating where the code is from if it is copied verbatim.

k2/csrc/cuda/debug.h
danpovey: @songmeixu there are some comments here you need to address: (1) using parentheses, (2) printing out the values of a and b and the line numbers if it fails.
I'm thinking that if we want a regular assert that just dies, we can use assert().

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,226 @@
...
+ *  error enum value. Otherwise, it does nothing.
+ *
+ * @param[in] e an enum type indicating CUDA errors.
+ *
+ * @code{.cpp}
+ * K2_CUDA_CHECK_ERROR(cudaGetLastError());
+ * @endcode
+ */
+#define K2_CUDA_CHECK_ERROR(e)                                    \
+  if (::k2::_K2CudaDebug((cudaError_t)(e), __FILE__, __LINE__)) { \
```
danpovey, line 131: This will evaluate e twice!   Not good!

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,226 @@
...
+ *  (May assgin to nvcc to take control through change cmake
+ *  compiler and options. Then, it need cmake-3.18 `FindCUDAToolkit`
+ *  or other config helpers to make things easy.)
+ */
+#include <cuda_runtime.h>
+#include <cuda_runtime_api.h>
+#include <device_launch_parameters.h>
+#include <driver_types.h>
+
+#include <stdio.h>
```
qindazhu, line 25: use `cstdio`?

songmeixu: Done.

danpovey:
Thanks!  Merging.

danpovey:
Thanks!  Merging.

k2/csrc/cuda/debug.h
```diff
@@ -0,0 +1,226 @@
...
+#include <assert.h>
```
qindazhu, line 26: Is this header file  from cuda? or from c++ library `<cassert>`? I know there's an `assert` function in cuda header file, but I don't know how `nvcc` decides which one to use.

k2/csrc/cuda/debug.h
qindazhu: Noticed that `assert` in cuda has printed some error messages, e.g. 
```
test.cu:19: void testAssert(): block: [0,0,0], thread: [0,0,0] Assertion `should_be_one` failed.
```
so maybe we need to avoid the duplicate of line numbers when printing error messages.

danpovey:
Meixu, I just noticed this is to the wrong branch!
Should be to cuda_draft :-(

Issue #96, opened 2020-08-28
A compile error of k2/k2/csrc/cuda/ops_test.cu


Jarvan-Wang:
OS version is Linux 3.10.0-957.el7.x86_64
nvcc is 10.2
gcc is 6.3.1

Error is here:
`<pre>[ 94%] <font color="#4E9A06">Building CUDA object k2/csrc/cuda/CMakeFiles/ops_test.dir/ops_test.cu.o</font>
cd /search/odin/wangjiawen/k2/build/k2/csrc/cuda &amp;&amp; /usr/local/cuda-10.2/bin/nvcc -forward-unknown-to-host-compiler -DGOOGLE_GLOG_DLL_DECL=&quot;&quot; -DGOOGLE_GLOG_DLL_DECL_FOR_UNITTESTS=&quot;&quot; -I/search/odin/wangjiawen/k2 -I/search/odin/wangjiawen/k2/build/_deps/cub-src -I/search/odin/wangjiawen/k2/build/_deps/glog_glog-build -I/search/odin/wangjiawen/k2/build/_deps/glog_glog-src/src -I/search/odin/wangjiawen/k2/build/_deps/googletest-src/googlemock/include -isystem=/search/odin/wangjiawen/k2/build/_deps/googletest-src/googletest/include -isystem=/search/odin/wangjiawen/k2/build/_deps/googletest-src/googletest --expt-extended-lambda -gencode arch=compute_30,code=sm_30 --expt-extended-lambda -gencode arch=compute_32,code=sm_32 --expt-extended-lambda -gencode arch=compute_35,code=sm_35 --expt-extended-lambda -gencode arch=compute_50,code=sm_50 --expt-extended-lambda -gencode arch=compute_52,code=sm_52 --expt-extended-lambda -gencode arch=compute_53,code=sm_53 --expt-extended-lambda -gencode arch=compute_60,code=sm_60 --expt-extended-lambda -gencode arch=compute_61,code=sm_61 --expt-extended-lambda -gencode arch=compute_62,code=sm_62 --expt-extended-lambda -gencode arch=compute_70,code=sm_70 --expt-extended-lambda -gencode arch=compute_72,code=sm_72 -g -std=c++14 -x cu -c /search/odin/wangjiawen/k2/k2/csrc/cuda/ops_test.cu -o CMakeFiles/ops_test.dir/ops_test.cu.o
[...]

PR #97, opened 2020-08-28
A compile error of k2/k2/csrc/cuda/ops_test.cu #96

Jarvan-Wang:
* Fix syntax error of ops_test.cu

* Add a const version of k2::Array1::Context to fix quantifier error of array.h


qindazhu:
It would be appreciated if you can change the target branch from `master` to `cuda_draft`, currently `cuda_draft` is the branch we are working on about cuda.

Jarvan-Wang:
apply the PR in #98

PR #102, opened 2020-08-31
Implement tensor

qindazhu:


k2/csrc/cuda/tensor.cu
```diff
@@ -0,0 +1,98 @@
...
+    elements *= dims_[i];
+  }
+  return elements;
+}
+
+bool Shape::CheckContiguous() {
+  int32_t z = 1;
+  for (int32_t i = ndim_ - 1; i >= 0; --i) {
+    if (dims_[i] != 1) {
+      if (strides_[i] != z) return false;
```
qindazhu, line 57: if `dims[i] == 1`, should we require `strides[i] == z` here? as there's only one element in axis `i`,  `index * strides[i]` will always be 0 as the only valid of index here is 0.

k2/csrc/cuda/tensor.h
```diff
@@ -1,53 +1,67 @@
...
 
  private:
+  static const int32_t kMaxDim = 4;  // Will increase this as needed
+
   int32_t ndim_;  // Must be >= 0
-  // elements of dims_ and strides_ >= ndim_ are currently not set;
-  // in future we may change this.
-  int32_t dims_[MAX_DIM];
-  int32_t strides_[MAX_DIM];  // Strides in bytes
+  int32_t num_element_;
```
danpovey, line 51: am asking Haowen to revert this change and do it as I drafted, with fixed maximum size (other toolkits do it like this + its more efficient)

qindazhu: reverted

k2/csrc/cuda/tensor.cu
qindazhu: there is a public method `bool IsContiguous()` which just returns `is_contiguous`.`bool CheckContiguous()` is just doing the real check and cache the result in `is_contiguous`, and it's private

qindazhu:
Tests added

danpovey:
Great!  Merging.

PR #103, opened 2020-08-31
update to review

songmeixu:
- fix a twice error code checking issue
- change macro name `K2_CUDA_CHECK_ERROR` to `K2_CUDA_PRINT_ERROR` as its effect
- add a optional arg `abort` to control abort if hit error
- trivals

### DIFF
--- a/k2/csrc/cuda/debug.h
+++ b/k2/csrc/cuda/debug.h
@@ -22,8 +22,8 @@
 #include <device_launch_parameters.h>
 #include <driver_types.h>
 
-#include <stdio.h>
-#include <assert.h>
+#include <cstdio>
+#include <cassert>
 
 namespace k2 {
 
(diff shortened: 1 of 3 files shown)


k2/csrc/cuda/debug.h
```diff
@@ -107,30 +111,40 @@ __host__ __device__ __forceinline__ cudaError_t _K2CudaDebug(
...
  * @code{.cpp}
- * K2_CUDA_CHECK_ERROR(cudaGetLastError());
+ * K2_CUDA_PRINT_ERROR(error = cudaGetLastError());
  * @endcode
  */
-#define K2_CUDA_CHECK_ERROR(e)                                    \
-  if (::k2::_K2CudaDebug((cudaError_t)(e), __FILE__, __LINE__)) { \
-    exit(e);                                                      \
-  }
+#define K2_CUDA_PRINT_ERROR(e, bAbort...)                            \
```
danpovey, line 146: In the normal case we want to check the error and die if the status is not zero.
So please leave it as K2_CUDA_CHECK_ERROR... later if we need we can create another macro that just prints and continues.

songmeixu: Done.

danpovey:
Thanks!

PR #104, opened 2020-09-01
add StorageSize() in tensor

qindazhu:


k2/csrc/cuda/context.cu
```diff
@@ -42,7 +42,7 @@ class CpuContext : public Context {
 class CudaContext : public Context {
  public:
   CudaContext(int gpu_id) : gpu_id_(gpu_id) {
-    if (gpu_id_ != 1) {
+    if (gpu_id_ != -1) {
```
qindazhu, line 45: seems some merge before breaks `CudaContext`, fixed now.

k2/csrc/cuda/tensor.cu
```diff
@@ -98,9 +111,9 @@ TensorPtr Tensor::Index(int32_t axis, int32_t index) const {
 }
 
 void Tensor::Init(ContextPtr c) {
-  int32_t nelement = shape_.Nelement();
+  int32_t storage_size = shape_.StorageSize();
```
qindazhu, line 114: As `tensor` may be not contiguous, we need to use `StorageSize()` instead of `Nelement()` to allocate memory.

danpovey:
Thanks!

PR #105, opened 2020-09-01
add cuda check

songmeixu:


songmeixu:
Wait, I found this PR have two files exists on master but not cuda_draft. Please note this Dan. And I am going to remove these two files here.

k2/csrc/cuda/types.h
```diff
@@ -0,0 +1,268 @@
+// k2/csrc/cuda/types.h
```
qindazhu, line 1: I thought we (Dan or I) have deleted types.h and ragged_shape.cc before, seems you forget to sync

songmeixu: oh, let me sync it.

.clang-format
```diff
@@ -3,7 +3,7 @@ BasedOnStyle: Google
 ---
 Language:               Cpp
 Cpp11BracedListStyle:   true
-Standard:               Cpp11
```
qindazhu, line 6: @songmeixu why do you change the value of `Standard`? I will get below error with value `c++11`
```
YAML:6:25: error: unknown enumerated scalar
Standard:               c++11

```

songmeixu: Oh, it's my clangformat suggested, it's due to versions?

qindazhu: mine is `9.0.0` installed with `pip`, which is suggested by @csukuangfj and it works well

songmeixu: image
I check the doc of clang-format, Çpp11 is alias of keyword latest (which may be c++20), which is deprecated, and as its meaning is changed, I vote the c++11.

songmeixu: Where the error comes from, may it cause by locally setting?

csukuangfj: https://releases.llvm.org/9.0.0/tools/clang/docs/ClangFormatStyleOptions.html

qindazhu: no, I changed nothing, just pull & rebase the code, error occurs. What's your version of `clang-format`

qindazhu: atually we are not using `Lastest`, we are just using `Cpp11`. That is the 2nd line in your screenshots, so I guess we should change back to `Cpp11`.

songmeixu: This is from official doc.. https://clang.llvm.org/docs/ClangFormatStyleOptions.html. Obviously, it's deprecated after 9.0, if keep to the old, use higher versions would get the latest standard , "LS_Latest (in configuration: Latest) Parse and format using the latest supported language version. Cpp11 is a deprecated alias for Latest". I guess when 9.0 version, latest == cpp11, but not right now.

songmeixu: > atually we are not using `Lastest`, we are just using `Cpp11`. That is the 2nd line in your screenshots, so I guess we should change back to `Cpp11`.

And the second line indicates it is "c++11 in configuration", the Ls_Cpp11 is a key (used to check internally maybe), not value.

qindazhu: OK, I will update my clang-format version and try

songmeixu: The error has to be caused by sth.. I doubt it's the clang-format version problem. Mine is:
image

qindazhu: according to the link @csukuangfj gives, `Cpp11` instead of `c++11` is the valid option in `9.0.0`, so I guess updating version may work.

PR #113, opened 2020-09-03
implemented resize of Array1 and fix some (compilation) issues

qindazhu:
`error.h` removed and renamed `K2_CUDA_CHECK_ERROR` to `K2_CHECK_CUDA_ERROR`


k2/csrc/cuda/array.h
```diff
@@ -61,22 +61,10 @@ class Array1 {
   // TODO(haowen): why do we need this version?
   Array1() : dim_(0), byte_offset_(0), region_(nullptr) {}
 
-  Array1(int32_t dim, RegionPtr region, int32_t byte_offset)
-      : dim_(dim), region_(region), byte_offset_(byte_offset) {}
-
-  Array1(ContextPtr ctx, int32_t size, T elem) {
```
qindazhu, line 67: @danpovey , I deleted this version of constructor, as it seems capturing `*this` in lambda in constructor is not valid. Instead, we can initialize data with `operator=`
```
Array1<int> array(context, size);
array = 3;
```

k2/csrc/cuda/array.h
```diff
@@ -54,29 +55,22 @@ class Array1 {
...
     Init(ctx, size);
-    T *data = Data();
-    // TODO(haowen): need capture *this, not sure it's valid in constructor
-    /*
-    auto lambda = [=] __host__ __device__(int32_t i) -> void {
-      data[i] = elem;
-    };
-    Eval(ctx, dim_, lambda);
-    */
+    *this = elem;
```
qindazhu, line 68: well, it's strange that it's OK to call `operator=` to do the same initialization things in the constructor...the code are totally same...

PR #119, opened 2020-09-06
using parameter pack to accept additional parameters in kernel lambda

qindazhu:
add more tests for `Array2`

### DIFF
--- a/k2/csrc/cuda/array.h
+++ b/k2/csrc/cuda/array.h
@@ -300,10 +300,13 @@ class Array2 {
       Array1<T> array(dim0_ * dim1_, region, 0);
       const T *this_data = Data();
       T *data = array.Data();
-      auto lambda_copy_elems = [=] __host__ __device__(int32_t i, int j)->void {
-        data[i * dim1_ + j] = this_data[i * elem_stride0_ + j];
+      auto lambda_copy_elems = [=] __host__ __device__(
+                                       int32_t i, int32_t j, int32_t dim1,
+                                       int32_t elem_stride0)->void {
+        data[i * dim1 + j] = this_data[i * elem_stride0 + j];
       };
-      Eval2(region_->context, dim0_, dim1_, lambda_copy_elems);
+      Eval2(region_->context, dim0_, dim1_, lambda_copy_elems, dim1_,
+            elem_stride0_);
       return array;
     }
   }
@@ -388,7 +391,8 @@ class Array2 {
       if (!copy_for_strides) {
         LOG(FATAL) << "non-unit stride on 2nd axis of tensor";
       }
-      region_ = NewRegion(region->context, dim0_ * dim1_ * ElementSize());
+      region_ =
+          NewRegion(region->context, dim0_ * elem_stride0_ * ElementSize());
       byte_offset_ = 0;
       CopyDataFromTensor(t);
     }
@@ -403,7 +407,7 @@ class Array2 {
     region_ = a.GetRegion();
   }
 
-  // Warning: user should never call this function, we declare
+  // Caution: user should never call this function, we declare
   // it as public just because the enclosing parent function for an
   // extended __host__ __device__ lambda
   // must have public access.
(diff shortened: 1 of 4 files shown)


k2/csrc/cuda/array.h
```diff
@@ -403,19 +407,22 @@ class Array2 {
...
-      this_data[i * elem_stride0_ + j] =
-          t_data[i * elem_stride0_ + j * elem_stride1];
+    auto lambda_copy_elems = [=] __host__ __device__(
+                                     int32_t i, int32_t j, int32_t elem_stride0,
+                                     int32_t elem_stride1)->void {
+      this_data[i * elem_stride0 + j] =
+          t_data[i * elem_stride0 + j * elem_stride1];
     };
-    Eval2(region_->context, dim0_, dim1_, lambda_copy_elems);
+    Eval2(region_->context, dim0_, dim1_, lambda_copy_elems, elem_stride0_,
```
qindazhu, line 424: Sometims there are additional parameters  required by `lambda` but they are on CPU memory (e.g. `elem_stride0_` and `elem_stride1` here), we need a way to convert them to GPU memory automatically. Here I use `parameter pack` to accept and forward those parameters. See my updates for `Eval2` below for details.

k2/csrc/cuda/context.h
```diff
@@ -369,17 +370,22 @@ void Eval(ContextPtrType c, int32_t n, LambdaT &lambda) {
...
+  Note if `lambda` requires additional parameters besides `i` and `j`,
+  user can pass those parameters with `Params... params`,
+  Eval2 will forward them to `lambda`.
+
 */
-template <typename LambdaT>
-void Eval2(cudaStream_t stream, int32_t m, int32_t n, LambdaT &lambda) {
+template <typename LambdaT, typename... Params>
+void Eval2(cudaStream_t stream, int32_t m, int32_t n, LambdaT &lambda,
+           Params... params) {
```
qindazhu, line 380: `params` here is to accept and forward additional parameters of `lambda`, CPU memory to GPU memory conversion may happen here.

danpovey:
No I dont want to merge this.  The whole point of lambdas is that variables are forwarded automatically.

danpovey:
Mm, my comment was lost somehow...
This is not necessary. The whole point of lambdas is they capture things like that and it all happens automagically.

danpovey:
.. these things will be passed by value, there is no problem with being in the wrong memory space, NVidia takes care of it for us.

qindazhu:
> The whole point of lambdas is that variables are forwarded automatically.

mm.. How? Suppose `elem_stride0` and `elem_stride1` are local variables (on CPU), the kernel will access CPU memory below..and crash..
```
auto lambda_copy_elems = [=] __host__ __device__(
                                     int32_t i, int32_t j)->void {
      this_data[i * elem_stride0 + j] =
          t_data[i * elem_stride0 + j * elem_stride1];
    };
```

qindazhu:
Maybe you should run the code to see what happens...

danpovey:
did you try?  Lambdas  on GPU would be useless if what you say were true.

qindazhu:
will get error `illegal memory access`. Will write a piece of code to show you the behavior.

qindazhu:
[test.zip](https://github.com/danpovey/k2/files/5179426/test.zip)

you can build with `nvcc --expt-extended-lambda -std=c++11 -o test test.cu`. And IMO, it does not matter there is such behavior for lambda in class, as we usually would not pass lots of paremter to lambda in class, as we can pass `*this` if those variables are class members.

danpovey:
What you are doing, as far as I can see, does not reflect how I am using lambdas.  I always avoid implicitly passing `*this`; I think it is passed by address.  That's why I never use class members in lambdas.

danpovey:
.. anyway there is no problem, you just need to make sure you have a local variable as a copy of any class member you want to use; and make use of that in the lambda.  It's less work than adding extra explicit args.

qindazhu:
so how should I implement copy data for `Array2`, as we need access `elem_stride0` which is on CPU. And I believe we will encounter the same problem in other classes. And I don't think `parameter pack` here will introduct any overhead, it just pass parameters to kernel which is exactly we do while we call kernel on host (such as pass `rows` and `cols` in `MatrixTranspose`)

danpovey:
Look at the kernels I already wrote, it's easy.  You just do:
  int32_t elem_stride0 = elem_stride0_;
  and use elem_stride0 in the lambda.

qindazhu:
I said before, `elem_stride0` here will be on CPU. you even don't see my example code seriously....

danpovey:
Haowen, it doesn't matter because it's passed by value, not by
pointer/reference, and it is not a pointer.  Show me what in your example
code failed and I'll show you why your argument doesn't apply.  Your
example code had many things.

danpovey:
Also, I think your example code is not quite right, you are accessing
elements past the end of `data` I think, you need to do
if (i < n) {
...
}
probably data[i] = data[i] got optimized out.

danpovey:
Oh i think the problem is that `data` itself is a class member, of A.

qindazhu:
so you mean i should do `T* data = data_;` for every class member?

danpovey:
Yes.

qindazhu:
OK, will revert back

PR #121, opened 2020-09-06
add more tests to Array2

qindazhu:
Reverted paramter pack changes of lambda.


danpovey:
kthanks!

danpovey:
meixu, do you have the time tonight to make a PR that moves the stuff in csrc/ into old/, cuda/ into csrc/, and renames .cc to .cu?

PR #123, opened 2020-09-06
fix compiling.

csukuangfj:
Fix compiling errors.

### DIFF
--- a/k2/csrc/cuda/array.h
+++ b/k2/csrc/cuda/array.h
@@ -3,7 +3,7 @@
 // Copyright (c)  2020  Xiaomi Corporation (authors: Daniel Povey
 //                                                   Haowen Qiu)
 
-// See ../../LICENSE for clarification regarding multiple authors
+// See ../../../LICENSE for clarification regarding multiple authors
 
 #ifndef K2_CSRC_CUDA_ARRAY_H_
 #define K2_CSRC_CUDA_ARRAY_H_
@@ -12,8 +12,8 @@
 #include <iostream>
 
 #include "k2/csrc/cuda/context.h"
-#include "k2/csrc/cuda/log.h"
 #include "k2/csrc/cuda/dtype.h"
+#include "k2/csrc/cuda/log.h"
 #include "k2/csrc/cuda/tensor.h"
 
 namespace k2 {
@@ -25,6 +25,7 @@ namespace k2 {
 template <typename T>
 class Array1 {
  public:
+  static_assert(std::is_pod<T>::value, "T must be POD");
   using ValueType = T;
   int32_t ElementSize() const { return sizeof(ValueType); }
   int32_t Dim() const { return dim_; }  // dimension of only axis (axis 0)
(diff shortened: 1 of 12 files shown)


csukuangfj:
The [colab notebook][1] shows that all tests are passed.

[1]: https://colab.research.google.com/drive/1qbHUhNZUX7AYEpqnZyf29Lrz2IPHBGlX?usp=sharing#scrollTo=2QRbpTm0BkEn

qindazhu:
Many thanks, Merge, will go forward based on this.

k2/csrc/cuda/array.h
```diff
@@ -322,16 +327,18 @@ class Array2 {
   /* Create new array2 with given dimensions.  dim0 and dim1 must be >0.
      Data will be uninitialized. */
   Array2(ContextPtr c, int32_t dim0, int32_t dim1)
-      : dim0_(dim0), dim1_(dim1), elem_stride0_(dim1), byte_offset_(0) {
+      : dim0_(dim0), elem_stride0_(dim1), dim1_(dim1), byte_offset_(0) {
+    K2_DCHECK_GT(dim0, 0);
```
qindazhu, line 331: BTW, for those index check, which version should we use? `DCHECK` or `CHECK`? I'd like to use `CHECK` if it's not in a loop, and `DCHECK` if it's in a loop. What do you think? I'm just thinking it's better if we can follow a same rule.

csukuangfj: I am fine with either case. I just replaced all `assert`  that I could find with debug check.

qindazhu: yes, I know, just to make sure we can follow the same convention in the future, @danpovey any suggestion?

danpovey: @qindazhu I'd like to use K2_CHECK if it's not in an inner loop or expensive.  If somewhat expensive then K2_CHECK, else K2_PARANOID_ASSERT (or we could implement K2_PARANOID_CHECK).

Also i want to remove glog and just use the K2_ macros, this will keep things simple and reduce chance of future problems.  We need to remove the glog include from cuda/context.h.

qindazhu: Yeah, by saying `CHECK` I mean `K2_CHECK` (sorry for the words as I suppose we will remove glog).

> If somewhat expensive then K2_CHECK, else K2_PARANOID_ASSERT

I don't get what you means here, when expensive, still use `K2_CHECK`? (I just thought it's `K2_DCHECK`)

danpovey:
Sorry I meant K2_DCHECK if expensive.
And if *very* expensive, K2_PARANOID_CHECK.  (Or perhaps we could call this
DDCHECK?)

qindazhu:
Sorry, a bit confused...`K2_DCHECK` now is to do nothing (`(void) 0`), so what is `DDCHECK` supposed to do?

danpovey:
There are 3 levels of debug:
  -DNDEBUG :    only K2_CHECK runs
  nothing defined:          K2_CHECK and K2_DCHECK run
 -DK2_PARANOID:       K2_CHECK and K2_DCHECK and
K2_DDCHECK/K2_PARANOID_CHECK run.

qindazhu:
Oh, OK

PR #126, opened 2020-09-07
Tensor ops implementation

danpovey:


danpovey:
Will prob. break compilation again... I'll figure out a way to test compilation though.

PR #128, opened 2020-09-08
fix compilation errors

qindazhu:
All tests passed


k2/csrc/CMakeLists.txt
```diff
@@ -2,16 +2,15 @@
...
-add_subdirectory(old)
+add_subdirectory(host)
 
 #---------------------------- Build K2 CUDA sources ----------------------------
 
 # the target
 add_library(context
     context.cu
     tensor.cu
-    tensor_ops.cu
```
qindazhu, line 13: note we have not added tests for `tensro_ops.cu`, so remove this line for now.

PR #132, opened 2020-09-09
implement some methods in RaggedShape and compile Ragged

qindazhu:
note there's still a compilation errors 
```
utils.cu(57): error: calling a __host__ function("k2::FillValues") from a __global__ function("k2::RowSplitsToRowIdsKernel") is not allowed
```
which requires to add flag in CMakeFiles to support separate compilation for cuda,  will do it later or if you guys @songmeixu @csukuangfj  can do this as you may be good at this.


qindazhu:
Note  I just fixed about 100 compilation errors to make it pass build (except the errors above), there are still a lot of functions which are not implemented for now, we can fix them and add tests one by one later.

k2/csrc/log.h
```diff
@@ -174,6 +174,13 @@ class Voidifier {
 #define K2_CHECK_CUDA_ERROR(x) \
   K2_CHECK_EQ(x, cudaSuccess) << " Error: " << cudaGetErrorString(x) << ". "
 
+#define K2_CUDA_SAFE_CALL(...)               \
```
qindazhu, line 177: This is from @songmeixu's code before

danpovey:
thx!!

csukuangfj:
> utils.cu(57): error: calling a __host__ function("k2::FillValues") from a __global__ function("k2::RowSplitsToRowIdsKernel") is not allowed


I guess it's not related to separate compilation.

The error log says 
https://github.com/danpovey/k2/blob/91f19f3cd2cc2a4452ceac8041e49cf3a13b4007/k2/csrc/utils.cu#L27

can only run on the host, but it is called from a function running on the GPU:
https://github.com/danpovey/k2/blob/91f19f3cd2cc2a4452ceac8041e49cf3a13b4007/k2/csrc/utils.cu#L37

danpovey:
It's "dynamic parallelism".  It's allowed but we may need special
compilation options.

qindazhu:
Sorry, I mis-deleted `__device__` prefix for function `FillValues`.  But ,we still need to add some compilation options to support dynamic parallelism

qindazhu:
Updated the error messages above, sorry for mis-pasting previous error logs

PR #133, opened 2020-09-09
enable separate compilation

qindazhu:


CMakeLists.txt
```diff
@@ -81,7 +81,9 @@ else()
   unset(CMAKE_CUDA_STANDARD CACHE)
   set(CMAKE_CUDA_STANDARD ${CMAKE_CXX_STANDARD})
 
-  set(K2_COMPUTE_ARCHS 30 32 35 50 52 53 60 61 62 70 72)
+  # set(K2_COMPUTE_ARCHS 30 32 35 50 52 53 60 61 62 70 72)
+  message(WARNING "arch 62/72 are not supported for now")
+  set(K2_COMPUTE_ARCHS 35 50 52 60 61 70)
```
qindazhu, line 86: Compiling and tests got passed, but there are still build errors for `compute 62/72` (note separate compilation is is only allowed on the compute_35 architecture or above)

```
nvlink error   : Undefined reference to 'cudaGetParameterBufferV2' in '../../lib/libcontext.a:utils.cu.o' (target: sm_62)
nvlink error   : Undefined reference to 'cudaLaunchDeviceV2' in '../../lib/libcontext.a:utils.cu.o' (target: sm_62)
```
so I just remove them for now and add a warning.

qindazhu: OK, on some machine, maybe `35` does not work either. We definitely need to add some link libraries (but I searched the symbol in my local cuda libraries and didn't find such symbols).

songmeixu:
Let me add more arch for cmake-3.18 and test. Running..

songmeixu:
I can compile with 62/72, and other archs. FYI. Will try with cmake3.12, I know there probably is problem, as cmake said so.

qindazhu:
OK, thanks, maybe it's because you use Cmake-3.18 and I use  3.14

PR #138, opened 2020-09-10
FSA implementation..

danpovey:


danpovey:
will prob. break the build but not too easy for me to check right now as cluster has problems.  will merge anyway

k2/csrc/fsa.cu
```diff
@@ -9,6 +9,134 @@
...
+  reachable = reachable.Range(0, num_states);
+  int32_t *neg_properties_data = properties.Data();
+  char *reachable_data = reachable.Data();  // access co_reachable via this.
+
+  auto lambda_get_properties = [=] __host__ __device__ (int32_t idx012) -> void {
+      Arc arc = arcs_data[idx012];
+      Arc prev_arc;
+      if (idx012 > 0) prev_arc = arcs_data[idx012 - 1];
+      int32_t idx01 = row_ids2_data[idx012],
+          idx01x = row_splits2_data[idx012],
```
csukuangfj, line 54: Is this a typo or an error?

Should it be
```cpp
idx01x = row_splits2_data[idx01];
```
?

danpovey: yes thanks, you are right.  (This is one of the benefits of this naming system.. would have been impossible to spot this error without it.)

PR #139, opened 2020-09-11
Update README

danpovey:


danpovey:
I think gitub is confused here.. i have verified in the command line that only README.md is changed.

PR #141, opened 2020-09-11
Tested RaggedShape and fixed some bugs

qindazhu:


k2/csrc/ragged.cu
```diff
@@ -143,10 +143,10 @@ int32_t RaggedShape::MaxSize(int32_t axis) {
   const auto &row_splits = axes_[axis - 1].row_splits;
   const int32_t num_rows = row_splits.Dim() - 1;
   if (num_rows == 0) return 0;
-  // TODO(haowen): write a new function to allocate just one element?
-  Array1<int32_t> max_array(row_splits.Context(), 1, 0);
+  Array1<int32_t> max_array(Context(), 1, 0);
   int32_t *max_value = max_array.Data();
   const int32_t *row_splits_data = row_splits.Data();
+  // TODO(haowen): use atomicMax or two reduce ops? not sure which is faster
```
qindazhu, line 149: @danpovey any suggestions?

danpovey: mm, this code isn't really right, you'd need a while loop as there would be contention.  The standard way to do these things is via a reduction operation.  See Max() in ops.h, although its not implemented yet.
It would internally use a generic reduction template from cub.

danpovey: See here:
https://nvlabs.github.io/cub/structcub_1_1_device_reduce.html#a974a241463ca892c8f5e73b879065e48

danpovey: Also, Haowen: I think the InputIteratorT can be anything that supports a [] operator, or at least + and dereferencing (not sure, check code), which you can implement with a custom struct that wraps the row_ids pointer and takes a difference, so you don't have to write the data out temporarily.
The Max() interface that I drafted in ops.h is not quite right for this, actually; don't use it.  Use cub directly.  Be careful, it writes to GPU memory, so you can create an Array1 with 1 element or re-use an existing one, as you choose, to store the output value.

qindazhu: Yeah, I know the current implementation is right, so I add a TODO to say use `atomicMax` in lambda or use reduce ops to implement it. So should I use `cub::DeviceReduce::Max` or cuda reduction to implement `Max`? 

And this PR is ready to merge, will add `Max` operation in next PR

danpovey:
OK.   Use cub::DeviceReduce::Max.

PR #143, opened 2020-09-11
block style check and build test

songmeixu:
For personal repo, to delete action/workflow need the ower account. This PR just commented all branchs-name that were required checks, rather than delete them.


songmeixu:
FYI, it's a request from dan, as the checks didn't work as expected.

danpovey:
Thanks!  Lemme wait to hear from the others whether this is still necessary... dunno if anything Fangjun committed would have affected this issue.

PR #144, opened 2020-09-11
implement Array2::To

csukuangfj:


k2/csrc/array.h
```diff
@@ -332,12 +332,14 @@ class Array2 {
   /* Could view this as num_cols */
   int32_t Dim1() const { return dim1_; }
 
-  ContextPtr &Context() const { return region_->context; }
+  ContextPtr &Context() { return region_->context; }
+
+  const ContextPtr &Context() const { return region_->context; }
```
danpovey, line 337: the returned ContextPtr should not be const, this may cause certain complications.

k2/csrc/array.h
```diff
@@ -410,10 +417,46 @@ class Array2 {
     that
      it will have a different memory layout than the input.
   */
-  Array2<T> To(ContextPtr ctx) {
-    // TODO
-    Array2<T> array;
-    return array;
+  Array2<T> To(ContextPtr ctx) const {
+    if (dim0_ * dim1_ == 0) {
+      if (!region_) return *this;
```
qindazhu, line 422: I think there should be no this line. As the parameter `ctx` is valid, we should always returen a valid `Array2` to users, even if the size is `0` and `this->region_ == nullptr`

k2/csrc/array.h
```diff
@@ -410,10 +417,46 @@ class Array2 {
...
+
+      if (ctx->IsCompatible(*Context())) return *this;
+
+      Array2 ans(ctx, dim0_, dim1_);
+      return ans;
+    }
+
+    if (ctx->IsCompatible(*Context()) && elem_stride0_ == dim1_) return *this;
```
danpovey, line 430: line break please

danpovey: ... and fewer blank lines.

k2/csrc/array.h
```diff
@@ -410,10 +417,46 @@ class Array2 {
...
+    if (elem_stride0_ == dim1_) {
+      // the current array is contiguous, use memcpy
+      auto kind = GetMemoryCopyKind(*Context(), *ctx);
+      auto *dst = ans.Data();
+      const auto *src = Data();
+      MemoryCopy(static_cast<void *>(dst), static_cast<const void *>(src),
+                 dim0_ * dim1_ * ElementSize(), kind);
+      return ans;
+    } else {
+      // do rowwise copy
```
danpovey, line 445: this will be super slow if you do sequential CUDA copies.  Instead, I'd return this->ToContiguous().To(ctx); solve it by recursion.

danpovey (commented):
few comments

k2/csrc/array.h
danpovey: We could perhaps require that region_ never be nullptr.  In fact that may already be a requirement.
Notice that NewRegion() allows zero bytes to be requested.

qindazhu: Well, sometimes we need a default constructor for Array, such as `Array1(): dim_(0), region_(nullptr)`. This has been used in a few algorithms written by you before. If we do not allow such a constructor, we need to make sure we always have a `Context` in all usage of Array, to call `NewRegion()`. Which one do you prefer?

csukuangfj: This follows the following comment:
https://github.com/danpovey/k2/blob/2a847823668699959e0be10687acb077a5c8b58c/k2/csrc/array.h#L401-L403

I think it should be an error for an empty Array2 to invoke the `To` method.

qindazhu: e.g. `RaggedShapeDim` which contains `Array1` using a default constructor

k2/csrc/array.h
qindazhu: Dan, do you mean here we need to get non-const `ContextPtr` in some algorithms with `const Array`? I know there's no such requirements for now (at least for files that have been compiled), but not sure for other algorithms that have not compiled.

k2/csrc/array.h
qindazhu: I guess the above documentation is something old and should be updated.

k2/csrc/array.h
csukuangfj: Could you give some hint to implement `ToContiguous` without sequential copy?

k2/csrc/array.h
```diff
@@ -410,10 +417,46 @@ class Array2 {
...
+      auto *dst = ans.Data();
+      const auto *src = Data();
+
+      auto kind = GetMemoryCopyKind(*Context(), *ctx);
+
+      for (int32_t i = 0; i != dim0_; ++i) {
+        MemoryCopy(static_cast<void *>(dst), static_cast<const void *>(src),
```
qindazhu, line 452: it could be done with kernel, see my implementation in   `Flatten`

k2/csrc/array.h
qindazhu: I guess something like using `kernel`, see my implementation in `Flatten`

k2/csrc/array.h
danpovey: yes, non-const.  It's OK for const this to return non-const Context, as it's not a member and not really owned by *this.

k2/csrc/array.h
csukuangfj: The source context and dest context may represent different devices, e.g., one for CPU and the other for GPU.
I am afraid in this case, we cannot use a kernel to replace memory copy.

k2/csrc/array.h
danpovey: > Well, sometimes we need a default constructor for Array, such as `Array1(): dim_(0), region_(nullptr)`. This has been used in a few algorithms written by you before. If we do not allow such a constructor, we need to make sure we always have a `Context` in all usage of Array, to call `NewRegion()`. Which one do you prefer?

We will consider that an Array1/Array2 constructed with default constructor is "not valid" and only assignment may be performed; other operations may segfault.

danpovey: mm yes, remove that old documentation.

k2/csrc/array.h
danpovey: yeah, kernel using Eval2(), a 2d kernel.

danpovey: .. eventually we can switch to a more specialized implementation that used the hardware in a smarter way.

k2/csrc/array_test.cu
```diff
@@ -233,7 +233,7 @@ void TestArray1() {
   }
 
   {
-    // test operator[](const Array1<int32_t> indexes)
+    // test operator[](const Array1<int32_t> &indexes) const
```
qindazhu, line 236: sorry for my laziness to spell complete function signature.. and thanks.

k2/csrc/array_test.cu
```diff
@@ -297,6 +297,87 @@ void TestArray2() {
...
+        // with operator [][]
+        EXPECT_EQ(cpu_array[r][c], k);
+        EXPECT_EQ(cuda_array[r][c], k);
+        ++k;
+      }
+
+    // test operator <<
+    std::ostringstream os;
+    os << array;
+    K2_LOG(INFO) << os.str().c_str();
```
qindazhu, line 334: Just use` K2_LOG(INFO) << array`, I redefine template function in `Logger `

https://github.com/danpovey/k2/blob/master/k2/csrc/log.h#L113-L118

k2/csrc/array.h
qindazhu: OK, seems we need to do `K2_CHECK(region_ != nullptr)`  in all algorithms for `Array1` and `Array2`

k2/csrc/array.h
qindazhu: And for non-compatiable context, it may be better to make it contiguous first with `kernel` in a same Context, and then do a one pass copy.

danpovey: @csukuangfj if the source array was not contiguous, we first make it contiguous *on the same device* by calling ToContiguous().  We need to define this function, e.g. can put it in ops.h:  Array2<T> ToContiguous(const Array2<T> &src);

danpovey: .. please see ToContiguous() in tensor_ops.h for reference.

k2/csrc/array.h
csukuangfj: fixed.

k2/csrc/array.h
csukuangfj: old documentation is removed.

The function will segfault when invoked with region_ == nullptr.

k2/csrc/array.h
csukuangfj: fixed.

---

I prefer more blanks as I think it is more readable. I know fewer blank lines save screen spaces.

k2/csrc/array.h
csukuangfj: @qindazhu 
thanks. fixed.

k2/csrc/array.h
csukuangfj: fixed. But I implemented it as a member function.

k2/csrc/array.h
```diff
@@ -410,10 +424,34 @@ class Array2 {
     that
      it will have a different memory layout than the input.
   */
-  Array2<T> To(ContextPtr ctx) {
-    // TODO
-    Array2<T> array;
-    return array;
+  Array2<T> To(ContextPtr ctx) const {
+    if (dim0_ * dim1_ == 0) {
```
qindazhu, line 428: seems we can remove this if now, below code can handle the case.

k2/csrc/array.h
```diff
@@ -391,17 +389,33 @@ class Array2 {
...
+    K2_CHECK_GE(elem_stride0_, dim1_);
+    K2_CHECK_GE(byte_offset_, 0);
+  }
+
+  Array2 ToContiguous() const {
+    if (elem_stride0_ == dim1_) return *this;
+    Array2 ans(Context(), dim0_, dim1_);
+    auto *dst = ans.Data();
+    const auto *src = Data();
+    auto dim1 = dim1_;
```
danpovey, line 404: please don't use `auto` except for complex types, it makes the code less easy to understand.

k2/csrc/array.h
```diff
@@ -410,10 +424,34 @@ class Array2 {
...
+  Array2<T> To(ContextPtr ctx) const {
+    if (dim0_ * dim1_ == 0) {
+      if (ctx->IsCompatible(*Context())) return *this;
+      Array2 ans(ctx, dim0_, dim1_);
+      return ans;
+    }
+
+    // clang-format off
+    if (ctx->IsCompatible(*Context()) &&
+          elem_stride0_ == dim1_)
```
danpovey, line 436: actually we shouldn't have to check that elem_stride_ == dim1_.  You can remove the documentation part that says the result will always be contiguous.

csukuangfj: Done.

k2/csrc/array.h
csukuangfj: Fixed.

k2/csrc/array.h
csukuangfj: Removed.

k2/csrc/array_test.cu
csukuangfj: Done.

k2/csrc/array.h
csukuangfj: Moved to ops.h and ops_inl.h

danpovey:
Thanks!!!  Merging.

k2/csrc/array.h
qindazhu: may be better if adding documentation to say we may make it contiguous if their cotext are not compatiable, as we called ToContiguous in that case.

danpovey:
sure.

k2/csrc/array.h
csukuangfj: I will add it in the next pullrequest.

csukuangfj: Fixed in https://github.com/danpovey/k2/pull/146

PR #149, opened 2020-09-12
require cpp14 as reported by cub warning

songmeixu:


songmeixu:
Please some kind reviewer help here. It just changes the cpp std compile flag from 11 to 14. As cub keep warn about it.

qindazhu:
I think we have discussed about it here https://github.com/danpovey/k2/pull/77#issuecomment-666895919. Maybe it's better to let @danpovey decide this. And BTW, I still use GCC 4.8.5 on the cluster..

danpovey:
We can leave the PR up for possible later merge, but for now just to make building easier I'd rather keep it as C++11.
The fact that cub only gives a warning, not an error, makes me think that it may not really be required.
Presumably we'll find out when something doesn't work properly, if it was really required.

songmeixu:
Got it. In this case, I want to close this and make a new one with only one change to here,  from 14 to 11. To simplify the codes.

csukuangfj:
I am afraid we need c++14 sooner or later when it comes to integration with PyTorch since PyTorch requires c++14.

Please refer to

- https://github.com/pytorch/pytorch/blob/master/README.md#from-source
> If you are installing from source, you will need Python 3.6 or later and a C++14 compiler.

Even if k2 is built against `libtorch_cpu.so`, it still needs c++14:
https://github.com/pytorch/pytorch/blob/e7d782e724c76bb0572023d52ee7438a40a7a262/torch/utils/cpp_extension.py#L296

danpovey:
OK, bumping to 14 is fine with me.

PR #150, opened 2020-09-12
replace glog with log.h

songmeixu:
- adjust `K2_CUDA_SAFE_CALL` and add `K2_STATIC_ASSERT`
- replace glog, as dan expected
- and fix some missing ops -> array_ops rename
- **make: commented csrc/host and python subdirectory, as they are deprecated**

### DIFF
--- a/k2/csrc/CMakeLists.txt
+++ b/k2/csrc/CMakeLists.txt
@@ -19,7 +19,6 @@ set_target_properties(context PROPERTIES CUDA_SEPARABLE_COMPILATION ON)
 
 # lib deps
 target_link_libraries(context PUBLIC cub)
-target_link_libraries(context PUBLIC glog)
 
 #---------------------------- Test K2 CUDA sources ----------------------------
 
--- a/k2/csrc/log.h
+++ b/k2/csrc/log.h
@@ -5,6 +5,7 @@
  *
  * @copyright
  * Copyright (c)  2020  Fangjun Kuang (csukuangfj@gmail.com)
+ *                      Xiaomi Corporation (authors: Meixu Song)
  *
  * @copyright
  * See LICENSE for clarification regarding multiple authors
@@ -13,9 +14,16 @@
 #ifndef K2_CSRC_LOG_H_
 #define K2_CSRC_LOG_H_
 
+#include <cassert>
 #include <cstdint>
 #include <cstdio>
 #include <sstream>
+
+#ifdef __CUDA_ARCH__ 
+#define K2_CUDA_HOSTDEV __host__ __device__
+#else
+#define K2_CUDA_HOSTDEV
+#endif
 
 namespace k2 {
 
@@ -35,7 +43,7 @@ enum class LogLevel {
   kFatal = 4,
 };
 
-// They are used in LOG(xxx), so their names
+// They are used in K2_LOG(xxx), so their names
 // do not follow the google c++ code style
 constexpr LogLevel DEBUG = LogLevel::kDebug;
 constexpr LogLevel INFO = LogLevel::kInfo;
@@ -45,8 +53,8 @@ constexpr LogLevel FATAL = LogLevel::kFatal;
 
 class Logger {
  public:
-  __host__ __device__ Logger(const char *filename, const char *func_name,
-                             uint32_t line_num, LogLevel level)
+  K2_CUDA_HOSTDEV Logger(const char *filename, const char *func_name,
+                         uint32_t line_num, LogLevel level)
       : filename_(filename),
         func_name_(func_name),
         line_num_(line_num),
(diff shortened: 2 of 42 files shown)


danpovey:
This is all great, except that I dont want to change log.h -> log.cuh.  It's inconsistent with the naming of other files.
I don't think header filenames matter to the compiler, only .cc/.cu files matter, to the build system.

https://stackoverflow.com/questions/24970602/can-i-just-define-cuda-kernels-in-h-files

Anyway even if they do matter, we'd have to do it consistently.

songmeixu:
rename log

> This is all great, except that I dont want to change log.h -> log.cuh. It's inconsistent with the naming of other files.
> I don't think header filenames matter to the compiler, only .cc/.cu files matter, to the build system.
> 
> [stackoverflow.com/questions/24970602/can-i-just-define-cuda-kernels-in-h-files](https://stackoverflow.com/questions/24970602/can-i-just-define-cuda-kernels-in-h-files)
> 
> Anyway even if they do matter, we'd have to do it consistently.

log.cuh, now it's log.h.
And compiled & gtest-passed.

k2/CMakeLists.txt
```diff
@@ -1,2 +1,2 @@
 add_subdirectory(csrc)
-add_subdirectory(python)
```
qindazhu, line 2: why do you remove `python` folder?

k2/csrc/CMakeLists.txt
```diff
@@ -2,8 +2,6 @@
 # The header should be included as `#include "k2/csrc/.."`, to avoid conflicts.
 include_directories(${CMAKE_SOURCE_DIR})
 
-add_subdirectory(host)
```
qindazhu, line 5: And why remove `host`? Actually I want to keep `host` and `python` for now. When we create tests for those Fsa-related algorithms with new Ragged interface, I may reuse those (test) code (at least test cases) in `python` and `host`, it's better if we can keep them compiled all the time.

k2/csrc/log.h
```diff
@@ -134,6 +136,23 @@ class Voidifier {
...
+ * @brief A macro to get checked at compile time, as static assertion
+ *
+ * @param[in] COND the compile-time boolean expression that must be true
+ * @param[in] MSG  an error message if exp is false
+ *
+ * @code{.cpp}
+ * K2_STATIC_ASSERT(DEFINED_SHAPE % DEFINED_X == 0);
+ * @endcode
+ */
+#define K2_STATIC_ASSERT4(COND, MSG) \
```
qindazhu, line 149: IIRC, we have decided to use the version `static_assert` in  `C++11`. (And define some macro like `#define K2_STATIC_ASSERT(x) static_assert(x, "")`)

k2/csrc/log.h
```diff
@@ -178,11 +197,12 @@ class Voidifier {
...
 
-#define K2_CUDA_SAFE_CALL(...)               \
-  do {                                       \
-    (__VA_ARGS__);                           \
-    cudaDeviceSynchronize();                 \
-    K2_CHECK_CUDA_ERROR(cudaGetLastError()); \
+#define K2_CUDA_SAFE_CALL(...)                                             \
+  do {                                                                     \
+    (__VA_ARGS__);                                                         \
+    ::k2::internal::kDisableDebug ? cudaSuccess : cudaDeviceSynchronize(); \
```
qindazhu, line 203: why is  `cudaSuccess` here?  Do you mean some code like `void (0)`?

csukuangfj: I would suggest
```cpp
::k2::internal::kDisableDebug ? (void)0 : (void)cudaDeviceSynchronize();
```

since we do not care about the result.

danpovey:
Thanks!  LGTM, merging.

songmeixu:
Wait a min. Sorry, I should have called wait here. I am changing things to get compiled.

songmeixu:
Please unmerge(?)

songmeixu:
If not convenient to unmerge, I could fix it in new PR. And replay the above review.

danpovey:
fix in new PR pls...

songmeixu:
Make a new PR to fix the nullptr issues.

k2/CMakeLists.txt
songmeixu: got it, added it back

k2/csrc/CMakeLists.txt
songmeixu: got it, added it back

k2/csrc/log.h
songmeixu: changed it as suggested

k2/csrc/log.h
songmeixu: changed it to a new sensible way. As `(void)0` give error: `operand types are incompatible ("void" and "cudaError_t")`. And if the return of cudaDeviceSynchronize gets checked, check cudaGetLastError again is excess.

songmeixu:
FYI, the log.h is added a macro at header:
```
#ifdef __CUDA_ARCH__ 
#define K2_CUDA_HOSTDEV __host__ __device__
#else
#define K2_CUDA_HOSTDEV
#endif
```

to make *.cc that include the log.h get compiled by gcc.

PR #153, opened 2020-09-14
fix compile error: `stringstream << nullptr` is ambiguous

songmeixu:


danpovey:
I think it would be better to let it accept nullptr.  Seems it has a type std::nullptr_t.
How about
std::ostream &operator<<(std::ostream &stream, std::nullptr_t null) {
   stream << "(null)";
}

qindazhu:
Tha's why I like maco `K2_CHECK_NOTNULL`, we can implement it with `K2_CHECK(x != nullptr)`, so it would just print `bool` (i.e. the bool value is  `x != nullptr`). There is no need to define `operator<<` for `nullptr`.

danpovey:
Too special-purpose for my taste.  I see no problem with defining operator
<< for nullptr.

songmeixu:
added a specialization for it，passed build, is it ok?

danpovey:
Thanks!  Merging.

PR #155, opened 2020-09-14
int -> int32_t

songmeixu:
- with typo fix, like int32 -> int32_t

### DIFF
--- a/k2/csrc/compose.cu
+++ b/k2/csrc/compose.cu
@@ -9,10 +9,12 @@
  * See LICENSE for clarification regarding multiple authors
  */
 
-#include "k2/csrc/fsa_algo.h"
+#include "k2/csrc/compose.h"
 
 #include <limits>
 #include <vector>
+
+#include "k2/csrc/fsa_algo.h"
 
 namespace k2 {
 
@@ -287,7 +289,7 @@ class MultiGraphDenseIntersect {
                     between min_active and max_active.
   */
   Array1<float> GetPruningCutoffs(const Ragged3<float> &arc_end_scores) {
-    int32 num_fsas = arc_end_scores.Dim0();
+    int32_t num_fsas = arc_end_scores.Dim0();
 
     // get the maximum score from each sub-list (i.e. each FSA, on this frame).
     // Note: can probably do this with a cub Reduce operation using an operator
@@ -401,7 +403,7 @@ class MultiGraphDenseIntersect {
     // sequence.
     const int32_t *b_fsas_row_ids1 = b_fsas_.shape.RowIds(1);
     const float *score_data = b_fsas_.scores.Data();
-    int score_num_cols = b_fsas_.scores.Dim1();
+    int32_t score_num_cols = b_fsas_.scores.Dim1();
 
     Ragged3<ArcInfo> ai(ai_shape);
     ArcInfo *ai_data = ai.values.Data();  // uninitialized
@@ -444,7 +446,7 @@ class MultiGraphDenseIntersect {
     Does the forward-propagation (basically: the decoding step) and
     returns a newly allocated FrameInfo* object for the next frame.
    */
-  FrameInfo *PropagateForward(int t, FrameInfo *cur_frame) {
+  FrameInfo *PropagateForward(int32_t t, FrameInfo *cur_frame) {
     Ragged<StateInfo> &states = cur_frame->states;
     // ai has 3 axes: fsa_id, state, arc.
     Ragged<ArcInfo> ai = GetUnprunedArcs(t, cur_frame);
--- a/k2/csrc/compose.cu
+++ b/k2/csrc/compose.cu
@@ -9,8 +9,6 @@
  * See LICENSE for clarification regarding multiple authors
  */
 
-#include "k2/csrc/compose.h"
-
 #include <limits>
 #include <vector>
 
@@ -641,11 +639,11 @@ class MultiGraphDenseIntersect {
       K2_CHECK_EQ(state_map_data[state_map_idx], state_idx01);
       state_map_data[state_map_idx] = -1;
     }
-  };
-  Eval(c_, ans->states.values.Dim(), lambda_reset_state_map);
 
-  return ans;
-}
+    Eval(c_, ans->states.values.Dim(), lambda_reset_state_map);
+
+    return ans;
+  }
 
   /*
     Does backward propagation of log-likes, which means setting the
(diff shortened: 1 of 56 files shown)


k2/python/csrc/dlpack.h
```diff
@@ -70,7 +70,7 @@ typedef struct {
   /*! \brief The device type used in the device. */
   DLDeviceType device_type;
   /*! \brief The device index */
-  int device_id;
+  int32_t device_id;
```
danpovey, line 73: not sure about this, it may have to match something declared elsewhere.  same with ndim.

songmeixu: reverted changes in dlpack.h, after check dlpack repo, they notice these int type issues. Keep the two vars int, I guess b/c it's impactless.

FYI
https://github.com/dmlc/dlpack/issues/1
https://github.com/dmlc/dlpack/pull/8

songmeixu:
Hi, @danpovey, this PR is ready to review, please. Thx!!

k2/csrc/utils.cu
```diff
@@ -137,9 +137,10 @@ void RowSplitsToRowIds(ContextPtr &c, int32_t num_rows,
       int32_t avg_elems_per_row = (num_elems + num_rows - 1) / num_rows,
               threads_per_row = RoundUpToNearestPowerOfTwo(avg_elems_per_row),
               tot_threads = num_rows * threads_per_row;
-      int block_size = 256;
-      int grid_size = NumBlocks(tot_threads, block_size);
-      K2_CUDA_SAFE_CALL(RowSplitsToRowIdsKernel<<<grid_size, block_size, 0,
```
qindazhu, line 142: Seems you reverted my changes in the last PR but actually you should not do this. `grid_size` is the number of blocks in a kernel and `block_size` is the number of threads in a block, so the current code you create has the wrong parameter order.

qindazhu: And I prefer name `block_size` and `grid_size` (or `num_blocks`), as many cuda example code also use these two names, I think it's better if we can follow the same convention.

csukuangfj: Did it pass the unit test?

qindazhu: I guess it may pass the tests for now as currently I don't add tests with hundreds of values  for `RaggedShape`, we may need to add those tests after get `RandomRaggedShape` tested.

qindazhu:
And it would be great if you can fix the style issues (seems only one such issue? Anyway, it could be easy with clang-format as @csukuangfj has put file `.clang-format` in the root path)

songmeixu:
> And it would be great if you can fix the style issues (seems only one such issue? Anyway, it could be easy with clang-format as @csukuangfj has put file `.clang-format` in the root path)

I see, would make it in next.

k2/csrc/utils.cu
songmeixu: changed this here and other places. Compiled and tested.

songmeixu:
FYI, it's ready for a fresh review. 

Changes to last:
- solve whole cpp code style issues
	- fix some typos during this progress
- commented two files in CMake, that added by dan today, to get compiled. And these two file issues are planned to get solved later.

danpovey:
Meixu, were the cosmetic changes (i.e. to whitespace) automatic or manual?
I'm concerned about a situation where every time you or Haowen touch a file your respective auto-formatters change half the lines, so we never know who really wrote any given line.

songmeixu:
> FYI, it's ready for a fresh review. 
> 
> Changes to last:
> - solve whole cpp code style issues, I mean all current existed cpp files.
> 	- fix some typos during this progress
> - commented two files in CMake, that added by dan today, to get compiled. And these two file issues are planned to get solved later.
> 

this time is automatically formated. Before is manually.

k2/csrc/compose.cu
```diff
@@ -9,11 +9,11 @@
  * See LICENSE for clarification regarding multiple authors
  */
 
-#include "k2/csrc/fsa_algo.h"
-
 #include <limits>
 #include <vector>
 
+#include "k2/csrc/fsa_algo.h"
```
danpovey, line 15: there are a lotof changes to this file, were these manual?  were they a reversion of something Haowen did?

k2/csrc/compose.cu
```diff
@@ -655,240 +727,166 @@ class MultiGraphDenseIntersect {
 
        @param [in]  cur_frame    The FrameInfo for the frame on which we want to
                                  set the forward log-like
-       @param [in]  next_frame  NULL if this is is the last frame of the sequence;
-                                otherwise the next frame's FrameInfo; arcs on
-                                `cur_frame` have transitions to states on `next_frame`.
-                                The `backward_loglike` values in `next_frame` are
-                                assumed to already be set.
+       @param [in]  next_frame  NULL if this is is the last frame of the
+    sequence; otherwise the next frame's FrameInfo; arcs on `cur_frame` have
```
danpovey, line 731: this seems to have gone bad.
I can fix it next time I see the code though.

songmeixu:
To get passed the style check, I think someone have to make these cosmic changes to other's code. For example the length greater than 80, it's should be the auto format tool job. I wanna setup a precommit require checking, which is handy for style and static check. Could i give a pr to show about it?

danpovey:
OK, I merged; but Haowen and Meixu, can  you please turn your auto-formatters off?  I don't want it that you have 2 different formatters that will change the files each time you load them.

songmeixu:
And a auto format tool follow the existed clangformat style setting, could be as one of precommit step. Then the code style should get immune of this problem.

danpovey:
I'd rather not have a precommit *requirement*.  Something that would be
better is some kind of script or tool that lets us know how
our code violates the style, then we can manually fix things.  We need to
know what the style is.
And sometimes there are valid reasons for not following the style.
I don't want it automatic; it can mess up code.  I saw examples in that
commit where the auto-formatter damaged comments and commented
code.

songmeixu:
> OK, I merged; but Haowen and Meixu, can  you please turn your auto-formatters off?  I don't want it that you have 2 different formatters that will change the files each time you load them.

About this issue happened between  @haowen  and me, I will compare the results of us tomorrow face to face. I think we both use the clangformat tool. The config is set, then the result should be same.

songmeixu:
> And a auto format tool follow the existed clangformat style setting, could be as one of precommit step. Then the code style should get immune of this problem. 

I see, as you pleasure, will not ask about it.

qindazhu:
> I think we both use the clangformat tool. The config is set, then the result should be same.

I think so, if we both use the same `.clang-format` file in the root path.

csukuangfj:
> I think so, if we both use the same .clang-format file in the root path.

You have to check that you use the identical version of `clang-format`.

BTW, to mark a region that should not be formatted by `clang-format`, you can use

```cpp
// clang-format off
some code;
// clang-format on
```

qindazhu:
And @csukuangfj what's your clang-format's version? `9.0.0`? Just make sure we can/will have the same version.

csukuangfj:
I am using 9.0.0. I can switch to whichever version you use.

If `clang-format` is installed via pip, I think the latest version is 9.0.0.
See https://pypi.org/project/clang-format/#history

PR #156, opened 2020-09-15
Put weight into arc for host code

danpovey:


danpovey:
Guys, I've merged this even though it breaks the host python tests; I'm hoping someone can fix this (meixu?),
it requires changing the python calling code to add the weights as arcs to the arc constructors.

This will make it easier to wrap the host code as the CPU-only implementations of some algorithms.

songmeixu:
I take this. And FYI, except the way you suggested, I would take it as a chance to try cppyy to handle it automatically. To be finished in two days.

danpovey:
Meixu, no cppyy please!  If there is going to be cppyy involved, don't take
it at all, let someone else do it!
(Anyway it's not a binding issue, it's an interface issue).

songmeixu:
Fine, then to finish it today.

danpovey:
Thanks!

PR #158, opened 2020-09-15
Show how to implement shims for old (host) code

danpovey:


danpovey:
wont' fully compile as I added compose.cc

PR #160, opened 2020-09-16
compiled compose

qindazhu:
Get compiled, but there are definitely tons of TODOs or bugs need to fix.

### DIFF
--- a/k2/csrc/array_ops_inl.h
+++ b/k2/csrc/array_ops_inl.h
@@ -27,6 +27,14 @@
 #include <vector>
 
 namespace k2 {
+// will be used in  MaxPerSublist
+template <typename T>
+struct MaxOp {
+  __device__ T operator()(T a, T b) { return (a > b ? a : b); }
+  __device__ MaxOp() {}
+  __device__ MaxOp(const MaxOp &src) {}
+};
+
 // Will be used in ExclusiveSumDeref to call ExclusiveSum (which calls
 // cub::DeviceScan::ExclusiveSum internally).
 template <typename T>
@@ -176,13 +184,13 @@ void ExclusiveSumDeref(Array1<T *> &src, Array1<T> *dest) {
 template <typename T>
 void MaxPerSublist(Ragged<T> &src, T default_value, Array1<T> *max_values) {
   K2_CHECK_EQ(src.NumAxes(), 2);
-  K2_CHECK_EQ(src.Dim0(), max_values->Dim());
-  K2_CHECK(IsCompatible(src, *max_values));
+  K2_CHECK_EQ(src.shape.Dim0(), max_values->Dim());
+  K2_CHECK(IsCompatible(src.shape, *max_values));
 
   ContextPtr c = src.Context();
 
-  const int32_t *row_splits = src.RowSplits(1);
-  int32_t num_rows = src.Dim0();
+  const int32_t *row_splits = src.shape.RowSplits(1).Data();
+  int32_t num_rows = src.shape.Dim0();
   const T *values_data = src.values.Data();
   T *output_data = max_values->Data();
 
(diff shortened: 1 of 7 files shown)


danpovey:
Great!!

k2/csrc/array_ops_inl.h
```diff
@@ -27,6 +27,14 @@
 #include <vector>
 
 namespace k2 {
+// will be used in  MaxPerSublist
+template <typename T>
+struct MaxOp {
+  __device__ T operator()(T a, T b) { return (a > b ? a : b); }
+  __device__ MaxOp() {}
```
csukuangfj, line 34: There are subtle differences between
```cpp
MaxOp() {}
```
and

```cpp
MaxOp() = default;
```

I would suggest
```cpp
MaxOp() = default;
MaxOp(const MaxOp &) = default;
MaxOp &operator=(const MaxOp &) = default;
```
or leave them out entirely since there is nothing to copy or initialize.

qindazhu: Actually I did this intentionally, nvcc will report a warning if you use `** = default`.

```
warning: __device__ annotation is ignored on a function("MaxOp") that is explicitly defaulted on its first declaration
```
But we really need copy constructor to be device code. Of cousre do `MaxOp {}` will make the struct not POD, but it does not matter in our case as we just use MaxOp in `Reduce`.

Check here for more details https://github.com/kokkos/kokkos/issues/1473#issuecomment-488844696. Note I don't add any comments in the code as other guys could see the warning easily if they try to do ` = default`.

qindazhu: Note it may be OK to not define any constructor at all (no `MaxOp() = default`, no `MaxOp() {}`), I will try this approach when really running the code.

k2/csrc/compose.cu
```diff
@@ -187,86 +190,91 @@ class MultiGraphDenseIntersect {
...
-            *oshapeu_row_splits1 = oshape_unpruned_.RowSplits(1);
-
-    int32_t *oshapep_row_ids3 = oshape_pruned_.RowIds(3),
-            *oshapep_row_ids2 = oshape_pruned_.RowIds(2),
-            *oshapep_row_ids1 = oshape_pruned_.RowIds(1),
-            *oshapep_row_splits3 = oshape_pruned_.RowSplits(3),
-            *oshapep_row_splits2 = oshape_pruned_.RowSplits(2),
-            *oshapep_row_splits1 = oshape_pruned_.RowSplits(1);
+    ContextPtr c_cpu = c_->GetCpuContext();
+    int32_t T = a_fsas_.shape.MaxSize(1);
```
csukuangfj, line 210: Should this line be
```cpp
int32_t T = b_fsas_.shape.MaxSize(1);
```
T is from the dense fsa.

k2/csrc/compose.cu
```diff
@@ -307,13 +315,12 @@ class MultiGraphDenseIntersect {
...
-      arc_map_a_data[pruned_idx0123] = arc_info.arc_idx;
+      arc_map_a_data[pruned_idx0123] = arc_info.a_fsas_arc_idx012;
 
       arc.src_state = pruned_src_state_idxx12;
       arc.dest_state = pruned_dest_state_idxx12;
-      arc.symbol = a_fsas_arc[arc_info.arc_idx].symbol;
-      int32_t fsa_id = unpruned_idx0, t = unpruned_idx1,
-              b_fsas_idx0x = b_fsas_row_ids1[fsa_id],
+      arc.symbol = a_fsas_arcs[arc_info.a_fsas_arc_idx012].symbol;
+      int32_t fsa_id = unpruned_idx0, b_fsas_idx0x = b_fsas_row_ids1[fsa_id],
```
csukuangfj, line 323: Should this be
```cpp
b_fsas_idx0x = b_fsas_row_splits1[fsa_id];
```

k2/csrc/compose.cu
```diff
@@ -187,86 +190,91 @@ class MultiGraphDenseIntersect {
...
               unpruned_idx01x = oshapeu_row_splits2[unpruned_idx01],
               unpruned_idx01xx = oshapeu_row_splits3[unpruned_idx01x],
               unpruned_idxxx23 = unpruned_idx0123 - unpruned_idx01xx,
               unpruned_idx0 = oshapeu_row_ids1[unpruned_idx01],  // fsa-id
-          unpruned_idx0x = oshapeu_row_splits1[unpruned_idx0],
-              unpruned_idx0xx = oshapeu_row_splits2[unpruned_idx0x],
-              unpruned_idx1 = unpruned_idx01 - unpruned_idx01,  // t
+          // unpruned_idx0x = oshapeu_row_splits1[unpruned_idx0],
+          // unpruned_idx0xx = oshapeu_row_splits2[unpruned_idx0x],
+          unpruned_idx1 = unpruned_idx01 - unpruned_idx01,  // t
```
csukuangfj, line 272: I this here it should be
```cpp
unpruned_idx1 = unpruned_idx01 - unpruned_idx0x,
```

danpovey:
thanks.. you could make a PR with those suggested changes so I dont forget
them, maybe?

k2/csrc/compose.cu
csukuangfj: fixed.

k2/csrc/compose.cu
csukuangfj: fixed.

k2/csrc/compose.cu
csukuangfj: fixed

PR #162, opened 2020-09-17
implement PytorchContext.

csukuangfj:
Make PytorchContext the default cuda context.

The [colab notebook][1] shows that all cuda related test cases are passed.

[1]: https://colab.research.google.com/drive/1qbHUhNZUX7AYEpqnZyf29Lrz2IPHBGlX?usp=sharing#scrollTo=Gl6iGTSVAgMh

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,8 +93,10 @@ endif()
 enable_testing()
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
-include(googletest)
 include(pybind11)
+include(torch)
 include(cub)
+include(googletest)
+
 
 add_subdirectory(k2)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,8 @@ set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -O0")
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 option(BUILD_SHARED_LIBS "Whether to build shared or static lib" ON)
+option(USE_PYTORCH "Whether to build with PyTorch" ON)
+
 
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib")
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib")
@@ -94,7 +96,9 @@ enable_testing()
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 include(pybind11)
-include(torch)
+if(USE_PYTORCH)
+  include(torch)
+endif()
 include(cub)
 include(googletest)
 
--- a/k2/csrc/context.cu
+++ b/k2/csrc/context.cu
@@ -14,6 +14,7 @@
 
 #include "k2/csrc/context.h"
 #include "k2/csrc/log.h"
+#include "k2/csrc/pytorch_context.h"
 
 namespace k2 {
 
@@ -101,7 +102,7 @@ class CudaContext : public Context {
 ContextPtr GetCpuContext() { return std::make_shared<CpuContext>(); }
 
 ContextPtr GetCudaContext(int32_t gpu_id /*= -1*/) {
-  return std::make_shared<CudaContext>(gpu_id);
+  return std::make_shared<PytorchContext>(gpu_id);
 }
 
 RegionPtr NewRegion(ContextPtr &context, std::size_t num_bytes) {
--- a/k2/csrc/context.cu
+++ b/k2/csrc/context.cu
@@ -3,107 +3,16 @@
  * context
  *
  * @copyright
- * Copyright (c)  2020  Fangjun Kuang (csukuangfj@gmail.com)
+ * Copyright (c)  2020  Mobvoi AI Lab, Beijing, China (authors: Fangjun Kuang)
  *                      Xiaomi Corporation (author: Haowen Qiu)
  *
  * @copyright
  * See LICENSE for clarification regarding multiple authors
  */
 
-#include <cstdlib>
-
 #include "k2/csrc/context.h"
-#include "k2/csrc/log.h"
-#include "k2/csrc/pytorch_context.h"
 
 namespace k2 {
-
-static constexpr std::size_t kAlignment = 64;
-
-// TODO(haowen): most of implementations below should be updated later.
-class CpuContext : public Context {
- public:
-  ContextPtr GetCpuContext() override { return nullptr; }
-  ContextPtr GetPinnedContext() override { return nullptr; }
-  DeviceType GetDeviceType() const override { return kCpu; }
-
-  void *Allocate(std::size_t bytes, void **deleter_context) override {
-    void *p = nullptr;
-    if (bytes) {
-      int32_t ret = posix_memalign(&p, kAlignment, bytes);
-      K2_CHECK_EQ(ret, 0);
-    }
-    if (deleter_context) *deleter_context = nullptr;
-    return p;
-  }
-
-  bool IsCompatible(const Context &other) const override {
-    return other.GetDeviceType() == kCpu;
-  }
-
-  void Deallocate(void *data, void * /*deleter_context*/) override {
-    free(data);
-  }
-};
-
-class CudaContext : public Context {
- public:
-  explicit CudaContext(int32_t gpu_id) : gpu_id_(gpu_id) {
-    if (gpu_id_ != -1) {
-      auto ret = cudaSetDevice(gpu_id_);
-      K2_CHECK_CUDA_ERROR(ret);
-    }
-    // TODO(haowen): choose one from available GPUs if gpu_id == -1?
-    // and handle GPU ids from multiple machines.
-    auto ret = cudaStreamCreate(&stream_);
-    K2_CHECK_CUDA_ERROR(ret);
-  }
-  ContextPtr GetCpuContext() override { return nullptr; }
-  ContextPtr GetPinnedContext() override { return nullptr; }
-  DeviceType GetDeviceType() const override { return kCuda; }
-  int32_t GetDeviceId() const override { return gpu_id_; }
-
-  void *Allocate(std::size_t bytes, void **deleter_context) override {
-    void *p = nullptr;
-    if (bytes) {
-      auto ret = cudaMalloc(&p, bytes);
-      K2_CHECK_CUDA_ERROR(ret);
-    }
-    if (deleter_context) *deleter_context = nullptr;
-    return p;
-  }
-
-  bool IsCompatible(const Context &other) const override {
-    return other.GetDeviceType() == kCuda && other.GetDeviceId() == gpu_id_;
-  }
-
-  void Deallocate(void *data, void * /*deleter_context*/) override {
-    auto ret = cudaFree(data);
-    K2_CHECK_CUDA_ERROR(ret);
-  }
-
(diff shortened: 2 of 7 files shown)


csukuangfj:
I think we can also replace the current `CpuContext` with `PytorchContext`. 

The `Allocator *allocator_` in `PytorchContext` can be made to point to either a cuda allocator or a cpu allocator depending on the given `gpu_id_`. 

 if `gpu_id_ == -1`, we assign it a cpu allocator, otherwise a cuda allocator.

k2/csrc/context.cu
```diff
@@ -101,7 +102,7 @@ class CudaContext : public Context {
 ContextPtr GetCpuContext() { return std::make_shared<CpuContext>(); }
 
 ContextPtr GetCudaContext(int32_t gpu_id /*= -1*/) {
```
qindazhu, line 104: I would like to leave `GetCudaContext` as before, i.e. returns a `raw` GPU context (and `GetCpuContext` returns a `raw` CPU Context). While returning `PyTorchContext` from a function like `GetPyTorchContext`, and as you said, the `PyTorchContext` could be on either GPU device or CPU device. We definitely will create other Context, e.g. `TensorFlowContext`.

CMakeLists.txt
```diff
@@ -93,8 +93,10 @@ endif()
 enable_testing()
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
-include(googletest)
 include(pybind11)
+include(torch)
```
qindazhu, line 97: We may also make `k2` avaiable from other DL toolkits (e.g. TensorFlow, MegEngine), I guess it's better if we can make this dependency configurable as users usually would not use  those toolkits together.

k2/csrc/context.cu
csukuangfj: My original plan was like the following:

```cpp
// context.cu
Leave as it is
```

```cpp
// PytorchContext.h
// like the current one

// PytorchContext.cu
ContextPtr GetCudaContext(int32_t gpu_id /*= -1*/) {
  return std::make_shared<PytorchContext>(gpu_id);
}
```

There would be an option in `CMakeLists.txt` to configure whether `PytorchContext.cu` **OR** `context.cu` should
be compiled.

To support other frameworks, e.g., tensorflow, we can create `TensorflowContext.h` and `TensorflowContext.cu`.

One of the advantages of this approach is that we can use `GetCudaContext`  in the code without worrying about
where it comes from. The drawback is that it supports only one framework.

CMakeLists.txt
csukuangfj: Yes, I agree. But I think the priority is first to make one framework work, i.e., PyTorch.  Support for multiple
framework in one go may be too complicated.

csukuangfj: Now the user can choose whether to include PyTorch.

k2/csrc/pytorch_context.h
```diff
@@ -15,10 +13,85 @@
...
+#include "c10/cuda/CUDACachingAllocator.h"
 #include "k2/csrc/context.h"
+#include "k2/csrc/log.h"
+#include "torch/torch.h"
+
+namespace k2 {
 
 class PytorchContext : public Context {
-  // ...
+ public:
```
danpovey, line 24: It would be more efficient to have separate classes for the GPU and CPU contexts, so we can avoid that if-statement.  Things like GetDeviceType() are called quite often so I think this is worthwhile.  But that's not a blocking change, can do it later.

k2/csrc/pytorch_context.h
```diff
@@ -15,10 +13,85 @@
...
+  void InitCpu() {
+    allocator_ = torch::GetAllocator(torch::kCPU);
+    K2_CHECK(allocator_->raw_deleter() != nullptr);
+  }
+
+  void InitCuda() {
+    auto ret = cudaSetDevice(device_id_);
+    K2_CHECK_CUDA_ERROR(ret);
+    // TODO(fangjun): invoke init only once
+    c10::cuda::CUDACachingAllocator::init(device_id_ + 1);
```
danpovey, line 84: we might have to introduce some static variables to avoid multiple calls to init..

danpovey (commented):
Looks great!  Some small comments but let me merge this now.

k2/csrc/pytorch_context.h
csukuangfj: I will split it to two classes: `PytorchCpuContext` and `PytorchCudaContext`.

k2/csrc/pytorch_context.h
csukuangfj: PyTorch uses `std::call_once` to achieve this.

Relevant codes are:

https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/Context.h#L78
```cpp
  THCState* lazyInitCUDA() {
    std::call_once(thc_init,[&] {
      thc_state = detail::getCUDAHooks().initCUDA();
    });
    return thc_state.get();
  }
```

https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/cuda/detail/CUDAHooks.cpp#L45
```cpp
std::unique_ptr<THCState, void (*)(THCState*)> CUDAHooks::initCUDA() const {
  THCudaInit(thc_state);
}
```

https://github.com/pytorch/pytorch/blob/master/aten/src/THC/THCGeneral.cpp#L41
```cpp
void THCudaInit(THCState* state) {
  c10::cuda::CUDACachingAllocator::init(numDevices);
}
```

https://github.com/pytorch/pytorch/blob/master/c10/cuda/CUDACachingAllocator.cpp#L816
```cpp
  void init(int device_count) {
    int size = device_allocator.size();
    if (size < device_count) {
      device_allocator.resize(device_count);
      for (int i = size; i < device_count; i++) {
        device_allocator[i] = std::unique_ptr<DeviceCachingAllocator>(new DeviceCachingAllocator());
      }
    }
  }
```

---

I think there are no negative consequences if `init` is called multiple times. Of course, we should invoke it only once if we can.

qindazhu:
Will get error below when building on the cluster
```
CMakeFiles/Makefile2:591: recipe for target 'k2/csrc/CMakeFiles/cu_array_test.dir/all' failed
make[1]: *** [k2/csrc/CMakeFiles/cu_array_test.dir/all] Error 2
/home/storage23/qiuhaowen/anaconda3/envs/py-k2/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so: undefined reference to `operator delete(void*, std::align_val_t)@CXXABI_1.3.11'
/home/storage23/qiuhaowen/anaconda3/envs/py-k2/lib/python3.7/site-packages/torch/lib/libtorch_cpu.so: undefined reference to `operator new(unsigned long, std::align_val_t)@CXXABI_1.3.11'
collect2: error: ld returned 1 exit status

```

csukuangfj:
what is your GCC version? PyTorch requires a compiler supporting c++14.

qindazhu:
```
(py-k2) (py-k2) [qiuhaowen@tj1-asr-dl87.kscn build]$ gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.11) 5.4.0 20160609
```

Should we build with `-std=c++14`?

qindazhu:
And seems we have set `-std=c++14` in CMakeFile

csukuangfj:
From https://github.com/pytorch/pytorch/blob/caea1adc35404b39fe4f2e3fa75f9a0b9e554bbc/cmake/TorchConfig.cmake.in#L126

```cmake
set_target_properties(torch PROPERTIES
    IMPORTED_LOCATION "${TORCH_LIBRARY}"
    INTERFACE_INCLUDE_DIRECTORIES "${TORCH_INCLUDE_DIRS}"
    CXX_STANDARD 14
)
```

It shows that any target depending on torch will have `-std=c++14` in the compilation flags. 
You can verify this with `make VERBOSE=1 context`.

---- 

Did `make VERBOSE=1` show that the compiler flags contain
`-D_GLIBCXX_USE_CXX11_ABI=0`?

csukuangfj:
It turns out that I am using the same version of gcc with you. But I did not encounter this error.

qindazhu:
No `D_GLIBCXX_USE_CXX11_ABI=0`, I even try to set `set(TORCH_CXX_FLAGS "-D_GLIBCXX_USE_CXX11_ABI=0")`, but still get the same error. Note I install pytorch with pip in conda env

qindazhu:
Oh, Sorry, It shows `D_GLIBCXX_USE_CXX11_ABI=0`

```
cd /home/storage23/qiuhaowen/k2/build/k2/csrc && /usr/local/cuda/bin/nvcc -forward-unknown-to-host-compiler  -I/home/storage23/qiuhaowen/k2 -I/home/storage23/qiuhaowen/k2/build/_deps/cub-src -isystem=/home/storage23/qiuhaowen/anaconda3/envs/py-k2/lib/python3.7/site-packages/torch/include -isystem=/home/storage23/qiuhaowen/anaconda3/envs/py-k2/lib/python3.7/site-packages/torch/include/torch/csrc/api/include -isystem=/usr/local/cuda/include --expt-extended-lambda -g --generate-code=arch=compute_61,code=[compute_61,sm_61] -D_GLIBCXX_USE_CXX11_ABI=0 -std=c++14 -x cu -dc /home/storage23/qiuhaowen/k2/k2/csrc/ragged.cu -o CMakeFiles/context.dir/ragged.cu.o

```

csukuangfj:
I am trying to install anaconda3 and to see  if I can reproduce your problem.

qindazhu:
Many thanks!

qindazhu:
I tried to download LibTorch https://download.pytorch.org/libtorch/cu102/libtorch-win-shared-with-deps-debug-1.6.0.zip from here https://pytorch.org/ , then I can build successfully with `cmake -DCMAKE_PREFIX_PATH=my-local-libtorch-path`.  So it seems there's some issue with my installtion of PyTorch (Well, my `PyTorch` itself works well..). Will look into it tomorrow.

csukuangfj:
@qindazhu 
I find a solution to your problem. Assume that your anaconda3 is installed in the default location, e.g., $HOME/anaconda3.

You can use

```bash
LIBRARY_PATH=$HOME/anaconda3/lib:$LIBRARY_PATH make cu_array_test   # NOTE that it is LIBRARY_PATH, NOT LD_LIBRRARY_PATH

LD_LIBRARY_PATH=$HOME/anaconda3/lib:$LIBRARY_PATH ./bin/cu_array_test
```

The PyTorch downloaded from conda requires the `libstdc++.so` from anancoda and we cannot use the one from the system,
so we have to use `LIBRARY_PATH` to tell the compiler to choose a different `libstdc++.so`. During the runtime,
we have to set `LD_LIBRARY_PATH`.

csukuangfj:
You can go to the directory  `/home/storage23/qiuhaowen/anaconda3/envs/py-k2/lib/python3.7/site-packages/torch/lib` and
run
```
ldd libtorch_cpu.so
```

You will find that it depends on `/home/storage23/qiuhaowen/anaconda3/lib/libstdc++.so.6`.

PyTorch installed from `pip install` depends on `/usr/lib/x86_64-linux-gnu/libstdc++.so.6`.

The two `libstdc++.so` will likely be different.

csukuangfj:
@qindazhu 
anaconda3 takes a lot of disk space. I would recommend you to use `virtualenv` and  to install `PyTorch` with `pip install`.

qindazhu:
Fixed, many thanks!

k2/csrc/pytorch_context.h
csukuangfj: Fixed in #163  by using `at::globalContext().lazyInitCUDA();`.

k2/csrc/pytorch_context.h
csukuangfj: Fixed in #163 .

PR #164, opened 2020-09-20
add implementation of Eval(ctx, data, size, lambda)

qindazhu:


.clang-format
```diff
@@ -3,7 +3,7 @@ BasedOnStyle: Google
 ---
 Language:               Cpp
 Cpp11BracedListStyle:   true
-Standard:               c++11
+Standard:               Cpp11
```
qindazhu, line 6: Reverted it to `Cpp11` as @songmeixu said he could install `clang-format 9.0.0` (the latest version we can install with `pip`).

k2/csrc/array_test.cu
```diff
@@ -108,6 +108,25 @@ void TestArray1() {
     }
   }
 
+  {
+    // created with Array1(ContextPtr, int32_t size, Callable &&callable)
+    auto lambda_set_values = [=] __host__ __device__(int32_t i) -> T {
+      return i * i;
+    };
+    Array1<T> array(context, 5, lambda_set_values);
+    ASSERT_EQ(array.Dim(), 5);
```
csukuangfj, line 117: The function `operator[](int32_t i)` of `Array1` is very helpful for testing. You can use
```cpp
for (int32_t i = 0; i != array.Dim(); ++i) {
  EXPECT_EQ(array[i], i * i);
}
 ```
This requires far less code.

File: k2/csrc/array_test.cu [hunk 3]
```diff
@@ -108,6 +108,25 @@ void TestArray1() {
     }
   }
 
+  {
+    // created with Array1(ContextPtr, int32_t size, Callable &&callable)
+    auto lambda_set_values = [=] __host__ __device__(int32_t i) -> T {
```
csukuangfj, line 113: This lambda does not need to capture anything.

k2/csrc/context.h
```diff
@@ -337,6 +337,14 @@ __global__ void eval_lambda(int32_t n, LambdaT lambda) {
...
+template <typename T, typename LambdaT>
+__global__ void eval_lambda(T *data, int32_t n, LambdaT lambda) {
+  int32_t i = blockIdx.x * blockDim.x + threadIdx.x;
+  if (i < n) {
+    data[i] = lambda(i);
+  }
+}
+
 template <typename LambdaT>
 __global__ void eval_lambda2(int32_t m, int32_t n, LambdaT lambda) {
```
csukuangfj, line 349: should we fix the naming style for such functions?

k2/csrc/array_test.cu
qindazhu: Oh, yes, as we have implemented `operator[](int32_t)`. Updated other tests as well. Thanks!

k2/csrc/array_test.cu
qindazhu: Fixed.

k2/csrc/context.h
qindazhu: We may prefer lowercase for those `lambda` kernels to make it intuitive. But I'm OK with uppercase.

csukuangfj:
+2 merging

PR #165, opened 2020-09-20
Implement ToContiguous for k2::Tensor.

csukuangfj:
[Colab notebook][1]
All tests passed.

[1]: https://colab.research.google.com/drive/1qbHUhNZUX7AYEpqnZyf29Lrz2IPHBGlX?usp=sharing#scrollTo=0BmNS7f8_uYC


k2/csrc/tensor.cu
```diff
@@ -131,9 +132,57 @@ void Tensor::Init(ContextPtr c) {
...
+  std::vector<int32_t> dims = tensor.Dims();
+  std::vector<int32_t> strides = tensor.Strides();
+  int32_t num_dims = static_cast<int32_t>(dims.size());
+
+  std::size_t num_dims_bytes = num_dims * sizeof(int32_t);
+
+  RegionPtr dims_ptr = NewRegion(tensor.Context(), num_dims_bytes);
+  RegionPtr strides_ptr = NewRegion(tensor.Context(), num_dims_bytes);
+
+  auto cpu_context = tensor.Context()->GetCpuContext();
```
danpovey, line 145: This is a bit heavy-handed for just transferring dims and strides.  The Shape object has a fixed size with no pointers.  I suggest you give it a device constructor and pass it in as a captured variable.
I'm not sure if it's necessary to make a local copy of it... dunno if the implicit-capture-of-this only happens
for *this* class object or for all class objects.

k2/csrc/tensor.cu
```diff
@@ -131,9 +132,57 @@ void Tensor::Init(ContextPtr c) {
...
+
+  Tensor ans(tensor.Context(), tensor.GetDtype(), dims);
+
+  char *ans_data = static_cast<char *>(ans.Data());
+  const char *this_data = static_cast<const char *>(tensor.Data());
+  int32_t element_size = tensor.ElementSize();
+
+  auto copy_data = [=] __host__ __device__(int32_t i) -> void {
+    char *dst = ans_data + i * element_size;
+    const char *src = this_data + to_offset(i) * element_size;
```
danpovey, line 178: This is super inefficient, doing things character by character.   Better to have this templated on the dtype.

k2/csrc/tensor.h
```diff
@@ -88,7 +89,7 @@ class Shape {
   // compute the number of elements
   int32_t ComputeNumElement() const;
   int32_t ComputeStorageSize() const;
-  bool CheckContiguous() const;
+  bool RefreshContiguous() const;
```
danpovey, line 92: Please dont add functions like this without documenttion!
IsContiguous() is obvious enough, but this function is unclear.

k2/csrc/tensor.h
```diff
@@ -147,26 +148,35 @@ class Tensor {
         reinterpret_cast<char *>(impl_->data->data) + impl_->bytes_offset);
   }
 
+  void *Data() {
+    return reinterpret_cast<char *>(impl_->data->data) + impl_->bytes_offset;
+  }
+
+  const void *Data() const {
```
danpovey, line 155: I think we can just make the other version const, and return non-const void.
It has turned out to be unworkable to let 'const' reflect the const-ness of the underlying data.
If anything it will just mean that the metadata is not changed.

k2/csrc/tensor.h
csukuangfj: I would like the following form:
```cpp
void RefreshContiguous() {
  //  update is_contiguous_;
}

File: k2/csrc/tensor.cu [hunk 5]
```diff
@@ -131,9 +132,57 @@ void Tensor::Init(ContextPtr c) {
...
+  // turn a linear index of a multi-dimensional array
+  // first into a coordinate representation, and then to offset
+  // in number of elements via strides information.
+  auto to_offset = [=] __host__ __device__(int32_t i) -> int32_t {
+    int32_t offset = 0;
+    for (int32_t k = num_dims - 1; k > 0; --k) {
+      int32_t index = i % dims_data[k];
+      offset += index * strides_data[k];
+      i = (i - index) / dims_data[k];
+      if (i == 0) break;
```
danpovey, line 164: I dont see the need for this if-statement.  If can slow things down quite a bit.

csukuangfj: Yes, you are right. I just wanted it to break the loop earlier. I will fix it.

k2/csrc/tensor.cu
danpovey: ... or at least we can have versions of the kernel that are 1-byte, 2-byte, 4-byte and 8-byte, and we select the largest one that's a multiple of the dtype's number of characters (that can be looked up from the type traits).

danpovey: ... and my preference would be to handle num_axes of (0,1) and (2) separately, using Eval() and Eval2().  That will be much more efficient than this.  Sorry, I know it will increase the amount of code that has to be written.

k2/csrc/tensor.h
danpovey: Please clarify exactly when it should be called.  Is it to be called only by member functions?
Does the user have an obligation to call it under certain circumstances?

csukuangfj: These functions are all private.

csukuangfj: I remember that you commented on this before and you felt that `Check` is not a good function name

danpovey:
Ah OK.  I think ComputeIsContiguous() might be clearer?

csukuangfj:
I'll fix all the comments this night.

danpovey:
Thanks!!

k2/csrc/tensor_test.cu
```diff
@@ -4,6 +4,7 @@
  *
  * @copyright
  * Copyright (c)  2020  Xiaomi Corporation (authors: Haowen Qiu)
+ *                      Fangjun Kuang (csukuangfj@gmail.com)
```
qindazhu, line 7: just to make sure you do this (not adding your company information here) intentionally :)

k2/csrc/tensor.cu
qindazhu: And if you have to copy `std::vector` to device, you can just call
```
Array1<int32_t> dims_array(ctx, dims);
const int32_t *dims_data = dims_array.Data();
```

k2/csrc/tensor.cu
qindazhu: Can we just implement cases `num_axes = 0,1,2,3` for now as we may only need those versions for now (add version num_axes>=4 if we really need)? As Cuda supports `dim3` for `block/thread`, we can implement `1,2,3` version with `Eval`, `Eval2`, `Eval3` straightforward.

danpovey:
I just realized I have already implemented it, see CopyTensorElements()
It has  not been tested though.
ToContiguous() just needs to reallocate the data, CopyTensorElements() can
copy them.

qindazhu:
Seems you have implemented `ToContiguous` as well.
https://github.com/k2-fsa/k2/blob/bc3d168d8908ad6761e24a0a612a73beb1e5a914/k2/csrc/tensor_ops.cu#L94-L101

For now we just support directly copy for `num_axes <=2`, should we add `Eval3` and support directly copy for `num_axes ==3` as Cuda can support dim3?
https://github.com/k2-fsa/k2/blob/bc3d168d8908ad6761e24a0a612a73beb1e5a914/k2/csrc/tensor_ops.cc#L59-L60

danpovey:
adding eval3 si fine with me.

csukuangfj:
Sorry, I'm busy tonight. I'll finish it ASAP.

csukuangfj:
> I just realized I have already implemented it, see CopyTensorElements()

I will switch to your implementation for `ToContiguous`.

k2/csrc/tensor.cu
csukuangfj: Fixed (removed).

k2/csrc/tensor.cu
csukuangfj: Replaced with Dan's implementation

k2/csrc/tensor.h
csukuangfj: Replaced with `ComputeIsContiguous`.

k2/csrc/tensor.h
csukuangfj: Fixed.

k2/csrc/tensor.cu
csukuangfj: Removed.

k2/csrc/tensor_test.cu
csukuangfj: Fixed.

csukuangfj:
Ready for review.

qindazhu:
LGTM, merging..

PR #166, opened 2020-09-21
Fix bugs in RowIdsToRowSplits and add tests

qindazhu:
All tests (except python tests) get passed.


danpovey:
LGTM!
BTW for testing I generally prefer if we can have randomized tests that include large sizes, which may of course require adding code that generates random entities.  Because in CUDA some bugs only show up at larger sizes.
Dunno if your tests have that property.  Anyway, will merge.

qindazhu:
Sure, as I said before, will add add tests for large size after test `RandomRaggedShape`, will do it tomorrow at the latest.

danpovey:
Great!

PR #168, opened 2020-09-21
implemented MaxValue and add tests

qindazhu:
Will will do things related to `(Random)RaggedShape` next.

### DIFF
--- a/k2/csrc/utils.h
+++ b/k2/csrc/utils.h
@@ -203,7 +203,7 @@ void ExclusiveSum(ContextPtr &c, int32_t n, SrcPtr src, DestPtr dest);
    additional cost.
  */
 template <typename T>
-T MaxValue(Context *c, size_t nelems, T *t);
+T MaxValue(ContextPtr &c, int32_t nelems, T *t);
 
 /*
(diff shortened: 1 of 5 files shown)


k2/csrc/utils.h
```diff
@@ -465,6 +465,14 @@ std::ostream &operator<<(std::ostream &os, const std::vector<T> &vec) {
   return os;
 }
 
+template <typename T>
+struct MaxOp {
+  __host__ __device__ __forceinline__ T operator()(const T &a,
+                                                   const T &b) const {
+    return (a > b) ? a : b;
```
danpovey, line 472: Great!  incidentally we may be able to use cub::Max(), not sure if it's the same.

qindazhu: `cub_Max` uses `std::numeric_limits<T>::lowest()` as the initial value. Here I use `T(0)` as the initial value (and return T(0) if `num_elems = 0`) according to the documentation you created before.

https://github.com/k2-fsa/k2/blob/e70ce8c7817fd0a8b2842df6d828d2c761e3fd17/k2/csrc/utils.h#L198-L199

PR #170, opened 2020-09-22
Mostly adding interfaces

danpovey:
Guys, I have't tried to make this compile.  Does anyone have time to do that?
Notice that I have added an interface for sorting sublists.  This will need a comparison-based sort,
which is not provided by cub, so we'll have to look elsewhere.  I'm asking around what is the best
approach.


qindazhu:
Just merge this and I'll compile it in my next PR ( late today)

PR #171, opened 2020-09-22
Fix some bugs and add tests for RandomRaggedShape

qindazhu:
Fixed compilation issues and get tests (except python tests) passed.


danpovey:
great!

PR #173, opened 2020-09-22
Wrap Array1<T> as torch::Tensor.

csukuangfj:
Fix k2host python test cases. 

There are two failed test cases: 
- host_fsa_equivalent_test_py, 
- host_weights_test_py
- @qindazhu  please help [fix them][1].

[1]: https://colab.research.google.com/drive/1qbHUhNZUX7AYEpqnZyf29Lrz2IPHBGlX?usp=sharing#scrollTo=Nsb5eqnV5olp


csukuangfj:
The binding code is now much simpler due to `LibTorch`. Please see
- https://github.com/csukuangfj/k2/blob/fangjun-pybind/k2/python/csrc/torch/array.cu
- https://github.com/csukuangfj/k2/blob/fangjun-pybind/k2/python/k2/array.py
- https://github.com/csukuangfj/k2/blob/fangjun-pybind/k2/python/tests/array_test.py

k2/python/csrc/torch.h
```diff
@@ -0,0 +1,18 @@
+/**
+ * @brief python wrappers for PyTorch.
```
danpovey, line 2: we need to make clear that these are wrappers for k2 for use in PyTorch.  Also variable names should make this obvious..

danpovey:
This looks great to me!  Any comments from anyone else?

k2/python/host/k2host/fsa.py
```diff
@@ -5,24 +5,26 @@
...
-        super().__init__(src_state, dest_state, label)
+    def __init__(self, src_state: int, dest_state: int, label: int,
+                 weight: float):
+        super().__init__(src_state, dest_state, label, weight)
 
     def to_tensor(self):
-        return torch.tensor([self.src_state, self.dest_state, self.label],
-                            dtype=torch.int32)
+        return torch.tensor(
+            [self.src_state, self.dest_state, self.label, self.weight],
```
pzelasko, line 22: isn't the weight accidentally getting truncated from float to int32?

csukuangfj: Yes, you are right. Another big issue is that `float` has a different binary representation than `int` (except for number 0).

I am looking for a workaround.

pzelasko: would it make sense to have `to_tensors` instead and return 2 tensors - one for state ids and labels, and the other one for the weights? I guess that change would affect other APIs too (e.g. `Fsa`).

qindazhu: > would it make sense to have to_tensors instead and return 2 tensors

Yes, that's one posibble approach. Another posibble approach is doing type cast, i.e. `float->int` either Python side or C++ side. That was what I did in the python wrapper code before for `LogSumTracebackState::DerivType`  (`std::pair<int, float>`), this approach is also prefered by @danpovey that time. See below code for the details.

https://github.com/k2-fsa/k2/blob/fde770de4b40d20265b5e4d92bbbf6fd53949230/k2/python/csrc/array.cc#L80-L84
https://github.com/k2-fsa/k2/blob/fde770de4b40d20265b5e4d92bbbf6fd53949230/k2/python/csrc/array.cc#L271-L273
https://github.com/k2-fsa/k2/blob/fde770de4b40d20265b5e4d92bbbf6fd53949230/k2/python/k2/array.py#L92-L98
https://github.com/k2-fsa/k2/blob/fde770de4b40d20265b5e4d92bbbf6fd53949230/k2/python/tests/determinize_test.py#L97-L102

note that the last step casts (actually it's `reinterpret`) `float` to `int`, but we usually would not do this frequently, as in model training process, all things related to derivs computation would be done on C++ side, so the reinterpretion is done automatically.

k2/python/host/tests/arcsort_test.py
```diff
@@ -45,9 +46,12 @@ def test_arc_sort(self):
         arc_map = k2host.IntArray1.create_array_with_size(fsa.size2)
         k2host.arc_sort(fsa, arc_map)
         expected_arc_indexes = torch.IntTensor([0, 3, 5, 6, 6, 6])
-        expected_arcs = torch.IntTensor([[0, 2, 0, 3], [0, 4, 0, 2],
-                                         [0, 1, 2, 1], [1, 3, 0, 5],
-                                         [1, 2, 1, 4], [2, 1, 0, 6]])
+        expected_arcs = torch.IntTensor([[0, 2, 0, float_to_int(3)],
```
qindazhu, line 49: I guess casting type one by one element is not so efficient.

qindazhu:
> Fix k2host python test cases.
> 
> There are two failed test cases:
> 
> * host_fsa_equivalent_test_py,
> * host_weights_test_py
> * @qindazhu  please help [fix them](https://colab.research.google.com/drive/1qbHUhNZUX7AYEpqnZyf29Lrz2IPHBGlX?usp=sharing#scrollTo=Nsb5eqnV5olp).

OK, will fix them. Just @songmeixu here in case he duplicated the effort as he was supposed to do fix those python test issues

k2/csrc/pytorch_context.h
```diff
@@ -50,8 +59,14 @@ class PytorchCpuContext : public Context {
     return p;
   }
 
-  void Deallocate(void *data, void * /*deleter_context*/) override {
-    allocator_->raw_deallocate(data);
+  void Deallocate(void *data, void *deleter_context) override {
+    if (deleter_context) {
```
qindazhu, line 63: use `deleter_context != null_ptr`? as we talking about the style before (for any type except bool,  we always use `!=` or `==` to check its value)

k2/python/host/tests/arcsort_test.py
danpovey: the time would mostly be an issue in test code.  we can work out more efficient ways to do this in real code, e.g. by reinterpreting entire tensors, which I think you can do as `cast` in torch.
I want to keep this design, it is efficient for production stuff.  
What we could actually do is, have various convenience functions that manipulate these things, as in,
   combine_arcs(int_tensor with 3 cols, float_tensor with 1 col) -> int tensor with 1 col
   separate_arcs(int_tensor with 1 col) -> (int_tensor with 3 cols, float_tensor with 1 col) 
and for test:
  arcs_from_float(float_tensor with 4 cols) -> int_tensor with 4 cols, except last field which is a float in binary terms.

qindazhu: Yeah, that is similar to the approach we used for old host python code https://github.com/k2-fsa/k2/pull/173#discussion_r493171104

k2/csrc/pytorch_context.h
csukuangfj: fxied.

qindazhu:
Merging, we can fix the type reinterpretation issue later.

PR #174, opened 2020-09-23
fix tests error in host python

qindazhu:


qindazhu:
Just fix left test errors, Merging..
Note all tests in the codebase now get passed.

PR #177, opened 2020-09-25
Implemented/Tested some array_ops: ExclusiveSum/MaxPerSublist..

qindazhu:
What this PR contains:

- Implemented and tested `ExclusiveSum/ExclusiveSumDeref` for `Array1` and `Array2`.
- Fixed bug in `MaxPerSublist` and tested .
- Re-implemented `RandInt` and `RandIntGeometric` with STL.
- Fixed bug in `ExclusiveSum` (version in `utils.h`) to support in-place prefix sum.
- Fixed bug in `Transpose` and `RowIdToRowSplits`.
- Other small fixes.

Note there are still some array_ops what are not implemented/tested, will fix them in following PR, just want to merge this now to aovid future conficts.


k2/csrc/math.h
```diff
@@ -24,9 +26,55 @@ namespace k2 {
   HighestBitSet(1) = 0,
   HighestBitSet(2,3) = 1
   ...
+
+  Note we may delete this function later if there's no other usage.
  */
 int32_t HighestBitSet(int32_t i);
 
+// Generate a uniformly distributed random variable of type int32_t.
+class RandIntGenerator {
```
qindazhu, line 35: This class is copied from https://github.com/k2-fsa/k2/blob/072853691083239fa93220de1956bc642fb2a8c5/k2/csrc/host/fsa_util.cc#L32-L49, we may need to delete that function in `fsa_util.cc` in the future.

k2/csrc/array_ops.h
```diff
@@ -21,49 +21,12 @@
...
-// different configuration depending on num_elements and data type.
-// block size for matrix transpose.
-static constexpr int32_t kTransTileDim = 32;
-static constexpr int32_t kTransBlockRows = 8;
-
 namespace k2 {
-// TODO(haowen): move the implementations to file `op_inl.h` or
-// `op.cu`(specialized on device and data type)?
-template <typename T>
-__global__ void TransposeKernel(int32_t rows, int32_t cols, const T *input,
```
qindazhu, line 38: moved to `array_ops_inl.h` and added `elementStride0` in the parameter list (my previous implementation does not support case stride0 != cols)

k2/csrc/default_context.cu
```diff
@@ -26,8 +26,8 @@ class CpuContext : public Context {
 
  public:
   static ContextPtr Make() {
-    auto p = new CpuContext();
-    return ContextPtr{p};
+    static ContextPtr ctx(new CpuContext());
```
csukuangfj, line 29: The destructor of `ContextPtr` is non-trivial.

According to https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables

> Objects with static storage duration are **forbidden** unless they are trivially destructible.

qindazhu: The previous implementation (withou `static`) is not a real singleton pattern, but we may only need one instance for CPU context. That's why I change it to `static`.
 I just think we can not follow the google style here to make it a singleton, as seems no potential risk for our usages, what do you think? And I think if we must follow, we can just not use singleton pattern (I mean, the previous one which is a wrong implemetation), just make the default constructor public is fine.

k2/csrc/array.h
```diff
@@ -325,6 +325,7 @@ struct ConstArray2Accessor {
   __host__ __device__ T operator()(int32_t i, int32_t j) const {
     return data[i * elem_stride0 + j];
   }
+  const T *Row(int32_t i) { return data + elem_stride0 * i; }
```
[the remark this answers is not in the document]
qindazhu: Fixed.

k2/csrc/default_context.cu
csukuangfj: The original implementation is not a singleton design. It is used to ensure that you cannot create a raw pointer to
`Context` since we want to use its `shared_from_this` method.

csukuangfj: A workaround is to put the declaration and definition of the `xxxContext` in the `*.cu` files
and expose only `GetCpuContext()` and `GetCudaContext()`. Then we can make the constructors public
since they will be called only by us.

qindazhu: That's a good idea, will update. BTW, according to google style, how do we do if we want a object constructed only once?

csukuangfj: Maybe `std::call_once`?

qindazhu: mm...I doubt it as that may mean Google style only supports C++11 or later..

csukuangfj: https://google.github.io/styleguide/cppguide.html#C++_Version

> Currently, code should target C++17, i.e., should not use C++2x features. The C++ version targeted by this guide will advance (aggressively) over time.
> Consider portability to other environments before using features from C++14 and C++17 in your project.

csukuangfj: Even PyTorch requires a C++14 compiler.

qindazhu: just wonder how they do this back to the time when most of compilers does not support C++11 perfectly. Anyway, updated (moved to `.cu`)

danpovey:
I'm OK to merge this.  You can do it, Haowen...

qindazhu:
OK, merging, will continute to work on other ops.

danpovey:
great, thanks!

k2/csrc/default_context.cu
csukuangfj: [`pthread_once`][1] is an alternative for `std::call_once`.

[1]: https://pubs.opengroup.org/onlinepubs/007908799/xsh/pthread_once.html

csukuangfj: You can find its usage in the early versions of OpenFST:

https://android.googlesource.com/platform/external/srec/+/9164572ac412a36d61d05fe206a7c2811e7c8a65/tools/thirdparty/OpenFst/fst/lib/register.h#54

```cpp

static FstRegister<A> *GetRegister() {
  pthread_once(&register_init_, &FstRegister<A>::Init);
  return register_;
}
```

PR #181, opened 2020-09-26
Add segsort from moderngpu

### LINKED ISSUE #176: Sort with operator
Guys,
I spoke with some guys at NVidia to get advice on how we'd implement an interface like
```
template <typename T, typename Op>
void SortSublists(Ragged<T> &src, Array1<int32_t> *order);
```
They advised this
  https://moderngpu.github.io/segsort.html

It will require adding it as another dependency (moderngpu) but I think it's header only.  Apparently it is likely to be added to thrust at some point (i.e. that sort thing).  Sorry i dont have much energy right now so rather than doing it myself I am putting the info here. 
Eventually we'll want to make it customizable with sorting function-object, actually we should make it
template <typename T, typename Op=LessThan<T> >
struct LessThan<T> { 
  inline __host__ __device__ __forceinline__ operator (const T &a, const T &b) const { return (a  < b); }
};
(+device copy constructor as needed)

csukuangfj:
Closes #176


k2/csrc/moderngpu_context.cu
```diff
@@ -0,0 +1,50 @@
...
+ * See LICENSE for clarification regarding multiple authors
+ */
+
+#include "k2/csrc/context.h"
+#include "k2/csrc/moderngpu_context.h"
+#include "moderngpu/context.hxx"
+
+namespace {
+
+class ModernGpuContext : public mgpu::standard_context_t {
```
danpovey, line 20: i wonder if this should be called ModernGpuAllocator to clarify that it's not an instance of k2::Context.

danpovey (commented):
That was fast!  Thanks!

csukuangfj:
There exists an issue for the current commit:

https://github.com/k2-fsa/k2/blob/5a9fc1c1d9939319b69e0cf5df4c72f2e201dab6/k2/csrc/ragged_inl.h#L114-L129

The cuda error code for `mgpu::segmented_sort_indices` is not cudaSuccess, but it sorts the data correctly.

k2/csrc/moderngpu_context.cu
csukuangfj: Fixed.

danpovey:
Thanks!!
At some point it should be possible to implement `ArcSort` (see fsa_algo.h).

k2/csrc/ragged_inl.h
```diff
@@ -95,6 +100,35 @@ Ragged<T> RandomRagged(T min_value, T max_value, int32_t min_num_axes,
...
+void SortSublists(Ragged<T> *src, Array1<int32_t> *order) {
+  K2_DCHECK(IsCompatible(src->values, *order));
+  K2_DCHECK_EQ(src->values.Dim(), order->Dim());
+  K2_DCHECK_EQ(src->Context()->GetDeviceType(), kCuda)
+      << "It supports only CUDA at present";
+
+  std::unique_ptr<mgpu::context_t> context =
+      GetModernGpuAllocator(src->Context()->GetDeviceId());
+
+  Array1<int32_t> &segment = src->shape.RowSplits(src->NumAxes() - 1);
```
qindazhu, line 113: it's better to add `K2_CHECK_EQ(src->shape.NumAxes() == 2` at entry as we just sort values in `src`

k2/csrc/ragged_inl.h
```diff
@@ -95,6 +100,35 @@ Ragged<T> RandomRagged(T min_value, T max_value, int32_t min_num_axes,
...
+  mgpu::segmented_sort_indices(src->values.Data(),  // keys
+                               order->Data(),       // indices
+                               src->values.Dim(),   // count
+                               segment.Data() + 1,  // segments
```
qindazhu, line 117: why `+1` here? the api of mgpu is strange,

k2/csrc/ragged_inl.h
csukuangfj: It supports ragged arrays with more than two axes.

k2/csrc/ragged_inl.h
csukuangfj: I also tried `segment.Data()` and got the same result.

According to https://github.com/moderngpu/moderngpu/blob/2b3985541c8e88a133769598c406c33ddde9d0a5/tests/test_segsort.cu#L9-L15
```cpp
 std::vector<int> cpu_segsort(const std::vector<int>& data,
  const std::vector<int>& segments) {

  std::vector<int> copy = data;
  int cur = 0;
  for(int seg = 0; seg < segments.size(); ++seg) {
    int next = segments[seg];
    std::sort(copy.data() + cur, copy.data() + next);
    cur = next;
  }
  std::sort(copy.data() + cur, copy.data() + data.size());
  return copy;
}
```

The start index of the first segment is 0. That's why I use `segment.Data() + 1`.

k2/csrc/ragged_inl.h
qindazhu: Well, at least check `>=2` here otherwise there will be an error when calling `shape.RowSplits(NumAxes() - 1)`, note that you should add such requirements in the documentation https://github.com/k2-fsa/k2/blob/a17852434c87f76e2dfc837918197c98b7cb947d/k2/csrc/ragged.h#L577-L584



BTW, in other APIs, such as `MaxPerSublist`, we require `NumAxes() ==2` now. @danpovey I think we should make those api consistent,  should we require exactly `NumAxes() == 2` or `NumAxes() >=2`

danpovey: In general I'd like to make the APIs as general as they can be made without requiring extra work.

csukuangfj: It has no requirement about the number of axes a ragged array has. It always sorts the last axis.
This function works as long as the ragged array is not empty.

qindazhu: mm.. definitely it requires NumAxes >=2...It's meaningless to call RowSplits(1) on a shape with `NumAxes <  2`, actually it will crash..

Then I suggest we add documentation and requirements for those APIs with `axes >= 2`

csukuangfj: Could you give an example to show what a ragged array with `NumAxes < 2` looks like?

qindazhu: mm, the point here is, now we allow user define a empty RaggedShape technically, https://github.com/k2-fsa/k2/blob/a17852434c87f76e2dfc837918197c98b7cb947d/k2/csrc/ragged.h#L137, so it's better if we can check this and report error as early as possible (i.e. at entry of the function).

danpovey:
Checking is fine.
I dont think we need to document it because any valid RaggedShape
automatically has NumAxes() > 2.  But of course clarify
what the function does if it's not already clear.

k2/csrc/ragged_inl.h
csukuangfj: I'll add a `K2_DCHECK` in the next commit.

csukuangfj:
> At some point it should be possible to implement ArcSort (see fsa_algo.h).

@danpovey 

Are arcs sorted by `ilabel` or by `olabel`?

danpovey:
There is only one label in the arc.

danpovey:
BTW I have declared ArcSort in fsa_algo.h.

k2/csrc/ragged_inl.h
csukuangfj: > it's better to add K2_CHECK_EQ(src->shape.NumAxes() == 2 at entry as we just sort values in src

@qindazhu 
fixed in #218 

https://github.com/k2-fsa/k2/blob/d291daf7a88d50907d2b1039ca6033f307995a65/k2/csrc/ragged_inl.h#L111

danpovey:
Guys, IMO it would have been more convenient to allow SortSublists to
accept arrays with any num-axes but operate on the last one.
So it would access the last row-ids or row-splits.
Not urgent though.

danpovey:
... for instance, the way it is now would require adding extra code in the
implementation of ArcSort (else it wouldn't work
for FsaVec).

csukuangfj:
> Guys, IMO it would have been more convenient to allow SortSublists to
accept arrays with any num-axes but operate on the last one.

This is what the current implementation does, except that it requires a NON-empty input.

PR #185, opened 2020-09-28
fixed some issues in ValidateRowSplits/Ids/RandUniformArray1 and tested

qindazhu:
Note all methods in `array_ops` are now tested.

### DIFF
--- a/k2/csrc/array_ops.cu
+++ b/k2/csrc/array_ops.cu
@@ -146,7 +146,8 @@ Array1<int32_t> SpliceRowSplits(int32_t num_arrays,
   return ans;
 }
 
-bool ValidateRowIds(Array1<int32_t> &row_ids, Array1<int32_t> *temp) {
+bool ValidateRowIds(const Array1<int32_t> &row_ids,
+                    Array1<int32_t> *temp /*=nullptr*/) {
   ContextPtr ctx = row_ids.Context();
   const int32_t *data = row_ids.Data();
   int32_t dim = row_ids.Dim();
@@ -158,6 +159,7 @@ bool ValidateRowIds(Array1<int32_t> &row_ids, Array1<int32_t> *temp) {
   if (temp == nullptr || temp->Dim() == 0) {
     temp_array = Array1<int32_t>(ctx, 1);
   } else {
+    K2_CHECK(IsCompatible(row_ids, *temp));
     temp_array = temp->Range(0, 1);
   }
   temp = &temp_array;
@@ -173,7 +175,8 @@ bool ValidateRowIds(Array1<int32_t> &row_ids, Array1<int32_t> *temp) {
   return (*temp)[0] == 0;
 }
 
-bool ValidateRowSplits(Array1<int32_t> &row_splits, Array1<int32_t> *temp) {
+bool ValidateRowSplits(const Array1<int32_t> &row_splits,
+                       Array1<int32_t> *temp /*=nullptr*/) {
   ContextPtr ctx = row_splits.Context();
   const int32_t *data = row_splits.Data();
   int32_t dim = row_splits.Dim();
@@ -184,6 +187,7 @@ bool ValidateRowSplits(Array1<int32_t> &row_splits, Array1<int32_t> *temp) {
   if (temp == nullptr || temp->Dim() == 0) {
     temp_array = Array1<int32_t>(ctx, 1);
   } else {
+    K2_CHECK(IsCompatible(row_splits, *temp));
     temp_array = temp->Range(0, 1);
   }
   temp = &temp_array;
@@ -199,13 +203,15 @@ bool ValidateRowSplits(Array1<int32_t> &row_splits, Array1<int32_t> *temp) {
   return (*temp)[0] == 0;
 }
 
-bool ValidateRowSplitsAndIds(Array1<int32_t> &row_splits,
-                             Array1<int32_t> &row_ids, Array1<int32_t> *temp) {
+bool ValidateRowSplitsAndIds(const Array1<int32_t> &row_splits,
+                             const Array1<int32_t> &row_ids,
+                             Array1<int32_t> *temp /*=nullptr*/) {
   // Check if their context are compatible or not while getting
   ContextPtr ctx = GetContext(row_splits, row_ids);
   int32_t num_rows = row_splits.Dim() - 1, num_elems = row_ids.Dim();
   if (num_rows < 0 || (num_rows == 0 && num_elems > 0)) return false;
   if (row_splits[0] != 0 || row_ids[0] < 0) return false;
+  if (num_elems != row_splits[num_rows]) return false;
 
   const int32_t *row_ids_data = row_ids.Data(),
                 *row_splits_data = row_splits.Data();
@@ -214,6 +220,7 @@ bool ValidateRowSplitsAndIds(Array1<int32_t> &row_splits,
   if (temp == nullptr || temp->Dim() == 0) {
     temp_array = Array1<int32_t>(ctx, 1);
   } else {
+    K2_CHECK(ctx->IsCompatible(*temp->Context()));
     temp_array = temp->Range(0, 1);
   }
   temp = &temp_array;
@@ -247,6 +254,7 @@ void RowSplitsToRowIds(const Array1<int32_t> &row_splits,
   K2_CHECK_GE(num_rows, 0);
   // if there are more than zero elems, there must be at least one row.
   K2_CHECK(num_elems == 0 || num_rows > 0);
+  K2_CHECK_EQ(num_elems, row_splits[num_rows]);
   RowSplitsToRowIds(c, num_rows, row_splits.Data(), num_elems, row_ids.Data());
 }
 
@@ -257,6 +265,7 @@ void RowIdsToRowSplits(const Array1<int32_t> &row_ids,
   K2_CHECK_GE(num_rows, 0);
   // if there are more than zero elems, there must be at least one row.
   K2_CHECK(num_elems == 0 || num_rows > 0);
+  if (num_elems > 0) K2_CHECK_GT(num_rows, row_ids[num_elems - 1]);
   RowIdsToRowSplits(c, num_elems, row_ids.Data(), false, num_rows,
                     row_splits.Data());
 }
--- a/k2/csrc/array_ops.cu
+++ b/k2/csrc/array_ops.cu
@@ -9,6 +9,7 @@
  * See LICENSE for clarification regarding multiple authors
  */
 
+#include <algorithm>
 #include <vector>
 
 #include "k2/csrc/array_ops.h"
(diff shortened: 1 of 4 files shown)


danpovey:
Great!  Sorry dont have time to look right now but you can merge.

k2/csrc/array_ops.cu
```diff
@@ -147,63 +148,55 @@ Array1<int32_t> SpliceRowSplits(int32_t num_arrays,
 
 bool ValidateRowIds(Array1<int32_t> &row_ids, Array1<int32_t> *temp) {
```
csukuangfj, line 149: I would suggest giving a better name for `temp`, e.g.,  `invalid`.

k2/csrc/array_ops.cu
```diff
@@ -147,63 +148,55 @@ Array1<int32_t> SpliceRowSplits(int32_t num_arrays,
...
   }
-
+  temp = &temp_array;
   *temp = 0;
+
   int32_t *temp_data = temp->Data();
   auto lambda_check_row_ids = [=] __host__ __device__(int32_t i) -> void {
-    int32_t this_val = data[i], next_val = data[i + 1];
-    if (this_val > next_val || this_val < 0) *temp_data = 1;  // means it's bad.
+    if (data[i] > data[i + 1]) *temp_data = 1;  // means it's bad.
```
csukuangfj, line 168: All row ids should be non-negative in addition to be non-decreasing.

k2/csrc/array_ops.cu
csukuangfj, line 149: The passed `row_ids` should be `const`, i.e.,  `const Array1<int32_t> &row_ids`.

k2/csrc/array_ops.cu
```diff
@@ -147,63 +148,55 @@ Array1<int32_t> SpliceRowSplits(int32_t num_arrays,
...
+  // note `row_ids[0]` may copy memory from device to host
+  if (row_ids[0] < 0) return false;
 
   Array1<int32_t> temp_array;
   if (temp == nullptr || temp->Dim() == 0) {
-    temp_array = Array1<int32_t>(row_ids.Context(), 1);
-    temp = &temp_array;
+    temp_array = Array1<int32_t>(ctx, 1);
+  } else {
+    temp_array = temp->Range(0, 1);
```
csukuangfj, line 161: If `temp != nullptr`, we have to check it is compatible with `row_ids`.

If the signature was `bool ValidateRowIds(const Array1<int32_t> &row_ids)`,
the implementation would be much simpler.

k2/csrc/array_ops.cu
```diff
@@ -147,63 +148,55 @@ Array1<int32_t> SpliceRowSplits(int32_t num_arrays,
...
+  const int32_t *data = row_splits.Data();
   int32_t dim = row_splits.Dim();
-  if (dim == 0) return true;  // will treat this as valid.a
-
-  if (ctx->GetDeviceType() == kCpu) {
-    if (data[0] != 0) return false;
-    for (int32_t i = 0; i + 1 < dim; i++)
-      if (data[i] > data[i + 1]) return false;
-  }
+  // must have at least one element and row_splits[0] == 0
```
csukuangfj, line 180: Does a valid `row_splits` have at least **two** elements?

k2/csrc/array_ops.cu
```diff
@@ -212,45 +205,56 @@ bool ValidateRowSplitsAndIds(Array1<int32_t> &row_splits,
...
+  return (*temp)[0] == 0;
 }
 
-void RowSplitsToRowIds(Array1<int32_t> &row_splits, Array1<int32_t> &row_ids) {
+void RowSplitsToRowIds(const Array1<int32_t> &row_splits,
+                       Array1<int32_t> &row_ids) {
   ContextPtr c = GetContext(row_splits, row_ids);
   int32_t num_elems = row_ids.Dim(), num_rows = row_splits.Dim() - 1;
-  K2_CHECK(num_rows >= 0);
+  K2_CHECK_GE(num_rows, 0);
```
csukuangfj, line 247: An extra check `K2_CHECK_EQ(num_elems, row_splits[-1])`. // psesudo code
may be also added.

Is it required that `num_rows >= 1`?

k2/csrc/array_ops.h
```diff
@@ -295,8 +295,10 @@ void Or(Array1<T> &src, T default_value, Array1<T> *dest) {
 
 /*
   Returns a random Array1, uniformly distributed betwen `min_value` and
-  `max_value`.  CAUTION: this will be randomly generated on the CPU, for now,
-  and transferred to the CPU, so it will be slow if c is not a CPU context.
+  `max_value`.  CAUTION: for now, this will be randomly generated on CPU and
+  then transferred to other devices if c is not a CPU context, so it will be
+  slow if c is not a CPU context.
+  Note T should be floating-pointer type or integral type.
```
csukuangfj, line 301: This can be enforced at compile time with `static_assert` and `type_trais`.

k2/csrc/array_ops_inl.h
```diff
@@ -104,6 +105,29 @@ void ExclusiveSumPerRow(const Array2<T> &src, Array2<T> *dest) {
   }
 }
 
+// called in RandUniformArray1
+template <typename T, typename std::enable_if<
+                          std::is_floating_point<T>::value>::type * = nullptr>
```
csukuangfj, line 110: People will usually not use the default type `void`. It took me a while to figure out
that the default type is `void`.

k2/csrc/array_ops_inl.h
```diff
@@ -104,6 +105,29 @@ void ExclusiveSumPerRow(const Array2<T> &src, Array2<T> *dest) {
...
+void RandArray1Internal(ContextPtr &c, int32_t dim, T min_value, T max_value,
+                        T *data) {
+  std::random_device rd;
+  std::mt19937 gen(rd());
+  std::uniform_real_distribution<T> dis(min_value, max_value);
+  for (int32_t i = 0; i < dim; i++) data[i] = dis(gen);
```
csukuangfj, line 116: `++i` is preferred.

k2/csrc/array_ops.cu
csukuangfj, line 168: All row ids should be less than `dim`.

k2/csrc/array_ops.cu
```diff
@@ -212,45 +205,56 @@ bool ValidateRowSplitsAndIds(Array1<int32_t> &row_splits,
   ContextPtr ctx = GetContext(row_splits, row_ids);
   int32_t num_rows = row_splits.Dim() - 1, num_elems = row_ids.Dim();
   if (num_rows < 0 || (num_rows == 0 && num_elems > 0)) return false;
-  if (num_rows == 0 && num_elems == 0) {
-    return row_splits[0] == 0;
-  }
-  int32_t *row_ids_data = row_ids.Data(), *row_splits_data = row_splits.Data();
+  if (row_splits[0] != 0 || row_ids[0] < 0) return false;
+
```
csukuangfj, line 209: We have to check

```cpp
K2_CHECK_EQ(num_elems,  row_splits[row_splits.Dim() - 1]);
```

k2/csrc/array_ops.cu
```diff
@@ -147,63 +148,55 @@ Array1<int32_t> SpliceRowSplits(int32_t num_arrays,
...
   if (temp == nullptr || temp->Dim() == 0) {
-    temp_array = Array1<int32_t>(row_ids.Context(), 1);
-    temp = &temp_array;
+    temp_array = Array1<int32_t>(ctx, 1);
+  } else {
+    temp_array = temp->Range(0, 1);
   }
-
+  temp = &temp_array;
   *temp = 0;
```
csukuangfj, line 164: For the following use case
```cpp
Array1<int32_t> row_ids = ...; // a non-empty row_ids
Array1<int32_t> temp;
ValidateRowIds(row_ids, &temp);
```

After returning from `ValidateRowIds`, `temp` is still empty. I don't see the need for passing `temp`
as an argument. It can be defined inside the function.

k2/csrc/array_ops_test.cu
```diff
@@ -1069,4 +1069,252 @@ TEST(OpsTest, SpliceRowSplitsTest) {
...
+                                              4, 5, 5, 5, 5, 7, 7, 9};
+    Array1<int32_t> row_ids(context, row_ids_vec);
+    Array1<int32_t> row_splits(context, row_splits_vec);
+    EXPECT_TRUE(ValidateRowSplits(row_splits));
+    EXPECT_TRUE(ValidateRowIds(row_ids));
+    EXPECT_FALSE(ValidateRowSplitsAndIds(row_splits, row_ids));
+  }
+}
+
+TEST(OpsTest, ValidateRowSplitsAndIdsTest) {
```
csukuangfj, line 1315: Are there test cases for empty `row_splits` and `row_ids`?

k2/csrc/array_ops_test.cu
```diff
@@ -1069,4 +1069,252 @@ TEST(OpsTest, SpliceRowSplitsTest) {
...
+  {
+    // another bad case that row ids and row splits don't agree with each other
+    // i > = row_splits[row_ids[i]]
+    const std::vector<int32_t> row_splits_vec = {0,  2,  3,  5,  8, 9,
+                                                 12, 13, 15, 15, 16};
+    const std::vector<int32_t> row_ids_vec = {0, 0, 1, 2, 2, 3, 3, 3,
+                                              4, 5, 5, 5, 5, 7, 7, 9};
+    Array1<int32_t> row_ids(context, row_ids_vec);
+    Array1<int32_t> row_splits(context, row_splits_vec);
+    EXPECT_TRUE(ValidateRowSplits(row_splits));
```
csukuangfj, line 1309: As `temp` is not used here, should we define it inside the function instead of passing
it as an argument?

k2/csrc/array_ops.cu
qindazhu: The result (valid or not) has been returned by the function, tmp is just used to save one allocation.

qindazhu: Fixed

k2/csrc/array_ops.cu
qindazhu: Well, the purpose we add `temp` parameter here is to save one memory allocation if the caller provide it. There may be some situation that the caller re-use some idle storage as tmp, or some storage has not been freed but will not be used anymore. Another possilbe usage is: we may allocate one extra element for Array1 finally (I mean, change Array1's implementation) to make code easy to use (e.g. the case in ExcusiveSum where dest.Dim() == src.Dim() + 1), so the user even can pass `row_splits.Range(-1, 1)` here. Anyway, we may remove parameter `temp` here if we finally find we would not use it frequently, but for now, let's just keep it.

k2/csrc/array_ops.cu
qindazhu: No, empty `row_splits` is valid, i.e. `now_rows == 0 && row_splits[0] == 0`

k2/csrc/array_ops.cu
qindazhu: fixed, return false in that case, thanks

k2/csrc/array_ops.h
qindazhu: added

k2/csrc/array_ops_inl.h
qindazhu: added `T` as the default type

k2/csrc/array_ops_inl.h
qindazhu: Sure

k2/csrc/array_ops.cu
csukuangfj: Is a `row_splits` with **one** element valid?

k2/csrc/array_ops.cu
qindazhu: No, such row ids would be valid, [0,0,1,8], i.e. there are a lot of empty rows.

k2/csrc/array_ops.cu
qindazhu: Fixed, retured false here

k2/csrc/array_ops.cu
qindazhu: yes, it means it's empty as long as the first element is 0;

k2/csrc/array_ops.cu
qindazhu: Will not fix, see my explanation above

k2/csrc/array_ops_test.cu
qindazhu: Added (Actually there is for functions in utils, IIRC)

k2/csrc/array_ops_test.cu
qindazhu: Not fix for now, see my explanation above

k2/csrc/array_ops.cu
qindazhu: not we just require num_rows >=0. Empty case is fine

qindazhu:
Merging..left your comments if there's any other issue and I can fix in the next PR.

PR #187, opened 2020-09-28
Some fixes/changes to how arrays are resized

danpovey:


danpovey:
haowen, you can merge this, assuming merging your stuff doesn't create conflicts.

k2/csrc/array.h
```diff
@@ -161,30 +161,29 @@ class Array1 {
...
+    } else {
+      size_t cur_bytes_used = byte_offset_ + sizeof(T) * (size_t)dim_,
+          new_bytes_used = byte_offset_ + sizeof(T) * (size_t)new_size;
+      // the following avoids a situation where we overwrite data shared with
+      // other Array objects.  You can just do *this = Array1<T>(...) and
+      // overwrite *this with a new region if that's what you want.
+      K2_CHECK_EQ(cur_bytes_used, region_->bytes_used);
+      region_->Extend(new_bytes_used);
     }
+    dim_ = new_size;
```
qindazhu, line 186: just to make sure, if `new_size < dim_`, here we will drop some elements in *this, right? But as `region_->bytes_used` is not changed, so the dropped memory before `region_->bytes_used` has a change that will not be used by any Array anymore (i.e., idle forever)?

k2/csrc/array.h
```diff
@@ -161,30 +161,29 @@ class Array1 {
...
-  // Resizes, copying old contents if we could not re-use the same memory
-  // location. It will always at least double the allocated size if it has to
-  // reallocate. See Region::num_bytes vs. Region::bytes_used.
-  // TODO(haowen): now we only support the case that the current array `curr`
-  // (i.e. the array that will be resized) covers the highest used index in
-  // the region, that is, for any array `a` uses this region,
-  // curr.byte_offset_ + curr.Dim() * curr.ElementSize() == region_->bytes_used
-  // >= a.byte_offset + a.Dim() * a.ElementSize()
+  /*
+   Extend size of array, copying old contents if we could not re-use the same memory
```
qindazhu, line 165: Maybe it's better to update documentation here as we support `reduce array size` in `Resize` here? As I just feel `Extend` means that making it longer or doing nothing (as you did in Region.Extend())

k2/csrc/context.h
```diff
@@ -277,6 +277,35 @@ struct Region : public std::enable_shared_from_this<Region> {
...
+                         if num_bytes < new_bytes_used this region will be
+                         reallocated according to some heuristic (e.g. the
+                         larger of double the current size, or
+                         the next power of 2 greater than `new_bytes_used`). */
+  void Extend(size_t new_bytes_used) {
+    if (new_bytes_used <= bytes_used) return;
+    if (num_bytes < new_bytes_used) {  // reallocate and copy
+      size_t new_size = std::max<size_t>(num_bytes * 2, new_bytes_used);
+      size_t i = 4;
+      while (i < new_size / 8) i <<= 3;
```
qindazhu, line 294: Not sure if you want to expose function 
https://github.com/k2-fsa/k2/blob/960abbcfa75655a858ae2c6e1f5daea0bb7b5470/k2/csrc/utils.cu#L66 
to be public as use it here.

k2/csrc/ragged.h
```diff
@@ -72,6 +72,10 @@ class RaggedShape {
     }
   }
 
+  // Append `other` to `*this`.  `other` must have the same number of axes as
+  // `this`.  The reason this is
+  void _Append(const RaggedShape &other);
```
qindazhu, line 77: why is there a `_` here? seems typo?

k2/csrc/context.h
danpovey: that has the wrong type... for now let's leave as-is.

k2/csrc/array.h
danpovey: OK, will do.

k2/csrc/array.h
danpovey: yes.  This is similar to `resize` in std::vector.  If needed we can later remove the constraint that the array must be
using the last byte.  The point is if the memory was used by other arrays, they may still be pointing to that memory.
This is basically to enable appending to arrays efficiently.

k2/csrc/ragged.h
danpovey: I'll remove it.  Was going for a Pytorch-style convention but it's not consistent right now.

danpovey:
please merge if OK.

qindazhu:
Merging

PR #189, opened 2020-09-29
Convert FSAs to/from strings

csukuangfj:
~~Still working in progress.~~

~~Only the interface is added. Any comments are appreciated.~~

### DIFF
--- a/k2/csrc/fsa_utils.cu
+++ b/k2/csrc/fsa_utils.cu
@@ -1,0 +1,238 @@
+/**
+ * @brief Utilities for reading, writing and creating FSAs.
+ *
+ * @copyright
+ * Copyright (c)  2020  Mobvoi Inc.        (authors: Fangjun Kuang)
+ *
+ * @copyright
+ * See LICENSE for clarification regarding multiple authors
+ */
+
+#include <sstream>
+#include <vector>
+
+#include "k2/csrc/context.h"
+#include "k2/csrc/fsa.h"
+
+namespace k2 {
+
+// field separator within a line for a text form FSA
+static constexpr const char *kDelim = " \t";
+
+// Convert a string to an integer. Abort the program on failure.
+static int32_t StringToInt(const std::string &s) {
+  K2_CHECK_EQ(s.empty(), false);
+
+  bool ok = false;
+  char *p = nullptr;
+  // std::strtol requires a `long` type
+  long n = std::strtol(s.c_str(), &p, 10);  // NOLINT
+  if (*p == '\0') ok = true;
+
+  auto res = static_cast<int32_t>(n);
+  if (n != res) ok = false;  // out of range
+
+  if (ok == false)
+    K2_LOG(FATAL) << "Failed to convert " << s << " to an integer";
+
+  return res;
+}
+
+// Convert a string to a float. Abort the program on failure.
+static float StringToFloat(const std::string &s) {
+  K2_CHECK_EQ(s.empty(), false);
+  char *p = nullptr;
+  float f = std::strtof(s.c_str(), &p);
+  if (*p != '\0') K2_LOG(FATAL) << "Failed to convert " << s << " to a float";
+  return f;
+}
+
+// Trim leading and trailing spaces of a string.
+static void TrimString(std::string *s) {
+  K2_CHECK_NE(s, nullptr);
+  auto not_space = [](int32_t c) { return std::isspace(c) == 0; };
+
+  s->erase(s->begin(), std::find_if(s->begin(), s->end(), not_space));
+  s->erase(std::find_if(s->rbegin(), s->rend(), not_space).base(), s->end());
+}
+
+/** Split a string to a vector of strings using a set of delimiters.
+
+   Example usage:
+
+   @code
+    std::string in = "1 2 3";
+    const char *delim = " \t";
+    std::vector<std::string> out;
+    SplitStringToVector(in, delim, &out);
+   @endcode
+
+   @param [in]  in    The input string to be split.
+   @param [in]  delim A string of delimiters.
+   @param [out] out   It saves the split result.
+*/
+static void SplitStringToVector(const std::string &in, const char *delim,
+                                std::vector<std::string> *out) {
+  K2_CHECK_NE(delim, nullptr);
+  K2_CHECK_NE(out, nullptr);
+  out->clear();
+  std::size_t start = 0;
+  while (true) {
+    auto pos = in.find_first_of(delim, start);
+    if (pos == std::string::npos) break;
+
+    auto sub = in.substr(start, pos - start);
+    start = pos + 1;
+
+    TrimString(&sub);
+    if (!sub.empty()) out->emplace_back(std::move(sub));
+  }
+
+  if (start < in.size()) {
+    auto sub = in.substr(start);
+    TrimString(&sub);
+    if (!sub.empty()) out->emplace_back(std::move(sub));
+  }
+}
+
+/* Check that arcs are valid.
+
+   It will abort the program if there is an invalid arc.
+   An arc is invalid if one of the following happens:
+
+    - its symbol is -1, but its dest state is not final state.
+    - its dest state is larger than final state
+
+  @param [in] state_arcs   Indexed by state numbers.
+  @param [in] final_state  The final state.
+ */
+static void CheckStateArcs(const std::vector<std::vector<Arc>> &state_arcs,
+                           int32_t final_state) {
+  int32_t num_states = static_cast<int32_t>(state_arcs.size());
+  K2_CHECK_EQ(num_states, final_state + 1);
+  K2_CHECK_EQ(state_arcs[final_state].empty(), true);
+
+  for (int32_t i = 0; i != num_states; ++i) {
+    const auto &arcs = state_arcs[i];
+    for (const auto &arc : arcs) {
+      K2_CHECK_EQ(arc.src_state, i);
+      K2_CHECK_LE(arc.dest_state, final_state);
+      if (arc.symbol == -1) K2_CHECK_EQ(arc.dest_state, final_state);
+    }
+  }
+}
+
+/* Build an Fsa from a list of lists of arcs.
+
+  @param [in] state_arcs Indexes by states.
+  @param [in] num arcs   It is the number of arcs in total.
+                         Provided for optimization.
+ */
+static Fsa AcceptorFromStateArcs(
+    const std::vector<std::vector<Arc>> &state_arcs, int32_t num_arcs) {
+  int32_t num_states = static_cast<int32_t>(state_arcs.size());
+  std::vector<int32_t> row_splits(num_states + 1);
+  std::vector<Arc> arcs;
+  arcs.reserve(num_arcs);
+
+  row_splits[0] = 0;
+  for (int32_t i = 1; i != num_states; ++i) {
+    row_splits[i] =
+        row_splits[i - 1] + static_cast<int32_t>(state_arcs[i - 1].size());
+    arcs.insert(arcs.end(), state_arcs[i - 1].begin(), state_arcs[i - 1].end());
+  }
+  // the final_state has no leaving arcs, so there is nothing
+  // to add to `arcs` and row_splits[num_states]
+  row_splits[num_states] = row_splits[num_states - 1];
+
+  auto cpu_context = GetCpuContext();
+  Array1<int32_t> _row_splits(cpu_context, row_splits);
+  std::vector<RaggedShapeDim> axes(1);
+  axes[0].row_splits = _row_splits;
+  axes[0].cached_tot_size = -1;
+
+  RaggedShape shape(axes);
+  Array1<Arc> values(cpu_context, arcs);
+
+  return Fsa(shape, values);
+}
+
+// Create an acceptor from a stream.
+static Fsa AcceptorFromStream(std::string first_line, std::istringstream &is,
+                              bool negate_scores) {
+  std::vector<std::vector<Arc>> state_arcs;  // indexed by states
+  std::string line = std::move(first_line);
+  std::vector<std::string> splits;
+
+  float scale = 1;
+  if (negate_scores) scale = -1;
+
+  int32_t final_state = -1;  // invalid state
+  int32_t num_arcs = 0;
+  do {
+    SplitStringToVector(line, kDelim,
+                        &splits);  // splits is cleared in the function
+    if (splits.empty()) continue;  // this is an empty line
+
+    auto num_fields = splits.size();
+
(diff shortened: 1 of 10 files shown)


k2/csrc/fsa_utils.h
```diff
@@ -0,0 +1,85 @@
...
+
+/*
+  Read an Fsa from file.
+
+  @param [in]   filename
+  @param [in]   binary    true to read in binary format;
+                          false to read in text format.
+  @param [out]  aux_labels
+                          If the file contains a transducer,
+                          it will contain the aux_labels if not NULL.
```
qindazhu, line 78: The documentation is a little confusing here. Do we read aux_labels to `aux_labels` or provided an existed `aux_labels`?

k2/csrc/fsa_utils.h
```diff
@@ -0,0 +1,85 @@
...
+ */
+Fsa ReadFsa(const std::string &filename, bool binary = true,
+            Array1<int32_t> *aux_labels = nullptr);
```
qindazhu, line 81: And it's better if we can add documentation, says that this function only works on CPU? So if `aux_labels != nullptr`, there would be `aux_labels->Context()->GetDeviceType() == kCpu`

k2/csrc/fsa_utils.h
```diff
@@ -0,0 +1,85 @@
...
+  Read an Fsa from file.
+
+  @param [in]   filename
+  @param [in]   binary    true to read in binary format;
+                          false to read in text format.
+  @param [out]  aux_labels
+                          If the file contains a transducer,
+                          it will contain the aux_labels if not NULL.
+ */
+Fsa ReadFsa(const std::string &filename, bool binary = true,
```
qindazhu, line 80: There's no parameter `neg_score`, so I suppose lines in the file can be positive or negative and we just read them and there's no conversion? It's better if we can add documentation about this.

danpovey: Mostly we will be reading from std::string or something else in memory, not a file.  I don't think it's a good idea to be opening files from C++.

danpovey:
great!
I am working on fsa.h and fsa.cu so you might want to avoid touching that code.
It includes creation of FSA from Array1<Arc> so you could consider just going to that point.

k2/csrc/fsa_utils.h
csukuangfj: The signature is `Array1<int32_t> *aux_labels = nullptr`. 

Note that it is NOT `const Array1<int32_t> *aux_labels = nullptr`. So `aux_labels` is an output parameter.

k2/csrc/fsa_utils.h
csukuangfj: >  I don't think it's a good idea to be opening files from C++.

Are you going to read FSAs from Python?

I think we need a way to save/load FSAs to/from disks.

csukuangfj:
> It includes creation of FSA from Array1 so you could consider just going to that point.

Have you finished this part? I need the code constructing an FSA from `Array1<Arc>`.

k2/csrc/fsa_utils.h
csukuangfj: > There's no parameter neg_score

I was thinking it read whatever you saved. Since there is no `neg_score` in the write API, so it is absent from the read API.

k2/csrc/fsa_utils.h
csukuangfj: I will add documentation to clarify that it returns an FSA on the CPU.

k2/csrc/fsa_utils.h
qindazhu: Yeah, I'm OK with read whatever the user saved, just thought it's better to clarify this in the documentation.

k2/csrc/fsa_utils.cu
```diff
@@ -0,0 +1,238 @@
...
+static constexpr const char *kDelim = " \t";
+
+// Convert a string to an integer. Abort the program on failure.
+static int32_t StringToInt(const std::string &s) {
+  K2_CHECK_EQ(s.empty(), false);
+
+  bool ok = false;
+  char *p = nullptr;
+  // std::strtol requires a `long` type
+  long n = std::strtol(s.c_str(), &p, 10);  // NOLINT
```
qindazhu, line 29: Maybe it will be easy to use `std::stoi` and convert the exception to FATAL.

csukuangfj: stoi will throw an exception on failure. I think we don't want exceptions in K2.

csukuangfj:
Serialization APIs of FSAs (read/write)  are not implemented. 

It supports converting strings to/from FSAs.

It is ready for review. @danpovey @qindazhu

k2/csrc/fsa_utils.h
csukuangfj: Fixed.

k2/csrc/fsa_utils.h
csukuangfj: Read/write APIs are to be implemented in later pullrequests if they are needed.

k2/csrc/fsa_utils.cu
```diff
@@ -0,0 +1,328 @@
...
+#include "k2/csrc/fsa_utils.h"
+
+namespace k2 {
+
+// field separator within a line for a text form FSA
+static constexpr const char *kDelim = " \t";
+
+// Convert a string to an integer. Abort the program on failure.
+static int32_t StringToInt(const std::string &s) {
+  K2_CHECK_EQ(s.empty(), false);
```
qindazhu, line 25: just `K2_CHECK(!s.empty())`?

k2/csrc/fsa_utils.cu
```diff
@@ -0,0 +1,328 @@
...
+  char *p = nullptr;
+  // std::strtol requires a `long` type
+  long n = std::strtol(s.c_str(), &p, 10);  // NOLINT
+  if (*p == '\0') ok = true;
+
+  auto res = static_cast<int32_t>(n);
+  if (n != res) ok = false;  // out of range
+
+  if (ok == false)
+    K2_LOG(FATAL) << "Failed to convert " << s << " to an integer";
```
qindazhu, line 37: `K2_CHECK(ok) << ...`  ?

k2/csrc/fsa_utils.cu
```diff
@@ -0,0 +1,328 @@
...
+  char *p = nullptr;
+  float f = std::strtof(s.c_str(), &p);
+  if (*p != '\0') K2_LOG(FATAL) << "Failed to convert " << s << " to a float";
+  return f;
+}
+
+// Trim leading and trailing spaces of a string.
+static void TrimString(std::string *s) {
+  K2_CHECK_NE(s, nullptr);
+  auto not_space = [](int32_t c) { return std::isspace(c) == 0; };
```
qindazhu, line 54: better to add return type for the lambda, i.e. `bool`

csukuangfj: Done.

k2/csrc/fsa_utils.cu
csukuangfj: Fixed.

k2/csrc/fsa_utils.cu
csukuangfj: Done

k2/csrc/fsa_utils.cu
```diff
@@ -0,0 +1,328 @@
...
+  s->erase(std::find_if(s->rbegin(), s->rend(), not_space).base(), s->end());
+}
+
+/* Split a string to a vector of strings using a set of delimiters.
+
+   Example usage:
+
+   @code
+    std::string in = "1 2 3";
+    const char *delim = " \t";
```
qindazhu, line 65: is there a  space before `\t`?

k2/csrc/fsa_utils.cu
```diff
@@ -0,0 +1,328 @@
...
+    std::vector<std::string> out;
+    SplitStringToVector(in, delim, &out);
+   @endcode
+
+   @param [in]  in    The input string to be split.
+   @param [in]  delim A string of delimiters.
```
qindazhu, line 71: seems we only support `delim` to be a character? or the code can be run with `char sequence`?

k2/csrc/fsa_utils.cu
```diff
@@ -0,0 +1,327 @@
...
+    - its dest state is larger than final state
+
+  @param [in] state_arcs   Indexed by state numbers.
+  @param [in] final_state  The final state.
+ */
+static void CheckStateArcs(const std::vector<std::vector<Arc>> &state_arcs,
+                           int32_t final_state) {
+  int32_t num_states = static_cast<int32_t>(state_arcs.size());
+  K2_CHECK_EQ(num_states, final_state + 1);
+  K2_CHECK_EQ(state_arcs[final_state].empty(), true);
```
qindazhu, line 113: `K2_CHECK(..empty())`

k2/csrc/fsa_utils.cu
csukuangfj: Yes, you are right.
Fields are separated by space(s) or TAB(s), or both. The corresponding header describes the format.

k2/csrc/fsa_utils.cu
csukuangfj: https://en.cppreference.com/w/cpp/string/basic_string/find_first_of

(3)

It is a set of characters.

k2/csrc/fsa_utils.cu
csukuangfj: Done

k2/csrc/fsa_utils.cu
```diff
@@ -0,0 +1,327 @@
...
+
+  *aux_labels = Array1<int32_t>(GetCpuContext(), labels);
+
+  CheckStateArcs(state_arcs, final_state);
+  return AcceptorFromStateArcs(state_arcs, num_arcs);
+}
+
+Fsa FsaFromString(const std::string &s, bool negate_scores /*= false*/,
+                  Array1<int32_t> *aux_labels /*= nullptr*/) {
+  std::istringstream is(s);
```
qindazhu, line 276: Check `aux_labels`'s device type at entry? i.e. it should be on CPU if `aux_labels != nullptr`

k2/csrc/fsa_utils.cu
```diff
@@ -0,0 +1,327 @@
...
+    K2_CHECK_EQ(aux_labels->Dim(), arcs.Dim());
+    p = aux_labels->Data();
+  }
+  float scale = 1;
+  if (negate_scores) scale = -1;
+
+  std::ostringstream os;
+
+  int32_t n = arcs.Dim();
+  char sep = ' ';
```
qindazhu, line 315: `kDelim`?

k2/csrc/fsa_utils.cu
csukuangfj: `aux_labels` is not for reading. It does not matter where it currently resides.
It is assigned if the FSA is a transducer.

k2/csrc/fsa_utils.cu
csukuangfj: For printing, I described in the documentation that fields are separated by only ONE space.
For parsing, I also described that it supports multiple spaces/tabs between fields.

k2/csrc/fsa_utils_test.cu
```diff
@@ -0,0 +1,179 @@
...
+ */
+
+#include <gtest/gtest.h>
+
+#include "k2/csrc/fsa_utils.h"
+
+namespace k2 {
+
+// clang-format off
+bool operator==(const Arc &a, const Arc &b) {
```
qindazhu, line 18: I guess the function may be used in other places, so it would be better if we can put it in `Arc`. Anyway, it's OK to leave it here for now.

k2/csrc/fsa_utils.cu
qindazhu: Thanks for the explanation.

qindazhu:
I'm OK to merge this, Many thanks to @csukuangfj .

k2/csrc/fsa_utils.cu
```diff
@@ -0,0 +1,327 @@
...
+  if (negate_scores) scale = -1;
+
+  int32_t final_state = -1;  // invalid state
+  int32_t num_arcs = 0;
+  do {
+    SplitStringToVector(line, kDelim,
+                        &splits);  // splits is cleared in the function
+    if (splits.empty()) continue;  // this is an empty line
+
+    auto num_fields = splits.size();
```
danpovey, line 177: Actually it should be possible to just read the arcs as a vector and construct the FSA from that, based on existing functions in fsa.h.  (My other PR will correct some issues in their implementations).  But we can leave this for now, and involve that other code later after this is tested.

k2/csrc/ragged.h
```diff
@@ -98,6 +98,13 @@ class RaggedShape {
     return axes_[axis - 1].row_splits;
   }
 
+  const Array1<int32_t> &RowSplits(int32_t axis) const {
```
danpovey, line 101: no need for a separate one for const and non-const: just mark the other one const.

csukuangfj: > k2/csrc/ragged.h(98): error: qualifiers dropped in binding reference of type "k2::Array1<int32_t> &" to initializer
of type "const k2::Array1<int32_t>"

The function returns a reference and the compiler will complain.

k2/csrc/fsa_utils.h
```diff
@@ -0,0 +1,85 @@
...
+  @param [in]   negate_scores
+                    If true, the string form has the weights as costs,
+                    not scores, so we negate as we read.
+  @param [out]  aux_labels
+                    Used only when it is a transducer. It contains the
+                    aux_label of each arc.
+
+  @return It returns an Fsa.
+ */
+Fsa FsaFromString(const std::string &s, bool negate_scores = false,
```
danpovey, line 59: What's more important, to me, than reading/writing to files, is printing to a string in the same format
that FsaFromString uses.  I think that's more likely to be used in practice.

My plan for file-based I/O was to not really do this from C++ but instead do it from Python via vectors of arcs,
i.e. write the tensor to disk using torch or TF mechanisms according to the frontend framework.  I'm open
to discussion on this though.
I'd rather not confuse matters by having un-used APIs around though, if we stick to my original plan.

csukuangfj: Removed from c++.

k2/csrc/fsa_utils.cu
csukuangfj: Fixed. Replaced with API `FsaFromArray1`.

csukuangfj:
Ready to merge @danpovey

qindazhu:
Merging..

PR #193, opened 2020-09-29
Wrap Fsa to Python.

csukuangfj:


k2/python/csrc/torch/fsa.cu
```diff
@@ -0,0 +1,64 @@
...
+
+static void PybindFsaImpl(py::module &m) {
+  using PyClass = Fsa;
+  py::class_<PyClass> pyclass(m, "Fsa");
+  pyclass.def(py::init<>());
+  pyclass.def(py::init([](torch::Tensor &arcs) {
+    Array1<Arc> array = FromTensor<Arc>(arcs);
+    bool error = true;
+    Fsa fsa = FsaFromArray1(array, &error);
+    K2_CHECK_EQ(error, false);
```
qindazhu, line 28: `K2_CHECK(!error)`

k2/python/csrc/torch/arc.cu
```diff
@@ -0,0 +1,46 @@
...
+      .def_readwrite("symbol", &PyClass::symbol)
+      .def_readwrite("score", &PyClass::score);
+
+  pyclass.def("__str__", [](const PyClass &self) -> std::string {
+    std::ostringstream os;
+    os << self;
+    return os.str();
+  });
+
+  m.def("float_as_int",
```
qindazhu, line 37: I think it would be great if we can define such function (i.e. reinterpret int as float or vice versa ) for a tensor (e.g. for arcs' weight in an Fsa which is a n*1 tensor, or even for an Fsa), as we usually would access a tensor instead of a particular value in the tensor.

k2/python/csrc/torch/fsa.cu
csukuangfj: Fixed.

k2/python/csrc/torch/arc.cu
csukuangfj: Done.

csukuangfj:
@qindazhu @danpovey 
ready for review.

k2/python/csrc/torch/arc.cu
danpovey: With these kinds of things I think the documentation might be a problem.  When a Python programmer sees this, who doesn't understand C++, what will they make of it?  It's even a little confusing to me, e.g. how would this be invoked?  As _k2.float_as_int(1.0) ->  [some integer] ?

danpovey: Also, I was thinking of separating the FSA-related things into a different namespace k2fsa, instead of k2, at the python level.  I don't know if you guys think that is a bad idea though.
Of course we may still use the same internal namespace, _k2, for all the C++-exported things.

csukuangfj: They are invoked like this:

https://github.com/k2-fsa/k2/blob/a6b25e8daf33d7ab04999109da6772d2b17c62c1/k2/python/tests/fsa_test.py#L58-L66

An arc contains 3 integers and 1 float but it is saved as a torch::Tensor with dtype==torch.int32. We want to reinterpret the last integer as a float.

When `int_as_float` is applied to a tensor, it is invoked like this:

https://github.com/k2-fsa/k2/blob/a6b25e8daf33d7ab04999109da6772d2b17c62c1/k2/python/tests/fsa_test.py#L68-L76

qindazhu: I was thinking those APIs would be just for internal usage, we can expose functions later such as `Fsa.get_weights`, 'Fsa.create_from_arc_and_weights' which may be called directly by the users, based on below reinterpretation version for tensor.

csukuangfj: There is a `view` method in `numpy`: https://numpy.org/doc/stable/reference/generated/numpy.ndarray.view.html

We should probably rename them to `view`.

PyTorch users have this kind of need, see https://discuss.pytorch.org/t/reinterpret-pytorch-array-as-a-different-dtype/24256

Unfortunately, PyTorch does support this kind of operation.

csukuangfj: > When a Python programmer sees this, who doesn't understand C++, what will they make of it?

If we rename them to `view`, I think a Python programmer who is familiar with `numpy` knows the meaning
of the method once he/she sees it.

csukuangfj: > Also, I was thinking of separating the FSA-related things into a different namespace k2fsa, instead of k2, at the python level. 

this is fine with me.

qindazhu: > Also, I was thinking of separating the FSA-related things into a different namespace k2fsa, instead of k2, at the python level. I don't know if you guys think that is a bad idea though.
 
Did you mean we suppose user just call Fsa-related code at python level? no Ragged and Array? If user still need to use Ragged and Array, I don't see the necessity to just put fsa code into a new namespace (as user will see two global namespace? i.e. k2 and k2fsa?)

k2/python/csrc/torch/fsa.cu
```diff
@@ -0,0 +1,89 @@
...
+#include "k2/csrc/fsa_utils.h"
+#include "k2/python/csrc/torch/fsa.h"
+#include "k2/python/csrc/torch/torch_util.h"
+#include "torch/extension.h"
+
+namespace k2 {
+
+static void PybindFsaImpl(py::module &m) {
+  using PyClass = Fsa;
+  py::class_<PyClass> pyclass(m, "Fsa");
```
danpovey, line 25: OK, what I was thinking was that we could export the Ragged template (maybe instantiated for a certain set of types) pretty directly, i.e one-for-one its members, and that the Fsa object that appears in Python would actually be a Python object, whose "arcs" member would be the Ragged<Arc> object.  We could perhaps make Fsa one object at the Python level that includes vectors of FSAs, I'm not convinced that it makes sense to have separate types for different levels of vector, we could  even support vector-of-vector-of-FSA which would be useful for n-best lists.

k2/python/tests/fsa_test.py
```diff
@@ -0,0 +1,157 @@
...
+            3 6 -1  -7.2
+            5 0  1  -8.2
+            6
+        '''
+
+        fsa, aux_labels = k2.fsa_from_str(s)
+        assert aux_labels is None
+
+        expected_str = '''0 1 2 -1.2
+0 2 10 -2.2
```
qindazhu, line 35: strange format here..

k2/python/csrc/torch/arc.cu
csukuangfj: `torch.Tensor` has a `view` method: https://pytorch.org/docs/stable/tensors.html#torch.Tensor.view
However, it only supports reshaping and does not allow to change the underlying data type.

k2/python/tests/fsa_test.py
csukuangfj: This is equivalent to
```
0 1 2 -1.2
0 2 10 -2.2
```

It has no leading spaces.

csukuangfj: If you can think of a better-looking format, I am willing to do a switch.

qindazhu: it's OK, not a big issue, just feel  aligned columns  would be easy to read.

k2/python/csrc/torch/fsa.cu
qindazhu: Not sure if I understand correctly, supposing there is a peoperty `ndim` in `Fsa` object, then an `Fsa` object with `ndim = n` is a vector of `Fsa`s with `ndim = n-1`, right?

csukuangfj: I'll make FSA a python class and wrap Ragged<T> instead.

danpovey:
Great, thanks, yes, that's what my plan was.

k2/python/tests/fsa_test.py
csukuangfj: Fixed.

csukuangfj:
Now `Fsa` is a Python class.

I will add more documentation to `fsa.py`.

Ready for review @danpovey  @qindazhu

k2/python/csrc/torch/arc.cu
csukuangfj: Now, these methods are used only **internally**.

k2/python/csrc/torch/fsa.cu
csukuangfj: Fixed.

qindazhu:
It's OK to me. We can add those `vector-of-vector-FSAs` support later.

k2/python/csrc/torch/ragged.cu
```diff
@@ -0,0 +1,95 @@
...
+  // in Python, so the only way to get an instance of
+  // RaggedShape is from Ragged<T>.
+
+  pyclass.def("dim0", &PyClass::Dim0);
+  pyclass.def("total_size", &PyClass::TotSize, py::arg("axis"));
+
+  // TODO(fangjun): Append has not been implemented yet
+  // pyclass.def("append", &PyClass::Append, py::arg("other"));
+
+  pyclass.def("numel", &PyClass::NumElements);
```
danpovey, line 34: We'll need to document the Python interface somehow, esp. as regards these naming differences.
One possibility is to document this by the C++ declarations, i.e. make a note of their names in python?

k2/python/csrc/torch/ragged.cu
```diff
@@ -0,0 +1,95 @@
...
+static void PybindRaggedTpl(py::module &m, const char *name) {
+  using PyClass = Ragged<T>;
+  py::class_<PyClass> pyclass(m, name);
+
+  pyclass.def_readwrite("shape", &PyClass::shape);
+  pyclass.def_readwrite("values", &PyClass::values);
+
+  pyclass.def("num_axes", &PyClass::NumAxes);
+  pyclass.def("index", &PyClass::Index, py::arg("axis"), py::arg("i"));
+  pyclass.def("remove_axis", &PyClass::RemoveAxis, py::arg("axis"));
```
danpovey, line 72: FYI, I'm questioning whether it was a good idea to make Index() and RemoveAxis() members;  might it have been more consistent to make them non-member templates and put them in [new file] ragged_ops.h?   However the members could still be convenient from Python.

k2/python/csrc/torch/ragged.cu
csukuangfj: I agree. I have been planning to add documentation to every wrapped API.
What is the progress of setting up the sphinx documentation environment for K2 @songmeixu ?

k2/python/csrc/torch/ragged.cu
```diff
@@ -0,0 +1,95 @@
...
+
+  pyclass.def(
+      "cuda",
```
danpovey, line 75: OK..  this will probably work given the current compilation setup, where we use only the PyTorch context if we're compiled with pytorch.  When I added the member-function GetCpuContext() to the Context class I assumed we would be generating these contexts from existing contexts, mostly.  I guess this is OK, but I just want to remind people that my original idea was that we'd mostly generate contexts from other contexts.

Also I don't know which GPU this would give us in cases where there were multiple GPUs ?

k2/python/csrc/torch/torch_util.cu
```diff
@@ -24,4 +24,53 @@ torch::DeviceType ToTorchDeviceType(DeviceType type) {
...
+  auto device = torch::Device(device_type, device_id);
+  auto scalar_type = ToScalarType<int32_t>::value;
+  // an Arc has 4 members
+  torch::IntArrayRef strides = {4, 1};  // in number of elements
+  auto options = torch::device(device).dtype(scalar_type);
+
+  // NOTE: we keep a copy of `array` inside the lambda
+  // so that `torch::Tensor` always accesses valid memory.
+  return torch::from_blob(
+      array.Data(), {array.Dim(), 4}, strides, [array](void *) {}, options);
```
danpovey, line 53: I'm a little unclear what the lambda
`[array] (void *) {}`
is doing, and why it needs to capture `array` if it is doing nothing like it seems to.

k2/python/csrc/torch/ragged.cu
csukuangfj: > Also I don't know which GPU this would give us in cases where there were multiple GPUs ?

This would be the GPU specified by the `gpu_id`. If no `gpu_id` is specified.
PyTorch will use the default GPU, like the following code:

```python
a = torch.arange(10).cuda()  // PyTorch select the `default gpu`.
```

> When I added the member-function GetCpuContext() to the Context class I assumed we would be generating these contexts from existing contexts

How do we get the first context?  I think there are only allocators associated with
`torch::Tensor`, but no `Context`.

danpovey: Mm, I hadn't mapped out the details, e.g. I don't know whether you can get this from the C++ or have to do it from Python.  Anyway the way it is now seems fine.  We could perhaps just remove that function GetCpuContext() from the Context interface.

k2/python/csrc/torch/torch_util.cu
csukuangfj: `torch::Tensor` shares the same memory with `array`.  What happens if  we try  to access the returned `tensor` after
`array` is freed?

`[array] (void *) {}` is a deleter for the `torch::Tensor`. In our case, it deletes nothing, since the memory
is managed by the `array`. But we have to ensure that the `array` outlives the returned `tensor`.

The lamda deleter captures a copy of `array` to keep it from being freed.

---

You can consider this as a trick, though I do not like tricks.

PyTorch also uses this kind of technique for its `from_numpy()` method. Refer to
https://github.com/pytorch/pytorch/blob/master/torch/csrc/utils/tensor_numpy.cpp#L178

```cpp
  Py_INCREF(obj);
  return at::from_blob(
      data_ptr,
      sizes,
      strides,
      [obj](void* data) {
        pybind11::gil_scoped_acquire gil;
        Py_DECREF(obj);
      },
      at::device(kCPU).dtype(numpy_dtype_to_aten(PyArray_TYPE(array)))
  );
```

csukuangfj:
I just managed to visualize an Fsa in [colab notebook][1]

Screen Shot 2020-10-02 at 17 55 47

danpovey:
Ah, I see, great.
It would be nice if you could document that at that point.

danpovey:
Fantastic!

csukuangfj:
> It would be nice if you could document that at that point.

I'll add more detailed documentation.

csukuangfj:
Merging

PR #196, opened 2020-10-01
Added a note in INSTALL regarding the cmake version

chenguoguo:
I started by using cmake v3.16.3, and I got error message `nvcc fatal : Unknown option '-Wall'`. Somehow cmake was passing the cxx flags to the cuda flags. The issue was fixed after upgrading to cmake v3.18.0. I added a note in the INSTALL file that we haven't tested all version of cmake and people should probably start with cmake v3.18.0.

At some point we may want to improve the installation process by (my two cents):
* Write dependency documentation (in INSTALL?). Add requirements.txt for Python dependencies.
* Let users choose their own Python environment. For now, cmake automatically picks up the Python environment. This is not ideal for Python developers as people may have a lot of different environments (through virtual environment), and the one picked up by cmake may or may not work.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -7,3 +7,8 @@ The installation instructions are
 
 You can run tests with
     make test
+
+Note: The current installation only works with certain versions of cmake, e.g.,
+      it works with cmake v3.18.0, but it does not work with cmake v3.16.3. We
+      have not done a thorough test on this yet. Try to use v3.18.0 if you get
+      issues running the installation.


chenguoguo:
Looks like I can't assign reviewers. Could @csukuangfj or @danpovey take a look?

csukuangfj:
RE CMake versions,  I just tried

- locally, with CMake 3.11.0 + Python 3.8 + PyTorch 1.6.0 + cuda 10.1
- with [colab notebook][1], with CMake 3.11.0 + Python 3.7 + PyTorch 1.6.0 + cuda 10.1

Both compile and have the test cases passed. I agree that we should document which CMake version is required by K2.


As for `requirements.txt`, I agree with you. Currently, we depend only on PyTorch. The latest stable version 1.6.0 is known
to work, though I have not tested other versions. I've committed a [requirements.txt][2] in #193 , but it only contains
packages for formatting purposes.  Maybe we can state explicitly which version of PyTorch is required by K2.

RE Python environment, you are right. The current setup is not ideal. It uses whichever version pybind11 finds.
It would be better if users can choose which Python to use. 

[2]: https://github.com/k2-fsa/k2/blob/b3ed9dd6ab08b1415f88782452010a1cc980a553/requirements.txt
[1]: https://colab.research.google.com/drive/1qbHUhNZUX7AYEpqnZyf29Lrz2IPHBGlX?usp=sharing#scrollTo=U0XpYZd4_6Od

csukuangfj:
> This is not ideal for Python developers as people may have a lot of different environments (through virtual environment), and the one picked up by cmake may or may not work.

If you activate your virtual environment before running `cmake build`, I think it will pick the Python you are currently using.

chenguoguo:
I was using a virtual environment (through pip). pybind11 first selected the conda environment that I had under my home directory. After deleting that environment, it selected the default python under /user/bin/... It didn't select python from my virtual environment.

I was using cmake 3.16.3 + python 3.6.9 + pytorch 1.5.0 + cuda 10.2 + ubuntu 1604, and building was complaining about `nvcc fatal : Unknown option '-Wall'`.  The error went away after upgrading cmake from 3.16.3 to 3.18.0, but still, some tests failed:

```
The following tests FAILED:
	  6 - Test.Cuda.cu_ragged_shape_test (Child aborted)
	  7 - Test.Cuda.cu_ragged_test (Failed)
	 24 - array_test_py (Failed)
```

I guess we may need a specific combination of pytorch + cuda. It's always tricky anyways for this kind of things...

I created this issue (https://github.com/k2-fsa/k2/issues/197) to keep track of the points that we have to improve. We can have further discussion there regarding the build process, and implement the things that we think are reasonable.

csukuangfj:
Could you use

```bash
cmake -DPYBIND11_PYTHON_VERSION=3.6 -DCMAKE_PREFIX_PATH=/path/to/your/virtual_env  ..
```
I assume that there is a binary named `python3.6` or `python3` inside `/path/to/your/virtual_env/bin`.

Will it still pick up your conda environment or your virtual environment with pip?

chenguoguo:
I tried:

```
cmake -DPYBIND11_PYTHON_VERSION=3.6.9 -DCMAKE_PREFIX_PATH=/my_virtual_env  ..
```

DPYBIND11_PYTHON_VERSION=3.6.9 is working. I have python 3.7 in my conda environment, and instead of using pyhon3.7 in the conda environment, it was using python3.6.9. But it was using /usr/bin/python3.6 instead of using the python from my virtual environment.

Regarding the test fail, I took a closer look by running the test in verbose mode. Test 6, 7 and 24 (when it was calling test_cuda_float_array1_to_tensor) failed because they were not able to find a proper cuda device. I have one job running on my GPU 0 (which takes almost all the memory on that card), I guess they failed because they were trying to use GPU 0 but it ran out of memory. I also have 2 idle GPU cards on the same machine. What is the GPU selection logic here?

chenguoguo:
I re-installed cmake 3.16.3 from source, and it looks like now cmake is working properly (without the `nvcc fatal : Unknown option '-Wall'` error). So I guess it was not a cmake version issue, but something else. I'm going to close this PR, and we can address the build process from issue #197

qindazhu:
> What is the GPU selection logic here

For now we just choose the first available GPU device without checking its memory usage. Definitely GPU selection strategy should be addressed later.
https://github.com/k2-fsa/k2/blob/866c6cf9c7440285678e7ca9a51ccc317a7162af/k2/csrc/pytorch_context.cu#L121-L123
https://github.com/k2-fsa/k2/blob/866c6cf9c7440285678e7ca9a51ccc317a7162af/k2/csrc/default_context.cu#L51-L59

danpovey:
Regarding GPU selection: for most use-cases we'll probably be using PyTorch
or TensorFlow as the outer-level code and will be piggybacking on their GPU
selection strategy.  This native Context object is mostly intended for
testing.  But of course we should still improve it.

Issue #180, opened 2020-09-25
Goal: composition to create supervision FSAs


danpovey:
Guys,

Here's something that it would be great to be able to finish within a couple of weeks.  I'd like to be able to do the
following:

  - Create a lexicon.fst in Kaldi and write it as a text file.
  - Convert that to a k2 FSA (with the olabels as a separate vector)
  - Do Invert() to swap the olabels and ilabels, as we need to compose on the word labels
  - Convert a string (a supervision) to a linear FSA, e.g. could just make the FSA as a string.
  - Compose the lexicon FSA and the linear supervision FSA
  - Use the arc mapping to replace the word labels on the linear supervision FSA with the phone labels

This will require writing quite a bit of code, but I think it's mostly straightforward.  I have not had enough energy to be extremely pro-active with some of this stuff.  If there's anything you guys need clarification on, let me know.

chenguoguo:
I started taking a closer look of the codebase and this might be something that I can help with if it's not assigned yet.

danpovey:
Fantastic!!
Bear in mind, though, this this is a medium-term goal as there are many
pieces that need to be finished before this is done.
For example:
   - Add Python wrapping for the code in csrc/ (currently we only have it
for csrc/host/)
   - Add and test interface to construct k2::Fsa from string and output to
string
Will add more detail in a bit...

danpovey:
A sub-task that you could start with, Guoguo, is to begin by adding support
for weights in FsaToString and StringToFsa in the host/
subdirectory.  A couple weeks ago I added the weights to the Arc in the
host/ dir (they had previously been separate).

I commented on an issue recently about what the options for the conversion
to/from strings would look like:

https://github.com/k2-fsa/k2/issues/178

Fangjun said he would look at that issue (178) more broadly, but you could
start with that small part of it to familiarize yourself with the workflow,
I don't think that will inconvenience him.

csukuangfj:
Glad to see a new contributor.

chenguoguo:
Sounds good to me. I'll spend some time next week to look into adding weights to FsaToString. I'll get back to you/Fangjun if I need some clarification.

csukuangfj:
@danpovey 
Do you want this to be implemented in Python or in C++?

danpovey:
Each bullet point would be invoked separately from Python, although of course  the appropriate wrappers would call C++ code.

qindazhu:
Closed as we have fininshed this.

Issue #178, opened 2020-09-25
Conversion to/from strings


danpovey:
Guys,
I dont know where we are with the I/O format right now, but I propose that we find a way to enable I/O to and
from FSAs and strings.
We could have the following options, with defaults:

   acceptor=True   # if true, format is: src_state dest_state label cost.  if false it's src_state dest_state label aux_label cost; the aux_label would be stored as a separate vector of int32.
  negate_scores=False   # if true, the string form has the weights as costs, not scores, so we negate as we read/write

The final-state is of course treated differently.  The format will be like OpenFST, as just:
final_state
(with no cost, since we don't support a cost on the final state, there should be an arc with -1 going to final state, which
has the costs on).

Eventually we'd have options related to symbol tables, i.e. allow the user to supply a symbol table.
 
This interface should probably exist at the C++ level in some form, since the aim is to basically have the same functionality
available from C++ as from Python.

qindazhu:
Actually @csukuangfj wrote a basic version before, so I guess he may be interested in this task.

https://github.com/k2-fsa/k2/blob/960abbcfa75655a858ae2c6e1f5daea0bb7b5470/k2/csrc/host/fsa_util.h#L238-L252

csukuangfj:
I would like to implement it.

danpovey:
Fantastic!

csukuangfj:
Can we assume that the `final_state` has the largest state number?

danpovey:
Yes, it would be an error if not.

qindazhu:
Closed as we have finished this.

PR #199, opened 2020-10-02
Added weight support for StringToFsa and FsaToString

chenguoguo:
Things I did:
   * Added weight support to both StringToFsa and FsaToString
   * Removed GetArcWeights as it's not used in other places and doesn't seem to be necessary
   * Moved the actual string read work out of StringToFsa::GetSizes(). The previous implementation assumes that user first run StringToFsa::GetSizes() and then run StringToFsa::GetOutput(), but they are both public functions. Assumptions like this will make programming more complicated.

Things to discuss:
   * How general should the input string FSA be? I did add support for final state with weight (typical OpenFst style WFSA) and convert that to K2 style (final state without weight), but I didn't extend it to support multiple final states (since the original code doesn't support it). Do we have to support all WFSAs that can be generated by e.g., fstprint?
   * I read weight as is, no conversions between scores/costs whatsoever. Do we have to convert it between scores/costs, and also check if that's valid weight?

### DIFF
--- a/k2/csrc/host/fsa_util.cc
+++ b/k2/csrc/host/fsa_util.cc
@@ -79,7 +79,7 @@ int32_t StringToInt(const std::string &s, bool *is_ok = nullptr) {
                       is successful; false otherwise
   @return The converted float number.
  */
-int32_t StringToFloat(const std::string &s, bool *is_ok = nullptr) {
+float StringToFloat(const std::string &s, bool *is_ok = nullptr) {
   K2_CHECK_EQ(s.empty(), false);
 
   bool ok = false;
@@ -100,21 +100,6 @@ std::vector<int32_t> StringVectorToIntVector(
   for (const auto &s : in) {
     bool ok = false;
     auto n = StringToInt(s, &ok);
-    K2_CHECK(ok);
-    res.push_back(n);
-  }
-  return res;
-}
-
-/** Convert `std::vector<std::string>` to `std::vector<float>`.
- */
-std::vector<float> StringVectorToFloatVector(
-    const std::vector<std::string> &in) {
-  std::vector<float> res;
-  res.reserve(in.size());
-  for (const auto &s : in) {
-    bool ok = false;
-    auto n = StringToFloat(s, &ok);
     K2_CHECK(ok);
     res.push_back(n);
   }
@@ -279,15 +264,14 @@ void StringToFsa::ReadArcsFromString() {
 
     K2_CHECK_EQ(finished, false);
 
-    auto fields = StringVectorToFloatVector(splits);
-    auto num_fields = fields.size();
+    auto num_fields = splits.size();
     if (num_fields == 3u || num_fields == 4u) {
       Arc arc{};
-      arc.src_state = static_cast<int32_t>(fields[0]);
-      arc.dest_state = static_cast<int32_t>(fields[1]);
-      arc.label = static_cast<int32_t>(fields[2]);
+      arc.src_state = StringToInt(splits[0]);
+      arc.dest_state = StringToInt(splits[1]);
+      arc.label = StringToInt(splits[2]);
       if (num_fields == 4u) {
-        arc.weight = fields[3];
+        arc.weight = StringToFloat(splits[3]);
       }
 
       auto new_size = std::max(arc.src_state, arc.dest_state);
@@ -299,16 +283,16 @@ void StringToFsa::ReadArcsFromString() {
       // TODO(guoguo): the original code only supports one final state. Is it
       //               necessary to supports multiple final states?
       finished = true;
-      K2_CHECK_EQ(static_cast<int32_t>(fields[0]) + 1,
+      K2_CHECK_EQ(StringToInt(splits[0]) + 1,
                   static_cast<int32_t>(arcs_.size()));
       if (num_fields == 2u) {
         arcs_.resize(arcs_.size() + 1);
         // If the final state bears a weight, we convert that into an arc.
         Arc arc{};
-        arc.src_state = static_cast<int32_t>(fields[0]);
-        arc.dest_state = static_cast<int32_t>(fields[0]) + 1;
+        arc.src_state = StringToInt(splits[0]);
+        arc.dest_state = StringToInt(splits[0]) + 1;
         arc.label = k2host::kFinalSymbol;
-        arc.weight = fields[1];
+        arc.weight = StringToFloat(splits[1]);
         arcs_[arc.src_state].push_back(arc);
         ++num_arcs_;
       }
--- a/k2/csrc/host/fsa_util.cc
+++ b/k2/csrc/host/fsa_util.cc
@@ -75,7 +75,7 @@ int32_t StringToInt(const std::string &s, bool *is_ok = nullptr) {
 /** Convert a string to a float number.
 
   @param [in]   s     The input string.
-  @param [out]  is    If non-null, it is set to true when the conversion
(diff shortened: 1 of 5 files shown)


chenguoguo:
To address the comment here: https://github.com/k2-fsa/k2/issues/180

k2/csrc/host/fsa_util.cc
```diff
@@ -251,46 +262,63 @@ void GetArcIndexes2(const Array2<int32_t *, int32_t> &arc_map,
...
-    auto fields = StringVectorToIntVector(splits);
+    auto fields = StringVectorToFloatVector(splits);
     auto num_fields = fields.size();
-    if (num_fields == 3u) {
+    if (num_fields == 3u || num_fields == 4u) {
       Arc arc{};
-      arc.src_state = fields[0];
-      arc.dest_state = fields[1];
-      arc.label = fields[2];
+      arc.src_state = static_cast<int32_t>(fields[0]);
```
csukuangfj, line 286: We have to check that these are integers, i.e., there are no truncations.

---

https://github.com/k2-fsa/k2/blob/866c6cf9c7440285678e7ca9a51ccc317a7162af/k2/csrc/fsa_utils.cu#L115-L122 may be another alternative.

k2/csrc/host/fsa_util.cc
```diff
@@ -251,46 +262,63 @@ void GetArcIndexes2(const Array2<int32_t *, int32_t> &arc_map,
...
       auto new_size = std::max(arc.src_state, arc.dest_state);
       if (new_size >= arcs_.size()) arcs_.resize(new_size + 1);
 
       arcs_[arc.src_state].push_back(arc);
-      ++num_arcs;
-    } else if (num_fields == 1u) {
+      ++num_arcs_;
+    } else if (num_fields == 1u || num_fields == 2u) {
+      // TODO(guoguo): the original code only supports one final state. Is it
+      //               necessary to supports multiple final states?
```
csukuangfj, line 300: I think there would be only one final state.
To support multiple final states, we can add a new state and let these final
states connect to the new state so the new state becomes the only final state.

k2/csrc/host/fsa_util.cc
```diff
@@ -251,46 +262,63 @@ void GetArcIndexes2(const Array2<int32_t *, int32_t> &arc_map,
...
       finished = true;
-      K2_CHECK_EQ(fields[0] + 1, static_cast<int32_t>(arcs_.size()));
+      K2_CHECK_EQ(static_cast<int32_t>(fields[0]) + 1,
```
csukuangfj, line 302: According to https://github.com/k2-fsa/k2/issues/178#issue-708879692

> The final-state is of course treated differently. The format will be like OpenFST, as just:
final_state
(with no cost, since we don't support a cost on the final state, there should be an arc with -1 going to final state, which
has the costs on).

I think there is only one field for the final state.

k2/csrc/host/fsa_util.cc
```diff
@@ -71,6 +72,25 @@ int32_t StringToInt(const std::string &s, bool *is_ok = nullptr) {
...
 }
 
+/** Convert a string to a float number.
+
+  @param [in]   s     The input string.
+  @param [out]  is    If non-null, it is set to true when the conversion
+                      is successful; false otherwise
+  @return The converted float number.
+ */
+int32_t StringToFloat(const std::string &s, bool *is_ok = nullptr) {
```
csukuangfj, line 82: the return type is `float`.

k2/csrc/host/fsa_util.cc
chenguoguo: Fixed.

k2/csrc/host/fsa_util.cc
chenguoguo: It will make things easier if there's only one field. I thought we were going to support more general OpenFst style WFSAs.

k2/csrc/host/fsa_util.cc
chenguoguo: Thanks, fixed.

csukuangfj:
+1

k2/csrc/host/fsa_util.cc
```diff
@@ -71,6 +72,25 @@ int32_t StringToInt(const std::string &s, bool *is_ok = nullptr) {
   return res;
 }
 
+/** Convert a string to a float number.
+
+  @param [in]   s     The input string.
+  @param [out]  is    If non-null, it is set to true when the conversion
```
qindazhu, line 78: `is_ok` instead of `is`?

k2/csrc/host/fsa_util.cc
```diff
@@ -251,46 +247,62 @@ void GetArcIndexes2(const Array2<int32_t *, int32_t> &arc_map,
...
-    } else if (num_fields == 1u) {
+      ++num_arcs_;
+    } else if (num_fields == 1u || num_fields == 2u) {
+      // TODO(guoguo): the original code only supports one final state. Is it
+      //               necessary to supports multiple final states?
       finished = true;
-      K2_CHECK_EQ(fields[0] + 1, static_cast<int32_t>(arcs_.size()));
+      K2_CHECK_EQ(StringToInt(splits[0]) + 1,
+                  static_cast<int32_t>(arcs_.size()));
+      if (num_fields == 2u) {
```
qindazhu, line 288: For Fsa in `k2`, definitely the final states would have no weight.

k2/csrc/host/fsa_util.cc
chenguoguo: Thanks, fixed.

k2/csrc/host/fsa_util.cc
chenguoguo: OK, I was trying to make the code smart, and let it convert the final state to an arc with weight if it sees OpenFST style final state. But I guess simple is good, as long as we have the same assumption of the input FST. I deleted this part of the code.

qindazhu:
BTW, @chenguoguo, I guess you may miss the PR https://github.com/k2-fsa/k2/pull/189 from @csukuangfj before which supports weights reading. 
Also note that we would delete those `/host/*.*` files finally (but for now we have to keep them to wrap some CPU fsa-related algorithms in that folder with struct `Fsa` in https://github.com/k2-fsa/k2/blob/master/k2/csrc/fsa.h, to keep up with the deadline..)
So if you want to add some functions later to read openFST format (e.g  `lexicon.fst`) to k2 Fsa, it would be better if can add those functions in `k2/csrc` (instead of `k2/csrc/host)`. Thanks again.

danpovey:
Oh, I'm sorry, it's my fault, I should have told you that Fangjun already did that reading stuff.
Regarding the final-weights, we support the same thing but the format is different.  The final weight goes on that special arc that has "-1" as the label, and our final-state is like a super-final state.  Regarding how to
support OpenFST-format input and output... we'd have to add the super-final state when reading; would involve going back and "fixing" final-arcs by adding the correct super-final state-id.
We previously discussed an option to negate the costs; it might make sense to just have a single option 'openfst' where if it's true, it negates the costs and also does this super-final-state stuff.
But this should be done in the code in csrc/, not csrc/host/ (as you noticed, the host stuff is old and not necessary in the long term).

chenguoguo:
Yeah I didn't notice #189 ...

I'll add support for the "openfst" option under csrc. Created the issue here: https://github.com/k2-fsa/k2/issues/200 . I listed 3 things that we have to support. Let me know if we have to support more.

Issue #200, opened 2020-10-02
add openfst support for fst read


chenguoguo:
- [x] replace the `negate_scores` option with `openfst`
- [x] support final states with weight when `openfst` is true
- [x] support multiple final states when `openfst` is true

danpovey:
Yes.  Bear in mind the equivalence between the two formats.  We represent final-probs as "special arcs" because it greatly simplifies algorithms.

chenguoguo:
Close this via https://github.com/k2-fsa/k2/pull/202

PR #202, opened 2020-10-02
Added openfst support for fst read

chenguoguo:
To address https://github.com/k2-fsa/k2/issues/200

### DIFF
--- a/k2/csrc/fsa_utils.cu
+++ b/k2/csrc/fsa_utils.cu
@@ -240,7 +240,11 @@ static Fsa TransducerFromStream(std::string first_line, std::istringstream &is,
                         super_final_state,
                         -1,             // kFinalSymbol
                         scale * original_final_weights[i]);
-      state_aux_labels.push_back(-1);   // kFinalSymbol
+      // TODO(guoguo) We are not sure yet what to put as the auxiliary label for
+      //              arcs entering the super final state. The only real choices
+      //              are kEpsilon or kFinalSymbol. We are using kEpsilon for
+      //              now.
+      state_aux_labels.push_back(0);    // kEpsilon
     }
   }
 
--- a/k2/csrc/fsa_utils.cu
+++ b/k2/csrc/fsa_utils.cu
@@ -144,7 +144,7 @@ static Fsa AcceptorFromStream(std::string first_line, std::istringstream &is,
       break;                         // finish reading
     } else {
       K2_LOG(FATAL) << "Invalid line: " << line
-                    << "\nIt expects a line with 4 fields";
+                    << "\nIt expects a line with 1, 2 or 4 fields";
     }
   } while (std::getline(is, line));
 
@@ -156,13 +156,16 @@ static Fsa AcceptorFromStream(std::string first_line, std::istringstream &is,
     K2_CHECK_EQ(openfst, true);
     K2_CHECK_EQ(original_final_states.size(), original_final_weights.size());
     int32_t super_final_state = max_state + 1;
-    for (int32_t i = 0; i != original_final_states.size(); ++i) {
+    for (auto i = 0; i != original_final_states.size(); ++i) {
       arcs.emplace_back(original_final_states[i],
                         super_final_state,
                         -1,     // kFinalSymbol
                         scale * original_final_weights[i]);
     }
   }
+
+  // Sort arcs so that source states are in non-decreasing order.
+  std::sort(arcs.begin(), arcs.end());
 
   bool error = true;
   Array1<Arc> array(GetCpuContext(), arcs);
@@ -223,7 +226,7 @@ static Fsa TransducerFromStream(std::string first_line, std::istringstream &is,
       break;  // finish reading
     } else {
       K2_LOG(FATAL) << "Invalid line: " << line
-                    << "\nIt expects a line with 5 fields";
+                    << "\nIt expects a line with 1, 2 or 5 fields";
     }
   } while (std::getline(is, line));
 
@@ -235,7 +238,7 @@ static Fsa TransducerFromStream(std::string first_line, std::istringstream &is,
     K2_CHECK_EQ(openfst, true);
     K2_CHECK_EQ(original_final_states.size(), original_final_weights.size());
     int32_t super_final_state = max_state + 1;
-    for (int32_t i = 0; i != original_final_states.size(); ++i) {
+    for (auto i = 0; i != original_final_states.size(); ++i) {
       arcs.emplace_back(original_final_states[i],
                         super_final_state,
                         -1,             // kFinalSymbol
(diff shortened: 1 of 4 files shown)


k2/csrc/fsa_utils.cu
```diff
@@ -120,7 +125,21 @@ static Fsa AcceptorFromStream(std::string first_line, std::istringstream &is,
...
+                        << "\nFinal state with weight detected in K2 format";
+      original_final_states.push_back(StringToInt(splits[0]));
+      original_final_weights.push_back(StringToFloat(splits[1]));
+      max_state = std::max(max_state, original_final_states.back());
     } else if (num_fields == 1u) {
+      //   0
+      // final_state
       (void)StringToInt(splits[0]);  // this is a final state
       break;                         // finish reading
     } else {
```
csukuangfj, line 145: The following output info also needs to be modified:  4 fields or 2 fields.

danpovey (commented):
Looks good to me from quick check!  One of you other guys can merge if you think it's OK.

k2/csrc/fsa_utils_test.cu
```diff
@@ -143,6 +178,53 @@ TEST(FsaFromString, Transducer) {
...
+  // src_state dst_state label aux_label cost
+  std::string s = R"(0 1 2 22  -1.2
+    0 2  10 100 -2.2
+    1 3  3  33  -3.2
+    1 6  4  16  -4.2
+    2 6  5  26  -5.2
+    2 4  2  22  -6.2
+    3 6  7  36  -7.2
+    5 7  1  50  -8.2
+    6 -1.2
```
csukuangfj, line 191: If I remember correctly, OpenFst does not require that the state number
of a final state has to be larger than that of a non-final state.

I'm afraid the current implementation cannot be used to read an FST in OpenFst format
as it will abort if the first column is not in non-decreasing order.

k2/csrc/fsa_utils.cu
```diff
@@ -168,7 +204,21 @@ static Fsa TransducerFromStream(std::string first_line, std::istringstream &is,
...
+      //   0            1
+      // final_state  score
+      // In this case, openfst is true. There could be multiple final states, so
+      // we first have to collect all the final states, and then work out the
+      // super final state.
+      K2_CHECK(openfst) << "Invalid line: " << line
+                        << "\nFinal state with weight detected in K2 format";
+      original_final_states.push_back(StringToInt(splits[0]));
+      original_final_weights.push_back(StringToFloat(splits[1]));
+      max_state = std::max(max_state, original_final_states.back());
```
csukuangfj, line 218: With `std::max`, I think you assume that the first column is not in non-decreasing order,
otherwise, you would use `original_final_states.back()` as the max state.

k2/csrc/fsa_utils.cu
```diff
@@ -177,7 +227,22 @@ static Fsa TransducerFromStream(std::string first_line, std::istringstream &is,
...
-  K2_CHECK(is) << "Failed to read fsa from string";
+  // Post processing on final states. When openfst is true, we may have multiple
+  // final states with weights associated with them. We will have to add a super
+  // final state, and convert that into the K2 format (final state with no
+  // weight).
+  if (original_final_states.size() > 0) {
+    K2_CHECK_EQ(openfst, true);
+    K2_CHECK_EQ(original_final_states.size(), original_final_weights.size());
+    int32_t super_final_state = max_state + 1;
+    for (int32_t i = 0; i != original_final_states.size(); ++i) {
```
csukuangfj, line 238: I would recommend to use either `std::size_t i = 0` or  `static_cast` of `size()`
to avoid compiler warnings about comparison between signed and unsigned integers.

----

An alternative is

```cpp

for (auto i : original_final_states) {
   // ....
}

File: k2/csrc/fsa_utils.cu [hunk 5]
```diff
@@ -120,7 +125,21 @@ static Fsa AcceptorFromStream(std::string first_line, std::istringstream &is,
       int32_t symbol = StringToInt(splits[2]);
       float score = scale * StringToFloat(splits[3]);
       arcs.emplace_back(src_state, dest_state, symbol, score);
+      max_state = std::max(max_state, std::max(src_state, dest_state));
```
csukuangfj, line 128: If `src_state` > `dest_state`, the program will crash.

k2/csrc/fsa_utils_test.cu
danpovey: we'd add a new super-final state regardless, i.e. the final-state becomes an arc like
6 8 -1   0  -1.2

csukuangfj: yes, I saw that. The problem is that if you switch
```
    6 -1.2
    7 -2.3
```
to
```
    7 -2.3
    6 -1.2
```

I believe the program will crash.  Since @chenguoguo is using `std::max` to compute the max state, he
is not assuming that the first column is in non-decreasing order.

csukuangfj: I think the code is adding
```cpp
6 8 -1 -1 -1.2
```

instead of

```cpp
6 8 -1 0 -1.2
```

danpovey:
Mm.  In the transducer case, it's actually not 100% clear to me what is the
right thing to do here.  The only real choices
are 0 and -1.  For now let's just make it 0.  We may decide to change this
later or make it overridable.

chenguoguo:
@csukuangfj I took a look at FsaFromArray1() and now I see what you meant by `crash`. I guess that's why you asked in the other thread if we could assume the source states will be in non-descending order. That's a very big assumption and ask though, if we want to work with FSTs created by other tools (and I think it's a fair assumption that people will want to load FSTs from other tools). The issue was `row_ids1`. I think we can sort arcs before creating the FSA. I'll add some code for that real quick.

Regarding the auxiliary label for arcs that go into the super final state, I thought about -1 and 0 and decided to go with -1 so that algorithms will be able to know those artificial arcs when necessary. But on another though, they will know them anyways since we already have -1 on the FSA. I'll change -1 to 0.

csukuangfj:
I have an implementation <https://github.com/k2-fsa/k2/pull/189/commits/a1eaae77fb67d3f0a0d150beba67b684a20b586c>
that does not require a sorted src_states. It may be used as a reference.

chenguoguo:
I added the sorting code. Fairly straightforward. It doesn't require a state to arc mapping so might be a little bit easier.

There might be other corner cases that we don't support yet for OpenFst style FSTs. I think they can be addressed when I work on https://github.com/k2-fsa/k2/issues/180

k2/csrc/fsa_utils.cu
chenguoguo: Fixed.

chenguoguo:
I'm seeing this build error, but the line seems fine to me. Something wrong with the CI?

image

csukuangfj:
+2

Not sure what is happening about the style check.

danpovey:
Let's not do the sorting of arcs by src_state if not needed.
OpenFST would always print in order except the start-state.
Anyway we require the start state to be state 0.

k2/csrc/fsa_utils.cu
```diff
@@ -120,16 +125,47 @@ static Fsa AcceptorFromStream(std::string first_line, std::istringstream &is,
...
+    for (std::size_t i = 0; i != original_final_states.size(); ++i) {
+      arcs.emplace_back(original_final_states[i],
+                        super_final_state,
+                        -1,     // kFinalSymbol
+                        scale * original_final_weights[i]);
+    }
+  }
+
+  // Sort arcs so that source states are in non-decreasing order.
+  std::sort(arcs.begin(), arcs.end());
```
csukuangfj, line 168: Oh, `std::sort` is not stable. The python test fails.

chenguoguo:
OK, let's address this in a separate PR. I created this issue here for separating the internal K2 FSA reading and OpenFST reading: https://github.com/k2-fsa/k2/issues/207

Let me know if have more comments on the new issue.

k2/csrc/fsa_utils.cu
```diff
@@ -168,16 +207,66 @@ static Fsa TransducerFromStream(std::string first_line, std::istringstream &is,
...
+      //              are kEpsilon or kFinalSymbol. We are using kEpsilon for
+      //              now.
+      state_aux_labels.push_back(0);    // kEpsilon
+    }
+  }
+
+  // Sort arcs so that source states are in non-decreasing order. We have to do
+  // this simultaneously for both arcs and auxiliary labels. The following
+  // implementation makes a pair of (Arc, AuxLabel) for sorting.
+  // TODO(guoguo) Optimize this when necessary.
```
csukuangfj, line 257: @chenguoguo 
I find a method with NO memory copying.

```cpp

auto sort_lambda = [&arcs](int32_t i, int32_t j) {
   return arcs[i] < arcs[j];
};

std::sort(arcs.begin(), arc.end(), sort_lambda);
std::sort(state_aux_labels.begin(), state_aux_labels.end(), sort_lambda);
```
This may be helpful if you are adding support for OpenFst.

chenguoguo:
Ok thanks, that was actually my first try but it would do the sorting
twice, so I ended up with the pair thing.

I think eventually we will remove the sort. Will do some test on the
lexicon fst and see what is the minimal requirement we need for openfst
support. Asking the users to do topological sorting on the input fst might
be enough, but I’ll like to test this on real fats.  I think I will be able
to get this (also the separation of internal reading functions) done by
tomorrow — have to finish some other tasks I have by hand now.

PR #206, opened 2020-10-03
add input/output symbol table.

csukuangfj:


k2/csrc/fsa_utils.cu
```diff
@@ -240,16 +238,15 @@ static Fsa TransducerFromStream(std::string first_line, std::istringstream &is,
     K2_CHECK_EQ(openfst, true);
     K2_CHECK_EQ(original_final_states.size(), original_final_weights.size());
     int32_t super_final_state = max_state + 1;
-    for (std::size_t = 0; i != original_final_states.size(); ++i) {
-      arcs.emplace_back(original_final_states[i],
-                        super_final_state,
-                        -1,             // kFinalSymbol
+    for (std::size_t i = 0; i != original_final_states.size(); ++i) {
```
csukuangfj, line 241: @chenguoguo 
FYI: we use `clang-format==9.0.0` for auto formatting.

k2/python/k2/symbol_table.py
```diff
@@ -0,0 +1,93 @@
...
+@dataclass(frozen=True)
+class SymbolTable(object):
+    _id2sym: Dict[int, str]
+    '''Map an integer to a symbol.
+    '''
+
+    _sym2id: Dict[str, int]
+    '''Map a symbol to an integer.
+    '''
+
```
chenguoguo, line 19: We may want to add a function for validation, and check if it's a valid symbol table. For example:
* If 0 is used for other symbols
* If we have <eps>, and if we don't, add that

Then when we initialize this symbol table, we should run the validation.

chenguoguo:
Otherwise looks good to me.

k2/csrc/fsa_utils.cu
chenguoguo: Thanks. Is that something that can be enforced? E.g., as part of the CI style checking?

danpovey:
I don't want to enforce the style as I think it will get in the way of more
substantive issues.  Also I can't figure out how to get emacs to format
that way.
(How does one apply the clang-format?)

csukuangfj:
I am using vim with <https://github.com/google/vim-codefmt> and it reformats the file automagically on save.

CI will not complain if the source files are not clang-format compliant. But it does complain if the code
does not pass `cpplint` check.

csukuangfj:
> I don't want to enforce the style

It is not enforced. Both @qindazhu and me will reformat the files that we are editing.

k2/python/k2/symbol_table.py
csukuangfj: Thanks, fixed.

csukuangfj:
Output with [colab notebook][1]

Screen Shot 2020-10-03 at 12 14 53


danpovey:
Great work!!  Merge if you think it's ready.

csukuangfj:
merging

PR #209, opened 2020-10-03
Added tests for ragged shape ops and fixed some bugs

qindazhu:


k2/csrc/ragged.cu
```diff
@@ -426,7 +426,7 @@ RaggedShape RaggedShape2(Array1<int32_t> *row_splits, Array1<int32_t> *row_ids,
     axes[0].row_splits = row_splits_array;
   }
   if (row_ids != nullptr) axes[0].row_ids = *row_ids;
-  axes[0].cached_tot_size = cached_tot_size;
+  axes[0].cached_tot_size = axes[0].row_splits.Back();
```
danpovey, line 429: The reason we cache the size is that GPU->CPU transfer is slow.  Therefore we should use every opportunity to avoid it.

danpovey: there is a K2_CHECK_EQ on this above.. can make this a K2_DCHECK_EQ.

danpovey: ... the only time we should do that transfer is if cached_tot_size == -1 and row_ids is not supplied.

danpovey: .. also, I believe we don't *require* cached_tot_size to be set, unless requested.  So I think it would be more consistent with the design to just leave it as -1 if row_ids is not set.

qindazhu: OK, will first use cached_tot_size then row_ids.Dim then row_splits.Back() if cached_tot_size and row_ids are null (-1)

qindazhu: >   just leave it as -1 if row_ids is not se
OK

qindazhu: Updated in RaggedShape2 and RaggedShape3

danpovey:
Thanks!  Feel free to merge..

qindazhu:
Merging.. thanks

PR #215, opened 2020-10-03
check cuda error for every kernel launch

qindazhu:


csukuangfj:
+2

k2/csrc/utils.cu
```diff
@@ -329,9 +330,10 @@ void RowIdsToRowSplits(ContextPtr &c, int32_t num_elems, const int32_t *row_ids,
               tot_threads = num_elems * threads_per_elem;
       int32_t block_size = 256;
       int32_t grid_size = NumBlocks(tot_threads, block_size);
-      K2_CUDA_SAFE_CALL(RowIdsToRowSplitsKernel<<<grid_size, block_size, 0,
-                                                  c->GetCudaStream()>>>(
-          num_elems, threads_per_elem, row_ids, num_rows, row_splits));
+      RowIdsToRowSplitsKernel<<<grid_size, block_size, 0, c->GetCudaStream()>>>(
+          num_elems, threads_per_elem, row_ids, num_rows, row_splits);
+      auto err = cudaGetLastError();
+      K2_DCHECK_CUDA_ERROR(err);
```
danpovey, line 336: Let's put cudaGetLastError() inside K2_DCHECK_CUDA_ERROR.  There's no reason to call it if not.  In fact, calling it will clear the error state (IIRC) so it's not good to call it but not check it.

qindazhu: As `K2_DCHECK_CUDA_ERROR(x)` accepts an error code for now, do you mean we should call `cudaGetLastError` (and check the returned error code) in it after  checking the passed error code? Or should we just create a new macro e.g. `K2_DCHECK_CUDA_LAST_ERROR()` with no parameter.

danpovey:
No I'm saying just write
      K2_DCHECK_CUDA_ERROR(cudaGetLastError());
so it only gets evaluated in debug mode.  Unfortunately the macro
      K2_DCHECK_CUDA_ERROR
would evaluate that twice which is not what we want.  It should be changed
to have a declared variable and do..while.

qindazhu:
@danpovey updated as we discussed offline, please have a review.

Issue #217, opened 2020-10-04
Clone() for Array1 and Array2


danpovey:
Can we please have Clone() functions for Array1 and Array2, that will give a new array on the same device?

BTW there is something I'm not very satisfied about the current design, that some functions are members but others are not, even though they might most naturally be members.  I'm thinking for now we could put them as non-member functions like ToContiguous(), in array_ops.h.

Also my intention was that any functions involving Ragged would go in ragged_ops.h, not array_ops.h.  Just FYI: not urgent or important.

qindazhu:
> Ragged would go in ragged_ops.h, not array_ops.

 Will do this after tested all methods in ragged.h. Actually there are a lot of ops functions now in `ragged.h` instead of `ragged_ops.h`

danpovey:
BTW, this is done for array1 but not array2.

PR #218, opened 2020-10-04
implement ArcSort.

csukuangfj:


csukuangfj:
The non-in-place version will be added after #217  is resolved.

danpovey:
Looks good to me!  One of u guys can merge if you think it's ready.

csukuangfj:
I'll merge it after fixing `FsaFromString`: Sort the src_states only if `openfst == true`.

csukuangfj:
merging.

k2/csrc/ragged_inl.h
```diff
@@ -103,23 +103,35 @@ Ragged<T> RandomRagged(T min_value, T max_value, int32_t min_num_axes,
...
+                                 *context);           // context
+  else
+    mgpu::segmented_sort(src->values.Data(),  // keys
+                         src->values.Dim(),   // count
+                         segment.Data(),      // segments
+                         segment.Dim() - 1,   // num_segments
+                         Op(),                // cmp
+                         *context);           // context
+
   auto err = cudaGetLastError();
```
qindazhu, line 135: wondering if there's still cuda error now?

csukuangfj: It disappears!

qindazhu: ok, then i guess the error before was from some other code (and fixed now) because we don't sync kernel call

PR #223, opened 2020-10-04
Add top-sorting code; a few other changes.

danpovey:
This needs some more work (not close to finished) but putting it up here as I am doing something to-morrow and may not do much work then.
I decided to debug this before Compose(), as this is simpler but has some similar elements to it.

### DIFF
--- a/k2/csrc/array.h
+++ b/k2/csrc/array.h
@@ -54,7 +54,7 @@ class Array1 {
 
   // Called when creating Array2 using Array1, users should not call this for
   // now.
-  RegionPtr &GetRegion() { return region_; }
+  const RegionPtr &GetRegion() const { return region_; }
 
   // generally Callable will be some kind of lambda or function object; it
   // should be possible to evaluate it on the CUDA device (if we're compiling
@@ -364,7 +364,7 @@ class Array2 {
   // Currently ByteOffset and GetRegion is for internal usage, user should never
   // call it for now.
   int32_t ByteOffset() const { return byte_offset_; }
-  RegionPtr &GetRegion() { return region_; }
+  const RegionPtr &GetRegion() const { return region_; }
 
   ContextPtr &Context() const { return region_->context; }
 
--- a/k2/csrc/array.h
+++ b/k2/csrc/array.h
@@ -306,6 +306,7 @@ class Array1 {
     dim_ = size;
     byte_offset_ = 0;
   }
+
 };
 
 // Could possibly introduce a debug mode to this that would do bounds checking.
--- a/k2/csrc/array.h
+++ b/k2/csrc/array.h
@@ -50,6 +50,9 @@ class Array1 {
                                        byte_offset_);
   }
 
+  // Return a copy of this array that does not share the same underlying data.
+  Array1<T> Clone();
+
   int32_t ByteOffset() const { return byte_offset_; }
 
   // Called when creating Array2 using Array1, users should not call this for
@@ -156,15 +159,12 @@ class Array1 {
     if (ctx->IsCompatible(*Context())) return *this;
 
     Array1 ans(ctx, Dim());
-    if (dim_ == 0) return ans;
-
-    auto kind = GetMemoryCopyKind(*Context(), *ctx);
-    T *dst = ans.Data();
-    const T *src = Data();
-    MemoryCopy(static_cast<void *>(dst), static_cast<const void *>(src),
-               Dim() * ElementSize(), kind);
+    ans.CopyFrom(*this);
     return ans;
   }
+
+  // Copy from another array of the same dimension and type.
+  void CopyFrom(const Array1<T> &src);
 
   /*
     Modify size of array, copying old contents if we could not re-use the same
@@ -224,8 +224,11 @@ class Array1 {
     } else {
       K2_CHECK_EQ(type, kCuda);
       T ans;
-      MemoryCopy(static_cast<void *>(&ans), static_cast<const void *>(data),
-                 ElementSize(), MemcpyDeviceToHost);
+      cudaError_t ret = cudaMemcpy(static_cast<void *>(&ans),
+                                   static_cast<const void *>(data),
+                                   ElementSize(),
+                                   cudaMemcpyDeviceToHost);
+      K2_CHECK_CUDA_ERROR(ret);
       return ans;
     }
   }
@@ -272,7 +275,7 @@ class Array1 {
     T *data = Data();
     auto kind = GetMemoryCopyKind(*GetCpuContext(), *Context());
     MemoryCopy(static_cast<void *>(data), static_cast<const void *>(src.data()),
-               src.size() * ElementSize(), kind);
+               src.size() * ElementSize(), kind, Context().get());
   }
 
   Array1(const Array1 &other) = default;
@@ -472,7 +475,7 @@ class Array2 {
       T *dst = ans.Data();
       const T *src = Data();
       MemoryCopy(static_cast<void *>(dst), static_cast<const void *>(src),
-                 dim0_ * dim1_ * ElementSize(), kind);
+                 dim0_ * dim1_ * ElementSize(), kind, ctx.get());
       return ans;
     } else {
       return ToContiguous(*this).To(ctx);
@@ -612,5 +615,9 @@ std::ostream &operator<<(std::ostream &stream, const Array2<T> &array) {
 }
 
 }  // namespace k2
+
+#define IS_IN_K2_CSRC_ARRAY_H_
+#include "k2/csrc/array_inl.h"
+#undef IS_IN_K2_CSRC_ARRAY_H_
 
 #endif  // K2_CSRC_ARRAY_H_
--- a/k2/csrc/array_inl.h
+++ b/k2/csrc/array_inl.h
@@ -1,0 +1,57 @@
+/**
+ * @brief
+ * array_inl
+ *
+ * @note
+ * Don't include this file directly; it is included by array.h.
+ * It contains implementation code.
+ *
+ * @copyright
+ * Copyright (c)  2020  Xiaomi Corporation (authors: Daniel Povey, Haowen Qiu)
+ *                      Mobvoi Inc.        (authors: Fangjun Kuang)
+ * @copyright
+ * See LICENSE for clarification regarding multiple authors
+ */
+
+#ifndef K2_CSRC_ARRAY_INL_H_
+#define K2_CSRC_ARRAY_INL_H_
+
+#ifndef IS_IN_K2_CSRC_ARRAY_H_
+#error "this file is supposed to be included only by array_ops.h"
+#endif
+
+#include <algorithm>
+#include <cassert>
+#include <cub/cub.cuh>  // NOLINT
+#include <random>
+#include <type_traits>
+#include <utility>
+#include <vector>
+#include "k2/csrc/utils.h"
+
+namespace k2 {
+
+template <typename T>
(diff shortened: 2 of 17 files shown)


k2/csrc/top_sort.cu
```diff
@@ -0,0 +1,1015 @@
...
+
+       @param [in] fsas    A vector of FSAs; must have 3 axes.
+   */
+  TopSorter(FsaVec &fsas)
+      : fsas_(fsas) {
+    c_ = GetContext(fsas_);
+
+    K2_CHECK_EQ(fsas_.NumAxes(), 3);
+    int32_t num_fsas = fsas_.shape.TotSize(0),
+        num_states = fsas_.shape.TotSize(0);
```
csukuangfj, line 40: a typo here, `TotSize(1);`

danpovey: thanks for noticing!

danpovey:
I have added changes here to rework FSA<-->tensor I/O.  Not tested.
Actually it is giving a compilation error that I can't figure out.  Would appreciate if someone would test.
It's as if I am trying to access an Array1 inside a lambda but I'm not doing that.  And even replacing that particular lambda invocation with an empty lambda doesn't help.  It's odd.
```
^[[35m^[[1mScanning dependencies of target context^[[0m
[ 18%] ^[[32mBuilding CUDA object k2/csrc/CMakeFiles/context.dir/fsa.cu.o^[[0m
/ceph-dan/k2/k2/csrc/context.h(365): error: calling a __host__ function("k2::Array1<int> ::operator [] const") from a __global__ function("k2::eval_lambda< ::> ") is not allow\
ed

/ceph-dan/k2/k2/csrc/context.h(365): error: identifier "k2::Array1<int> ::operator [] const" is undefined in device code

/ceph-dan/k2/k2/csrc/fsa.cu(260): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::__shared_count") from a __host__ __device__ functi\
on("") is not allowed

/ceph-dan/k2/k2/csrc/fsa.cu(260): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::~__shared_count") from a __host__ __device__ funct\
ion("") is not allowed

2 errors detected in the compilation of "/tmp/tmpxft_0003c541_00000000-6_fsa.cpp1.ii".
k2/csrc/CMakeFiles/context.dir/build.make:133: recipe for target 'k2/csrc/CMakeFiles/context.dir/fsa.cu.o' failed
make[2]: *** [k2/csrc/CMakeFiles/context.dir/fsa.cu.o] Error 1
make[2]: Target 'k2/csrc/CMakeFiles/context.dir/build' not remade because of errors.
CMakeFiles/Makefile2:587: recipe for target 'k2/csrc/CMakeFiles/context.dir/all' failed
make[1]: *** [k2/csrc/CMakeFiles/context.dir/all] Error 2

```

danpovey:
Forget compilation problem, I found it... just the compiler was not giving the right line number for it.

danpovey:
Guys, is it expected to get some test failures for some of the python tests?  
```
Total Test time (real) =  65.02 sec

The following tests FAILED:
     25 - arc_test_py (Failed)
     26 - array_test_py (Failed)
     27 - fsa_test_py (Failed)
Errors while running CTest


de-74279-k2-dev-0929153814-fcbbc64c5-h7vkf:build: python3  ../k2/python/tests/fsa_test.py
EEE
======================================================================
ERROR: test_acceptor_from_str (__main__.TestFsa)
----------------------------------------------------------------------
  File "../k2/python/tests/fsa_test.py", line 37, in test_acceptor_from_str
    fsa = k2.Fsa(_remove_leading_spaces(s))
AttributeError: module 'k2' has no attribute 'Fsa'

======================================================================
ERROR: test_symbol_table_and_dot (__main__.TestFsa)
----------------------------------------------------------------------
  File "../k2/python/tests/fsa_test.py", line 154, in test_symbol_table_and_dot
    isym = k2.SymbolTable.from_str(isym_str)
AttributeError: module 'k2' has no attribute 'SymbolTable'

======================================================================
ERROR: test_transducer_from_str (__main__.TestFsa)
----------------------------------------------------------------------
  File "../k2/python/tests/fsa_test.py", line 105, in test_transducer_from_str
    fsa = k2.Fsa(_remove_leading_spaces(s))
AttributeError: module 'k2' has no attribute 'Fsa'

----------------------------------------------------------------------
Ran 3 tests in 0.000s

FAILED (errors=3)
```

danpovey:
Guys, I have got this top-sorting code to compile but have not tested it.
(It's complex enough that it's likely to have a few bugs; I can help debug once there is some test code, if needed).

danpovey:
LMK if you think I should merge.

csukuangfj:
> de-74279-k2-dev-0929153814-fcbbc64c5-h7vkf:build: python3  ../k2/python/tests/fsa_test.py

@danpovey 
To run Python tests like this, we have to add
```bash
export PYTHONPATH=$PWD/lib:$PWD/../k2/python:${PYTHONPATH}
```

Normally, we are running python tests using ONE of the following:

```bash
cd build
ctest
make test
ctest -E arc_test_py  # to run only tests for arc_test
```

and CMake will set up PYTHONPATH for us.

csukuangfj:
> Guys, is it expected to get some test failures for some of the python tests?


I just run all the test cases against your branch `dan_temp` and there are NO failures:
> 100% tests passed, 0 tests failed out of 39

k2/csrc/array.h
```diff
@@ -224,8 +224,11 @@ class Array1 {
     } else {
       K2_CHECK_EQ(type, kCuda);
       T ans;
-      MemoryCopy(static_cast<void *>(&ans), static_cast<const void *>(data),
-                 ElementSize(), MemcpyDeviceToHost);
+      cudaError_t ret = cudaMemcpy(static_cast<void *>(&ans),
```
qindazhu, line 227: using `MemoryCopy` instead of `cudaMemcpy`? I think it's better if we can use consistent API for memory copy. (And if we want to  use cudaMemcpy directly, then we should delete MemoryCopy)

k2/csrc/array_inl.h
```diff
@@ -0,0 +1,57 @@
...
+Array1<T> Array1<T>::Clone() {
+  Array1<T> ans(Context(), Dim());
+  ans.CopyFrom(*this);
+  return ans;
+}
+
+template <typename T>
+void Array1<T>::CopyFrom(const Array1<T> &src) {
+  K2_CHECK_EQ(dim_, src.dim_);
+  if (dim_ == 0)
```
qindazhu, line 44: if `dim_ == 0`, is array `*this` valid or not? i.e. Can we access its context or not? And in the case `src.dim == 0` but we can access its context, then does copy semantics imply that we'll copy a valid empty array here?

qindazhu: As we can construct an empty non-valid array with the default constructor, then do copy form an empty valid array.

k2/csrc/array_ops_inl.h
```diff
@@ -477,6 +477,27 @@ Array2<T> ToContiguous(const Array2<T> &src) {
   return ans;
 }
 
+template <typename T>
+bool Equal(const Array1<T> &a, const Array1<T> &b) {
+  if (a.Dim() != b.Dim()) {
+    K2_LOG(FATAL) << "Dimension mismatch: " << a.Dim() << " vs. "
```
qindazhu, line 483: just do `K2_CHECK_EQ(a.Dim(), b.Dim()) << ...`

k2/csrc/array_ops_inl.h
```diff
@@ -477,6 +477,27 @@ Array2<T> ToContiguous(const Array2<T> &src) {
...
+                  << b.Dim();
+  }
+  ContextPtr c = GetContext(a, b);
+  const T *a_data = a.Data(), *b_data = b.Data();
+  if (c->GetDeviceType() == kCpu) {
+    return memcmp((void*)a_data, (void*)b_data, sizeof(T) * a.Dim()) == 0;
```
qindazhu, line 489: seems there is a const_cast here? from `const T*` to `void*`?

k2/csrc/ragged.h
```diff
@@ -594,6 +608,11 @@ RaggedShape TrivialShape(ContextPtr &c, int32_t num_elems);
 RaggedShape RaggedShapeFromTotSizes(ContextPtr &c, int32_t num_axes,
                                     int32_t *tot_sizes);
 
+inline RaggedShape RaggedShapeFromTotSizes(ContextPtr &c, std::vector<int32_t> &tot_sizes) {
+  return RaggedShapeFromTotSizes(c, tot_sizes.size(), &(tot_sizes[0]));
```
qindazhu, line 612: use `tot_sizes.data()` instead of get the address of the first element?

qindazhu, line 612: seems there is a static_cast from `size_t` to `int32_t` for `tot_sizes.size()`, not sure if there's compilation warning.

qindazhu:
I'm fine to merge except a few small issues. We can add tests later.

k2/csrc/array.h
danpovey: I wasn't crazy about the idea of wrapping the cudaMemcpyKind in another enum.  I don't like doing these kinds of things before they are actually necessary (i.e. having one-to-one maps like this).
BTW I slightly modified MemoryCopy so it uses asynchronous mode for device-to-device transfers, for speed,
and uses the appropriate stream in that case.

danpovey:
Test works now.. turns out I hadn't installed graphviz.  I found the error by doing ctest -V

csukuangfj:
This file lists all the dependencies except `PyTorch` needed by k2.

https://github.com/k2-fsa/k2/blob/8f24f75e6a8ee7d41981b72fb5bf911be3ab4b0f/requirements.txt#L1-L4

k2/csrc/array_ops_inl.h
csukuangfj, line 489: Actually, there is no need for casting.

```cpp
return memcmp(a_data, b_data, sizeof(T) * a.Dim()) == 0;
```
is enough.

k2/csrc/top_sort.cu
```diff
@@ -0,0 +1,533 @@
...
+class TopSorter {
+ public:
+  /**
+     Topological sorter object.  You should call TopSort() after
+     constructing it.
+
+       @param [in] fsas    A vector of FSAs; must have 3 axes.
+   */
+  TopSorter(FsaVec &fsas)
+    : c_(fsas_.Context()), fsas_(fsas) {
```
chenguoguo, line 36: You probably meant fsas.Context() instead of fsas_.Context() since fsas_ is not initialized yet at this point?

k2/csrc/array_ops_inl.h
```diff
@@ -477,6 +477,27 @@ Array2<T> ToContiguous(const Array2<T> &src) {
...
+  } else {
+    Array1<int32_t> is_same(c, 1, 1);
```
csukuangfj, line 491: Should we add an additional check here?

```cpp
K2_CHECK_EQ(c->GetDeviceType(), kCuda);
```

danpovey: that may be checked inside eval(), I think I am not going to put these checks in places like this.

k2/csrc/top_sort.cu
danpovey: yes thx

PR #224, opened 2020-10-05
Refactored FSA reading code

chenguoguo:
This is WIP because I still have to test some real FSTs, e.g., lexicon.fst from Kaldi.

- [x] Separated reading code for K2 and OpenFST
- [x] Removed sorting but does not require source states in non-descending order for OpenFST style FSTs
- [x] Support omitted weights in OpenFST (treat it as 0.0f)
- [x] Test this for a real lexicon fst

@csukuangfj I added one argument `acceptor` (default true) for the Python fsa initialization OpenFST style WFSTs can have 1, 2, 4, 5 fields, which makes it difficult to figure acceptor/transducer automatically. Also, asking users to explicitly set acceptor/transducer may lead to less coding errors. But let me know your thoughts.

### DIFF
--- a/k2/csrc/fsa_utils.cu
+++ b/k2/csrc/fsa_utils.cu
@@ -99,7 +99,7 @@ static void SplitStringToVector(const std::string &in, const char *delim,
   }
 }
 
-/* Create an acceptor from a stream, assuming the acceptor is in the K2 format:
+/* Create an acceptor from a stream, assuming the acceptor is in the k2 format:
 
    src_state1 dest_state1 label1 score1
    src_state2 dest_state2 label2 score2
@@ -143,8 +143,8 @@ static Fsa K2AcceptorFromStream(std::istringstream &is) {
       finished = true;               // set finish
     } else {
       K2_LOG(FATAL) << "Invalid line: " << line
-                    << "\nK2 acceptor expects a line with 1 (final_state) or "
-                    << "4 (src_state dest_state label score) fields";
+                    << "\nk2 acceptor expects a line with 1 (final_state) or "
+                       "4 (src_state dest_state label score) fields";
     }
   }
 
@@ -209,8 +209,8 @@ static Fsa K2TransducerFromStream(std::istringstream &is,
       finished = true;               // set finish
     } else {
       K2_LOG(FATAL) << "Invalid line: " << line
-                    << "\nK2 transducer expects a line with 1 (final_state) or "
-                    << "5 (src_state dest_state label aux_label score) fields";
+                    << "\nk2 transducer expects a line with 1 (final_state) or "
+                       "5 (src_state dest_state label aux_label score) fields";
     }
   }
 
@@ -246,7 +246,7 @@ static Fsa K2TransducerFromStream(std::istringstream &is,
 
    @return It returns an Fsa on CPU.
 */
-static Fsa OpenFSTAcceptorFromStream(std::istringstream &is) {
+static Fsa OpenFstAcceptorFromStream(std::istringstream &is) {
   std::vector<Arc> arcs;
   std::vector<std::vector<Arc>> state_to_arcs;            // indexed by states
   std::vector<std::string> splits;
@@ -298,8 +298,8 @@ static Fsa OpenFSTAcceptorFromStream(std::istringstream &is) {
     } else {
       K2_LOG(FATAL) << "Invalid line: " << line
                     << "\nOpenFST acceptor expects a line with 1 (final_state),"
-                    << " 2 (final_state score), 3 (src_state dest_state label) "
-                    << "or 4 (src_state dest_state label score) fields.";
+                       " 2 (final_state score), 3 (src_state dest_state label) "
+                       "or 4 (src_state dest_state label score) fields.";
     }
   }
 
@@ -308,7 +308,7 @@ static Fsa OpenFSTAcceptorFromStream(std::istringstream &is) {
 
   // Post processing on final states. We may have multiple final states with
   // weights associated with them for OpenFST style FSTs. We will have to add a
-  // super final state, and convert that into the K2 format (final state with no
+  // super final state, and convert that into the k2 format (final state with no
   // weight).
   if (original_final_states.size() > 0) {
     K2_CHECK_EQ(original_final_states.size(), original_final_weights.size());
@@ -363,7 +363,7 @@ static Fsa OpenFSTAcceptorFromStream(std::istringstream &is) {
 
    @return It returns an Fsa on CPU.
 */
-static Fsa OpenFSTTransducerFromStream(std::istringstream &is,
+static Fsa OpenFstTransducerFromStream(std::istringstream &is,
                                        Array1<int32_t> *aux_labels) {
   K2_CHECK(aux_labels != nullptr);
 
@@ -431,9 +431,9 @@ static Fsa OpenFSTTransducerFromStream(std::istringstream &is,
     } else {
       K2_LOG(FATAL) << "Invalid line: " << line
                     << "\nOpenFST transducer expects a line with "
-                    << "1 (final_state), 2 (final_state score), "
-                    << "4 (src_state dest_state label aux_label) or "
-                    << "5 (src_state dest_state label aux_label score) fields.";
+                       "1 (final_state), 2 (final_state score), "
+                       "4 (src_state dest_state label aux_label) or "
+                       "5 (src_state dest_state label aux_label score) fields.";
     }
   }
 
@@ -442,7 +442,7 @@ static Fsa OpenFSTTransducerFromStream(std::istringstream &is,
 
   // Post processing on final states. We may have multiple final states with
   // weights associated with them for OpenFST style FSTs. We will have to add a
-  // super final state, and convert that into the K2 format (final state with no
+  // super final state, and convert that into the k2 format (final state with no
   // weight).
   if (original_final_states.size() > 0) {
     K2_CHECK_EQ(original_final_states.size(), original_final_weights.size());
@@ -501,9 +501,9 @@ Fsa FsaFromString(const std::string &s, bool openfst /*= false*/,
   else if (openfst == false && aux_labels != nullptr)
     return K2TransducerFromStream(is, aux_labels);
   else if (openfst == true && aux_labels == nullptr)
-    return OpenFSTAcceptorFromStream(is);
+    return OpenFstAcceptorFromStream(is);
   else if (openfst == true && aux_labels != nullptr)
-    return OpenFSTTransducerFromStream(is, aux_labels);
+    return OpenFstTransducerFromStream(is, aux_labels);
 
   return Fsa();  // unreachable code
 }
--- a/k2/csrc/fsa_utils.cu
+++ b/k2/csrc/fsa_utils.cu
@@ -210,7 +210,7 @@ static Fsa K2TransducerFromStream(std::istringstream &is,
     } else {
       K2_LOG(FATAL) << "Invalid line: " << line
                     << "\nK2 transducer expects a line with 1 (final_state) or "
-                    << "4 (src_state dest_state label aux_label score) fields";
+                    << "5 (src_state dest_state label aux_label score) fields";
     }
   }
 
--- a/k2/csrc/fsa_utils.cu
+++ b/k2/csrc/fsa_utils.cu
@@ -252,6 +252,7 @@ static Fsa OpenFSTAcceptorFromStream(std::istringstream &is) {
   std::vector<std::string> splits;
   std::string line;
 
+  bool has_final_state = false;
   int32_t max_state = -1;
   int32_t num_arcs = 0;
   std::vector<int32_t> original_final_states;
@@ -289,10 +290,11 @@ static Fsa OpenFSTAcceptorFromStream(std::istringstream &is) {
       // final_state  score
       float score = 0.0f;
       if (num_fields == 2u)
-        score = StringToFloat(splits[1]);
+        score = -1.0f * StringToFloat(splits[1]);
       original_final_states.push_back(StringToInt(splits[0]));
       original_final_weights.push_back(score);
       max_state = std::max(max_state, original_final_states.back());
+      has_final_state = true;
     } else {
       K2_LOG(FATAL) << "Invalid line: " << line
                     << "\nOpenFST acceptor expects a line with 1 (final_state),"
@@ -300,6 +302,9 @@ static Fsa OpenFSTAcceptorFromStream(std::istringstream &is) {
                     << "or 4 (src_state dest_state label score) fields.";
     }
   }
+
+  K2_CHECK(is.eof());
+  K2_CHECK_EQ(has_final_state, true) << "No final state detected.";
 
   // Post processing on final states. We may have multiple final states with
   // weights associated with them for OpenFST style FSTs. We will have to add a
@@ -313,13 +318,13 @@ static Fsa OpenFSTAcceptorFromStream(std::istringstream &is) {
       state_to_arcs[original_final_states[i]].emplace_back(
           original_final_states[i], super_final_state,
           -1,  // kFinalSymbol
-          -1.0f * original_final_weights[i]);
+          original_final_weights[i]);
       ++num_arcs;
     }
   }
 
   // Move arcs from "state_to_arcs" to "arcs".
-  std::size_t arc_index = 0;
+  int32_t arc_index = 0;
   arcs.resize(num_arcs);
   for (std::size_t s = 0; s < state_to_arcs.size(); ++s) {
     for (std::size_t a = 0; a < state_to_arcs[s].size(); ++a) {
@@ -328,8 +333,7 @@ static Fsa OpenFSTAcceptorFromStream(std::istringstream &is) {
       ++arc_index;
     }
   }
-  K2_CHECK_EQ(num_arcs, static_cast<int32_t>(arc_index));
-  state_to_arcs.resize(0);
+  K2_CHECK_EQ(num_arcs, arc_index);
 
   bool error = true;
   Array1<Arc> array(GetCpuContext(), arcs);
@@ -370,6 +374,7 @@ static Fsa OpenFSTTransducerFromStream(std::istringstream &is,
   std::vector<std::string> splits;
   std::string line;
 
+  bool has_final_state = false;
   int32_t max_state = -1;
   int32_t num_arcs = 0;
   std::vector<int32_t> original_final_states;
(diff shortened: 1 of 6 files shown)


k2/csrc/fsa_utils.cu
```diff
@@ -99,75 +99,237 @@ static void SplitStringToVector(const std::string &in, const char *delim,
   }
 }
 
-// Create an acceptor from a stream.
-static Fsa AcceptorFromStream(std::string first_line, std::istringstream &is,
-                              bool openfst) {
+/* Create an acceptor from a stream, assuming the acceptor is in the K2 format:
```
csukuangfj, line 102: In the original design without taking openfst into account, Dan wanted
a version accepting a parameter `negate_scores`. I am not sure
whether the parameter is still needed here.

k2/csrc/fsa_utils.cu
```diff
@@ -99,75 +99,237 @@ static void SplitStringToVector(const std::string &in, const char *delim,
...
+      aux_labels_internal.push_back(aux_label);
+    } else if (num_fields == 1u) {
+      //   0
+      // final_state
+      (void)StringToInt(splits[0]);
+      finished = true;               // set finish
+    } else {
+      K2_LOG(FATAL) << "Invalid line: " << line
+                    << "\nK2 transducer expects a line with 1 (final_state) or "
+                    << "4 (src_state dest_state label aux_label score) fields";
```
csukuangfj, line 213: The output info is incorrect. 5 vs 4.

k2/csrc/fsa_utils.cu
chenguoguo: @danpovey do we want to have the option for negating score when openfst == false? Or do we only want that when openfst == true?

k2/csrc/fsa_utils.cu
```diff
@@ -99,75 +99,237 @@ static void SplitStringToVector(const std::string &in, const char *delim,
...
+   src_state2 dest_state2 label2 score2
+   ... ...
+   final_state final_score
+
+   We will negate the cost/score when we read them in. Also note, OpenFST may
+   omit the cost/score if it is 0.0.
+
+   If there are multiple final states, we will create a super final state, and
+   create arcs from the old final states to the new super final state, with the
+   (negated) old final state cost/score as its cost/score, and -1 as its label.
```
csukuangfj, line 243: 0 as its aux_label.

k2/csrc/fsa_utils.cu
chenguoguo: Good catch, fixed, thanks.

k2/csrc/fsa_utils.cu
chenguoguo: We don't have aux_label for acceptors. I do have the comment for transducer, see here: https://github.com/k2-fsa/k2/blob/04872431289b5b3e9c154ef389d094d08a1a6e9c/k2/csrc/fsa_utils.cu#L356

k2/csrc/fsa_utils.cu
```diff
@@ -99,75 +99,237 @@ static void SplitStringToVector(const std::string &in, const char *delim,
...
       //   0            1
       // final_state  score
-      // In this case, openfst is true. There could be multiple final states, so
-      // we first have to collect all the final states, and then work out the
-      // super final state.
-      K2_CHECK(openfst) << "Invalid line: " << line
-                        << "\nFinal state with weight detected in K2 format";
+      float score = 0.0f;
+      if (num_fields == 2u)
+        score = StringToFloat(splits[1]);
```
csukuangfj, line 292: the scale -1 is missing.

k2/csrc/fsa_utils.cu
```diff
@@ -99,75 +99,237 @@ static void SplitStringToVector(const std::string &in, const char *delim,
...
+  std::vector<std::vector<Arc>> state_to_arcs;            // indexed by states
+  std::vector<std::string> splits;
+  std::string line;
 
   int32_t max_state = -1;
+  int32_t num_arcs = 0;
   std::vector<int32_t> original_final_states;
   std::vector<float> original_final_weights;
-  do {
+  while (std::getline(is, line)) {
```
csukuangfj, line 259: we should either add a `bool finish` flag or check outside the loop
that the stream `is` is still in a good state.

k2/csrc/fsa_utils.cu
chenguoguo: The variable is called `original_final_weights` so I didn't negate it here. I did it later: https://github.com/k2-fsa/k2/blob/04872431289b5b3e9c154ef389d094d08a1a6e9c/k2/csrc/fsa_utils.cu#L316

But I can move it up if that's easier to read.

k2/csrc/fsa_utils.cu
```diff
@@ -99,75 +99,237 @@ static void SplitStringToVector(const std::string &in, const char *delim,
...
+          -1.0f * original_final_weights[i]);
+      ++num_arcs;
     }
   }
-  if (openfst) {
-    // Sort arcs so that source states are in non-decreasing order.
-    std::sort(arcs.begin(), arcs.end());
+
+  // Move arcs from "state_to_arcs" to "arcs".
+  std::size_t arc_index = 0;
```
csukuangfj, line 322: `int32_t arc_index;` since `num_arcs` is of type `int32_t`

k2/csrc/fsa_utils.cu
```diff
@@ -99,75 +99,237 @@ static void SplitStringToVector(const std::string &in, const char *delim,
...
+  arcs.resize(num_arcs);
+  for (std::size_t s = 0; s < state_to_arcs.size(); ++s) {
+    for (std::size_t a = 0; a < state_to_arcs[s].size(); ++a) {
+      K2_CHECK_GT(num_arcs, arc_index);
+      arcs[arc_index] = state_to_arcs[s][a];
+      ++arc_index;
+    }
   }
+  K2_CHECK_EQ(num_arcs, static_cast<int32_t>(arc_index));
+  state_to_arcs.resize(0);
```
csukuangfj, line 332: `resize(0)` will not free the underlying memory.
This statement can be deleted.

k2/csrc/fsa_utils.cu
csukuangfj: Please add a comment here or move it up. That's easier to read.

k2/csrc/fsa_utils.cu
chenguoguo: For the openfst input, we will read till the end of the file since the final state may appear in the middle of the file. I added check for has_final_state as well as is.eof()

csukuangfj:
> I added one argument acceptor (default true) for the Python fsa initialization OpenFST style WFSTs can have 1, 2, 4, 5 fields, which makes it difficult to figure acceptor/transducer automatically

Can we define another python interface for reading OpenFST format? It will make the interface and user's life simpler.

chenguoguo:
Are you suggesting adding a new Python class, e.g., FsaOpenFst?

If yes, I'm inclined to not adding such a new class. The difference is really the reading, and once they are loaded, they will be equivalent to K2 Fsas. Having two classes Fsa and FsaOpenFst may make people feel they are different type of finite state acceptors.

But if you guys think it's a good idea to have a separate Python class for OpenFST, I can add that. Should be fairly straightforward.

csukuangfj:
```python
def read_openfst(...) -> Fsa:
   ...
```

chenguoguo:
I guess it makes sense to add that function. Will do that.

csukuangfj:
I would propose to leave the current interface as it is. To add support for a new format, we just add another interface.
A single interface to handle multiple formats makes things complicated.

chenguoguo:
So you'd like to keep the code to automatically figure out K2 FSAs/Transducers right?

csukuangfj:
> So you'd like to keep the code to automatically figure out K2 FSAs/Transducers right?

For the K2 part, I would say yes. Every field in the K2 format has to be present, so it is possible to figure out
automatically whether it is an acceptor or a transducer. I think the k2 format will be more popular than
OpenFST in K2, so a simpler interface for the k2 format makes life easier.

Not sure if @danpovey and @qindazhu have something to say about this.

csukuangfj:
I think you can leave your current C++ interface unchanged. The auto detection for the k2 format can be
supported in Python.

qindazhu:
> ```python
> def read_openfst(...) -> Fsa:
> ```

+1 ,
For k2 format, I prefer the approach as well which figures out FSA/FST automatically. For openFST format, I also would like to just provide a function to users and let them call the function to convert openFST format to k2 format.

csukuangfj:
If we are going to add I/O support for `FsaVec`, the extra support for OpenFst complicates things further.

chenguoguo:
OK sounds good. I'll make some changes later today when I get a chance.

danpovey:
Agreed: separate function for OpenFST I/O makes sense.

danpovey:
... Incidentally, I did not have in mind reading FsaVec from text.  The
plan for reading Fsa and FsaVec efficiently is to use, say, PyTorch I/O to
read/write tensors of dimension N by 4 or N by 5 (for transducers) and to
use those to construct the FsaVec using functions I've already written (not
tested perhaps)... for writing you'd just store the arcs as a PyTorch
tensor of int32.  So when possible we'd use this fast binary I/O.

chenguoguo:
Pushed more changes. Tested on Lexicon FST generated by Kaldi and worked. Removing "WIP"

k2/csrc/fsa_utils.cu
```diff
@@ -99,75 +99,241 @@ static void SplitStringToVector(const std::string &in, const char *delim,
...
+
+   If there are multiple final states, we will create a super final state, and
+   create arcs from the old final states to the new super final state, with the
+   (negated) old final state cost/score as its cost/score, and -1 as its label.
+
+   @param [in]  is    The input stream that contains the acceptor.
+
+   @return It returns an Fsa on CPU.
+*/
+static Fsa OpenFSTAcceptorFromStream(std::istringstream &is) {
```
danpovey, line 249: Google style guide says this should be OpenFst not OpenFST

k2/python/k2/fsa.py
```diff
@@ -18,7 +19,7 @@
 
 class Fsa(object):
 
-    def __init__(self, s: str, openfst: bool = False):
+    def __init__(self, s: str):
```
danpovey, line 22: This is a preexisting issue, but I want to remind you guys that string input won't be the normal method; more commonly I anticpate reading a tensor in from disk, e.g. N by 4 or N by 5 of int32_t.

k2/csrc/fsa_utils.cu
```diff
@@ -99,75 +99,241 @@ static void SplitStringToVector(const std::string &in, const char *delim,
...
+      float score = StringToFloat(splits[3]);
+      arcs.emplace_back(src_state, dest_state, symbol, score);
+    } else if (num_fields == 1u) {
+      //   0
+      // final_state
+      (void)StringToInt(splits[0]);  // this is a final state
+      finished = true;               // set finish
+    } else {
+      K2_LOG(FATAL) << "Invalid line: " << line
+                    << "\nK2 acceptor expects a line with 1 (final_state) or "
```
danpovey, line 146: BTW, we've been using lowercase k2 in text... this is open to discussion but I felt it would get complicated with directory names and variable names and the like if we consistently use uppercase.

k2/csrc/fsa_utils.cu
```diff
@@ -99,75 +99,241 @@ static void SplitStringToVector(const std::string &in, const char *delim,
...
+                    << "4 (src_state dest_state label score) fields";
+    }
+  }
+
+  K2_CHECK_EQ(finished, true) << "The last line should be the final state";
```
danpovey, line 151: Actually it's not a requirement for OpenFST that the last line is a final state; also, in principle it's possible to have an FSA with no final states (although its not so useful).  So maybe remove this check?

k2/csrc/fsa_utils.cu
```diff
@@ -99,75 +99,241 @@ static void SplitStringToVector(const std::string &in, const char *delim,
...
-    } else if (num_fields == 1u) {
-      //   0
-      // final_state
-      (void)StringToInt(splits[0]);  // this is a final state
-      break;                         // finish reading
+      has_final_state = true;
     } else {
       K2_LOG(FATAL) << "Invalid line: " << line
-                    << "\nIt expects a line with 1, 2 or 4 fields";
+                    << "\nOpenFST acceptor expects a line with 1 (final_state),"
```
danpovey, line 300: may not need the <<'s here..

k2/python/k2/fsa.py
chenguoguo: I was also wondering about that. What about we have
```
def from_string(...) -> 'Fsa':
```
for reading from strings, and let the default `__init__` function read from a tensor, something like
```
def __init__(self, tensor: torch.Tensor, aux_labels: Optional[torch.Tensor] = None) -> 'Fsa':
```

chenguoguo:
I addressed most of the comments. For the `__init__` function of the `Fsa` class, I'd propose we make the default `__init__` function read from tensors, and have another function `from_string(...) -> Fsa` to read from strings. What do you guys think? If you think that would be a good change, I'll make some changes.

danpovey:
Thanks!
Yes, I agree init from tensors as default makes more sense.
If you want you can merge with my PR and test like that?  I haven't fully
tested that and today I'm not sure if I'll have time.

chenguoguo:
Sounds good, I'll find some time to make the change and test.

danpovey:
Thanks!!

chenguoguo:
I was distracted by my other tasks so getting back to this late. But I have pushed changed to initialize Fsa() from a tensor, and also added some python test code for initializing from tensor.

k2/csrc/fsa_utils.cu
```diff
@@ -99,75 +99,241 @@ static void SplitStringToVector(const std::string &in, const char *delim,
...
+    K2_CHECK_EQ(finished, false);
+
+    auto num_fields = splits.size();
+    if (num_fields == 4u) {
+      //   0            1          2      3
+      // src_state  dest_state   label  score
+      int32_t src_state = StringToInt(splits[0]);
+      int32_t dest_state = StringToInt(splits[1]);
+      int32_t symbol = StringToInt(splits[2]);
+      float score = StringToFloat(splits[3]);
```
danpovey, line 137: Not urgent, but we'll probably at some point get into  locale problems here, with comma vs. period for decimals.
 :-(  These things are hard to avoid.  I don't know whether the C code will behave the same w.r.t. locale as Python does.  In Kaldi we always recommended LC_ALL=C.  We should decide at some point what the best practice for this toolkit is.

k2/csrc/fsa_utils.cu
```diff
@@ -177,101 +343,145 @@ static Fsa AcceptorFromStream(std::string first_line, std::istringstream &is,
...
+
+   src_state1 dest_state1 label1 aux_label1 score1
+   src_state2 dest_state2 label2 aux_label2 score2
+   ... ...
+   final_state final_score
+
+   We will negate the cost/score when we read them in. Also note, OpenFST may
+   omit the cost/score if it is 0.0.
+
+   If there are multiple final states, we will create a super final state, and
```
danpovey, line 357: I don't believe it is actually conditional.
The code as you have it requires there to be at least one final state, no?
I think it would be better to always create a superfinal state, but not require there to be any final states.
(would just be disconnected in that case).

k2/python/k2/fsa.py
```diff
@@ -18,7 +19,48 @@
 
 class Fsa(object):
 
-    def __init__(self, s: str, openfst: bool = False):
+    def __init__(self, tensor: torch.Tensor,
```
danpovey, line 22: OK... I notice we only have an Fsa object and not FsaVec.  FsaVec is actually going to be the more common case.  For now I think we can have both (not for this PR!) but we may eventually decide to unify them.

k2/csrc/fsa_utils.cu
chenguoguo: I thought about it for a second, but felt an FSA without a final state will be use less?

Let's do this then. We will always add a super final state. If there's at least one final state in the original FSA, we create an arc(s) from the original final state(s) to the super final state, with label -1; if there's no final state in the original FSA, we add the super final state (max_id + 1) but with no arcs connecting it with the original FSA. Or do you think we should add an arc from max_id to the super final state with label -1 and weight 0?

k2/python/k2/fsa.py
chenguoguo: That's right, I planned to add FsaVec as a separate PR. But I'm not super clear yet whether or not I should add it as a separate class or just have one class. I'll create an issue for that and we can discuss from there. Once we have a conclusion, and can then implement it.

k2/python/k2/fsa.py
```diff
@@ -50,60 +92,58 @@ def __init__(self, s: str, openfst: bool = False):
...
+            acceptor = False
 
-    @classmethod
-    def _create(cls, fsa: _Fsa,
-                aux_labels: Optional[torch.Tensor] = None) -> 'Fsa':
-        '''For internal use only.
-        '''
         ans = cls.__new__(cls)
         super(Fsa, ans).__init__()
+        fsa, aux_labels = _fsa_from_str(s, acceptor, False)
```
danpovey, line 106: Instead of having aux_labels as a member directly, we may have it as one of the `per_arc` pseudo-members.
Basically, if we have an Fsa or FsaVec called f, we will have a bunch of fields like f.aux_labels, f.scores, f.foo, f.bar and so on.  They are not actually members but will correspond to keys in some kind of dict.  We'll have fairly general purpose code for propagating them; the main constraint is that the 1st index into them will correspond to the arc-index. 
In the specific case of aux_labels, it may end up getting treated specially for certain operations, like Invert() which would swap those labels with the `symbol` in the arc.

k2/csrc/fsa_utils.cu
danpovey: Always add super-final state but no special cases, we don't want to guess at what the user wants.  Better to have predictable behavior.  Non-coaccessible FSAs do sometimes arise naturally.

k2/python/k2/fsa.py
danpovey: Yeah it's not clear to me either.  I kind of assumed Fangjun would be doing this part (FsaVec).  I will follow up with some other comments on the design.  In fact I'll create an issue.

k2/csrc/fsa_utils.cu
chenguoguo: I added a TODO to the StringToFloat function regarding the locale issue.

k2/csrc/fsa_utils.cu
chenguoguo: I removed the constrain of the presence of the final state in the OpenFST format. Super final state will always be added.

k2/python/k2/fsa.py
chenguoguo: I'm not sure if I understand the pseudo-member part. Is this on the Python side, or on the C++ side accessed by Python through pybind? Do you have an example on this?

csukuangfj: I think `pseudo-member` is via `__getattr__` and `__setattr__`, which is mentioned in one of the issues Dan just created.

danpovey:
Yes, pseudo-member would not be supported at the C++ level for now, only in
python.

chenguoguo:
Oh I see issue #231 now. Looks like we will have a re-design on the k2.Fsa class. Shall we keep the aux_labels as is for now, and fix the pseudo-member issues when we re-do the k2.Fsa class in a separate PR?

k2/python/k2/fsa.py
csukuangfj: I did not add `FsaVec` in Python because the construction of `FsaVec` had bugs which were fixed by #226 .

danpovey:
Keeping them as-is is fine with me.

PR #226, opened 2020-10-07
Implemented EvalRedirect and fixed bug in Append

qindazhu:
@danpovey please have a review. I'm not 100% sure about my implementation of ParallelRunner (and the context stuff). Please leave your comments and I'll fix tomorrow.

### DIFF
--- a/k2/csrc/context.cu
+++ b/k2/csrc/context.cu
@@ -28,4 +28,52 @@ RegionPtr NewRegion(ContextPtr &context, std::size_t num_bytes) {
   return ans;
 }
 
+ParallelRunner::ParallelRunner(ContextPtr c) : c_(c) {
+  if (c_->GetDeviceType() == kCuda) {
+    auto ret = cudaEventCreate(&event_);
+    K2_CHECK_CUDA_ERROR(ret);
+    // record event on `c_->GetCudaStream` and will be waited on `NewStream`
+    ret = cudaEventRecord(event_, c_->GetCudaStream());
+    K2_CHECK_CUDA_ERROR(ret);
+  }
+}
+cudaStream_t ParallelRunner::NewStream() {
+  DeviceType d = c_->GetDeviceType();
+  if (d == kCpu) {
+    return kCudaStreamInvalid;
+  } else {
+    K2_CHECK_EQ(d, kCuda);
+    cudaStream_t stream;
+    auto ret = cudaStreamCreate(&stream);
+    K2_CHECK_CUDA_ERROR(ret);
+    streams_.push_back(stream);
+
+    ret = cudaStreamWaitEvent(stream, event_, 0);
+    K2_CHECK_CUDA_ERROR(ret);
+    return stream;
+  }
+}
+
+ParallelRunner::~ParallelRunner() {
+  if (c_->GetDeviceType() == kCuda) {
+    for (std::size_t i = 0; i != streams_.size(); ++i) {
+      // create and record event on `stream_[i]`, and wait on c_->GetCudaStream
+      cudaEvent_t event;
+      auto ret = cudaEventCreate(&event);
+      K2_CHECK_CUDA_ERROR(ret);
+      ret = cudaEventRecord(event, streams_[i]);
+      K2_CHECK_CUDA_ERROR(ret);
+      ret = cudaStreamWaitEvent(c_->GetCudaStream(), event, 0);
+      K2_CHECK_CUDA_ERROR(ret);
+      ret = cudaEventDestroy(event);
+      K2_CHECK_CUDA_ERROR(ret);
+      ret = cudaStreamDestroy(streams_[i]);
+      K2_CHECK_CUDA_ERROR(ret);
+    }
+    // destroy event_
+    auto ret = cudaEventDestroy(event_);
+    K2_CHECK_CUDA_ERROR(ret);
+  }
+}
+
 }  // namespace k2
--- a/k2/csrc/context.h
+++ b/k2/csrc/context.h
@@ -533,7 +533,7 @@ class With {
 */
 class ParallelRunner {
  public:
-  explicit ParallelRunner(ContextPtr c) : c_(c) {}
+  explicit ParallelRunner(ContextPtr c);
 
   // create a new stream, that first syncs with stream of c_ via an event.  The
   // destructor will cause the stream of c_ to wait on this stream in the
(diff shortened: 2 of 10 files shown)


k2/csrc/context.h
```diff
@@ -550,15 +550,33 @@ class ParallelRunner {
...
   cudaStream_t NewStream() {
-    // TODO
-    return kCudaStreamInvalid;
+    DeviceType d = c_->GetDeviceType();
+    if (d == kCpu) {
+      return kCudaStreamInvalid;
+    } else {
+      K2_CHECK_EQ(d, kCuda);
+      cudaStream_t stream;
+      auto ret = cudaStreamCreate(&stream);
```
danpovey, line 559: after creating the stream we need to record an event in the stream c_.Stream(), and have stream `stream` wait on that event.

k2/csrc/context.h
```diff
@@ -550,15 +550,33 @@ class ParallelRunner {
...
+      K2_CHECK_CUDA_ERROR(ret);
+      q_.push_back(stream);
+      return stream;
+    }
   }
 
-  void Finish();  // like calling destructor manually.
+  ~ParallelRunner() {
+    for (std::size_t i = 0; i != q_.size(); ++i) {
+      auto ret = cudaStreamSynchronize(q_[i]);
```
danpovey, line 568: cudaStreamSynchronize is not the way we synchronize streams... we create an even in the stream q_[i], and have stream c_.Stream() wait on that event.

k2/csrc/context.h
```diff
@@ -550,15 +550,33 @@ class ParallelRunner {
...
+    for (std::size_t i = 0; i != q_.size(); ++i) {
+      auto ret = cudaStreamDestroy(q_[i]);
+      K2_CHECK_CUDA_ERROR(ret);
+    }
+  }
 
  private:
   ContextPtr c_;
-  // TODO: list of events to wait on, maybe CUDA streamss.
+  std::vector<cudaStream_t> q_;
```
danpovey, line 579: I'd call this streams_; I'm not convinced that it's a queue.

k2/csrc/utils.h
```diff
@@ -412,7 +431,33 @@ void EvalWithRedirect(cudaStream_t stream, int32_t num_jobs,
                       TaskRedirect *redirect, int32_t min_threads_per_job,
                       int32_t tot_work, int32_t target_num_loops,
                       LambdaT &lambda) {
-  // TODO:
+  if (num_jobs <= 0 || tot_work <= 0) return;
```
danpovey, line 434: EvalWithDirectt() may not be that useful for host code, and in fact we may just avoid calling it for host code.  We may eventually have a version of Eval() that  is device-only, since this is a common pattern.  But we dont have to worry about this just yet.

danpovey (commented):
Made a couple comments directly.  LMK if there is anything else you weren't sure of.

k2/csrc/context.h
qindazhu: Updated.

k2/csrc/utils.h
qindazhu: Added a `TODO` for now, will get back when we get other functions calling `TaskRedirect` tested.

k2/csrc/context.h
qindazhu: Updated with events, Please help to review again.

qindazhu:
@danpovey, updated with Cuda event and added tests for random large size...Please review again, many thanks!

k2/csrc/context.h
```diff
@@ -550,15 +550,51 @@ class ParallelRunner {
   // will call CudaStreamOverride::OverrideStream() and replace it
   // with this stream automatically.
   cudaStream_t NewStream() {
-    // TODO
-    return kCudaStreamInvalid;
+    DeviceType d = c_->GetDeviceType();
```
danpovey, line 553: not urgent, but I'd prefer this to be implemented in a .cu file not the header, for readability.

k2/csrc/context.h
```diff
@@ -550,15 +550,51 @@ class ParallelRunner {
...
+    if (d == kCpu) {
+      return kCudaStreamInvalid;
+    } else {
+      K2_CHECK_EQ(d, kCuda);
+      // create event
+      cudaEvent_t event;
+      auto ret = cudaEventCreate(&event);
+      K2_CHECK_CUDA_ERROR(ret);
+      events_.push_back(event);
```
danpovey, line 562: Mm, this event stuff is not quite right.
Firstly, you can create an even in c_->GetCudaStream() when this object is created, and have all streams that you create wait on it as soon as they are created. 
Next, you only need to create events in the spawned streams in the destructor of ParallelRunner.  The idea is to force c_->GetCudaStream() to wait on the jobs done in the spawned streams, so the event must be created after those jobs have been submitted.

danpovey: .. so you don't need an events_ member, but you can have an event_ member containing just one event.

qindazhu: Why do we need to wait event on `c_->GetCudaStream` in created streams when the object is created? Can't we suppose all jobs on `c_->GetCudaStream` have been done before creating these streams?

danpovey:
No, because jobs run asynchronously, the CPU doesn't wait for them to
finish, they are just put in a queue.
It's invisible to the user if you use a single stream but with multiple
streams you need to explicitly record these things.

k2/csrc/context.h
qindazhu: Moved to `context.cu`

k2/csrc/context.h
qindazhu: Updated, please review again, thanks

k2/csrc/context.cu
```diff
@@ -28,4 +28,47 @@ RegionPtr NewRegion(ContextPtr &context, std::size_t num_bytes) {
...
+ParallelRunner::~ParallelRunner() {
+  if (c_->GetDeviceType() == kCuda) {
+    for (std::size_t i = 0; i != streams_.size(); ++i) {
+      // record event on `stream_[i]` and wait on c_->GetCudaStream
+      auto ret = cudaEventRecord(event_, streams_[i]);
+      K2_CHECK_CUDA_ERROR(ret);
+      ret = cudaStreamWaitEvent(c_->GetCudaStream(), event_, 0);
+      K2_CHECK_CUDA_ERROR(ret);
+      ret = cudaStreamDestroy(streams_[i]);
+      K2_CHECK_CUDA_ERROR(ret);
```
danpovey, line 66: I think we need to destroy these events too, inside the loop.
I had a look at the documentation, it doesn't matter that they are not finished with yet.

qindazhu: There is only one event_ now and I destroy it at the end of loop. Do you mean we should create and destroy the `event_` again and again in the loop？

danpovey:
Oh I see... it needs to be a different event, I believe events aren't
really reusable like that.  Yeah you have to create one every time you go
around the loop.

qindazhu:
OK,  will update later (eating out currently..).  I just thought you means using one event in the last review, sorry for the misunderstanding.

k2/csrc/context.cu
qindazhu: Fixed.

k2/csrc/context.cu
```diff
@@ -28,4 +28,52 @@ RegionPtr NewRegion(ContextPtr &context, std::size_t num_bytes) {
...
+  if (d == kCpu) {
+    return kCudaStreamInvalid;
+  } else {
+    K2_CHECK_EQ(d, kCuda);
+    cudaStream_t stream;
+    auto ret = cudaStreamCreate(&stream);
+    K2_CHECK_CUDA_ERROR(ret);
+    streams_.push_back(stream);
+
+    ret = cudaStreamWaitEvent(stream, event_, 0);
```
qindazhu, line 51: @danpovey I'm still doubting this: as we only record event for `c_>GetCudaStream` once in the construction, does't only the first call `NewStream` wait the event? 
But If we record event every time in `NewStream`, as `With` will override the default stream of `c`, we may make those streams waiting one another, so we should not do this, am I right?

danpovey: cudaStreamWaitEvent causes stream `stream` to block until even `event` has finished.
We don't cause the default stream to wait for the streams until the object's destructor (or its
Finish() call, which should do what the destructor would normally do), so it's doing the right thing.

qindazhu: mm. What I was asking is about `event_` created on`c_->GetCudaStream` in the constructor (not those events created in the destructor, I know that's right.)

Anyway, forgot that, I just realized that we need to wait on `event_` in every streams (not only int the first call) created in `NewStream()` as they run asynchronously and we don't know which one will set the value of `event_` first.

PR #233, opened 2020-10-09
Declare a function related to sparse matrix transposition

danpovey:
Guys, this is a bit more time sensitive than some of the other things because it's needed so I can write code to get posteriors of a lattice, which of course would have to be tested.


danpovey:
I'll merge this and make an issue.

PR #236, opened 2020-10-09
Add couple more declarations 

danpovey:
...  we need to wrap cub things to implement these.


danpovey:
I am hoping someone can implement these and the other one within the next few days...
I'll be writing some code that requires it, e.g. (fixed) top-sorting code and code to get forward-backward probabilities over a lattice.

danpovey:
Merging.. this is just declarations.

k2/csrc/ragged_ops.h
```diff
@@ -347,6 +347,70 @@ inline Ragged<T> RaggedFromTotSizes(ContextPtr &c,
                    Array1<T>(c, tot_sizes.back()));
 }
 
+/*
+  Transpose a ragged tensor as if it were the index information of a CSR-format
+  sparse matrix (but with possibly repeated elements!).  This is easiest to
+  explain if we assume `src` has 2 axes.  We view `src` as a list of nonzero
+  elements of a matrix, indexed first by row, and containing column-indexes
+  (but possibly repeated column indexes, which violates the assumptions of
+  the cusparse library).  This function returns an array of dimension
```
csukuangfj, line 356: @danpovey 
I would like to give an example to show my understanding of the comment.
Please correct me if I am wrong.

For the following CSR-format matrix:
```
0 0 0 0
5 8 0 0
0 0 3 0
0 6 0 0
```


- row splits: 0 0 2 3 4
- values:       5 8 3 6

Transpose of the matrix is:

```
0 5 0 0
0 8 0 6
0 0 3 0
0 0 0 0
```
The values are `5 8 6 3`

So the returned `Array<int32_t>` contains: 0 1 3 2, which
maps `5 8 3 6` to `5 8 6 3`.

csukuangfj: If the above example is correct, I would like to implement it
by using the transpose algorithm from this paper:
[Two Fast Algorithms for Sparse Matrices: Multiplication and Permuted Transposition][1]

[gustavson1978.pdf](https://github.com/k2-fsa/k2/files/5355804/gustavson1978.pdf)


[1]: https://scholar.google.com/scholar?hl=en&as_sdt=0%2C5&q=Two+Fast+Algorithms+for+Sparse+Matrices%3A+Multiplication+and+Permuted+Transposition&btnG=

danpovey:
That seems right, but remember we need the algorithm to work even if there are 2 copies of the same element (i.e. the same (row, column) pair.  I'm afraid I can't see that paper from here (paywall).  
I was thinking you could perhaps do it via sorting-- although another option is to try the cusparse routine I mentioned and see whether it actually works in practice (even though the repeated elements might be a problem in theory).

csukuangfj:
I've uploaded the paper to GitHub: https://github.com/k2-fsa/k2/files/5355804/gustavson1978.pdf

csukuangfj:
> I was thinking you could perhaps do it via sorting

How is a transpose operation implemented via sorting?

k2/csrc/ragged_ops.h
qindazhu: why is `values` here so large? I was thinking they are column indexes (so any value should be `<4`)

csukuangfj: > The values are 5 8 6 3

This is for readability.

---

The following is how it looks like in code

https://github.com/k2-fsa/k2/blob/a531d3e0b2f6ecda1fd1ca45227289613ca4fa09/k2/csrc/ragged_test.cu#L1020-L1025

qindazhu: aha, OK.

danpovey:
I looked at that paper briefly but it doesn't seem to be a *parallel* algorithm.  I don't think it's applicable.

RE how to implement via sorting: you can view the nonzero entries of a sparse matrix as a list of (row,column) pairs.
In a CSR format they are sorted by row; we want them sorted by column.  So it's just a question of re-sorting them.  In fact, if we consider row as idx0 and column as idx1, the most natural way to do this would be to sort the (idx1,idx01) pairs which gives you the reordering vector we want.  I'd recommend to try that first; right now we are on a schedule so it's important to get things working fast.

csukuangfj:
> I looked at that paper briefly but it doesn't seem to be a parallel algorithm. I don't think it's applicable.


Thanks, understood. I will switch to sorting based transpose.

qindazhu:
@csukuangfj Have you ever started implementing `GetCountsPartitioned` and `GetCounts`, if not, I'll do this.

csukuangfj:
No, not yet.




Saturday, 10 October 2020, 13:10 +0800 from Haowen Qiu  <notifications@github.com>:

qindazhu:
OK, then I'll do this, thanks.

PR #237, opened 2020-10-10
implement GetTransposeReordering.

### LINKED ISSUE #234: Implement transpose function for ragged tensor
see GetTransposeReordering(), it's here 
https://github.com/k2-fsa/k2/pull/233
(already merged, but only declared)

csukuangfj:
This fixes #234 

~~It supports CPU only due to the lack of synchronizations in the lambda.~~

### DIFF
--- a/k2/csrc/ragged_ops.cu
+++ b/k2/csrc/ragged_ops.cu
@@ -10,6 +10,8 @@
  * See LICENSE for clarification regarding multiple authors
  */
 
+#include <algorithm>
+#include <memory>
 #include <vector>
 
 #include "cub/cub.cuh"
--- a/k2/csrc/ragged_ops.cu
+++ b/k2/csrc/ragged_ops.cu
@@ -4,6 +4,7 @@
  *
  * @copyright
  * Copyright (c)  2020  Xiaomi Corporation (authors: Daniel Povey, Haowen Qiu)
+ *                      Mobvoi Inc.        (authors: Fangjun Kuang)
  *
  * @copyright
  * See LICENSE for clarification regarding multiple authors
--- a/k2/csrc/ragged_ops.cu
+++ b/k2/csrc/ragged_ops.cu
@@ -10,13 +10,16 @@
  * See LICENSE for clarification regarding multiple authors
  */
 
-#include <cub/cub.cuh>
 #include <vector>
 
+#include "cub/cub.cuh"
 #include "k2/csrc/array_ops.h"
 #include "k2/csrc/math.h"
+#include "k2/csrc/moderngpu_allocator.h"
 #include "k2/csrc/ragged.h"
 #include "k2/csrc/ragged_ops.h"
+#include "moderngpu/kernel_mergesort.hxx"
+
 namespace {
 
 /*
--- a/k2/csrc/ragged_ops.cu
+++ b/k2/csrc/ragged_ops.cu
@@ -812,6 +812,26 @@ Ragged<int32_t> GetCountsPartitioned(Ragged<int32_t> &src,
   return Ragged<int32_t>(ans_ragged_shape, counts);
 }
 
+static Array1<int32_t> GetTransposeReorderingCpu(Ragged<int32_t> &src,
+                                                 int32_t num_cols) {
+  std::vector<std::vector<int32_t>> column_indexes(num_cols);  // [column][row]
+  const int32_t *values_data = src.values.Data();
+  int32_t n = src.values.Dim();
+
+  for (int32_t i = 0; i != n; ++i) {
+    int32_t bucket = values_data[i];
+    column_indexes[bucket].push_back(i);
+  }
+
+  Array1<int32_t> ans(src.Context(), n);
+  int32_t *ans_data = ans.Data();
+  for (int32_t i = 0; i != num_cols; ++i) {
+    std::copy(column_indexes[i].begin(), column_indexes[i].end(), ans_data);
+    ans_data += column_indexes[i].size();
+  }
+  return ans;
+}
+
 Array1<int32_t> GetTransposeReordering(Ragged<int32_t> &src, int32_t num_cols) {
   ContextPtr &context = src.Context();
   if (src.NumAxes() < 2) {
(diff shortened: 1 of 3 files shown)


csukuangfj:
I think it supports only  2-D matrices.

k2/csrc/ragged_ops.cu
```diff
@@ -793,4 +794,72 @@ RaggedShape TrivialShape(ContextPtr &c, int32_t num_elems) {
...
+  Array1<int32_t> column_counts(context, num_cols + 1, 0);
+  int32_t *col_ptr = column_counts.Data();
+
+  const Array1<int32_t> &values = src.values;
+  const int32_t *values_ptr = values.Data();
+
+  auto lambda_column_counts = [=] __host__ __device__(int32_t i) -> void {
+    const int32_t *begin = values_ptr + row_ptr[i];
+    const int32_t *end = values_ptr + row_ptr[i + 1];
+    for (; begin != end; ++begin) {
```
qindazhu, line 827: As `end-begin` may be large, I'm afraid the code here will be relatively too slow. At least I think we can launch a kernel and get the index (`*begin` here) in each thread, instead of in a loop.

qindazhu: I mean, at least we can do (in a kernel) 
```
int32_t row = threadIdx / num_threads_per_row;
int32_t thread_in_this_row = threadIdx % num_threads_per_row;
```

danpovey: yeah, the way you would do this is to have the kernel parameter i be the idx01, then use row_ids to get the idx0 then row_splits to get the idx0x and then the idx1.  Of course incrementing the column pointer would need to be protected: could use atomicAdd but would need to write host version as well to avoid compilation error.  (Or we can make a version of Eval() that's GPU only.
However there are certain inputs for which that algorithm would be very slow, because the atomicAdd makes it sequential.  That's why it's a bit safer to do it based on sorting.

csukuangfj:
@danpovey @qindazhu 
ready for review and merge.

csukuangfj:
I am using mergesort from moderngpu: https://moderngpu.github.io/mergesort.html

k2/csrc/ragged_ops.cu
```diff
@@ -793,4 +799,86 @@ RaggedShape TrivialShape(ContextPtr &c, int32_t num_elems) {
...
+  Array1<int32_t> ans(src.Context(), src.values.Dim());
+  int32_t *ans_data = ans.Data();
+  int32_t n = ans.Dim();
+  for (int32_t i = 0; i != n; ++i) ans_data[i] = i;
+
+  std::sort(ans_data, ans_data + n, lambda_comp);
+  return ans;
+}
+
+Array1<int32_t> GetTransposeReordering(Ragged<int32_t> &src, int32_t num_cols) {
```
csukuangfj, line 834: It turns out `num_cols` is not needed for sorting based transpose.

k2/csrc/ragged_ops.cu
```diff
@@ -793,4 +799,86 @@ RaggedShape TrivialShape(ContextPtr &c, int32_t num_elems) {
...
+  const int32_t *row_ids1_data = src.RowIds(src.NumAxes() - 1).Data();
+  const int32_t *value_data = src.values.Data();
+
+  auto lambda_set_value = [] __device__(int32_t idx01) -> int32_t {
+    return idx01;
+  };
+  int32_t n = src.values.Dim();
+  Array1<int32_t> ans(context, n, lambda_set_value);
+
+  auto lambda_comp = [=] __device__(int32_t a_idx01, int32_t b_idx01) -> bool {
```
qindazhu, line 858: seems duplicate code with above cpu version? Can't we use `__host__ __device__` in mgpu?

k2/csrc/ragged_ops.cu
```diff
@@ -793,4 +799,86 @@ RaggedShape TrivialShape(ContextPtr &c, int32_t num_elems) {
...
+
+  const int32_t *row_splits1_data = src.RowSplits(src.NumAxes() - 1).Data();
+  const int32_t *row_ids1_data = src.RowIds(src.NumAxes() - 1).Data();
+  const int32_t *value_data = src.values.Data();
+
+  auto lambda_set_value = [] __device__(int32_t idx01) -> int32_t {
+    return idx01;
+  };
+  int32_t n = src.values.Dim();
+  Array1<int32_t> ans(context, n, lambda_set_value);
```
qindazhu, line 817: With https://github.com/k2-fsa/k2/blob/d4ca5714316774d80be5f7ea713892d087b17ae0/k2/csrc/array_ops.h#L319

You can just do Array1<int32_t> ans = Range(c,n,0);

k2/csrc/ragged_ops.cu
csukuangfj: fixed via removal.

k2/csrc/ragged_ops.cu
csukuangfj: fixed, thanks.

k2/csrc/ragged_ops.cu
```diff
@@ -806,4 +812,52 @@ Ragged<int32_t> GetCountsPartitioned(Ragged<int32_t> &src,
...
+    if (a_idx0 > b_idx0) return false;
+
+    // now we have a_idx0 == b_idx0 && a_col_index == b_col_index
+    // this entry is duplicated in the sparse matrix.
+    return false;  // we can return either true or false here.
+  };
+
+  DeviceType device_type = context->GetDeviceType();
+  if (device_type == kCpu) {
+    std::sort(ans.Data(), ans.Data() + n, lambda_comp);
```
danpovey, line 850: Not *super* urgent, but on CPU we can do much better than sort (linear time, not log(n) time) by using, say, a vector of vectors.

k2/csrc/ragged_ops.cu
danpovey: OK, let's leave num_cols in the interface anyway, for future extensibility.

danpovey:
Great, thanks!!  Will merge.

k2/csrc/ragged_ops.cu
csukuangfj: Thanks, fixed.

qindazhu:
Great, merging.

PR #238, opened 2020-10-10
Implemented GetCounts and tested

qindazhu:


k2/csrc/array_ops.cu
```diff
@@ -270,4 +270,44 @@ void RowIdsToRowSplits(const Array1<int32_t> &row_ids,
...
+    std::size_t temp_storage_bytes = 0;
+    // Note we will get below warning here:
+    //   /k2/build/_deps/cub-src/cub/device/device_histogram.cuh(451):
+    //   warning: calling a constexpr __host__ function("max") from a __host__
+    //   __device__ function("MultiHistogramEven") is not allowed. The
+    //   experimental flag '--expt-relaxed-constexpr' can be used to allow this.
+    // That's because cub calls `std::numeric_limits<int>::max` in
+    // `MultiHistogramEven`, see here
+    // https://github.com/NVlabs/cub/blob/c3cceac115c072fb63df1836ff46d8c60d9eb304/cub/device/device_histogram.cuh#L451.
+    // Hopes they can fix it in the future.
```
qindazhu, line 298: Seems in branch `main` of cub, they have fixed this issue by using `INT_MAX` instead of `std::numeric_limits<int>::max`
https://github.com/NVIDIA/cub/blob/a39e385cc6be20754f859dd266021ab1d88459d3/cub/device/device_histogram.cuh#L451

qindazhu: `1.10.0` is fine too https://github.com/NVIDIA/cub/blob/1.10.0/cub/device/device_histogram.cuh#L451.

@csukuangfj is there any special reason that you download `1.9.10` in cmake but I don't know?  https://github.com/k2-fsa/k2/blob/9c84418bfc9bf6f8263a57c5530c85ec9ca8fc78/cmake/cub.cmake#L11

qindazhu:
Anyway, ready to merge, we can fix the cub issues later.

danpovey:
Fantastic!  Merging.

k2/csrc/array_ops.cu
csukuangfj: 1.9.10 was the latest version when it was added.

I will update it to the current latest version 1.10.0

csukuangfj: The warnings will disappear once #237 is merged.

PR #239, opened 2020-10-10
Implemented GetCountsPartitioned and tested

qindazhu:


k2/csrc/ragged_ops.cu
```diff
@@ -793,4 +793,18 @@ RaggedShape TrivialShape(ContextPtr &c, int32_t num_elems) {
...
 
+Ragged<int32_t> GetCountsPartitioned(Ragged<int32_t> &src,
+                                     RaggedShape &ans_ragged_shape) {
+  K2_CHECK_EQ(src.NumAxes(), 2);
+  K2_CHECK_EQ(ans_ragged_shape.NumAxes(), 2);
+  K2_CHECK(IsCompatible(src, ans_ragged_shape));
+  K2_CHECK_EQ(src.Dim0(), ans_ragged_shape.Dim0());
+  const Array1<int32_t> &values = src.values;
+  const Array1<int32_t> &row_splits = ans_ragged_shape.RowSplits(1);
+  int32_t n = row_splits.Back() - row_splits[0];
```
danpovey, line 804: Just use ans_ragged_shape.NumElements().  BTW, row_splits[0] is always 0.

k2/csrc/ragged_ops.cu
```diff
@@ -793,4 +793,18 @@ RaggedShape TrivialShape(ContextPtr &c, int32_t num_elems) {
...
+  Array1<int32_t> counts = GetCounts(values, n);
+  Ragged<int32_t> ans(ans_ragged_shape, counts);
+  return ans;
```
danpovey, line 807: I'd just do: `return Ragged<int32_t>(ans_ragged_shape, counts);`

k2/csrc/ragged_ops.cu
qindazhu: aha, thanks for reminding, updated!

k2/csrc/ragged_ops.cu
qindazhu: +1, fixed.

danpovey:
Thanks, you can merge.
BTW, it will be good if you can get in the habit of using FSA equivalence testing for checking algorithms, when the algorithm's output is an FSA.  Some of the algorithms I am implementing will use non-deterministic numbering of the
output states.

qindazhu:
OK, BTW, I think we should add LOG in those checking algorithms as we usually wants to print the data to be checked, especially when there are not equal as expected.

qindazhu:
@danpovey ,what's the function name of  `FSA equivalence testing` you mentioned, I didn't find such method in `k2/csrc`, did you mean the fsa equivalence method in `k2/csrc/host`? (Note all ragged_ops methods for now I tested are all output-deterministic so Fsa random equivalence checking is not so applicable to those methods.)

danpovey:
I meant code in host/.
It doesnt matter that the ops are deterministic.. some of the algorithms I'm writing use lambdas that have race conditions.

PR #240, opened 2020-10-11
Tested Transpose and Stack

qindazhu:
Note all `ragged_ops`  now have been tested (except a few simple wrapper methods based on these ops)

### DIFF
--- a/k2/csrc/array_ops.cu
+++ b/k2/csrc/array_ops.cu
@@ -291,15 +291,6 @@ Array1<int32_t> GetCounts(const Array1<int32_t> &src, int32_t n) {
     K2_CHECK_EQ(d, kCuda);
     void *d_temp_storage = NULL;
     std::size_t temp_storage_bytes = 0;
-    // Note we will get below warning here:
-    //   /k2/build/_deps/cub-src/cub/device/device_histogram.cuh(451):
-    //   warning: calling a constexpr __host__ function("max") from a __host__
-    //   __device__ function("MultiHistogramEven") is not allowed. The
-    //   experimental flag '--expt-relaxed-constexpr' can be used to allow this.
-    // That's because cub calls `std::numeric_limits<int>::max` in
-    // `MultiHistogramEven`, see here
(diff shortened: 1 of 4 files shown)


danpovey:
Great! 
BTW I think we need to get in the habit of implementing randomized tests.
Can you implement a test like the following, for Transpose?
  - Generate random ragged array
  - transpose
  - convert both source and dest to CPU
  - use RaggedShapeIndexIterator to test: for each element of src, get the std::vector that's the index, swap elements 0 and 1 to get the index for the transposed one, and check that src[orig_index] == dest[transposed_index] ?
I'll merge this though.

qindazhu:
Just to make sure, what do you mean by saying `generate random ragged array`? it's `Ragged<T>` or just `RaggedShape`? Note my current implementation just checks if the swapped indexes are valid or not, no elements equality check (as I thought for a random large size `RaggedShape`, index checking failure may equal to elements check failure with a very great probability?)

This is my current implementation of `Transpose Test` (Note `transposed` is a `RaggedShape` with random size):
https://github.com/k2-fsa/k2/blob/3cdc00fa061dec199264ac9d94311342365dd2fa/k2/csrc/ragged_test.cu#L356-L368
And this is my current implementation of `Stack Test` (Note `cpu_shapes` is a vector of `RaggedShape` with random sizes):
https://github.com/k2-fsa/k2/blob/3cdc00fa061dec199264ac9d94311342365dd2fa/k2/csrc/ragged_test.cu#L1210-L1218

qindazhu:
BTW, all tests I implemented for `ragged_ops` are all with two test cases:

-  one is `simple case` (well, it's mainly for debug purpose as we can just replace bad case in the `simple case` to debug, that's exactly what I was doing when I fixed bugs in those algorithms). 

- one is `random case` with random size (e.g. RandInt(0, 10000)), I run the random test with hundreds times locally and change back to `2 times` when committing code.

danpovey:
Oh yes, your test looks fine, sorry.

PR #242, opened 2020-10-11
Refactor Fsa in Python.

csukuangfj:
Implement `__getattr__` and `__setattr__` of `Fsa` in Python.

### DIFF
--- a/k2/csrc/array.h
+++ b/k2/csrc/array.h
@@ -51,7 +51,7 @@ class Array1 {
   }
 
   // Return a copy of this array that does not share the same underlying data.
-  Array1<T> Clone();
+  Array1<T> Clone() const;
 
   int32_t ByteOffset() const { return byte_offset_; }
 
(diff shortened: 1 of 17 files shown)


csukuangfj:
No tests for `FsaVec` since `FsaVecFromArray1` has not been implemented in C++.

csukuangfj:
> The `__setattr__` will just check that the first index equals the `arcs.numel()`, and set it in the dict. However, if it is the .weight attribute it will also overwrite the weights in the FSA with it, i.e. the .scores field, in addition to setting the class member.

> Note: we will always create an attribute called '.scores' when we construct an FSA, e.g. from a tensor. This .scores attribute will be parallel to the .score field of the arcs, the idea is that they will always be the same (but separately stored). If the user overwrites the .scores field, we will also automatically overwrite the .score elements of the arcs.

The implementation does not follow the above description striclty. However, I think it achieves the same effect.

danpovey:
Great work!  Had a quick look and didn't see any problems.  Let me leave this up a little in case of other comments, then merge.

qindazhu:
+1

danpovey:
OK to merge if you think its OK

PR #243, opened 2020-10-12
Declare MonotonicLowerBound

danpovey:
Would be nice if someone can implement this.  Will be needed for the forward-backward algorithm, to compute batches.


qindazhu:
OK, will do later today.

danpovey:
Wait, just realized I don't need this right now after all.  Do it if you
want anyway though.

qindazhu:
mm. Sorry, didn't check the messages...I have finished the code..

danpovey:
Fantastic!

PR #246, opened 2020-10-12
implemented MinPerSubList and moved some ragged_ops from array_op.h to ragged_ops.h

qindazhu:
Just notice there are some tedious code in current test code which can be simplified with APIs added in these days (e.g. RaggedShape2.), but let us fix those issues after the first release instead of now.


danpovey:
Great!  Will merge.

Issue #175, opened 2020-09-23
Pseudocode for confidence estimation


danpovey:
Guys, I am creating this issue just as a way to show this pseudocode, it demonstrates parts of where we are going
with k2.  One feature is that the fsa objects will have fields 'per_arc' which contain arbitrary tensors whose first
dimension is the values.Dim() of the arcs.  (They are accessed as if they were class members but are really 
members of a dict).  When we do operations on FSAs, these `per_arc` quantities are propagated.  (This is
easy given the arc_map objects).

The python-level fsa object is going to be a more complicated object than the C++ one.   You can think of it as containing the C++ object as just one member (maybe we could make the C++ object a member called `arcs` or something).
```
loglikes = eval_nnet(minibatch.data)
dense_fsas = k2fsa.dense_fsas(loglikes)
lattices = k2fsa.pruned_compose(dense_fsas, decoding_graph)
if train:
   oracle = k2fsa.pruned_compose(dense_fsas, minibatch.supervision_graphs)
   # `oracle` inherited per_arc.word_labels from `supervision_graphs`
    oracle.per_arc.frames = dense_fsas.get_idx01(oracle.src_arcs_a())
    oracle = k2fsa.best_path(oracle)
    ref_phone_labels = torch.zeros([dense_fsas.shape.TotSize(1)], dtype=torch.long)
    ref_word_labels = ref_phone_labels.clone()
    ref_phone_labels[oracle.per_arc.frames] = oracle.get_labels()
    # note: word_labels were propagated from `supervision_graphs.per_arc.word_labels`
    ref_word_labels[oracle.per_arc.frames] = oracle.per_arc.word_labels

post = k2fsa.arc_post(lattices)
lattices.per_arc.frames = dense_fsas.get_idx01(lattices.src_arcs_a())
      # gives unique identifiers for
      # frames of input, different for different utterances.
arc_post = k2fsa.arc_post(lattices)

num_arcs = lattices.arcs.shape[0]
phones_one_hot = torch.zeros(num_arcs, num_phones)
phones_one_hot[range(num_arcs), lattices.get_labels()] = 1.0
# initial features are (acoustic, LM, posterior, LLR vs. best path, one-hot phone labels
lattices.per_arc.feats = torch.cat(torch.stack(lattices.inputs.scores_a(),
      lattices.inputs.scores_b(), log(arc_post), k2fsa.llr(lattices)),
               phones_one_hot)
# augment with the features above but averaged on each frame, over all paths.
[...]

PR #248, opened 2020-10-12
Wrap DenseFsaVec to Python.

csukuangfj:


csukuangfj:
Ready for review.

k2/python/k2/dense_fsa.py
```diff
@@ -0,0 +1,53 @@
...
+    '''Construct a DenseFsa from neural net log-softmax outputs.
+
+    Args:
+      log_probs:
+        A 3-D tensor of dtype ``torch.float32`` with shape ``(N, T, C)``, where
+        ``N`` is the batch size, ``T`` the maximum input length, and ``C`` the
+        number of output classes.
+      input_lengths:
+        A 1-D tensor of dtype ``torch.int32`` with size ``(N)``, where ``N`` is
+        the batch size. It represents the length of the input (must each NOT
```
qindazhu, line 19: each must?

k2/python/k2/dense_fsa.py
```diff
@@ -0,0 +1,53 @@
...
+    '''
+    assert log_probs.ndim == 3
+    assert log_probs.dtype == torch.float32
+    assert input_lengths.ndim == 1
+    assert input_lengths.dtype == torch.int32
+
+    N = log_probs.size(0)
+    assert input_lengths.size(0) == N
+
+    num_rows = input_lengths.sum().item() + N
```
qindazhu, line 33: Is `+N` here for the last row? (`+1` for every one?), maybe it's better to add comments here.

k2/python/k2/dense_fsa.py
csukuangfj: I think both are fine.

this is from
https://pytorch.org/docs/stable/generated/torch.nn.CTCLoss.html

> Input_lengths: Tuple or tensor of size (N)(N) , where N = \text{batch size}N=batch size . It represent the lengths of the inputs (must each be \leq T≤T ).

k2/python/k2/dense_fsa.py
csukuangfj: The length of every sequence is incremented.
https://github.com/k2-fsa/k2/blob/9b679bdbfd803fdd1e6fc4c5813f251a2b8ebf19/k2/csrc/fsa.h#L118-L120

I'll add more comments for the Python code.

k2/python/k2/dense_fsa.py
qindazhu: OK, just feel a little hard for me (a non-native) to read.

k2/python/k2/dense_fsa.py
qindazhu: Yeah, as Python users may not check C++ code. Thanks!

k2/csrc/compose.cu
```diff
@@ -507,7 +508,7 @@ class MultiGraphDenseIntersect {
               a_fsas_arc_idx012 = a_fsas_arc_idx01x + ai_arc_idxxx2;
       Arc arc = arcs[a_fsas_arc_idx012];
 
-      int32_t scores_idx0x = b_fsas_row_ids1[ai_fsa_idx0],
+      int32_t scores_idx0x = b_fsas_row_ids1[ai_fsa_idx0], // FIXME(fangjun): this is probably b_fsa_row_splits1  // NOLINT
```
danpovey, line 511: you are right, should be the row-splits.

k2/python/csrc/torch/fsa.cu
```diff
@@ -70,9 +61,42 @@ static void PybindFsaUtil(py::module &m) {
...
 
+static void PybindDenseFsaVec(py::module &m) {
+  using PyClass = DenseFsaVec;
+  py::class_<PyClass> pyclass(m, "DenseFsaVec");
+  // We do not need to access its members in Python
+
+  // TODO(fangjun): add docstring for this funciton
+  pyclass.def(
+      py::init([](torch::Tensor scores, torch::Tensor sizes) -> DenseFsaVec * {
+        // remove the contiguous check once the following comment
```
danpovey, line 72: OK.  Bear in mind that on the Python side we'll need some convenience wrappers to get things into this format.  Look at the comments in the C++ to understand stuff about the -infinity's.  Basically the original format will be: a tensor of nnet outputs (say: shape=(batch_size,T,dim)); and 3 pieces of information for each supervised segment: batch_index, start_time, duration.

k2/python/k2/dense_fsa.py
```diff
@@ -0,0 +1,53 @@
...
+
+def dense_fsa(log_probs: torch.Tensor,
+              input_lengths: torch.Tensor) -> _k2.DenseFsaVec:
+    '''Construct a DenseFsa from neural net log-softmax outputs.
+
+    Args:
+      log_probs:
+        A 3-D tensor of dtype ``torch.float32`` with shape ``(N, T, C)``, where
+        ``N`` is the batch size, ``T`` the maximum input length, and ``C`` the
+        number of output classes.
```
danpovey, line 16: This is not quite right, the format is more general.  Instead of `input_lengths`, we'll have a list of (sequence_index,start_time, duration).
And you haven't made much of an effort here to avoid loops!  Loops that access the GPU, in particular, are not
a great idea.  If you need to do small writes, better to do them on the CPU and transfer things to the GPU in one operation.  I'm open to adding some code at the C++ level to do this formatting, also.  (But that's not urgent, we can leave it for later, assuming we can't find a way to do it efficiently in Python).

k2/python/csrc/torch/fsa.cu
csukuangfj: I was thinking that it might be better to provide a single interface to construct `DenseFsaVec` and the supervision `FsaVec` together, which is less error-prone.

The `SupervisionSegment` from `lhotse` is defined as
https://github.com/pzelasko/lhotse/blob/master/lhotse/supervision.py#L8
```python
class SupervisionSegment:
    start: Seconds
    duration: Seconds
    text: Optional[str] = None
```

I guess `k2` is not going to depend on `lhotse`. I'd propose to add a supervision class for k2.

k2/python/k2/dense_fsa.py
csukuangfj: I was following: https://github.com/k2-fsa/k2/issues/175#issuecomment-702571462
> Note, something would need to be done in Python in general, when constructing it, to add the - infinitys in the right place. (This needs to be done at the Python level, using the base toolkit, so that autograd will work).

I don't think `autograd` plays a role here. We'll calculate the gradient via arc maps manually.

I agree it is more efficient to do this in C++.  I'll move the formatting process to C++.

csukuangfj: >  Instead of input_lengths, we'll have a list of (sequence_index,start_time, duration).

Do you mean we should also pass the supervision as an argument? Without supervisions, I cannot understand
how to make use of `start_time, duration`. Could you give some hints?

danpovey: Perhaps you were following the "regular" case there.  The regular case was more regular than you realized, I think: i just meant completely rectangular, no raggedness.
The irregular case may have zero or more than one supervision segments per sequence.

Regarding autograd: it still plays a role.  The arc_maps will propagate the gradient back to the special matrix (containing -infinitys and zeros as well as scores) that we construct, but (a) actually in most instances this will happen automatically using PyTorch autograd, because we always create a "parallel" version of the scores, separate from the FSA, and that's used to handle the gradients; and (b) the autograd back from the special matrix containing -infinity's, to the original matrix of scores, will rely on the PyTorch autograd.  [Incidentally, the arc_map will rely on a linearized/flattened version of the special matrix, and we will assume that matrix is contiguous.]
We should ensure that the copying of the scores from the original nnet output into the special matrix is done as a single indexing operation: this will be fastest in both forward and backward passes.

k2/python/csrc/torch/fsa.cu
danpovey: Mm.  The supervision will really just be an FsaVec.  (Its 1st index, the fsa_idx, corresponds to the 1st index in the DenseFsaVec).

danpovey: ... The creation of the supervision FSAs, and their I/O, will rely on fairly general FSA functions.  The only special thing is that in many cases we won't need to retain autograd info for the graph creation.

danpovey:
No, we dont need the supervision FSAs just yet...
The regular matrix of scores is (N, T, C).  But each sequence of length T
may have zero, one, or more than one supervised sub-segments in it, and
those sub-segments will not necessarily start at t=0.  So we have a list of
supervised sub-segments, and each one has (n, t, dur) where 0 <= n < N, 0
<= t < T, 0 <= t+dur < T, and dur > 0.

k2/python/k2/dense_fsa.py
csukuangfj: RE autograd: I reread the pseudocode from https://github.com/k2-fsa/k2/issues/175. Yes, you're right.
I'll try to figure out how to use PyTorch's index method to do the formatting process.

danpovey:
RE indexing, it may be easier to reshape the input from (N, T, C) to (N*T,
C).  Then you'll be able to use a one-dimensional LongTensor as the index
vector.

csukuangfj:
> The regular matrix of scores is (N, T, C).  But each sequence of length T
may have zero, one, or more than one supervised sub-segments in it

Aha, I see. You are going to split a long sequence into smaller segments and decode them in parallel.

>  So we have a list of supervised sub-segments, and each one has (n, t, dur) 

If the list size is `S`,  in the end, the resulting `DenseFsaVec` will contain `S` FSAs. The i-th FSA contains `dur_i + 1` states
and its `scores` has `sum of all dur` `+ S` rows. Is this correct?

danpovey:
> The regular matrix of scores is (N, T, C). But each sequence of length T
> may have zero, one, or more than one supervised sub-segments in it
>
> Aha, I see. You are going to split a long sequence into smaller segments
> and decode them in parallel.
>
> So we have a list of supervised sub-segments, and each one has (n, t, dur)
>
> If the list size is S, in the end, the resulting DenseFsaVec will contain
> S FSAs. The i-th FSA contains dur_i + 1 states
> and its scores has sum of all dur + S rows. Is this correct?
>
Correct except the i'th FSA actually has dur_i + 2 states, including the
final state.  But its part of the `scores` matrix
has dur_i + 1 rows.

csukuangfj:
Understood, thanks.

k2/python/k2/dense_fsa.py
csukuangfj: > Loops that access the GPU, in particular, are not
a great idea. 

Fixed. Please have a review.

k2/python/k2/dense_fsa.py
csukuangfj: Comments are updated.

k2/python/k2/dense_fsa.py
csukuangfj: C++ comments are copied to Python.

k2/csrc/compose.cu
csukuangfj: Fixed.

PR #249, opened 2020-10-13
Implemented Transpose for Ragged<T>

qindazhu:
And moved `Stack<T>` to `ragged.h`

### DIFF
--- a/k2/csrc/ragged.h
+++ b/k2/csrc/ragged.h
@@ -338,6 +338,53 @@ struct Ragged {
 template <typename T>
 std::ostream &operator<<(std::ostream &stream, const Ragged<T> &r);
 
+/*
+  Stack a list of Ragged arrays to create a Ragged array with one more axis.
+  Similar to TF/PyTorch's Stack.  The result will have Dim0 == num_srcs.  All
+  the source Ragged arrays' shapes must have the same NumAxes().
+
+     @param [in] axis   The new axis whose dimension will equal num_srcs.
+                        CAUTION: only axis == 0 is supported right now.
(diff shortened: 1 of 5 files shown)


k2/csrc/ragged.h
```diff
@@ -338,6 +338,53 @@ struct Ragged {
 template <typename T>
 std::ostream &operator<<(std::ostream &stream, const Ragged<T> &r);
 
+/*
+  Stack a list of Ragged arrays to create a Ragged array with one more axis.
```
danpovey, line 342: Thanks, can you please remove the `const` from Ragged and RaggedShape inputs?  I want to get out of the habit of using that.  Getting row-ids/row-splits is not always const.

k2/csrc/ragged.h
```diff
@@ -338,6 +338,53 @@ struct Ragged {
...
+  Similar to TF/PyTorch's Stack.  The result will have Dim0 == num_srcs.  All
+  the source Ragged arrays' shapes must have the same NumAxes().
+
+     @param [in] axis   The new axis whose dimension will equal num_srcs.
+                        CAUTION: only axis == 0 is supported right now.
```
danpovey, line 347: actually the implementation seems to support axis == 1 right now so this comment isnt right currently

qindazhu: Oh yes, we have added `Transpose` in it after the first implementation.

k2/csrc/ragged.h
qindazhu: Done in PR https://github.com/k2-fsa/k2/pull/250

PR #251, opened 2020-10-13
fix compilation and style issues

qindazhu:
All tests get passed.

### DIFF
--- a/k2/csrc/algorithms.h
+++ b/k2/csrc/algorithms.h
@@ -33,8 +33,12 @@ class Renumbering {
   int32_t NumOldElems();
   int32_t NumNewElems();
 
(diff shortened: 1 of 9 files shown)


danpovey:
Great!
I added a few functions in that PR I merged before this.. would be great if you can test them at some point.

qindazhu:
Sure, will do in the following PR.

Seems I need to test `GetBatches` and `GetDestStates` and implement `Append` for `Ragged<T>` (should be easy to call Stack when axis ==1?), anything else?

BTW, we need to implement class `Renumbering` as well.

danpovey:
That sounds right.  Thanks!!

qindazhu:
OK, will do those stuff from now. Merging this PR.

PR #253, opened 2020-10-14
Implemented Append for Ragged<T>

qindazhu:
And tested Stack for Ragged<T>

### DIFF
--- a/k2/csrc/ragged_ops_inl.h
+++ b/k2/csrc/ragged_ops_inl.h
@@ -82,7 +82,7 @@ void ApplyOpPerSublist(Ragged<T> &src, T default_value, Array1<T> *dst) {
 
 template <typename T>
 Ragged<T> Stack(int32_t axis, int32_t num_srcs, Ragged<T> **src) {
-  K2_CHECK_EQ(axis, 0);
+  K2_CHECK(axis == 0 || axis == 1);
   K2_CHECK_GT(num_srcs, 0);  // can later relax this, maybe
   std::vector<RaggedShape *> src_shapes(num_srcs);
   std::vector<const Array1<T> *> src_values(num_srcs);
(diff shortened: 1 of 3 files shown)


danpovey:
Great!!  Merging.

k2/csrc/ragged_ops_inl.h
```diff
@@ -123,16 +150,14 @@ Ragged<T> Transpose(Ragged<T> &src) {
...
   Array1<int32_t *> src_row_splits_ptr = GetRowSplitsPtr(src_shape);
   int32_t **src_row_splits_data = src_row_splits_ptr.Data();
-  int32_t num_axes = src.NumAxes();
   Array1<T> ans_values(c, src_values.Dim());
   const T *src_data = src_values.Data();
   T *ans_data = ans_values.Data();
   auto lambda_set_values = [=] __host__ __device__(int32_t new_idx) {
-    int32_t *idx_data =
-        reinterpret_cast<int32_t *>(malloc(sizeof(int32_t) * num_axes));
+    int32_t idx_data[kMaxNumAxes];
```
qindazhu, line 157: Replace dynamic malloc with a static array. For now I suppose max-num-axes <= 6 (actually the code just requires size 
 with NumAxes() - 1)

danpovey: That must have slipped past review last time, I would have objected if I saw the malloc in there.
Even having every warp doing malloc would be super slow.  Every thread is a complete no-go .

qindazhu: Yeah, that's why I update it this time. So do you think if the current approach is fine or not? Or any other suggestion?

danpovey:
Yes, current approach is fine.

PR #257, opened 2020-10-14
Add a bunch of declarations

danpovey:


danpovey:
I hope it should be reasonably clear how to implement these; none of them are really complicated.
In the forward and backward routines you'll probably want to store those scores in an array indexed by
real state_idx01 (i.e. into fsas), so after using the LogSum-based reduction over a ragged array to
compute the scores per state for one batch, you'd probably use the state-indexes for that batch
to place them in the array indexed by state_idx01 (i.e. into fsas).  So each batch/iter will probably
have (one kernel that does something per arc), then (call log-sum based [or max-based] reduction on a ragged array),
then (one kernel that places the totaled scores into the correct location indexed by idx01 into fsas).

qindazhu:
I will do this after getting `GetBatches` and `GetDestStates` fixed and tested tomorrow. But I'm just getting a little worried about if we can get the first version done before Nov. as we still add new interfaces (even if it's necessary to add them..but that may means we didn't  practically know we need those functions before implemeting some algorithms..just afraid that there are other similar interfaces needed to be add in the following days. )

danpovey:
They are parts of the algorithm that computes posteriors over a lattice.  I
am breaking it up.
Don't worry, they aren't so complicated.
Similarly, with the other things I asked for, it was generally because I
knew it would be needed for something on the critical path.

PR #258, opened 2020-10-14
Wrap TopSort and LinearFsa to Python

csukuangfj:
- [x] add test cases for vector of FSAs
- [x] wrap TopSort to Python


k2/csrc/top_sort_test.cu
```diff
@@ -0,0 +1,49 @@
...
+  /* top sorted fst is
+  0 2 1 1    // src arc 0
+  0 1 2 2    // src arc 1
+  1 2 1 21   // src arc 3
+  1 3 -1 23  // src arc 4
+  2 3 -1 13  // src arc 2
+  3
+  */
+
+  CheckArrayData(arc_map, {0, 1, 3, 4, 2});
```
qindazhu, line 46: Note that we definitely need to create test cases with random large size for those functions containing complex kernels. According to my experience, most of bugs could only be found with random large size (and empty cases)

k2/csrc/ragged_ops.h
```diff
@@ -70,7 +70,7 @@ void MinPerSublist(Ragged<T> &src, T default_value, Array1<T> *min_values) {
 template <typename T>
 void LogSumPerSublist(Ragged<T> &src, T default_value, Array1<T> *dst_values) {
   static_assert(std::is_same<float, T>::value ||
-                std::is_same<double, T>::value);
+                std::is_same<double, T>::value, "");
```
qindazhu, line 73: It's strange that my compiler does not complain this without "", but anyway, thanks! (Just thinking of that we have defined `K2_STATIC_ASSERT` in log.h, maybe we can use that.)

k2/csrc/top_sort_test.cu
csukuangfj: Yes, you're right. I will generate random acyclic FSAs to-night.

k2/csrc/ragged_ops.h
csukuangfj: Fixed.

csukuangfj:
A [Colab notebook][1] is provided to demonstrate:

- Install K2 with `pip install`
- Create, TopSort, and visualize FSAs
- Autograd of TopSort

csukuangfj:
@qindazhu  @danpovey 
Please have a review. Ready to merge.

k2/csrc/top_sort_test.cu
csukuangfj: This will be added in the next commit.

k2/csrc/ragged_test.cu
```diff
@@ -1430,53 +1391,148 @@ TEST(RaggedShapeOpsTest, TestRenumber) {
...
+                                std::vector<int32_t>{0, 3, 5, 6, 8, 11});
+    RaggedShape src =
+        RaggedShape3(&row_splits1, nullptr, -1, &row_splits2, nullptr, -1);
+
+    int32_t size_delta = 2;
+    RaggedShape dst = ChangeSublistSize(src, size_delta);
+    CheckArrayData(dst.RowSplits(2), std::vector<int32_t>{0, 5, 9, 12, 16, 21});
+
+    // it is an error to use -2 here
+    // because the state (state_idx01 == 2) has only 1 entry
```
qindazhu, line 1524: The program will crash if use `-2` here? I thought the implementation was supposed to use something like `max(0, size + delta)` to handle such case

k2/csrc/ragged_test.cu
```diff
@@ -1430,53 +1391,148 @@ TEST(RaggedShapeOpsTest, TestRenumber) {
...
+        RaggedShape3(&row_splits1, nullptr, -1, &row_splits2, nullptr, -1);
+    Ragged<int32_t> ragged(shape, col_indexes);
+    Array1<int32_t> order = GetTransposeReordering(ragged, 6);
+    CheckArrayData(
+        order, {5, 6, 9, 0, 1, 8, 10, 11, 7, 2, 3, 4, 12, 13, 14, 15, 16, 17});
     EXPECT_TRUE(context->IsCompatible(*order.Context()));
   }
 }
 
+TEST(ChangeSublistSize, TwoAxes) {
```
qindazhu, line 1486: And I think we can test `ChangeSublistSize` with a random size, e.g. `RandomRaggedShape(false, 2,4,0,1000)` and `size_delta = RandInt(0, random_shape.MaxSize(NumAxes())'.

k2/python/csrc/torch/fsa_algo.cu
```diff
@@ -0,0 +1,101 @@
...
+      },
+      py::arg("src"), py::arg("need_arc_map") = true);
+}
+
+static void PybindLinearFsa(py::module &m) {
+  // TODO(fangjun): make std::vector an opaque type for Pybind11
+  // to avoid copying data from Python to C++.
+  m.def(
+      "linear_fsa",
+      [](const std::vector<int32_t> &symbols, int32_t gpu_id = -1) -> Fsa {
```
danpovey, line 45: So what kind of Python objects will get converted to std::vector?  
I was thinking this might be a Tensor, which could be converted to Array1 with zero-copy.

k2/python/csrc/torch/fsa_algo.cu
```diff
@@ -0,0 +1,101 @@
...
+      },
+      py::arg("symbols"), py::arg("device_id") = -1,
+      R"(
+  If gpu_id is -1, the returned FSA is on CPU.
+  If gpu_id >= 0, the returned FSA is on the specified GPU.
+  )");
+
+  m.def(
+      "linear_fsa",
+      [](const std::vector<std::vector<int32_t>> &symbols,
```
danpovey, line 62: Again, what binds to std::vector<std::vector<int32_t> > ?
If we are to have conversion routines from list-of-list to ragged array, I prefer that they be centralized.
In the longer term, I  think we should create a type called RaggedTensor, which is like Ragged<T> but 
contains a Tensor whose first dim must be the NumElements() of the shape; and wrap that to Python;
and have generic methods of creating it from Python.
This may be OK for now though.

k2/python/csrc/torch/fsa_algo.cu
csukuangfj: A python list object [1, 2, 3] is **copied** to construct a `std::vector<int>` in C++.

A Python list object [ [1, 2, 3], [4, 5, 6] ] is **copied** to construct a `std::vector<std::vector<int>>` in C++.

---

Using a `torch::Tensor` as arguments involves NO memory copy.

For a list of list of integers, can we pass two 1-D tensors:
- one with the integer ids
- another with size information about the first tensor?

k2/python/csrc/torch/fsa_algo.cu
csukuangfj: > what binds to std::vector<std::vector<int32_t> >

Pybind11 will copy construct `std::vector<std::vector<int32_t> >` from a list of list of integers from Python.

csukuangfj: > In the longer term, I think we should create a type called RaggedTensor, which is like Ragged but
contains a Tensor whose first dim must be the NumElements() of the shape; and wrap that to Python;
and have generic methods of creating it from Python.

Great idea. I will implement it.

k2/csrc/ragged_test.cu
csukuangfj: I'll add tests with random sizes.

k2/csrc/ragged_test.cu
csukuangfj: The current implementation
https://github.com/k2-fsa/k2/blob/c3306ca4cdc5528f32ae85e521c412a3ee09e918/k2/csrc/ragged_ops.cu#L954

assumes that no entry will have a negative size after adding `size_delta`.

https://github.com/k2-fsa/k2/blob/c3306ca4cdc5528f32ae85e521c412a3ee09e918/k2/csrc/ragged_ops.h#L156-L157

danpovey:
OK, but if this current PR doesn't have bugs I'm OK to merge it.  We don't
need to get distracted by things that are not on the critical path right
now, we have a deadline.

csukuangfj:
All tests passed. I'm also OK to merge this.

PR #260, opened 2020-10-15
Tested GetDestStates and fixed some bugs in GetStateBatches

qindazhu:
Still need to fix bugs in `GetStateBatches` (about `Transpose` issue), just to merge now to avoid future conflicts.

### DIFF
--- a/k2/csrc/fsa_utils.cu
+++ b/k2/csrc/fsa_utils.cu
@@ -550,6 +550,7 @@ std::string FsaToString(const Fsa &fsa, bool openfst /*= false*/,
 }
 
 Array1<int32_t> GetDestStates(FsaVec &fsas, bool as_idx01) {
+  K2_CHECK(fsas.NumAxes() == 3);
   ContextPtr c = fsas.Context();
   int32_t num_arcs = fsas.NumElements();
   Array1<int32_t> ans(c, num_arcs);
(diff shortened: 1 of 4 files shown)


k2/csrc/fsa_utils_test.cu
```diff
@@ -213,4 +218,132 @@ TEST(FsaToString, Transducer) {
...
+  )";
+
+    Fsa fsa1 = FsaFromString(s1);
+    Fsa fsa2 = FsaFromString(s2);
+    Fsa fsa3 = FsaFromString(s3);
+    Fsa *fsa_array[] = {&fsa1, &fsa2, &fsa3};
+    FsaVec fsa_vec = CreateFsaVec(3, &fsa_array[0]);
+    fsa_vec = fsa_vec.To(context);
+
+    Ragged<int32_t> result = GetStateBatches(fsa_vec, false);
```
csukuangfj, line 338: Should we check the correctness of `result` here?

qindazhu: That has not be done as I said above, will do in next PR, just wants to merge to avoid conflicts.

k2/csrc/fsa_utils.cu
```diff
@@ -603,45 +604,35 @@ Ragged<int32_t> GetStateBatches(FsaVec &fsas, bool transpose) {
...
     // If this state has arcs, let its `dest_state` be the largest `dest_state`
     // of any of its arcs (which is the last one); otherwise, take the
     // `dest_state` from the 1st arc of the next state, which is the largest
     // value we can take (if the definition is: the highest-numbered state s for
     // which neither this state nor any later-numbered state has an arc to a
     // state lower than s).
-    int32_t arc_idx012 = max(arc_idx01x, next_arc_idx01x - 1);
+    int32_t arc_idx012 =
+        max(arc_idx01x,
+            next_arc_idx01x - 1);  // if this state has arcs, next_arc_idx01x -
```
csukuangfj, line 625: Does this assume that arcs of a state are sorted by its dest_state ?
The comment of this function only  says that the input fsas is topologically sorted.

qindazhu: No, we didn't assume it's sorted by dest_state. We just get the last arc index  in this state (well, if it has arcs). 
Then `arc_dest_states_data[arc_idx012]` will give the next state we should go forward (in the chain). Notice that we get lower bound of original dest_state to get `arc_dest_states`.

PR #261, opened 2020-10-15
Wrap the Intersect() code from host/, which works for generic FSAs

danpovey:


danpovey:
Guys,
This needs to be tested, and probably has bugs.  Anyway, merging because I don't *think*
it breaks anything.  Compiles at least.

csukuangfj:
I am adding tests for it.

danpovey:
Great!

PR #267, opened 2020-10-17
Fixes/updates to Renumbering

danpovey:


danpovey:
The behavior of Renumbering wasn't quite what I had in mind.  It was OK but a little harder to implement than what I had planned, and also not correct the way I was using it, because I was using things like renumbering.Old2New().Data(), which wasn't right since it was returning a local variable whose Data() would get deleted by that.

PR #271, opened 2020-10-20
Wrap ArcSort to Python.

csukuangfj:
Ready for review.

### DIFF
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -248,10 +248,14 @@ void ArcSort(Fsa *fsa) {
 }
 
 void ArcSort(Fsa &src, Fsa *dest, Array1<int32_t> *arc_map /*= nullptr*/) {
-  K2_CHECK_GE(src.NumAxes(), 2) << "Input FSA is empty!";
-  *dest = src.Clone();
-  if (arc_map) *arc_map = Array1<int32_t>(src.Context(), src.values.Dim());
-  SortSublists<Arc, ArcComparer>(dest, arc_map);
+  if (!src.values.IsValid()) return;
+
+  if (arc_map != nullptr)
+    *arc_map = Array1<int32_t>(src.Context(), src.NumElements());
+
+  Fsa tmp(src.shape, src.values.Clone());
+  SortSublists<Arc, ArcComparer>(&tmp, arc_map);
+  *dest = tmp;
 }
 
 }  // namespace k2
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -247,4 +247,11 @@ void ArcSort(Fsa *fsa) {
   SortSublists<Arc, ArcComparer>(fsa);
 }
 
+void ArcSort(Fsa &src, Fsa *dest, Array1<int32_t> *arc_map /*= nullptr*/) {
+  K2_CHECK_GE(src.NumAxes(), 2) << "Input FSA is empty!";
+  *dest = src.Clone();
+  if (arc_map) *arc_map = Array1<int32_t>(src.Context(), src.values.Dim());
+  SortSublists<Arc, ArcComparer>(dest, arc_map);
+}
+
 }  // namespace k2
--- a/k2/csrc/ragged.cu
+++ b/k2/csrc/ragged.cu
@@ -46,14 +46,6 @@ struct iterator_traits<::RowSplitsDiff> {
 
 namespace k2 {
 
-RaggedShapeDim RaggedShapeDim::Clone() const {
-  RaggedShapeDim ans;
-  ans.row_splits = row_splits.Clone();
-  ans.row_ids = row_ids.Clone();
-  ans.cached_tot_size = cached_tot_size;
-  return ans;
-}
-
 // Recursive function that prints (part of) a ragged shape.
 // 0 <=  begin_pos <= end_pos < shape.TotSize(axis).
 
@@ -211,16 +203,6 @@ RaggedShape RaggedShape::To(ContextPtr ctx) const {
     axes[i - 1].cached_tot_size = -1;
   }
   return RaggedShape(axes);
-}
-
-RaggedShape RaggedShape::Clone() const {
-  RaggedShape ans;
-  ans.axes_.reserve(axes_.size());
-  const_cast<RaggedShape *>(this)->Populate();
-  for (const auto &axis : axes_) {
-    ans.axes_.emplace_back(axis.Clone());
-  }
-  return ans;
 }
 
 RaggedShapeIndexIterator RaggedShape::Iterator() {
--- a/k2/csrc/ragged.cu
+++ b/k2/csrc/ragged.cu
@@ -46,6 +46,13 @@ struct iterator_traits<::RowSplitsDiff> {
 
 namespace k2 {
 
+RaggedShapeDim RaggedShapeDim::Clone() const {
+  RaggedShapeDim ans;
+  ans.row_splits = row_splits.Clone();
+  ans.row_ids = row_ids.Clone();
+  ans.cached_tot_size = cached_tot_size;
+  return ans;
+}
 
 // Recursive function that prints (part of) a ragged shape.
 // 0 <=  begin_pos <= end_pos < shape.TotSize(axis).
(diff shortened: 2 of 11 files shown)


danpovey:
Fantastic!  Will look to-morrow.

k2/csrc/fsa_algo.cu
```diff
@@ -247,4 +247,11 @@ void ArcSort(Fsa *fsa) {
   SortSublists<Arc, ArcComparer>(fsa);
 }
 
+void ArcSort(Fsa &src, Fsa *dest, Array1<int32_t> *arc_map /*= nullptr*/) {
+  K2_CHECK_GE(src.NumAxes(), 2) << "Input FSA is empty!";
```
qindazhu, line 251: Empty fsa is with NumAxes ==2 and Dim0 = 0, this algorthims should work correctly for empty FSA (and seems your implementation works well for empty FSA). I think we should update the log info here to `Invalid FSA`, or just remove the text.

k2/csrc/ragged.cu
```diff
@@ -208,11 +213,20 @@ RaggedShape RaggedShape::To(ContextPtr ctx) const {
   return RaggedShape(axes);
 }
 
+RaggedShape RaggedShape::Clone() const {
+  RaggedShape ans;
+  ans.axes_.reserve(axes_.size());
+  const_cast<RaggedShape *>(this)->Populate();
```
qindazhu, line 219: If src has no valid row_ids, maybe it's better to let dest has no valid row_ids either? User can populate dest if they want.

k2/csrc/fsa_algo.cu
csukuangfj: `Array1<T>::Clone` will crash for empty arrays in that their region is NULL and `Context()` will segfault.

If the input FSA is empty, what will we put in `dest` and `arc_map`?

k2/csrc/ragged.cu
```diff
@@ -46,6 +46,13 @@ struct iterator_traits<::RowSplitsDiff> {
 
 namespace k2 {
 
+RaggedShapeDim RaggedShapeDim::Clone() const {
+  RaggedShapeDim ans;
+  ans.row_splits = row_splits.Clone();
+  ans.row_ids = row_ids.Clone();
```
qindazhu, line 52: not should if `Array.Clone` checks itself valid or not. It's better if we can just copy row_splits or row_ids when they are valid.

k2/csrc/fsa_algo.cu
```diff
@@ -247,4 +247,11 @@ void ArcSort(Fsa *fsa) {
...
+  *dest = src.Clone();
+  if (arc_map) *arc_map = Array1<int32_t>(src.Context(), src.values.Dim());
```
qindazhu, line 253: src.values.Dim() -> src.NumElements?

qindazhu, line 253: use `if(arc_map != nullptr)` as we have discussed the style in if statement?

k2/csrc/ragged_test.cu
```diff
@@ -1872,4 +1872,72 @@ TEST(RaggedShapeOpsTest, TestMakeTransposable) {
...
+    Array1<int32_t> values(context, std::vector<int32_t>{10, 20});
+    Ragged<int32_t> ragged(shape, values);
+
+    Ragged<int32_t> copied = ragged.Clone();
+    Increment(&ragged.RowSplits(1));
+    Increment(&ragged.RowIds(1));
+    Increment(&ragged.values);
+
+    for (int32_t i = 0; i != ragged.RowSplits(1).Dim(); ++i)
+      EXPECT_EQ(copied.RowSplits(1)[i] + 1, ragged.RowSplits(1)[i]);
```
qindazhu, line 1933: it's OK to user `operator[]` on a small array, but it's better if we can notice it as it may call multiple device-to-host-copy in a loop. It's relatively OK here as the array is not so long, but we definitely need to convert to cpu first for a long array.

csukuangfj: Yes, you're right. This is only for unittests.

k2/csrc/fsa_algo.cu
qindazhu: Empty array is different from Invalid array, you can have a check of method `IsValid` in Array1.

> If the input FSA is empty, what will we put in dest and arc_map
Just leave them empty as well. I think your implementation of SortSublists work well for empty cases.

k2/csrc/ragged_test.cu
qindazhu: I mean, even for unintests we should try to avoid copy too many times (RE a long array). Anyway, it's OK here as this array just has a small size.

csukuangfj:
Will fix all comments this night.

k2/csrc/fsa_algo.cu
danpovey: It's possible for arrays to have zero elements but have a non-NULL region.  You can allocate a region of zero bytes, I believe.
Instead of saying "is empty", you can just say "has X axes" or change it to K2_CHECK_GE(x.NumAxes(), 2);

We don't need to waste too many cycles thinking about this though.  Although I try to make these things, where possible, not crash for empty input, in many cases it won't work for this or that reason.  And I think expecting things to work when we provide empty inputs from Python is maybe unreasonable, and in any case not a short-term priority.

k2/csrc/ragged.cu
danpovey: Agreed, it's not necessary to populate here, I think.
(Although you could make an argument that it might save some duplicate work later).

k2/csrc/ragged.cu
danpovey: Looks like Clone() will crash for non-valid input.  How about do: 
`if (row_ids.IsValid()) ans.row_ids = row_ids.Clone();`
I avoided making the Context() call of Array1 check whether the region is non-NULL, because it's called so often (for speed reasons).  Most calls implicity assume the arg is valid.

k2/csrc/fsa_algo.cu
danpovey: We shouldn't do duplicate or unnecessary work.  Axes whose row_splits and row_ids are unchanged should be copied, not cloned; and we shouldn't copy arrays twice.  (The function that sorts the arc should accept input-array != output-array).

danpovey: .. actually, in this case the output has the same shape as the input; we don't need to do Clone() at all.
There is an implicit assumption that row-ids and row-splits won't be modified in-place, so we freely share them between objects.

csukuangfj: > We shouldn't do duplicate or unnecessary work. Axes whose row_splits and row_ids are unchanged should be copied

Sharing the underlying `shape` memory of two FSAs has the risk that changing one FSA will affect another.

---


> The function that sorts the arc should accept input-array != output-array)

Agreed. We should check that `&src  != dest`.

k2/csrc/ragged.cu
csukuangfj: Agreed.

danpovey: Actually I think we shouldn't give RaggedShape a Clone() function at all, because I can't see where it would be
necessary.  If it's ever necessary to modify a RaggedShape in-place (which might invalidate other RaggedShapes it shared arrays with), then just the members that need to be copied can be cloned.

k2/csrc/fsa_algo.cu
danpovey: What I meant is the function that sorts the arcs should accept both input-array == output-array and input-array != output-array.  This is how sorts generally work.

danpovey: Regarding: `Sharing the underlying shape memory of two FSAs has the risk that changing one FSA will affect another.`... there is intended to be a convention that memory used in RaggedShape is `read-only` (unless you just created it).  I.e. you should generally not modify the memory of a RaggedShape unless you were specifically asked to.

danpovey: Oh I see now what you meant.  
I think it's not necessary to require that &src != dest in ArcSort() and similar functions.
The way we generally do this is do all the work internally without modifying *dest;
we create an Fsa (or its components) internally,
and then right at the end, do something like
` *dest = ans;`
or
`*dest = Ragged<Arc>(shape, arcs);`
(which just copies the metadata).  So actually it won't matter if dest == &src.

k2/csrc/ragged.cu
csukuangfj: Will remove `RaggedShape::Clone()` and `RaggedShapeDim::Clone()`.

k2/csrc/fsa_algo.cu
csukuangfj: I see.

csukuangfj:
Updated. @qindazhu  @danpovey  Please have a review.

qindazhu:
LGTM, +1.

danpovey:
Great, thanks!

PR #272, opened 2020-10-21
implemented GetForwardScores and fixed a bug in GetStatesBatches

qindazhu:


k2/csrc/fsa_utils_inl.h
```diff
@@ -0,0 +1,238 @@
...
+        RaggedShape2(&sum_sub_range, nullptr, sub_scores_values.Dim());
+    Ragged<FloatType> sub_scores(sub_scores_shape, sub_scores_values);
+    // we always use the first num_rows elements in score_cache.
+    Array1<FloatType> sub_state_scores = score_cache.Range(0, num_rows);
+    // get scores per state in this batch
+    MaxPerSublist(sub_scores, negative_infinity, &sub_state_scores);
+    // TODO(haowen): fix compilation issu of kMinLogDiffFloat/Double in utils.h
+    /*
+  if (log_semiring)
+    LogSumPerSublist(sub_scores, negative_infinity, &sub_state_scores);
```
qindazhu, line 184: There is a compilation failure here as `kMinLogDiffFloat` is a host value, but we want to access it in kernel. Which approach do you prefer to fix this issue?  just use `#define kMinLogDiffFloat the_corresponding_value`, or with different definition for host and device (e.g, for device, use `__device__` or `__constant__` and initalized in a function which called only once when the program starts?)
https://github.com/k2-fsa/k2/blob/c2c26ab54ae5c7fba08736cd78f9fe9874c67f86/k2/csrc/utils.h#L583
https://github.com/k2-fsa/k2/blob/c2c26ab54ae5c7fba08736cd78f9fe9874c67f86/k2/csrc/utils.h#L612-L633

csukuangfj: > Which approach do you prefer to fix this issue?

Neither.

The approach that I would propose is to remove `static`.

```cpp
const float kMinLogDiffFloat = logf(FLT_EPSILON);   // negative! 
```

qindazhu: Sorry, don't understand it, you mean removing `static` then kernel can access const variable of host directly?

csukuangfj: Looks like my approach does not work.

qindazhu: Yes, I guess so, that's why I said I don't understand it...

csukuangfj: Can we borrow the implementation from GTN which does not require `kMinLogDiffFloat`:

https://github.com/facebookresearch/gtn/blob/a810690546df258b6df157298ac68bf1178d1439/gtn/functions/shortest.cpp#L20

```cpp
constexpr float kInf = std::numeric_limits<float>::infinity();
constexpr float kNegInf = -std::numeric_limits<float>::infinity();

inline float logadd(float a, float b) {
  if (a == kNegInf) {
    return b;
  }
  if (b == kNegInf) {
    return a;
  }
  return std::max(a, b) + std::log1p(std::exp(-std::abs(a - b)));
}
```

We don't need `kNegInf`. CUDA provides `isinf`:
https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__SINGLE.html#group__CUDA__MATH__SINGLE_1g0a62e45f335a23ee64ecad3fb87a72e3

> __device__​ __RETURN_TYPE isinf ( float  a )

There is also a function `isinf()` on the host side.

qindazhu: I think it's fine, let's wait for @danpovey 's suggestion.

k2/csrc/fsa_utils_inl.h
```diff
@@ -0,0 +1,238 @@
...
+#include <utility>
+#include <vector>
+
+#include "k2/csrc/array.h"
+#include "k2/csrc/fsa.h"
+#include "k2/csrc/ragged.h"
+
+namespace k2 {
+template <typename FloatType>
+Array1<FloatType> GetForwardScores(FsaVec &fsas, Ragged<int32_t> &state_batches,
```
qindazhu, line 34: BTW, should we returen Array1<double> for those forward/backward scores? no matter `FloatType` is `float` or `double`

k2/csrc/fsa_utils_inl.h
danpovey: I prefer #define, you may have to figure out the numerical value.
Comparisons with infinity do not work when you add -Ofast: they are turned off, so that's a bit dangerous

k2/csrc/fsa_utils_inl.h
danpovey: Dunno what you are asking, but yes, return array1.  The function will need to be called with explicit template argument.

danpovey: If implementing the template is a hassle, do just double for now.

qindazhu: Sorry, I mean return `Array1<double>`. Anyway, will leave as it is as we talked offline.

k2/csrc/fsa_utils_inl.h
qindazhu: Ok, will do `#define` in the next PR.

qindazhu:
Ready to merge if there's no other issues, will fix the issue above (with `#define`) in the next PR.

qindazhu:
Merge, will go forward based on this. You could leave your comments and I can fix them in the next PR.

PR #273, opened 2020-10-21
Test TopSort with random generated FSAs.

csukuangfj:
It seems that there are bugs in TopSort.

The unittest fails most of the time for random FSAs.  Test logs are given below:

```
/root/k2/build$ ./bin/cu_top_sort_test --gtest_filter="*Random*Of*"
Running main() from /root/k2/build/_deps/googletest-src/googletest/src/gtest_main.cc
Note: Google Test filter = *Random*Of*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from TopSort
[ RUN      ] TopSort.RandomVectorOfFsas
[F] /root/k2/k2/csrc/ragged_ops.cu:Transpose:776 Check failed: src_tot_size1 % src_dim0 == 0 (5 vs. 0) Transpose(): all dims on axis 0 must be the same.
src_tot_size1: 74
src_dim0: 23

Aborted (core dumped)
/root/k2/build$ ./bin/cu_top_sort_test --gtest_filter="*Random*Of*"
Running main() from /root/k2/build/_deps/googletest-src/googletest/src/gtest_main.cc
Note: Google Test filter = *Random*Of*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from TopSort
[ RUN      ] TopSort.RandomVectorOfFsas
[F] /root/k2/k2/csrc/ragged_ops.cu:Transpose:776 Check failed: src_tot_size1 % src_dim0 == 0 (3 vs. 0) Transpose(): all dims on axis 0 must be the same.
src_tot_size1: 63
src_dim0: 20

Aborted (core dumped)
/root/k2/build$ ./bin/cu_top_sort_test --gtest_filter="*Random*Of*"
Running main() from /root/k2/build/_deps/googletest-src/googletest/src/gtest_main.cc
Note: Google Test filter = *Random*Of*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from TopSort
[ RUN      ] TopSort.RandomVectorOfFsas
[       OK ] TopSort.RandomVectorOfFsas (29884 ms)
[----------] 1 test from TopSort (29884 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (29884 ms total)
[  PASSED  ] 1 test.
```

PR #274, opened 2020-10-22
TopSort fixes (on top of #273)

danpovey:


k2/csrc/top_sort_test.cu
```diff
@@ -106,11 +106,91 @@ TEST(TopSort, VectorOfFsas) {
...
+    FsaVec fsa_vec = CreateFsaVec(num_fsas, &fsa_array[0]);
+    fsa_vec = fsa_vec.To(context);
+
+    int32_t gt = kFsaPropertiesTopSorted | kFsaPropertiesTopSortedAndAcyclic;
+    Array1<int32_t> properties;
+    int32_t p;
+    GetFsaVecBasicProperties(fsa_vec, &properties, &p);
+
+    /*
+    EXPECT_NE(p & gt, gt);
```
danpovey, line 168: Fangjun, was removing these lines OK?  I assume we can't guarantee these aren't top sorted.

danpovey:
Anyway I'll merge this, we'll fix any issues later.

k2/csrc/top_sort_test.cu
csukuangfj: Yes, it's fine. Will delete the commented code in the next commit.

PR #275, opened 2020-10-22
Implemented GetBackwardScores and GetTotScores

qindazhu:


k2/csrc/fsa_utils_inl.h
```diff
@@ -203,26 +199,228 @@ Array1<FloatType> GetForwardScores(FsaVec &fsas, Ragged<int32_t> &state_batches,
...
+          int32_t start_state = fsa_row_splits1[fsa_idx],
+                  start_state_next_fsa = fsa_row_splits1[fsa_idx + 1];
+          if (start_state_next_fsa - start_state > 0)
+            state_scores_data[start_state_next_fsa - 1] = 0;
+        };
+    Eval(c, num_fsas, lambda_set_final_state_score);
+  }
+
+  // compute leaving arc numbers in each batch
+  Array1<int32_t> leaving_arc_nums(c, num_batches);
```
danpovey, line 257: I'd prefer to call this num_arcs instead of arc_nums , for consistency with other code.

qindazhu: I used `num_arcs` above for the number of arcs in the `fsas`, will use `num_arcs_leaving`

k2/csrc/fsa_utils_inl.h
```diff
@@ -203,26 +199,228 @@ Array1<FloatType> GetForwardScores(FsaVec &fsas, Ragged<int32_t> &state_batches,
...
+  Array1<int32_t> leaving_arc_start_index(c, num_batches);
+  int32_t *leaving_arc_nums_data = leaving_arc_nums.Data();
+  int32_t *leaving_arc_start_index_data = leaving_arc_start_index.Data();
+  const int32_t *fsa_in_batch_row_splits =
```
danpovey, line 261: This could would be, IMO, easier to understand if you were more rigorous about using the naming convention.  E.g. how about using names arc_batches_row_splits{1,2,3}, and batch_idx0, this_state_idx0x, next_state_idx0x, this_arc_idx0xx, next_arc_idx0xx?  I think that would make the code more understandable.

k2/csrc/fsa_utils.h
```diff
@@ -219,16 +219,17 @@ Array1<FloatType> GetForwardScores(FsaVec &fsas, Ragged<int32_t> &state_batches,
 /*
   Return array of total scores (one per FSA), e.g. could be interpreted as
   the data probability or partition function.
-         @param [in] fsas   Input Fsa or FsaVec (must have 2 or 3 axes)
+         @param [in] fsas   Input FsaVec (must have 3 axes)
```
csukuangfj, line 222: Can we build an FsaVec from the input FSA and do a recursive call?

k2/csrc/fsa_utils_inl.h
qindazhu: OK, will do.

k2/csrc/fsa_utils_inl.h
```diff
@@ -203,26 +199,228 @@ Array1<FloatType> GetForwardScores(FsaVec &fsas, Ragged<int32_t> &state_batches,
...
+  const int32_t *states_data = state_batches.values.Data();
+  const Arc *arcs = fsas.values.Data();
+  Array1<FloatType> leaving_arc_score_values(
+      c, num_arcs);  // leaving arc_scores in batches
+  FloatType *arc_scores_data = leaving_arc_score_values.Data();
+  // convert leaving_arc_nums to cpu as we will access its elements in below
+  // Eval function for `lambda_set_leaving_arc_scores`
+  leaving_arc_nums = leaving_arc_nums.To(GetCpuContext());
+  const int32_t *cpu_leaving_arc_nums = leaving_arc_nums.Data();
+  Array1<int32_t> cpu_leaving_arc_start_index =
```
danpovey, line 296: since you already have this leaving_arc_start_index, it's not clear to me that you need leaving_arc_nums (could just take the difference in this).  Up to you though, no need to change.

k2/csrc/fsa_utils_inl.h
```diff
@@ -203,26 +199,228 @@ Array1<FloatType> GetForwardScores(FsaVec &fsas, Ragged<int32_t> &state_batches,
...
+      leaving_arc_start_index.To(GetCpuContext());
+  const int32_t *cpu_leaving_arc_start = cpu_leaving_arc_start_index.Data();
+  // copy row_splits2 to cpu
+  Array1<int32_t> cpu_row_splits2 =
```
danpovey, line 300: I think you are transferring too much to CPU here.  how about `leaving_arc_batches.RowSplits(2)[leaving_arc_batches.RowSplits(1)].To(GetCpuContext())` ?

k2/csrc/fsa_utils_inl.h
```diff
@@ -203,26 +199,228 @@ Array1<FloatType> GetForwardScores(FsaVec &fsas, Ragged<int32_t> &state_batches,
...
+      leaving_arc_batches.RowSplits(2).To(GetCpuContext());
+  const int32_t *cpu_row_splits2_data = cpu_row_splits2.Data();
+  Array1<int32_t> arc_sum_row_splits(c, num_states + 1);
+  Array1<FloatType> score_cache(c, num_states + 1);
+  // process batch sequentially.
+  for (int32_t i = num_batches - 1; i >= 0; --i) {
+    // get the range we would call Max/LogSum per sub list
+    int32_t row_begin =
+        cpu_row_splits2_data[i * num_fsas];  // row_start is idx01x
```
danpovey, line 309: is this even different from cpu_leaving_arc_start_index?

danpovey: ... and row_end isn't a great name, I'd prefer something like this_leaving_arc_idx0xx and next_leaving_arc_idx0xx, assuming that's acurate.

qindazhu: `cpu_row_splits2` is for state idx in batch (as I noted, row_begin is `idx01x`). `leaving_arc_start_index` is for arc idx in batch. they are on different levels.

qindazhu: OK, will change the name.

k2/csrc/fsa_utils.h
qindazhu: yeah, we can, but we don't need that version for now (If we do, we should do the same thing to GetForwardScores as well.)

k2/csrc/fsa_utils_inl.h
danpovey: But you seem to be only accessing leaving_arc_start_index for the 1st state of each FSA, and the other elements aren't used.  FSA<-->CPU transfer is expensive!  I think this is the same information that is in cpu_leaving_arc_start[i].

k2/csrc/fsa_utils_inl.h
```diff
@@ -203,26 +199,228 @@ Array1<FloatType> GetForwardScores(FsaVec &fsas, Ragged<int32_t> &state_batches,
...
+    {
+      ParallelRunner pr(c);
+      // get leaving arc scores
+      {
+        cudaStream_t stream = pr.NewStream();
+        With w(stream);
+        auto lambda_set_leaving_arc_score = [=] __host__ __device__(
+                                                int32_t idx3) {
+          // all idx** in below code are the indexes to leaving_arc_batches
+          int32_t idx0123 = leaving_arc_start_index_data[i] + idx3;
```
qindazhu, line 328: @danpovey I used `leaving_arc_start_index_data` here

k2/csrc/fsa_utils_inl.h
danpovey: Sorry I meant you seem to be only accessing      cpu_row_splits2_data for the 1st state of each FSA.

qindazhu: > Sorry I meant you seem to be only accessing cpu_row_splits2_data for the 1st state of each FSA.

I guess you mean `1st state of each batch`. 
 Yeah, that's true, so you mean I should just copy the 1st state of each batch to a cpu array?

danpovey: You could do that (as a single array obviously), but I think the information you need is exactly the info in leaving_arc_start_index, so you can just use that, it's already on CPU.

k2/csrc/fsa_utils_inl.h
qindazhu: Removed leaving_arc_nums. Get the value from leaving_arc_start_index

k2/csrc/fsa_utils_inl.h
qindazhu: Updated

k2/csrc/fsa_utils_inl.h
qindazhu: Removed leaving_arc_nums

k2/csrc/fsa_utils_inl.h
qindazhu: Updated as you said

k2/csrc/fsa_utils_inl.h
qindazhu: Updated with RowSplits(2)[RowSplits(1)].To(cpu)

qindazhu:
@danpovey Fixed teh name convention issue and memory copy issues, pleave have a review, thanks!

qindazhu:
Also updated the same issues in `GetForwardScores`

qindazhu:
Merging as you have talked offline.

PR #280, opened 2020-10-24
Wrap GetForwardWeights and GetBackwardWeights in host for test purpose

qindazhu:
Will use those functions to test `GetForwardScores` and `GetBackwardScores` in `fsa_utils.h` with random large size. May delete those functions finally.


k2/csrc/host_shim_inl.h
```diff
@@ -0,0 +1,131 @@
...
+// convert Array1<double> to Array1<float> (then return it) or return itself.
+// May delete this finally as we just use it for current test purpose.
+template <typename FloatType>
+Array1<FloatType> ConvertArrayType(const Array1<double> &src);
+template <>
+inline Array1<double> ConvertArrayType<double>(const Array1<double> &src) {
+  K2_CHECK_EQ(src.Context()->GetDeviceType(), kCpu);
+  return src;
+}
+template <>
```
danpovey, line 46: seems to me this could go in array_ops{,_inl}.h.

k2/csrc/host_shim_inl.h
```diff
@@ -0,0 +1,131 @@
...
+    if (log_semiring) {
+      k2host::ComputeForwardLogSumWeights(host_fsa, this_fsa_state_scores_data);
+    } else {
+      k2host::ComputeForwardMaxWeights(host_fsa, this_fsa_state_scores_data);
+    }
+  }
+  return internal::ConvertArrayType<FloatType>(state_scores);
+}
+
+template <typename FloatType>
```
danpovey, line 86: Not important, but if it were me I would have put this in the .cu file and just instantiated it for float and double, as (IIRC):
```
template
Array1<float> GetBackwardScores(
    FsaVec &fsas, const Array1<float *tot_scores, bool log_semiring);
```

k2/csrc/host_shim_inl.h
qindazhu: is this really needed by other functions? I was thinking it's only used here so i put it here in a internal namespace.

k2/csrc/host_shim_inl.h
qindazhu: ok, will do the same thing for those score computation functions in fsa_utils.h

k2/csrc/host_shim_inl.h
danpovey: Actually if we were to implement this I think it would be better to have it a function Astype(), in class Array.
But not necessary right now; just for the future.

qindazhu: Added AsType in Array1

k2/csrc/host_shim_inl.h
qindazhu: Fixed.

k2/csrc/array_inl.h
```diff
@@ -49,6 +49,37 @@ void Array1<T>::CopyFrom(const Array1<T> &src) {
...
+    ans_data[i] = static_cast<S>(this_data[i]);
+  };
+  Eval(Context(), Dim(), lambda_set_values);
+  return ans;
+}
+
+// if S is same with T, just return *this. But in C++, we cannot explicitly
+// specialize a class member template unless its enclosing class templates are
+// also explicitly specialized. Thus here we specialized Array1 as well.
+template <>
```
danpovey, line 68: This is unnecessary, you can just do
```
  if (std::is_same<S, T>) return *this;
```
at the top.

qindazhu: mm. we can't do this，`if` is runtime code, not compiling time. If T is double and S is float, You are trying to return double array for a function returns float array. noted the code  `if (true) return *this` and below codes always get into the generated code

danpovey: you could try: return *(reinterpret_cast<Array1<S> >(this));

danpovey: Also, you should be able to do the template specialization as 
`template <typename S> void Array1<S>::AsType<S>();`
but I think the if thing is maybe clearer, and should be optimized to the same thing.

qindazhu: OK, use cast here. Noted we cannot specialize, that's exactly what I commented before:
https://github.com/k2-fsa/k2/blob/d5e8626c51d7f21d55cc3c0e1516efdf832c6a99/k2/csrc/array_inl.h#L65-L67

qindazhu:
Ready to merge.

danpovey:
Thanks!  You can merge.

k2/csrc/array_inl.h
csukuangfj: Actually, you can use SIFNA

```cpp
  template <typename S>
  Array1<typename std::enable_if<!std::is_same<S, T>::value, S>::type>
  AsType() const {
    //  S != T
    // .... 
  }
  template <typename S>
  Array1<typename std::enable_if<std::is_same<S, T>::value, S>::type> AsType() const {
    // S == T
    return *this;
  }
```
No cast is needed.

csukuangfj: The cast results in the following warning:

```
k2/k2/csrc/array_inl.h(56): warning: dynamic initialization in unreachable code
          detected during:
            instantiation of "k2::Array1<S> k2::Array1<T>::AsType<S>() [with T=double, S=double]"
```

qindazhu: Can you paste the code your are compiling with? It's strange that I got no such warning with T=double and S=double.

BTW, seems I missed your last comments. I'm OK with using SFINA as long as Dan is fine with this.

csukuangfj: I am using the default settings and compiling against the master branch.

Maybe the warning depends on the compiler we are using.

qindazhu: Here is my local version: `gcc 5.5.0` `nvcc 10.1`. Anyway, I'm OK with either SFINA or just supressing the warning (as we know it's fine to supress it). It would be good if @danpovey can give his comment.

danpovey: I'm OK with either.

qindazhu: Updated to use SFINA in PR https://github.com/k2-fsa/k2/pull/286/files

Issue #281, opened 2020-10-24
Implement `compose`


csukuangfj:
I'd like to add a linear-chain CRF example for k2, like the one in GTN <https://github.com/facebookresearch/gtn/blob/master/examples/notebooks/linear_crf.ipynb>

But we don't have `compose` yet. I would like to add `csrc/host/compose.cc` borrowing code from `csrc/host/intersect.cc`.

danpovey:
We don't need special code for composition.  It can be done with
composition, plus the arc_map info it outputs, which can be used to
map the aux_labels.

danpovey:
I mean it can be done with intersection.

csukuangfj:
To `compose(a_fsa, b_fsa)`, we can do
- `a_fsa_projected = project_output(a_fsa, a_aux_labels)`, now `a_fsa_projected` is an acceptor
- `out, arc_map_a, arc_map_b = intersect(a_fsa_projected, b_fsa)`
- use `arc_map_a`, `arc_map_b`, `input labels` of `a_fsa` to set the `input labels` and `aux_labels` of `out`

Is the above process for computing `compose` via `intersect` correct ?

qindazhu:
Yeah. Actually in k2 you can view aux_labels as either ilabels or olabels (the one which is not participating in the operation you are doing is olable. )

RE `project`, which is called `Invert` in k2, you can check the code in https://github.com/k2-fsa/k2/blob/master/k2/csrc/host/aux_labels.h for `Invert` and `AuxMapper`

danpovey:
I don't think you need that many steps.  It's easier to not talk about
composition at all, but just talk about intersection, where the aux_labels
are naturally propagated (in fact, they are treated like any other quantity
that is indexed by arc).

So you just intersect A and B, and when the intersection happens, all
per-arc quantities in A and B are copied to the output via the arc_map.
There is no concept of input labels and output labels, just labels.
Suppose B is L.fst, we can have the labels on the arcs be the words, and
the aux_labels be the phones.  (If aux_labels were a ragged tensor we could
have multiple phones per arc, but for now just assume it's a 1-d tensor
with integers on it).  All the quantities in A and B are copied.  If the
same labels exists in both A and B, if it's a float we add; if it's an
integer, for now we just die if both have it.  In our specific case, if B
is L.fst, the aux_labels are the phones, and they are just copied to the
composed result (A.B).  We can then do Invert() to swap the aux_labels and
the regular labels on the composed result.

PR #282, opened 2020-10-25
Wrap properties and shortest path to Python

csukuangfj:
An example of linear-chain CRF is added based on this pullrequest.

The example follows https://github.com/facebookresearch/gtn and is available at
https://colab.research.google.com/drive/1NDNcEK39JYWGRNOJMCtUz1Z2skOS2var?usp=sharing

~~I will remove the code about `compose` and use `intersect` based on the comment
from https://github.com/k2-fsa/k2/issues/281~~


k2/csrc/fsa_algo.h
```diff
@@ -158,6 +163,21 @@ Fsa LinearFsa(const Array1<int32_t> &symbols);
  */
 FsaVec LinearFsas(Ragged<int32_t> &symbols);
 
+/* Compute the shortest distance in the tropical semiring, but
+   with the opposite sign. That is, it uses `max` instead of `min.
+
+   @param [in] fsa           The input FSA. It is on CPU and must be
```
danpovey, line 169: OK, but in general supporting FsaVec is as important, or more important, than supporting Fsa.

k2/csrc/fsa_algo.cu
```diff
@@ -258,4 +302,50 @@ void ArcSort(Fsa &src, Fsa *dest, Array1<int32_t> *arc_map /*= nullptr*/) {
...
+  K2_CHECK_EQ(fsa.NumAxes(), 2);
+  K2_CHECK_EQ(context->GetDeviceType(), kCpu);
+  int32_t num_states = fsa.Dim0();
+  k2host::Fsa host_fsa = FsaToHostFsa(fsa);
+  Array1<double> state_weights(context, num_states);
+  std::vector<int32_t> tmp_arc_map;
+  ComputeForwardMaxWeights(host_fsa, state_weights.Data(), &tmp_arc_map);
+  if (arc_map != nullptr) *arc_map = Array1<int32_t>(context, tmp_arc_map);
+
+  return state_weights.Back();
```
qindazhu, line 316: Calling `Back` on an empty vector will cause undefined behavior.

csukuangfj: I think the behavior is well defined

https://github.com/k2-fsa/k2/blob/4b50f519e0e6a7de5214b52a32741c1f8cbbdb8c/k2/csrc/array.h#L243-L246

k2/csrc/fsa_algo.h
```diff
@@ -158,6 +163,21 @@ Fsa LinearFsa(const Array1<int32_t> &symbols);
...
+                             topologically sorted.
+   @param [out] arc_map      The arc indexes of the best path from the start
+                             state to the final state. It is empty if there
+                             is no such path.
+   @return The shortest distance from the start state to the final state.
+           Returns negative infinity if there is no such path.
+ */
+double ShortestDistance(Fsa &fsa, Array1<int32_t> *arc_map = nullptr);
+
+void ShortestPath(Fsa &src, Fsa *out, Array1<int32_t> *arc_map = nullptr);
```
qindazhu, line 179: For `ShortestPath`, I see you create a new FSA with those arcs on the best path. It's better to add documentation here as it is different from ShortestDistance.

qindazhu, line 179: BTW, I think we may not need this version `ShortestPath`, `ShortestDistance` with `arc_map` is good enough. @danpovey what do you think?

k2/csrc/fsa_algo.h
```diff
@@ -158,6 +163,21 @@ Fsa LinearFsa(const Array1<int32_t> &symbols);
...
+
+   @param [in] fsa           The input FSA. It is on CPU and must be
+                             topologically sorted.
+   @param [out] arc_map      The arc indexes of the best path from the start
+                             state to the final state. It is empty if there
+                             is no such path.
+   @return The shortest distance from the start state to the final state.
+           Returns negative infinity if there is no such path.
+ */
+double ShortestDistance(Fsa &fsa, Array1<int32_t> *arc_map = nullptr);
```
qindazhu, line 177: Better to add documentation to say that we only get one of those best paths (if there are multiple best paths)

k2/csrc/fsa_algo.cu
qindazhu: We should not die here, we should return negative_infinity. Or we should at least add documentation to say what will happen when it's an empty FSA. (For empty FSA, I believe we should not die)

k2/csrc/fsa_algo.h
danpovey: I think this function is useful enough to keep, since there isn't, I believe, a simple way to create an FSA from an FSA and an arc_map.
Incidentally, in ShortestDistance, I don't think we should call the list of arcs `arc_map` as it doen't really map from anything specific.  I think maybe `best_path_arcs` would make more sense.  In fact, the 2 functions could be merged into ShortestPath and the fsa output made optional.

Eventually we can add an efficient implementation for FsaVec on GPU.  Its core will be similar to GetForwardScores, except it will keep track of the best traceback.

qindazhu: > I think this function is useful enough to keep, since there isn't, I believe, a simple way to create an FSA from an FSA and an arc_map.

The implementation, I believe, is just to create an Fsa from a list of arcs (instead of create an Fsa from an FSA and an arc_amp), that's why I said we may not need this version.
https://github.com/k2-fsa/k2/blob/9147dbcee8295bcce4e03e400a92dd1d3334aced/k2/csrc/fsa_algo.cu#L338

Thus I think we should define the interface either as
1. return arc_map in the best path, i.e. Array1<int32_t> for arc_map, or as
2. return an linear Fsa which is the best path.

I think we can just create a function to create an FSA from an src FSA and an arc_map, so the interface here could be just returning arc_map.

danpovey: Sure.  Bear in mind that eventually we'll need this to work for FsaVec, and in that case ShortestDistance would have to output a `Ragged<int32_t>`; and we'd need a function (working on GPU!) to create an FsaVec from an input FsaVec and a `Ragged<int32_t>`.  That's very easy to do.

danpovey:
All that's fine, but I'd rather not stress about these interfaces right now
because they are just for one FSA and the more important
one is for FsaVec.  I'd say we can just merge this with small changes and
focus more on the FsaVec version.  I'm working on extending
GetForwardScores() to return backtrace info.

k2/csrc/fsa_algo.h
qindazhu: Dan, why do we need to return `Ragged<int32_t>` here as arc_map, instead of `Array1<int32_t>`? For FsaVec version, what index is in the returned arc_map? idx2 or idx012?

qindazhu: I mean `best_arc_paths`, as it's not arc_map.

csukuangfj: >  what index is in the returned arc_map? idx2 or idx012?

I think the output argument `Ragged<int32_t> *best_path_arcs ` will have two axes containing:

- fsa 0, arc indexes [x x x x]
- fsa 1, arc indexes [x x x]
- ...

danpovey:
What I was thinking is, it wouldn't be obvious which FSAs those arcs
belonged to, in the FsaVec case.
I mean, it could be worked out, but it just doesn't seem like the right
interface to me.

k2/csrc/fsa_algo.h
qindazhu: mm. what `x` here it is? if it's arc-index in the original FsaVec,why not just return an Array1 with `x` is `arc_idx012`

csukuangfj: It's for a FsaVec.

danpovey: Mm.  I still think the right interface is to have it output a Ragged<T>.   (E.g. it makes it easier to look up the output for an individual FSA).  We'd have to have a separate version for FsaVec vs. Fsa, if we support Fsa.  It makes me think maybe we should have supported a Ragged<T> with only one axis, but that's not urgent.

danpovey: ... the Fsa version, i.e. the GPU implementation, could just add one extra axis and then forward the contents of the FsaVec version of the output (the elements of the Ragged<int32_t>).

csukuangfj: Do you mean the interface should look like the following:

```cpp
Ragged<int32_t> ShortestPath(FsaVec &src);

FsaVec CreateFsaVec(FsaVec &src, Ragged<int32_t> &arc_indexes);
```

`arc_indexes` has two axes:
- fsa 0, [arc_idx012 of fsa 0]
- fsa 1, [arc_idx012 of fsa 1]
- ...

`Ragged<int32_t>` is preferred over `Array1<int32_t>` since it is much easier to get the number of states per FSA
and to distribute the work of rebuilding the final FsaVec to different threads.

danpovey:
Couple of comments on your linear-chain CRF:
 -  When the Intersect() with DenseFsa is finished, we won't want to use Fsa to store the array of potentials, we'd use DenseFsa for that.
 -  It should never be necessary to set the aux_labels to a value that's the same as the labels themselves.
 - In the objective function, we should probably be using soft, rather than hard, alignments.  The way we'll be doing this is to do Intersect() to get the composed FsaVec, then compute the total-scores of that FsaVec, and the sum of the total-scores would be the objective.  The total-scores will appear to PyTorch as a Tensor.  This tensor will be created by a PyTorch Function that will cache certain things in the forward pass, if requires_grad is true.  (Note: I am thinking of actually creating a C++ object that represents a cache of various things computed from an Fsa, such as properties, forward scores, backward score, total scores, so that we could have a simplified c++<-->Python interface, and more easily reproduce code in pure C++; but one option is to cache those things in Python).

k2/python/k2/fsa_algo.py
```diff
@@ -145,3 +164,100 @@ def arc_sort(fsa: Fsa) -> Fsa:
...
+        setattr(out_fsa, name, a_value)
+
+    for name, b_value in b_fsa.named_non_tensor_attr():
+        if not hasattr(out_fsa, name):
+            setattr(out_fsa, name, b_value)
+
+    return out_fsa
+
+
+def shortest_distance(fsa: Fsa) -> torch.Tensor:
```
danpovey, line 220: Just a note: when we compute the total-score in the log semiring, there won't be a way to do it as a pure function, we'll need a python Function object to implement the backprop.

We need to do a good job of indicating, in function documentation, what properties are required of the inputs.  For instance, here, does the Fsa have to have 2 axes (single FSA) and be on CPU?

k2/python/k2/fsa_algo.py
```diff
@@ -145,3 +164,100 @@ def arc_sort(fsa: Fsa) -> Fsa:
...
+    for name, value in fsa.named_tensor_attr():
+        setattr(out_fsa, name, value.index_select(0, arc_map))
+    for name, value in fsa.named_non_tensor_attr():
+        setattr(out_fsa, name, deepcopy(value))
+    return out_fsa
+
+
+def project_output(fsa: Fsa) -> Fsa:
+    out_fsa = Fsa.from_ragged_arc(fsa.arcs)
+    indexes = out_fsa.arcs.values()[:, 2] != -1
```
danpovey, line 254: can you please comment here, IDK what this is doing.

danpovey: ... if you are trying to exclude final-states here; is that necessary?  I thought we already had -1's in the aux_labels.

k2/python/k2/fsa_algo.py
```diff
@@ -145,3 +164,100 @@ def arc_sort(fsa: Fsa) -> Fsa:
...
+    out_fsa.arcs.values()[:, 2][indexes] = fsa.aux_labels.to(
+        torch.int32)[indexes]
+    for name, value in fsa.named_tensor_attr():
+        setattr(out_fsa, name, value)
+    for name, value in fsa.named_non_tensor_attr():
+        setattr(out_fsa, name, deepcopy(value))
+    if hasattr(out_fsa, 'osym'):
```
danpovey, line 261: I don't like this 'osym' attribute being treated specially.  We are not reproducing GTN here; these objects are FSAs not FSTs, and they just have various attributes indexed by arc.  If it ever becomes necessary to implement something like this, it would probably just be a question of either copying a vector of labels to the arcs, or swapping the arcs' labels with the contents of some attribute.  But this could be done in-place, it shouldn't be necessary to create a separate object in general.  (We can have a separate function to clone the Fsa object, for where that's needed).

k2/python/k2/fsa_algo.py
```diff
@@ -145,3 +164,100 @@ def arc_sort(fsa: Fsa) -> Fsa:
...
+        out_fsa.isym = out_fsa.osym
```
danpovey, line 262: Ah I see now, 'isym' and 'osym' are the input and output symbol tables.  That was not documented anywhere so should have been commented on!  Again, I don't want to treat these objects as FSTs, let's just call these 'symbol_table' and 'aux_symbol_table' which I think is more self-explanatory.

k2/csrc/fsa_algo.h
danpovey: Yes that's what I mean.   However I think the function should optionally take the 'entering_arcs' vector that's output by GetForwardScores(), as this would be more friendly to future use of caching code.  If it is not supplied, the function can create it internally (and probably then recurse to itself).

k2/python/k2/fsa_algo.py
csukuangfj: For an arc:
```
src_state  dest_state label score
```

This function replaces `label` with `aux_label`, so the arc becomes

```
src_state dest_state aux_label score
```

But for arcs entering the final state, their labels are `-1` and `aux_label` 0. we do not want to change these arcs.

k2/python/k2/fsa_algo.py
csukuangfj: Will rename the symbol tables. They are for visualization only.

k2/python/k2/fsa_algo.py
csukuangfj: > when we compute the total-score in the log semiring, there won't be a way to do it as a pure function, we'll need a python Function object to implement the backprop.

GTN uses `log_semiring` in the crf example and I changed it to `tropical semiring` in k2.

I think the backprop for `log_semiring` has to be implemented in C++ and wrapped to
Python.

k2/python/k2/fsa_algo.py
danpovey: Regarding symbol tables: one possibility that is more general, is that the symbol table for the symbols on the FSA is stored as a non-tensor attribute 'symbols'; and symbol tables associated with other kinds of labels are stored as 'xxx_symbols' where 'xxx' is the name of the attribute.  Then we could just use names like "aux" or "phones" or whatever we want for the auxiliary symbols, and being int32_t, they'll be treated as symbols.

k2/python/k2/fsa_algo.py
danpovey: So it would be easier to set the aux_label for those arcs to -1.

danpovey: .. then there will be no mismatch.  we can add code to check that -1's were not added or removed.

csukuangfj: > I thought we already had -1's in the aux_labels.

Yes, we are using `-1` currently because I use `compose` instead of `intersect`.
I remember that you suggested the aux_labels for arcs entering the final state should
be 0, not -1. I'll will change the aux_labels to 0 once compose is replaced by intersect.

danpovey: ... anyway, I think this can just be done as an in-place modification of the FSA, doesn't have to be an operation.  We can have GetSymbols() and SetSymbols(), that would respectively get and set the symbols on the arcs; and we could just do `fsa.SetSymbols(fsa['aux']);`

k2/python/k2/fsa_algo.py
csukuangfj: Will update the documentation.

danpovey: The code needed for log-semiring backprop already exists, it involves GetForwardScores(), GetBackwradScores(), GetTotScores() to get the forward-pass output; and GetArcScores() [you'd have to exponentiate that to get the derivative].  Needs to be done as a Function in PyTorch.

danpovey: BTW, I think the Python-level Fsa should have a cache, indexed by string, which can store attributes like the batches and the forward and backward scores.  Some of them, like the forward and backward scores, should be stored with names that indicate which semiring we used, e.g. 'forward_log' vs. 'forward_tropical'.  This way we can re-use things where necessary.

k2/python/k2/fsa_algo.py
danpovey: I wasn't sure about that decision.  Let's set them to -1 for those final-arcs, now that it seems to be a problem to have them as zero.

csukuangfj:
> When the Intersect() with DenseFsa is finished, we won't want to use Fsa to store the array of potentials, we'd use DenseFsa for that.

Updated.  `torch.optim.SGD` is now used to do the optimization:

```python
optimizer = SGD([potentials.score, transitions.score], lr=learning_rate)
for it, (X, Y) in enumerate(sampler(train)):
    optimizer.zero_grad()
    model = k2.compose(potentials, transitions)
    loss = crf_loss(X, Y, model)
    loss.backward()
    optimizer.step()
```

---

> It should never be necessary to set the aux_labels to a value that's the same as the labels themselves.

You are right. I am refactoring the symbol table as you suggested.

csukuangfj:
> It should never be necessary to set the aux_labels to a value that's the same as the labels themselves.

Fixed.

I've removed `compose.cc` and use `intersect` for `compose`. A new method `invert_()` is added for `Fsa`
to swap its `labels` and `aux_labels` in-place.

The drawback of using `intersect` for `compose` is that the user has to do more work.

If we have native support for `compose`, to compose `A` and `B`, a simple call `C = k2.compose(A, B)` is enough (suppose
both A and B are arc sorted)

But now we have to do the following:

```python
A = A.invert_()
A = k2.arc_sort(A)
C = k2.compose(A, B).invert_()
```

That is, we have to call `invert_()` and `arc_sort()` before `compose`. Another call to `invert_()` is also needed after `compose`.

k2/python/k2/fsa_algo.py
csukuangfj: Fixed. `Fsa.invert_()` is added to swap labels and aux_labels in-place.

k2/python/k2/fsa_algo.py
csukuangfj: Fixed.

- `symbols` is for `labels`
- `aux_symbols` is for `aux_labels`.

k2/python/k2/fsa.py
```diff
@@ -59,34 +101,66 @@ def __init__(self,
...
+        '''Swap the ``labels`` and ``aux_labels``.
+
+        CAUTION:
+          The function name ends with an underscore which means this
+          is an **in-place** operation.
+        '''
+        if hasattr(self, 'aux_labels'):
+            aux_labels = self.aux_labels
+            self.aux_labels = self.labels
+            self.labels = aux_labels
```
qindazhu, line 153: > A new method invert_() is added for Fsa to swap its labels and aux_labels in-place.

Actually it's not what `Invert` is supposed to do. A naive `Swap` may make the FSA invalid, thinking about the case where the `aux_labels` are sequences (e.g. phones). See function https://github.com/k2-fsa/k2/blob/e9889fa3b202b1fea4b76c370b0f9f0f7efe4c43/k2/csrc/host/aux_labels.h#L149

danpovey: We can give Invert() a different implementation in the case when `aux_labels` has type RaggedTensor(); and we can choose to make Invert_ just crash in that case

danpovey:
> > It should never be necessary to set the aux_labels to a value that's the same as the labels themselves.
> 
> Fixed.
> 
> I've removed `compose.cc` and use `intersect` for `compose`. A new method `invert_()` is added for `Fsa`
> to swap its `labels` and `aux_labels` in-place.
> 
> The drawback of using `intersect` for `compose` is that the user has to do more work.
> 
> If we have native support for `compose`, to compose `A` and `B`, a simple call `C = k2.compose(A, B)` is enough (suppose
> both A and B are arc sorted)
> 
> But now we have to do the following:
> 
> ```python
> A = A.invert_()
> A = k2.arc_sort(A)
> C = k2.compose(A, B).invert_()
> ```
> 
> That is, we have to call `invert_()` and `arc_sort()` before `compose`. Another call to `invert_()` is also needed after `compose`.

How about this: what we rename the current 'compose' to 'intersect', and mak k2.compose be a wrapper that does those 3 steps you just showed?

k2/python/csrc/torch/fsa_algo.cu
```diff
@@ -157,6 +158,37 @@ static void PybindArcSort(py::module &m) {
       py::arg("src"), py::arg("need_arc_map") = true);
 }
 
+static void PybindShortestDistance(py::module &m) {
+  m.def(
```
danpovey, line 162: I'm concerned about how we'll extend this to the FsaVec case, given that the interfaces have to be different.
Perhaps we could call them {fsa_shortest_distance,fsavec_shortest_distance} and so on?

k2/python/k2/fsa_algo.py
```diff
@@ -145,3 +158,137 @@ def arc_sort(fsa: Fsa) -> Fsa:
...
+def shortest_distance(fsa: Fsa) -> torch.Tensor:
+    '''Return the shortest distance from the start state to the final state
+    in the tropical semiring.
+
+    Note:
+      It uses the opposite sign. That is, It uses `max` instead of `min`.
+
+    Args:
+      fsa:
+        The intpu FSA.
```
danpovey, line 260: typo

k2/python/k2/fsa_algo.py
```diff
@@ -145,3 +158,137 @@ def arc_sort(fsa: Fsa) -> Fsa:
...
+
+def shortest_path(fsa: Fsa) -> Fsa:
+    need_arc_map = True
+    ragged_arc, arc_map = _k2.shortest_path(fsa.arcs, need_arc_map)
+    arc_map = arc_map.to(torch.int64)  # required by index_select
+    out_fsa = Fsa.from_ragged_arc(ragged_arc)
+    for name, value in fsa.named_tensor_attr():
+        setattr(out_fsa, name, value.index_select(0, arc_map))
+    for name, value in fsa.named_non_tensor_attr():
+        setattr(out_fsa, name, deepcopy(value))
```
danpovey, line 279: I don't think it should be necessary to do a deep copy by default.  Currently this is just symbol tables, right?  This could get very slow in certain situations.  Can't we just have the user do something like `f.symbols = deepcopy(f.symbols)` if they are going to want to change elements of it?

k2/python/k2/fsa_algo.py
```diff
@@ -145,3 +158,137 @@ def arc_sort(fsa: Fsa) -> Fsa:
...
+    return out_fsa
+
+
+def project_output(fsa: Fsa) -> Fsa:
```
danpovey, line 283: I'd rather delete this function because IMO it doesn't make sense in this formulation.  project_output means setting the labels and aux_labels to the same value, but in that case we don't need aux_labels.  Also there is no assumption that the aux_labels are the output and the labels are the input.  The labels are whatever we're operating on right now; the aux_labels are whatever we are not operating on.

k2/python/k2/fsa_algo.py
csukuangfj: Yes, I agree.

k2/python/k2/fsa_algo.py
danpovey: .. we could perhaps rename it though, e.g. to with_aux_labels().
Does fsa.from_ragged_arc deepcopy the arcs?  I'd rather not have this happen by default; if we need a deepcopy
it would be clearer to do it directly here.

k2/python/k2/fsa_algo.py
```diff
@@ -145,3 +158,137 @@ def arc_sort(fsa: Fsa) -> Fsa:
...
+    out_fsa = Fsa.from_ragged_arc(fsa.arcs)
+    out_fsa.labels = fsa.aux_labels
+    for name, value in fsa.named_tensor_attr():
+        setattr(out_fsa, name, value)
+    for name, value in fsa.named_non_tensor_attr():
+        if name != 'symbols':
+            setattr(out_fsa, name, deepcopy(value))
+
+    if hasattr(out_fsa, 'aux_labels'):
+        out_fsa.symbols = out_fsa.aux_labels
```
danpovey, line 293: I think this may be a typo: surely should be out_fsa.aux_symbols?  
But since now the labels and aux_labels are the same there is no sense in keeping the aux_labels, so it would make sense to delete out_fsa.aux_symbols and out_fsa.aux_labels.  Anyway maybe the simplest fix right now it just to delete this.

k2/python/csrc/torch/fsa_algo.cu
csukuangfj: I think Fsa and FsaVec can share the same interface.

If we assume that there is only one best path, then what we need in Python
is just the arc indexes of the best path. If it is a FsaVec, the arc indexes would
be idx012. On the Python side, we would use `score.index_select(0, arc_indexes).sum()`
to compute the distance.

csukuangfj:
> How about this: what we rename the current 'compose' to 'intersect', and mak k2.compose be a wrapper that does those 3 steps you just showed?

The advantage of putting `invert_()` outside `compose` is that it can save the number of calls to `invert_()` when we
are composing multiple FSAs.

For example, to compose C L G:

```python
C = C.invert_()
CLG = k2.compose(k2.arc_sort(k2.compose(C, L)), G).invert_()
```

When `invert_()` is a cheap operation, I agree that we put `arc_sort` and `invert_` inside the `compose`.

k2/python/k2/fsa_algo.py
csukuangfj: I agree to delete it.

k2/python/k2/fsa_algo.py
csukuangfj: Yes, you are right. Will remove this function.

k2/python/csrc/torch/fsa_algo.cu
danpovey: No, because the natural output of ShortestDistance for a vector of FSAs is a Tensor of distances of shape (num_fsas,), and for that we need to know which distances come from which FSAs.

danpovey:
OK.  Let's rename compose to intersect though, because that's more  reflective of what it does.

k2/python/csrc/torch/fsa_algo.cu
csukuangfj: > natural output of ShortestDistance for a vector of FSAs is a Tensor of distances of shape (num_fsas,)

But this tensor cannot help backprop. For backprop, I think only the arc indexes
matter (for tropical semiring). We can compute the distances from the returned arc_indexes.

csukuangfj:
> OK. Let's rename compose to intersect though, because that's more reflective of what it does.

Will rename it.

k2/python/csrc/torch/fsa_algo.cu
danpovey: For this particular application it doesn't matter because we just want to sum the result, but the point is we are creating a general-purpose interface, and I don't want to have to change it more than necessary.

danpovey:
Please remove WIP when you think it's ready to merge, and flag any issue that you think we might still have been discussing or that there were doubts about.

csukuangfj:
Will fix the comments and remove WIP this night.

k2/csrc/fsa_algo.cu
csukuangfj: Fixed. Return negative infinity on empty FSAs.

k2/csrc/fsa_algo.h
csukuangfj: `ShortestDistance` is removed.

k2/python/k2/fsa_algo.py
csukuangfj: I have added two attributes: `arc_labels_sorted` and `arc_aux_labels_sorted` to avoid unnesseary `arc_sort` when
the fsa is already arc sorted.

k2/python/k2/fsa_algo.py
csukuangfj: Now arcs entering the final state have -1 for its labels and aux_labels.

project_output is removed.

k2/python/k2/fsa_algo.py
csukuangfj: Thanks! Fixed.

k2/python/k2/fsa_algo.py
csukuangfj: Removed.

k2/python/k2/fsa_algo.py
csukuangfj: Fixed.

csukuangfj:
Ready for review. WIP is removed.

k2/csrc/fsa_algo.cu
```diff
@@ -258,38 +259,27 @@ void ArcSort(Fsa &src, Fsa *dest, Array1<int32_t> *arc_map /*= nullptr*/) {
...
-  K2_CHECK_EQ(context->GetDeviceType(), kCpu);
-  int32_t num_states = fsa.Dim0();
-  k2host::Fsa host_fsa = FsaToHostFsa(fsa);
-  Array1<double> state_weights(context, num_states);
-  std::vector<int32_t> tmp_arc_map;
-  ComputeForwardMaxWeights(host_fsa, state_weights.Data(), &tmp_arc_map);
-  if (arc_map != nullptr) *arc_map = Array1<int32_t>(context, tmp_arc_map);
+double ShortestPath(Fsa &src, Fsa *out,
+                    Array1<int32_t> *best_path_arcs /* = nullptr*/) {
+  if (!src.values.IsValid()) return -std::numeric_limits<double>::infinity();
```
qindazhu, line 264: Actually, I think usually we can suppose arcs are always valid, so I suggest we don't check this every time.

k2/python/k2/fsa.py
```diff
@@ -26,6 +26,57 @@ class Fsa(object):
...
+                FSA. The first column represents the source states, second
+                column the destination states, third column the labels and the
+                fourth column is the score. Note that the score is actually
+                a float number but it is **reinterpreted** as an integer.
+
+    - ``score``: A 1-D ``torch.Tensor`` of dtype ``torch.float32``. It has
+                 as many entries as the number of arcs representing the score
+                 of every arc.
+
+    - ``labels``: A 1-D ``torch.Tensor`` of dtype ``torch.int32``. It has as
```
qindazhu, line 45: just a question, so now the `symbol` in `Arc` definition in C++ is actually a `label`?

https://github.com/k2-fsa/k2/blob/6acfe3d54140889a8456a4002cb393607c165c64/k2/csrc/fsa.h#L24-L28

k2/python/k2/fsa.py
```diff
@@ -26,6 +26,57 @@ class Fsa(object):
...
     for a vector of FSAs to have zero or one elements).
+
+    An instance of FSA has the following attributes:
+
+    - ``arcs``: You will NOT use it directly in Python. It is an instance
+                of ``_k2.RaggedArc`` with only one method ``values()`` which
+                returns a 2-D `torch.Tensor`` of dtype ``torch.int32`` with 4
+                columns. Its number of rows indicates the number of arcs in the
+                FSA. The first column represents the source states, second
+                column the destination states, third column the labels and the
```
qindazhu, line 37: as we declared here, `third column the labels`, instead of `third column the symbols`

k2/python/k2/fsa.py
csukuangfj: Sorry for the confusion. `labels` is used interchangeably with the `symbol`.

k2/python/k2/fsa.py
```diff
@@ -26,6 +26,57 @@ class Fsa(object):
...
+
+    - ``score``: A 1-D ``torch.Tensor`` of dtype ``torch.float32``. It has
+                 as many entries as the number of arcs representing the score
+                 of every arc.
+
+    - ``labels``: A 1-D ``torch.Tensor`` of dtype ``torch.int32``. It has as
+                  many entries as the number of arcs representing the label of
+                  every arc.
+
+    - ``arc_labels_sorted``: A boolean. True if arcs of every state are sorted
```
qindazhu, line 49: Can someone remind me why we sort arcs on labels first then dest_state in host code? I believe there was a reason that time. And seems now in k2 code, we just sort on labels?
https://github.com/k2-fsa/k2/blob/6acfe3d54140889a8456a4002cb393607c165c64/k2/csrc/fsa_algo.cu#L236-L250

k2/python/k2/fsa.py
```diff
@@ -26,6 +26,57 @@ class Fsa(object):
...
+                             by labels in increasing order. False means arcs
+                             are not sorted by ``labels``.
+
+    It MAY have the following attributes:
+
+    - ``symbols``: An instance of ``k2.SymbolTable``. It maps an entry in
+                   ``labels`` to an integer and vice versa. It is used for
+                   visualization only.
+
+    - ``aux_labels`: A 1-D ``torch.Tensor`` of dtype ``torch.int32``. It has the
```
qindazhu, line 59: not sure if we should add a TODO here to say, it maybe a `RaggedTensor` in the future.

k2/python/k2/fsa.py
```diff
@@ -26,6 +26,57 @@ class Fsa(object):
...
+                     same shape as ``labels``.
+
+    - ``aux_symbols``: An instance of ``k2.SymbolTable. It maps an entry in
+                       ``aux_labels`` to an integer and vice versa.
+
+    - ``arc_aux_labels_sorted``: A boolean. True if arcs of every state are
```
qindazhu, line 65: Actually it seems to me a little strange that arcs sorted by `axu_labels`, as I was thinking we just sort arcs on labels (though they may be inverted)

k2/python/k2/fsa.py
csukuangfj: As far as I know, sorting is to make `intersect` run faster. `dest` state does not affect the intersection process, only the labels matter. The current implementation assumes both the two input FSAs are arc sorted.

k2/python/k2/fsa_algo.py
```diff
@@ -50,37 +51,55 @@ def top_sort(fsa: Fsa) -> Fsa:
...
-            # TODO(fangjun): this does not support aux_labels!
+
+            # a_arc_map and b_arc_map have been offset by 1
+            # so we need a padding here
+            padding = a_value.new_zeros((1, *a_value.shape[1:]))
+            a_value = torch.cat((padding, a_value), dim=0)
+            b_value = torch.cat((padding, b_value), dim=0)
+
             assert a_value.dtype == b_value.dtype == torch.float32
             value = a_value.index_select(0, a_arc_map) \
```
qindazhu, line 104: So here we are getting the score on the result FSA's arcs? I believe we have set the scores on arcs in `Intersect`, why do we need to `add` again here? Did I miss some PR before?

k2/python/k2/fsa.py
csukuangfj: If an `fsa` is already arc sorted. Do you have a better method to determine that `fsa.invert_().invert_()` is also arc sorted?

k2/python/k2/fsa_algo.py
csukuangfj: This is for backprop. The add in C++ does not participate in the autograd mechanism of Pytorch.

qindazhu: Oh, thanks!

k2/python/k2/fsa.py
danpovey: The most direct way, IMO, to tell whether things are arc-sorted, is the properties.  
There is a flag for that when we call Properties().
One of the things the Fsa object can cache is the 'properties' tensor from GetFsaBasicProperties() (one property per FSA) and the 'properties' scalar which is the "and" of the properties tensor.

It might be simplest just to recompute the properties when we invert (?), because I don't think inversion will happen veyr often.

k2/python/k2/fsa_algo.py
```diff
@@ -26,7 +25,9 @@ def linear_fsa(symbols: Union[List[int], List[List[int]]]) -> Fsa:
       A vector of FSAs if the input is a list of list of integers.
     '''
     ragged_arc = _k2.linear_fsa(symbols)
-    return Fsa.from_ragged_arc(ragged_arc)
+    fsa = Fsa.from_ragged_arc(ragged_arc)
+    fsa.arc_labels_sorted = True
```
danpovey, line 29: My instinct here is to just compute and cache the properties and not treat `arc_labels_sorted` specially.
Although we can of course provide a convenient interface to test individual properties.
It may be sufficient to always deal with the scalar properties for now; we can add the vector properties as and when we need them.

k2/python/k2/fsa.py
qindazhu: I mean, if we just need a value to cache the property of `arc_sorted`, why not just use a local variable to do this? Exposing this    at the Fsa interface level is confusing as we never really sort an Fsa on aux_lables (what we do is:  invert then sort)

k2/python/k2/fsa_algo.py
```diff
@@ -50,37 +51,55 @@ def top_sort(fsa: Fsa) -> Fsa:
...
-      Intersection is also known as compose in FST.
-
     Args:
       a_fsa:
         The first input FSA. It can be either a single FSA or a vector of FSAs.
       b_fsa:
         The second input FSA. It can be either a single FSA or a vector of FSAs.
 
+    CAUTION:
+      ``aux_labels`` are discarded since it is not clear how to set the
```
danpovey, line 68: We don't need to treat aux_labels specially here.  The rules will be:
    For attributes where only one source (a or b) has that attribute:
       Copy via arc_map, or use zero if arc_map has -1.  This rule works for both 
       floating point and integer.
    For attributes where both sources (a and b) have that attribute:
       For floating point attributes: sum via arc_maps, or use zero if arc_map has -1.  
       For integer attributes, it's not supported for now.
Later we can add an optional argument to `intersect` that is a dict from an attribute-name to some
kind of object or enum or string that says how to process that attribute.

k2/python/k2/fsa_algo.py
```diff
@@ -50,37 +51,55 @@ def top_sort(fsa: Fsa) -> Fsa:
...
+      aux_labels of the output FSA.
+
     Returns:
       The result of intersecting a_fsa and b_fsa.
     '''
+    if not a_fsa.arc_labels_sorted:
```
danpovey, line 74: I'd rather have the user explicitly do the arc sorting, like it is done in OpenFST.
When we test whether an Fsa is arc-sorted, it will be done by:
   - if the properties are cached, use the cached properties;  otherwise compute and cache the properties first.

k2/python/k2/fsa_algo.py
```diff
@@ -50,37 +51,55 @@ def top_sort(fsa: Fsa) -> Fsa:
...
+
     need_arc_map = True
     ragged_arc, a_arc_map, b_arc_map = _k2.intersect(a_fsa.arcs, b_fsa.arcs,
                                                      need_arc_map)
-    a_arc_map = a_arc_map.to(torch.int64)  # required by index_select
-    b_arc_map = b_arc_map.to(torch.int64)
+
+    # Some of entries in a_arc_map and b_arc_map may be -1.
+    # The arc_maps are incremented so that every entry is non-negative.
+    a_arc_map = a_arc_map.to(torch.int64) + 1
```
danpovey, line 86: This is OK in the short term, but I think it would be significantly more efficient to create a k2 operation that does something
of the form:
`   c[:] = a[b]`
where c, a and b are tensors, and that sets the element to zero if the index was -1.  Assuming PyTorch doesn't already have an operation like this.

k2/python/k2/fsa.py
csukuangfj: For `c = k2.compose(a, b)`, we need to ensure that both `a.aux_labels`  and `b.labels` are sorted. I'm afraid it's difficult to know whether an `fsa` is arc sorted with a single variable `arc_sorted`.

The intention of these two variables is to avoid calling `arc_sort`  when an `fsa` is already arc sorted.
If `k2.compose()` is called `N` times, then `invert_()` will be invoked `3N` times. I am not sure how often `k2.compose` will be invoked.

> One of the things the Fsa object can cache

I think we have to figure out which properties are changed by which operations and update the cache accordingly.

k2/python/k2/fsa_algo.py
csukuangfj: Will fix it.

k2/python/k2/fsa_algo.py
csukuangfj: The problem is that we are calling `invert_()` inside `k2.compose()` and `arc_sort()`
is called after `invert_()`. Users do not have the chance to call `arc_sort` before `k2.compose()`.

k2/python/k2/fsa.py
csukuangfj: > what we do is: invert then sort)

`arc_aux_labels_sorted` is to avoid `arc_sort` after calling `invert_()`.

qindazhu: mm, I mean, can't we just cache `arc_sorted` in a local variable and then, after `a.invert().invert()`, set the local variable to the result fsa's `arc_sorted`. Or just as Dan said, check the properits again after `invert().invert()`, it's not so expensive considering we don't call `invert` frequently.  Anyway, I believe exposing `aux_labels_sorted` at the global interface level will be confusing as users would never hold an FSA sorted by `aux_labels`.  Actually, your code don't hold such an FSA either, you just use `arc_aux_labels_sorted` as a local cache of the orginal FSA's `arc_sorted`.

k2/python/k2/fsa.py
qindazhu: Seems that time we wanted to use this feature in `determinize`, but finally we didn't use it. https://github.com/k2-fsa/k2/pull/17#issuecomment-620587443
@danpovey Should we change the implementation in k2 for the `operator<` in Arc to just sort on labels?

https://github.com/k2-fsa/k2/blob/6acfe3d54140889a8456a4002cb393607c165c64/k2/csrc/fsa.h#L37-L42

k2/python/k2/fsa_algo.py
qindazhu: Sounds it would be simpler by removing `compose`... we just support `intersect`, uses should call `invert` if they need.

k2/python/k2/fsa.py
danpovey: I think we can actually rename `symbol` in C++ to label, for consistency, and reserve `symbol` for the actual symbol table.

k2/python/k2/fsa.py
danpovey: Just sorting on labels should be OK for now.  But it will make very little difference to the speed.

k2/csrc/fsa_algo.cu
csukuangfj: So is it checked only in debug build?

qindazhu: it's fine to not check this even in debug build as we have checked `src.values.Context()` in the constructor of `Ragged<T>` (if it is invaid, then `src.shape` is invalid as well, that would usually not happen considering the way we create `Ragged` or `RaggedShape`, we always created row_splits with a context)

The main reason we add `IsValid` in `Array1` is to let us know the `Array1` object is not created with the default constructor (no `context`, compared with empty case which also has `Dim() == 0`). Fow now it's just be useful in `RowIds` things since we leave `row_ids` in `RaggedShapeDim` as invalid as default.

BTW, sometimes it's worthy to check if the Ragged object empty or not (e.g. in Transpose, as we will do `TotSize(1) / dim0`, we need make sure it's not empy, i.e. dim0 >=0). But it's OK here to not check this as you have handled such cases with below code.

k2/python/k2/fsa_algo.py
csukuangfj: `compose` is removed. We give more freedom to users.

k2/python/k2/fsa_algo.py
csukuangfj: Fixed.

k2/python/k2/fsa_algo.py
csukuangfj: An attribute `properties` is added. Users can pre-compute the properties of an FSA and cache it before calling
any fsa algorithms.

k2/python/k2/fsa.py
csukuangfj: `arc_sorted` and `aux_labels_sorted `  are superseded by `properties`.

k2/python/k2/fsa.py
csukuangfj: Added.

csukuangfj:
Ready for review.

I also updated the [Colab notebook][1].

[1]: https://colab.research.google.com/drive/1NDNcEK39JYWGRNOJMCtUz1Z2skOS2var?usp=sharing

qindazhu:
Great, +1!

k2/python/csrc/torch/fsa_algo.cu
```diff
@@ -157,6 +158,22 @@ static void PybindArcSort(py::module &m) {
       py::arg("src"), py::arg("need_arc_map") = true);
 }
 
+static void PybindShortestPath(py::module &m) {
+  m.def(
+      "shortest_path",
+      [](Fsa &src, bool need_arc_indexes = true)
+          -> std::tuple<Fsa, torch::optional<torch::Tensor>, double> {
```
danpovey, line 165: I'd like to figure out how we'll deal with FsaVec before merging this.  The problem is, the input type is the same, but the return types would need to be different (i.e. FsaVec, Ragged<int32_t>, torch::Tensor).

k2/python/k2/fsa.py
```diff
@@ -70,23 +117,78 @@ def __setattr__(self, name: str, value: Any) -> None:
...
+                # NOTE: we **reinterpret** the float patterns
+                # to integer patterns here.
                 self.arcs.values()[:, -1] = _as_int(value.detach())
         else:
-            # TODO(fangjun): do we need to copy value?
             self._non_tensor_attr[name] = value
 
     def __getattr__(self, name: str) -> Any:
-        if name in self._tensor_attr:
+        if name == 'labels':
```
danpovey, line 136: I'm not a big Python expert, but wouldn't this kind of thing normally be done via a property?  (I.e. the labels and scores getters/setters?)

k2/python/csrc/torch/fsa_algo.cu
qindazhu: Wondering if we really need to support Fsa version, maybe it would be simpler if we just support version `FsaVec` as we may only need `FsaVec` version in production code? We can just wrap a funtion at Python level to convert Fsa to FsaVec with size 1 and do the simliar thing to the output (But I still doubt if we really need single Fsa version)

k2/python/k2/fsa.py
```diff
@@ -70,23 +117,78 @@ def __setattr__(self, name: str, value: Any) -> None:
...
+            return self.arcs.values()[:, 2]
+        elif name in self._tensor_attr:
             return self._tensor_attr[name]
         elif name in self._non_tensor_attr:
             return self._non_tensor_attr[name]
 
         raise AttributeError(f'Unknown attribute {name}')
 
+    def __delattr__(self, name: str) -> None:
+        assert name not in ('arcs', 'score', 'labels')
```
danpovey, line 146: I think 'score' should be 'scores', since all of the other attributes are plurals.

k2/python/k2/fsa.py
```diff
@@ -70,23 +117,78 @@ def __setattr__(self, name: str, value: Any) -> None:
             object.__setattr__(self, name, value)
         elif isinstance(value, torch.Tensor):
             assert value.shape[0] == self.arcs.values().shape[0]
-            # TODO(fangjun): do we really need a copy here?
+            if name == 'labels':
+                assert value.dtype == torch.int32
+                self.arcs.values()[:, 2] = value
+                return
+
             self._tensor_attr[name] = value.clone()
```
danpovey, line 125: I don't see why we need to do clone() here.

csukuangfj: My intention was to retain a copy inside FSA so that its attribute is not modified accidentally.

k2/python/csrc/torch/fsa_algo.cu
csukuangfj: How about adding `RaggedTensor` to Python which contains a C++ object `RaggedArray<T>` (for T == int32_t or float)?

```python
class RaggedTensor(object):
  def __init__(self, ragged_tensor: Union[_k2.RaggedIntArray, _k2.RaggedFloatArray]):
     pass
  def values() -> torch.Tensor:
     # return Ragged<T>::values
  def get(idx0: int) -> torch.Tensor:
    # if the ragged tensor has two axes, return the elements of `idx0`
```

We can return a `RaggedIntArray` to Python for a single FSA as well as a FsaVec.

k2/python/k2/fsa.py
csukuangfj: I just tried to add `@property` and `@labels.setter` for reading and writing `labels`. But it seems
that `__setatrr__` has higher precedence than `@labels.setter`. That is, the `setter` of `labels` is not called. 

There is a question on stackoverflow about `properties` and `__setattr__` in Python:
https://stackoverflow.com/questions/15750522/class-properties-and-setattr

Not sure whether `properties` are preferred over `__setattr__` and `__getattr__` or not.

k2/python/k2/fsa.py
danpovey: I'd rather just warn the user that it's not cloned so the user should be careful about modifying it.
In PyTorch in-place operations are fairly rare anyway.

csukuangfj: A caution is added and `clone` is removed.

k2/python/k2/fsa.py
csukuangfj: Fixed.

k2/python/csrc/torch/fsa_algo.cu
danpovey: What I was thinking is, it would be easier in the long term if we created a RaggedTensor type in C++, which would be like `Ragged<X>` but the values would be a Tensor, and we expect its 1st axis to have dim == shape.NumElements().
Then the Python code can wrap that directly.
And we can define a template like this:
```
template <typename T>
Ragged<T> AsRagged(RaggedTensor &r);
```
which would expect the tensor to be convertible to that type and would crash if not.  Then when we need Ragged<X> in python we can use RaggedTensor, and convert in C++.

csukuangfj: Is the `Tensor` a `k2::Tensor` or a `torch::Tensor`.

Does the `RaggedTensor` in C++ look like the following:

```cpp
struct RaggedTensor {
  RaggedShape shape;
  Tensor values;  // REQUIRES: 1-D; continguous
};
```

> we expect its 1st axis to have dim == shape.NumElements().

Did you mean `TotSize(0)` or `TotSize(1)` ?

danpovey: The Tensor would be a k2::Tensor, because at this level we want the code to be independent of the outer toolkit.
By "we expect its 1st axis to have dim == shape.NumElements()", what I mean that the `tensor.Dim(0)` would equal the `shape.NumElements()`.

k2/csrc/fsa_algo.h
danpovey: Please be very clear if it must have 2 axes, i.e. if FsaVec is not supported.  The implementation code calls the 'fsa' arg FsaVec.

k2/csrc/fsa_algo.cu
```diff
@@ -288,4 +289,127 @@ void ArcSort(Fsa &src, Fsa *dest, Array1<int32_t> *arc_map /*= nullptr*/) {
...
+      num_best_arcs_per_fsa_data[fsas_idx0] = 0;
+      return;
+    }
+
+    int32_t final_state_idx01 = state_idx01_next - 1;
+    int32_t cur_state = final_state_idx01;
+    int32_t cur_index = entering_arcs_data[cur_state];
+    int32_t num_arcs = 0;
+    while (cur_index != -1) {
+      cur_state = arcs_data[cur_index].src_state + state_idx01;
```
danpovey, line 376: I don't like that you have 2 kernels that both have a loop.  You could, in this kernel, store the best traceback in a ragged array that was "too large" (e.g. had the total num-states in the input FSAs), and then copy the info to the output within a kernel that was loop-free.

danpovey: I'm OK to merge this anyway, but please put it on your TODO list to fix this.

k2/csrc/fsa_algo.h
csukuangfj: This function has been removed as it does not support FsaVec.

k2/csrc/fsa_algo.cu
csukuangfj: Will change it as you are willing to trade space for speed.

csukuangfj: Fixed.

csukuangfj: It still has a loop that iterates over the number of states.

csukuangfj:
Ready for review.

k2/csrc/fsa_algo.cu
```diff
@@ -288,4 +289,92 @@ void ArcSort(Fsa &src, Fsa *dest, Array1<int32_t> *arc_map /*= nullptr*/) {
...
+    int32_t state_idx01 = row_splits1_data[fsas_idx0];
+    int32_t state_idx01_next = row_splits1_data[fsas_idx0 + 1];
+    if (state_idx01_next == state_idx01) return;
+
+    int32_t ans_idx01 = ans_row_splits_data[fsas_idx0];
+    int32_t *p = best_path_arc_indexes_data + ans_idx01;
+    for (int32_t s = state_idx01; s != state_idx01_next; ++s) {
+      int32_t index = state_best_arc_index_array_data[s];
+      if (index != -1) {
+        *p = index;
```
danpovey, line 369: This point is we shouldn't have *two* kernels with a loop.  There are still 2 kernels with a loop.
The idea was, you could store the best arc index in an array with the same dimensions as the num-states for each FSA, in a location like `values[row_splits1[fsa_idx0] + num_arcs]`.  Then you could extract them in a kernel that doesn't have a loop.

danpovey: ... I know in this particular case it's only a factor of 2, but this kind of thing is central about how the toolkit is intended to be used.  It'a all about removing the need for loops by using ragged tensors.

csukuangfj: I see. Will update it.

danpovey: ... incidentally, it's possible to further optimize this so the run time is less than linear, by using methods similar to what I use in GetStateBatches(); imagine computing a table that instead of the best traceback, is the best 2-step traceback; and then the 4-step traceback, and so on.  There's no need for this right now, since the forward-pass algorithm is already at least linear-time in the length of this path.  But we can consider it for the future.

danpovey:
Great, merging.  Thanks!

k2/csrc/fsa_algo.cu
csukuangfj: Fixed. The second loop is eliminated and a TODO item is added to optimize it further as you suggested.

PR #284, opened 2020-10-26
Add (and test) ability to return traceback to GetForwardScores().

danpovey:


danpovey:
The testing is not very thorough but it seems to work.
Here is what I recommend to implement the FsaVec version of GetBestPath():
  inside the kernel it would trace back from the final state.  We don't know how long the best path will be but we know it will not be longer than the number of batches, so we could allocate an Array2 of num_fsas by num_batches, to store the best path, and then consolidate it afterward.  The kernel would have a loop and would return when it hits the start state.

PR #285, opened 2020-10-27
Add bidirectional I/O (std::istream,std::ostream) for ragged tensors and arrays

danpovey:


k2/csrc/array_inl.h
```diff
@@ -62,6 +62,136 @@ Array1<S> Array1<T>::AsType() {
...
+  }
+}
+
+
+template <typename T>
+Array2<T>::Array2(const std::string &str): Array2() {
+  std::istringstream is(str);
+  is >> *this;
+  if (!is.good()) {
+    K2_LOG(FATAL) << "Failed to initialize Array1 from string: "
```
csukuangfj, line 189: A typo here. Should be Array2.

PR #286, opened 2020-10-28
Added random tests for GetForward/BackwardScores and Fixed some bugs

qindazhu:


k2/csrc/array.h
```diff
@@ -173,7 +174,24 @@ class Array1 {
   // Convert this array to another array with type S;
   // if S is same with T, then it just returns *this
   template <typename S>
-  Array1<S> AsType();
+  Array1<typename std::enable_if<!std::is_same<S, T>::value, S>::type> AsType()
```
danpovey, line 177: would be better in array_inl.h IMO, but OK.

k2/csrc/array.h
```diff
@@ -173,7 +174,24 @@ class Array1 {
...
+    S *ans_data = ans.Data();
+    const T *this_data = Data();
+    auto lambda_set_values = [=] __host__ __device__(int32_t i) {
+      ans_data[i] = static_cast<S>(this_data[i]);
+    };
+    Eval(Context(), Dim(), lambda_set_values);
+    return ans;
+  }
+  template <typename S>
+  Array1<typename std::enable_if<std::is_same<S, T>::value, S>::type> AsType()
```
danpovey, line 190: this one OK to be here for sure though.

danpovey: for later though; for now lets just make it work and ignore this.

k2/csrc/array_test.cu
```diff
@@ -705,22 +696,16 @@ TEST(ArrayTest, TestArray1Io) {
 TEST(ArrayTest, TestArray2Io) {
   for (int i = 0; i < 20; i++) {
     ContextPtr c = GetCpuContext();
-    int32_t dim0 = RandInt(0, 4),
-        dim1 = RandInt(0, 4);
-    if (dim0 == 0)
-      dim1 = 0;
-    Array1<int32_t> a = RandUniformArray1<int32_t>(
-        c, dim0 * dim1, 0, 100);
+    int32_t dim0 = RandInt(1, 4), dim1 = RandInt(1, 4);
```
danpovey, line 699: I was trying to test arrays where one dim is zero, which I believe are allowed, no?  Not a super big deal but I like to test all cases.

k2/csrc/array.h
qindazhu: Will get `function incompatible declaration` error if it's outside of the Array class declaration with enable_if. It may be fixed with tricks with template, but keep in it here would be much simpler. So let's keep it here.

k2/csrc/array_test.cu
qindazhu: For now, we don't allow to create an empty Array2 from Array1, should we change that? Or I can just add a test for empty array, but not create from Array1.

https://github.com/k2-fsa/k2/blob/6acfe3d54140889a8456a4002cb393607c165c64/k2/csrc/array.h#L569-L575

danpovey:
Ah I see.   Let's make it possible to create empty Array2 from Array1.  We
already support valid-but-zero-sized arrays.

qindazhu:
OK, will do.

k2/csrc/array_test.cu
qindazhu: Updated, reverted the tests.

PR #291, opened 2020-10-29
Fixed some bugs in TopSort(GPU) and change the implementenation of host TopSort

qindazhu:
- Fixed some bugs in TopSort(GPU version), but there are still some bugs, will look into later (mege now to avoid future conflicts)
- Our origninal version of host TopSort requires the input Fsa connected which is not necessary, re-implemented it by removing this requirement.
- Added parameter `epsilon_free` in `RandomFsa(Vec)` as we need (at least one of) input Fsas of `Intersect` be epsilon-free (as I will call `IsRandEquivalent` in random tests which calls host `Intersect` internally)

### DIFF
--- a/k2/csrc/fsa_utils_test.cu
+++ b/k2/csrc/fsa_utils_test.cu
@@ -747,8 +747,7 @@ TEST_F(StatesBatchSuiteTest, TestBackwardScores) {
   {
     // random case
     for (int32_t i = 0; i != 2; ++i) {
-      // for (auto &context : {GetCpuContext(), GetCudaContext()}) {
-      for (auto &context : {GetCpuContext()}) {
+      for (auto &context : {GetCpuContext(), GetCudaContext()}) {
         FsaVec random_fsas = RandomFsaVec();
         FsaVec fsa_vec = random_fsas.To(context);
         int32_t num_fsas = fsa_vec.Dim0(), num_states = fsa_vec.TotSize(1),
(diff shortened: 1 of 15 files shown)


k2/csrc/fsa_utils_test.cu
```diff
@@ -772,8 +771,6 @@ TEST_F(StatesBatchSuiteTest, TestBackwardScores) {
               GetBackwardScores<float>(cpu_fsa_vec, nullptr, false);
           CheckArrayData(scores, cpu_scores);
         }
-        // TODO(haowen): will get -nan in below tests, may be related to use
```
qindazhu, line 775: Ignore this, as I have fixed this issue in the last PR but forgot to remove this comment. (we get -nan because we may add `inf` with `-inf` before.)

k2/csrc/top_sort_test.cu
```diff
@@ -194,4 +195,61 @@ TEST(TopSort, RandomVectorOfFsas) {
   }
 }
 
+// another random test which uses IsRandEquivalent to check the result
+TEST(TopSort, RandomVectorOfFsas1) {
+  ContextPtr cpu = GetCpuContext();
+  for (int32_t n = 0; n != 0; ++n) {
```
qindazhu, line 201: Ignore this, use `n!=0` on purpose as there are still bugs in GPU version of TopSort, will continue debugging it.

qindazhu:
Noted I did not fix style issues in `Intersect.cu` in case of conflicts with Dan.

danpovey (commented):
LGTM!  Thanks!

qindazhu:
Merging, will continue debugging TopSort(GPU version)

PR #293, opened 2020-10-31
 Fix multiple bugs in TopSort(GPU)

qindazhu:


k2/csrc/top_sort.cu
```diff
@@ -91,15 +96,58 @@ FsaVec RenumberFsaVec(FsaVec &fsas, const Array1<int32_t> &order,
...
+      int32_t ans_arc_idx012 = valid_arc_new2old_data[valid_arc_idx012];
+      valid_row_ids2_data[valid_arc_idx012] = ans_row_ids2_data[ans_arc_idx012];
+      valid_arcs_data[valid_arc_idx012] = ans_arcs_data[ans_arc_idx012];
+      if (arc_map_data != nullptr)
+        valid_arc_map_data[valid_arc_idx012] = arc_map_data[ans_arc_idx012];
+    };
+    Eval(c, valid_num_arcs, lambda_set_valid_arcs_and_row_ids2);
+
+    Array1<int32_t> valid_row_splits2(c, new_num_states + 1);
+    RowIdsToRowSplits(valid_row_ids2, &valid_row_splits2);
```
qindazhu, line 142: @danpovey , I want to call `SumPerSublist` here on `keep_data` to get `valid_row_splits2` (in another stream), but the type of `keep_data` is `Array1<char>` and  the type of `row_splits` is `Array1<int32_t>`. So from a perspective of long term usage, should we change the type of `keep` in `Renumbering` to `Array1<int32_t>`? or should we support `ApplyOpPerSublist` with a template `dest` type, e.g.
```
template <typename T, typename S, typename Op>
void ApplyOpPerSublist(Ragged<T> &src, T default_value, Array1<S> *dst);
```
I'm not sure about which is better for general usage in a long term.

qindazhu:
There's a bug in `TopSort` because we processed those inaccessible states, I will fix it in the next PR (i.e. we have to start from only start state, but also propagate those states whose in-degree are zeros, and detect cycle in the process (access back to start state)). See below error examples:

```
[ [ ] [ 1 0 x x 1 2 -1 x] [ ] ],  our current implementation will generate arc [0 1 x x] [0 2 -1 x] as we start from state `1` which is inaccessible

 [ [ 0 3 -1 1.55377 0 1 8 4.39623 0 1 5 6.84274 ] [ ] [ 2 0 40 0.0509953 ] [ ] ], the correct result should be [ [ 0 1 -1 1.55377 ] [ ] ], note state `0`'s in-degree is not zero in such case, so our algorithms will fail to process it.
```

Anyway, will go forward to fix these bugs.

k2/csrc/top_sort.cu
danpovey: I'm open to changing the type of `keep`, and possible even merging it with the old2new_ vector (we can do the summation in-place).  But let's not do that now.
I don't think it should be necessary to do the sum per sub-list.  Can't you just take differences between the elements of the exclusive-sum?

danpovey:
What does the algorithm do for such input, currently?

k2/csrc/top_sort.cu
qindazhu: Of course we can take the differences, but that also needs we do exclusive-sum then take differences (two kernels).

So I'll keep the current code as-is, let's update the type of `keep` in the future.

qindazhu:
> What does the algorithm do for such input, currently?

You mean for the second example, right? (as I have pasted the result for the first example).

For input `[ [ 0 3 -1 1.55377 0 1 8 4.39623 0 1 5 6.84274 ] [ ] [ 2 0 40 0.0509953 ] [ ] ]`,
our current algorithm will output `[ [ 0 1 40 0.0509953 ] [ 1 3 -1 1.55377 1 2 8 4.39623 1 2 5 6.84274 ] [ ] [ ] ]`
as it first process state `2` (in-degree is 0), then it finds state `0`'s in-degree becomes `0`, so then it processes state `0`

danpovey:
We can make it an error if state 0 is not the start-state, and also if there are cycles.  In that case we can die with an error message I think.

danpovey:
... an early way to detect this error would be:  if state 0 is not in the 1st batch of states, we can make that an error, and we can just return the empty FsaVec.
Actually, right now, if the returned FsaVec has a different num-states than the input, that should be considered an error (we should emphasize this in the documentation).  So just returning the empty FsaVec should be OK.

qindazhu:
Sure, will fix it in the next PR (and may fix other bugs if there're). This PR is ready to merge if you think it's OK.

k2/csrc/fsa_utils.h
```diff
@@ -111,6 +111,18 @@ std::string FsaToString(const Fsa &fsa, bool openfst = false,
...
+                    not exist in the output Fsa (i.e. the dest_state is not in
+                    `order`), then the arc will be removed in the returned
+                    FsaVec.
+                    For each Fsa to be renumbered, if the corresponding number
+                    of states in `order` for that Fsa is not zero, then it must
+                    be at least 2 (one of is the final state of that FSA);
+                    Otherwise, the algorithms will generate invalid FSA, i.e.
+                    the generated Fsa would have only one state. Noted the
+                    algorithm never check this, the caller should make sure
+                    that it must be at least of 2 (if not zero) and one of
```
danpovey, line 124: Haowen, this isn't the desired behavior, this algorithm should make sure never to generate an FSA with exactly one state.  We can detect an error and return  the empty FSA if the start state has arcs entering it that are not from the start state.

k2/csrc/top_sort.cu
```diff
@@ -67,14 +67,19 @@ FsaVec RenumberFsaVec(FsaVec &fsas, const Array1<int32_t> &order,
...
     *arc_map = Array1<int32_t>(c, ans_num_arcs);
     arc_map_data = arc_map->Data();
   } else {
     arc_map_data = nullptr;
   }
+
   const Arc *fsas_arcs = fsas.values.Data();
   Arc *ans_arcs_data = ans_arcs.Data();
+  Renumbering arcs_renumbering(c, ans_num_arcs);
+  // we will remove those arcs whose dest_state is not in `order`,
```
danpovey, line 80: @qindazhu as the algorithm is written, it would be an error if any of the input states were not present in the output, i.e. it would indicate cycles in the input or that the start-state had arcs entering it from a non-start state.  
Therefore I am concerned to see you spending a bunch of effort to deal with this situation.   We can just return the empty FsaVec.  We need to clarify that a result whose size differs from that of the input is an error; we can just always return the empty FsaVec in this situation.

danpovey:
Mm it looks to me that this is more like a rewrite of the algorithm than fixing bugs. 

It is not an error that it processes inaccessible states; that was the behavior I intended, and I think I documented it in the header.   We can separately create a Connect() algorithm that operates on top-sorted input and that removes inaccessible or no-coaccessible states.

If there is an error at all, it is probably in failing to detect an error condition such as an input with cycles or with the start-state having arcs entering it from higher numbered states.

k2/csrc/top_sort.cu
qindazhu: I'm confused here, the documentation said 
https://github.com/k2-fsa/k2/blob/a377250db4b232bd8801c86a2973939325057d6d/k2/csrc/fsa_utils.h#L110
Suppose we only have one Fsa in `fsas`, and the order is `[0 2 3]`, there is an arc [0, 1, x ,x], but as state `1` is not present in the order, don't we need remove this arc? Did I misunderstand the documentation?

qindazhu:
Actually for TopSort, it is valid if there's an arc entering start state, as long as this arc is not in a cycle. Given example below
`[[0 3 x x] [2 0 x x]]`
it's fine here that has an arc from 2 to 0, as state `2` is inaccessible, so it's fine here.

So my original plan to fix it is: add a batch only contains start states as the first batch, then propagate it and together with other states whose in-degree are zeros. so the result will only contain accessible states. Correct me if I'm wrong.

qindazhu:
So we cannot just say it has cycles if the start state is not in the first batch (in current implementation), as the above example shows, there may be inaccessible state entering the start state.

k2/csrc/top_sort.cu
danpovey: Ah OK, this is RenumberFsaVec().

k2/csrc/top_sort.cu
danpovey: Something seems to me not right about what you are saying.  A row-splits vector would naturally come from an exclusive-sum and then taking certain elements...  taking differences doesn't seem right.

k2/csrc/top_sort.cu
danpovey: @qindazhu couldn't we just say it is an error if there are any arcs from states that are kept to those that are not kept, for now, though?  I don't see us ever needing this in the situation where some of the non-kept states have arcs going to them; normally when we do things like pruning we'll do that after top-sorting anyway and we'll do the renumbering ourselves.

danpovey: I think it would be better if when you find bugs that are not trivial, you ask me about it.  I don't tend to write code that needs fundamental rewrites, just code with small bugs.

danpovey:
That is not a cycle per se, but it is a situation that would never happen in an FSA that was generated by an algorithm that only generated top-sorted output (e.g. like a lattice).  So it's safe to treate it as an error condition.  There is no point doing a more complex implementation to handle a case that would never naturally arise.

danpovey:
Actually, including the start-state (and also the final-state, if it has in-degree zero) in the 1st batch might be reasonable- it would naturally remove inaccessible states.  But in this case we will need to add a special mechanism to detect cycles and report an error, because the start-state will be processed even if it has arcs entering it that arise from a cycle.
I'd prefer it if we leave that for later though, as we're making things more complicated than they need to be.  It would be much faster if you ask me what my intention was, when you find things like this, because the gap between the code I wrote and the behavior that I intended tends to be very small and this is the easiest path.

danpovey:
... the problems that you are having, like FSAs that have exactly one states, are problems that I previously thought about while considering algorithms like the one you are implementing.  You are going over ground that I went over before writing this.

qindazhu:
OK, so could we limit that For Fsa in k2: there is no arcs entering the start state (except self-loops)? I think it would be clear if we can add it in documentation so every algorithm can following it (even algorithms in host, and also RandFsaVec, etc.)

k2/csrc/top_sort.cu
qindazhu: OK. And just wondering why not we just require that the orders.size must be equal to fas.TotSize(1), I just feel it's hard for users to remove those arcs with dest-state not kept, I don't see the situation.

danpovey:
No, we can't require that of all FSAs because in FSAs that are not naturally top-sorted, such as decoding graphs, it can quite legitimately happen.

qindazhu:
so we are supposing the TopSort will be just used in lattice related Fas? as For other cases, an input `[[0 3 -1 x] [2 0 x x]]` is valid and should be sorted to `[[0 1 -1 x]]`

k2/csrc/top_sort.cu
danpovey: I suppose I wanted it as general as possible.  How about for now we don't require that order.Dim() == fsas.TotSize(1), but do say that it's an error if there are arcs from kept to not-kept states.  We can relax that restriction later if we ever have a case where that can happen.

danpovey:
Well it's only used for things that are acyclic.  The example you are using is somehow "un-physical"... I can't think of a situation where we would ever get that.  Anyway I thought about all of these issues and possible problems before deciding to write the algorithm in this way -> better to just fix any small bugs and keep moving, if we ever need to handle your case we can deal with it later.

k2/csrc/top_sort.cu
qindazhu: OK, will do (and also update the documentation)

danpovey:
Thanks!

qindazhu:
> Well it's only used for things that are acyclic. The example you are using is somehow "un-physical"... I can't think of a situation where we would ever get that. Anyway I thought about all of these issues and possible problems before deciding to write the algorithm in this way -> better to just fix any small bugs and keep moving, if we ever need to handle your case we can deal with it later.

Ok, will just return error to indicate there is cyclic (not self-loops), i.e. when `all_states.values.Dim()` in `TopSort` is not equal to `fsas_.TotSize(1)`,  is that OK?

danpovey:
OK.

qindazhu:
@danpovey ,Updated, please have a review, thanks!

k2/csrc/top_sort.cu
```diff
@@ -337,6 +363,47 @@ class TopSorter {
 
     std::vector<std::unique_ptr<Ragged<int32_t>>> iters;
     iters.push_back(GetInitialBatch());
+    // the start state in each Fsa must be present in the first batch
```
danpovey, line 366: If this extra code is purely checking code it should be clearly marked as such, so that a person trying to understand this function does not waste time trying to understand it.
I would recommend that you (preferably) move it to a separate function, or at least put it in a code block like as follows:
```
  {  // This block just checks that all non-empty FSAs in the input have their start state in their
     // first batch.
  }
```
However, I think it would be easy to check this in a single kernel, probably with num_fsas as its 'n' parameter
(you could locate the 1st state in the batch for each FSA and see if it's the start state).
You could just do a K2_CHECK_EQ inside the kernel if you want to make it easy to write.

k2/csrc/top_sort_test.cu
```diff
@@ -198,14 +198,17 @@ TEST(TopSort, RandomVectorOfFsas) {
...
       K2_LOG(INFO) << n;
       // We have to generate epsilon-free Fsas here as we will call Intersect
       // below (in `IsRandEquivalent`) which requires at least one of the two
       // input Fsas is epsilon-free.
-      FsaVec random_fsas = RandomFsaVec(1, 1000, false, true);
+      // TODO(haowen): here we generate acyclic Fsas for testing as current
+      // implementation of GPU TopSort supposes the input Fsas are acyclic, but
+      // definitely I need to `reorder` states randomly and add some self-loops
+      // for robust test.
+      FsaVec random_fsas = RandomFsaVec(1, 1000, true, true);
```
danpovey, line 211: We could use RenumberFsa to do this, perhaps?  Would be easiest to generate the reordering randomly on CPU (not touching start+final states, of course).

qindazhu: yes, that's exactly what I was planning to do, but not in this PR.

k2/csrc/top_sort.cu
qindazhu: Oh, yes, as our implementation of `GetInitBatch` will always keep the start state (if its indegree is zero) as the 1st state. Will update, thanks!

qindazhu: Updated as you said, OK to merge if there's no other issues, will go to fix the bug in the general `IsRandEquivalent` you metioned

k2/csrc/top_sort_test.cu
```diff
@@ -241,15 +244,16 @@ TEST(TopSort, RandomVectorOfFsas1) {
...
-        K2_LOG(INFO) << sorted_i;
-        K2_LOG(INFO) << host_sorted_i;
         EXPECT_EQ(properties[i] & gt, gt);
         EXPECT_TRUE(is_top_sorted[i]);
+        if (!IsRandEquivalent(sorted_i, host_sorted_i, false)) {
+          K2_LOG(INFO) << cpu_fsa_vec_i;
+          K2_LOG(INFO) << sorted_i;
+          K2_LOG(INFO) << host_sorted_i;
+        }
         EXPECT_TRUE(IsRandEquivalent(sorted_i, host_sorted_i, false));
```
danpovey, line 254: I'd prefer to have a KALDI_LOG(FATAL).... inside the if-statement.    IsRandEquivalent() is slow, so doing it twice is a bit wasteful.

qindazhu: mm, as I commented in the last PR about bug fixes in TopSort, these code is just for debugging, I definitely will remove those LOG after debugging. So please ignore this and I will remove those LOG stuff after add tests with `RenumberFsaVec`

PR #296, opened 2020-10-31
Change label comparisons to uint32_t as fix for host intersect; bug f…

danpovey:
…ix in intersect.

### DIFF
--- a/k2/csrc/host/intersect.cc
+++ b/k2/csrc/host/intersect.cc
@@ -132,7 +132,8 @@ void Intersection::GetSizes(Array2Size<int32_t> *fsa_size) {
       auto b_arc_range =
           std::equal_range(b_arc_iter_begin, b_arc_iter_end, *a_arc_iter_begin,
                            [](const Arc &left, const Arc &right) {
-                             return left.label < right.label;
+                             return static_cast<uint32_t>(left.label) <
+                                    static_cast<uint32_t>(right.label);
                            });
       for (auto it_b = b_arc_range.first; it_b != b_arc_range.second; ++it_b) {
         Arc curr_a_arc = *a_arc_iter_begin;  // copy here as we may swap later
--- a/k2/csrc/host/intersect.cc
+++ b/k2/csrc/host/intersect.cc
@@ -52,11 +52,9 @@ void Intersection::GetSizes(Array2Size<int32_t> *fsa_size) {
 
   if (IsEmpty(a_) || IsEmpty(b_)) return;
   // either `a` or `b` must be epsilon-free, both of them should be arc-sorted
-  status_ = IsArcSorted(a_) && IsArcSorted(b_) &&
-      (IsEpsilonFree(a_) || IsEpsilonFree(b_));
-  if (!status_) {
-    K2_LOG(WARNING) << "Either one of the inputs is not arc-sorted or "
-        "both are not epsilon-free.";
+  if (!IsArcSorted(a_) || !IsArcSorted(b_)) {
+    K2_LOG(WARNING) << "One of the inputs is not arc-sorted";
+    status_ = false;
     return;
   }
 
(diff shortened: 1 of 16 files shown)


danpovey:
Guys 
I need help with this.
I changed the arc-sorting in both host and GPU so that we treat labels as uint32_t for purposes of arc-sorting.
That fixes a likely bug in the host intersection code where it expected the epsilon arcs to be first.
However it creates a test failure in rmepsilon_test.  I suspect a pre-existing problem.  Do you think I should merge this and you try to fix the rmepsilon issue later?

qindazhu:
Sure, just merge it and I'll fix it in my next PR. I also think it's a old bug, as IIRC I don't rely on the property (i.e. kFinalSymbol is the first or the last) when I implemented rm-epsilon.

k2/csrc/host/intersect.cc
```diff
@@ -91,35 +89,36 @@ void Intersection::GetSizes(Array2Size<int32_t> *fsa_size) {
...
-      arcs_.emplace_back(curr_state_index, new_state_index, kEpsilon,
-                         b_arc_iter_begin->weight);
-      arc_map_a_.push_back(arc_map_none);
-      arc_map_b_.push_back(
-          static_cast<int32_t>(b_arc_iter_begin - arc_b_begin));
-    }
+    if (treat_epsilons_specially_) {
+      // As both `a` and `b` are arc-sorted, we first process epsilon arcs.
+      // Noted that at most one for-loop below will really run as either `a` or
+      // `b` is epsilon-free.
```
qindazhu, line 95: @danpovey , if `treat_epsilons_specially_ = true`, I think we should make sure here that at least one of `a` and `b` is epsilon-free.

qindazhu:
@danpovey,need your help here. I'm definitely sure the test failure above in rm_epsilon_test is because of our implementation of IsRandomEquivalentAfterRmEpsPrunedLogSum (I have checked the result of `rmepsilon`, it's correct).

I just feel that our documentation on `IsRandEquivalentAfterRmEpsPrunedLogSum` is not correct?

https://github.com/k2-fsa/k2/blob/762beb588e7d29c582658024718220981683ca09/k2/csrc/host/fsa_equivalent.h#L75-L81

For `1`, shouldn't it be :
any path in `b` that is within `beam` of the total weight of `b` should  be present in `a` with a weight that is not **less** than its weight in `b`?

And also, seems my implementation is not correct, as I generated a path, then intersect, then get the path weight and to check if it is in the beam. I think I should get (and check if it's in beam) the weight when generating the path in `a` or `b`, am I right?

qindazhu:
And also I wondering if we can implement it with `Prune` first then call the normal `IsRandEquivalent`?

k2/csrc/host/intersect.cc
danpovey: I decided to remove that check because there are situations where it produces the correct output, e.g. with max semiring you don't mind the extra paths with different orderings of epsilons.

danpovey:
> @danpovey,need your help here. I'm definitely sure the test failure above in rm_epsilon_test is because of our implementation of IsRandomEquivalentAfterRmEpsPrunedLogSum (I have checked the result of `rmepsilon`, it's correct).
> 
> I just feel that our documentation on `IsRandEquivalentAfterRmEpsPrunedLogSum` is not correct?
> 
> https://github.com/k2-fsa/k2/blob/762beb588e7d29c582658024718220981683ca09/k2/csrc/host/fsa_equivalent.h#L75-L81
> 
> For `1`, shouldn't it be :
> any path in `b` that is within `beam` of the total weight of `b` should be present in `a` with a weight that is not **less** than its weight in `b`?
That sounds right.

 
> And also, seems my implementation is not correct, as I generated a path, then intersect, then get the path weight and to check if it is in the beam. I think I should get (and check if it's in beam) the weight when generating the path in `a` or `b`, am I right?

You should always check the weight only after intersection.  We care about the weight assigned to a symbol sequence, not the weight assigned to a path (a WFSA is a weighted collection of symbol sequences and for equivalence we care about the weights for symbol sequences).

danpovey:
> And also I wondering if we can implement it with `Prune` first then call the normal `IsRandEquivalent`?

I'm not sure about Prune() then IsRandEquivalent(), it doesn't seem right.  The intersection is using a kind of heuristic.. I don't think it gives the kinds of guarantees we'd need for that test to work.

PR #297, opened 2020-11-01
Various fixes in intersecton code

danpovey:


danpovey:
Guys, I'm going to merge this; let me know if you see any problems though.
The test for intersection is failing, but I believe it's  a bug in RandEquivalentTest.  I am hoping someone else can figure that out.
It is saying that the two FSAs below are not equivalent:
```
[ [ ] ]
 [ [ [ 0 1 0 8.27126 ] [ 1 2 10 9.2197 ] [ 2 3 10 8.5597 ] [ ] [ ] ] ]
```
which if you look carefully, you'll see they are both equivalent to the empty FSA; note the [] [] in the second one, meaning the final state is state 4 with no incoming arcs.
Sorry to have not been that quick to merge your PRs.

danpovey:
I have made other changes here, including making it so arc-sorting does not sort on the weights (this is so that our ArcSort() routine in csrc/ does not create an FSA that IsArcSorted() fails in the host code).

qindazhu:
> Guys, I'm going to merge this; let me know if you see any problems though.
> The test for intersection is failing, but I believe it's a bug in RandEquivalentTest. I am hoping someone else can figure that out.
> It is saying that the two FSAs below are not equivalent:
> 
> ```
> [ [ ] ]
>  [ [ [ 0 1 0 8.27126 ] [ 1 2 10 9.2197 ] [ 2 3 10 8.5597 ] [ ] [ ] ] ]
> ```
> 
> which if you look carefully, you'll see they are both equivalent to the empty FSA; note the [] [] in the second one, meaning the final state is state 4 with no incoming arcs.
> Sorry to have not been that quick to merge your PRs.

Well, actualy it's not a bug, it's an design issue because now do allow there's epsilon arc in intersect. 
You call the normal case `IsRandEqivalent` here (I mean, the version for Fsa instead of WFsa), which calls below host version internally,
https://github.com/k2-fsa/k2/blob/e527c3c9f1b2abd76be8b30d81ea164a71f1d534/k2/csrc/host/fsa_equivalent.cc#L141-L146

The above test case has epsilon arc, so `IsRandEquivalent` will fail at the first line.

PR #303, opened 2020-11-03
Reorder Fsas in TopSort tests and fixed some other issues

qindazhu:
- Reorder states in TopSort tests and randomly add some self-loops.
- Fixed a small bug in `Array.Arange`
- Remvoed parameter `epsilon_free` in `RandomFsa(Vec)` as now we don't require epsilon-free input Fsas for `intersection`.
- Top-sort the composed path in `IsRandEquivalent` (the WFST version) as we may get non-top-sorted output if both inputs are non-epsilon-free.


csukuangfj:
+1

PR #304, opened 2020-11-03
Implement backprop for log-seming in Python

csukuangfj:
TODO lists:

- [x] Update `ShortestPath` to reuse entries `Fsa._cache`.
- [x] Implement a `Function` class supporting

```python
def get_tot_scores(fsa: Fsa, is_log_semiring: bool)
```

which will update/reuse `fsa._cache` if needed.

### DIFF
--- a/k2/csrc/fsa_algo.h
+++ b/k2/csrc/fsa_algo.h
@@ -230,23 +230,18 @@ FsaVec LinearFsas(Ragged<int32_t> &symbols);
                  property kFsaPropertiesTopSortedAndAcyclic if you were
                  to compute properties.
 
-   @param [out,optional] entering_arcs   If non-NULL, it will be set to
-                a new Array1, indexed by state_idx01 into `fsas`,
-                saying which arc_idx012 is the best arc entering it,
-                or -1 if there is no such arc.
-
-   @param [out,optional]
-                    If non-NULL, it will contain array of total scores,
-                    of dimension fsas.Dim0(), which will contain the scores
-                    in the final-states of `forward_scores`, or -infinity for
-                    FSAs that had no states.
+   @param [in] entering_arcs  An array indexed by state-index (idx01 into
+                              fsas), i.e.
+                              `entering_arcs.Dim()==fsas.TotSize(1)`,
+                              containing forward scores. (these will be zero
+                              for the start-states). It is returned by
+                              GetForwardScores with log_semiring==false.
 
    @return returns a tensor with 2 axes indexed by [fsa_idx0][arc_idx012]
            containing the best arc indexes of each fsa.
  */
 Ragged<int32_t> ShortestPath(FsaVec &fsas,
-                             Array1<float> *total_scores = nullptr,
-                             Array1<int32_t> *entering_arcs = nullptr);
+                             const Array1<int32_t> &entering_arcs);
 
 }  // namespace k2
 
--- a/k2/csrc/fsa_algo.h
+++ b/k2/csrc/fsa_algo.h
@@ -98,7 +98,6 @@ void TopSort(FsaVec &src, FsaVec *dest, Array1<int32_t> *arc_map = nullptr);
  */
 bool HostTopSort(Fsa &src, Fsa *dest, Array1<int32_t> *arc_map = nullptr);
 
-
 /*
   Add epsilon self-loops to an Fsa or FsaVec (this is required when composing
   using a composition method that does not treat epsilons specially, if the
@@ -230,12 +229,9 @@ FsaVec LinearFsas(Ragged<int32_t> &symbols);
                  property kFsaPropertiesTopSortedAndAcyclic if you were
                  to compute properties.
 
-   @param [in] entering_arcs  An array indexed by state-index (idx01 into
-                              fsas), i.e.
-                              `entering_arcs.Dim()==fsas.TotSize(1)`,
-                              containing forward scores. (these will be zero
-                              for the start-states). It is returned by
-                              GetForwardScores with log_semiring==false.
+   @param [in] entering_arcs   An array indexed by state_idx01 into `fsas`,
+                saying which arc_idx012 is the best arc entering it,
+                or -1 if there is no such arc.
 
    @return returns a tensor with 2 axes indexed by [fsa_idx0][arc_idx012]
            containing the best arc indexes of each fsa.
--- a/k2/csrc/fsa_algo_test.cu
+++ b/k2/csrc/fsa_algo_test.cu
@@ -340,7 +340,19 @@ TEST(FsaAlgo, ShortestPath) {
     FsaVec fsa_vec = CreateFsaVec(3, &fsa_array[0]);
     fsa_vec = fsa_vec.To(context);
 
-    Ragged<int32_t> best_path_arc_indexes = ShortestPath(fsa_vec);
+    Ragged<int32_t> state_batches = GetStateBatches(fsa_vec, true);
+    Array1<int32_t> dest_states = GetDestStates(fsa_vec, true);
+    Ragged<int32_t> incoming_arcs = GetIncomingArcs(fsa_vec, dest_states);
+    Ragged<int32_t> entering_arc_batches =
+        GetEnteringArcIndexBatches(fsa_vec, incoming_arcs, state_batches);
+
+    bool log_semiring = false;
+    Array1<int32_t> entering_arcs;
+    GetForwardScores<float>(fsa_vec, state_batches, entering_arc_batches,
+                            log_semiring, &entering_arcs);
+
+    Ragged<int32_t> best_path_arc_indexes =
+        ShortestPath(fsa_vec, entering_arcs);
     CheckArrayData(best_path_arc_indexes.values,
                    std::vector<int32_t>{1, 3, 5, 10, 13, 16, 17, 19, 21, 23});
 
--- a/k2/csrc/fsa_utils.cu
+++ b/k2/csrc/fsa_utils.cu
@@ -1090,7 +1090,7 @@ Array1<FloatType> GetForwardScores(FsaVec &fsas, Ragged<int32_t> &state_batches,
       entering_arc_start_index.To(GetCpuContext());
   const int32_t *cpu_entering_arc_start = cpu_entering_arc_start_index.Data();
   // copy the index of start state in each fsa to CPU
-  Array1<int32_t> arc_batches_row_splits1_array =
+  Array1<int32_t> &arc_batches_row_splits1_array =
       entering_arc_batches.RowSplits(1);
   Array1<int32_t> cpu_state_idx0xx =
       entering_arc_batches.RowSplits(2)[arc_batches_row_splits1_array].To(
--- a/k2/csrc/fsa_utils.cu
+++ b/k2/csrc/fsa_utils.cu
@@ -1092,11 +1092,12 @@ Array1<FloatType> GetForwardScores(FsaVec &fsas, Ragged<int32_t> &state_batches,
   // copy the index of start state in each fsa to CPU
   Array1<int32_t> &arc_batches_row_splits1_array =
       entering_arc_batches.RowSplits(1);
-  Array1<int32_t> cpu_state_idx0xx =
+  Array1<int32_t> arc_batches_row_splits12_cpu =
       entering_arc_batches.RowSplits(2)[arc_batches_row_splits1_array].To(
           GetCpuContext());
-  K2_CHECK_EQ(cpu_state_idx0xx.Dim(), num_batches + 1);
-  const int32_t *cpu_state_idx0xx_data = cpu_state_idx0xx.Data();
+  K2_CHECK_EQ(arc_batches_row_splits12_cpu.Dim(), num_batches + 1);
+  const int32_t *arc_batches_row_splits12_cpu_data =
+      arc_batches_row_splits12_cpu.Data();
   Array1<int32_t> arc_row_splits_mem(c, num_states + 1);
   Array1<FloatType> score_cache(c, num_states + 1);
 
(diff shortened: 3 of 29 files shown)


k2/python/k2/fsa.py
```diff
@@ -109,9 +109,50 @@ def __init__(self,
...
+        #
+        # - `_tensor_attr`
+        #     It saves attribute values of type torch.Tensor. `shape[0]` of
+        #     attribute values have to be equal to the number of arcs
+        #     in the FSA.
+        #
+        # - `_non_tensor_attr`
+        #     It saves non-tensor attributes, e.g., :class:`SymbolTable`.
+        #
+        # - `_cache`
```
qindazhu, line 122: how about using `_grad_cache` or other more specifical name?

k2/csrc/fsa_utils.cu
```diff
@@ -1090,7 +1090,7 @@ Array1<FloatType> GetForwardScores(FsaVec &fsas, Ragged<int32_t> &state_batches,
       entering_arc_start_index.To(GetCpuContext());
   const int32_t *cpu_entering_arc_start = cpu_entering_arc_start_index.Data();
   // copy the index of start state in each fsa to CPU
-  Array1<int32_t> arc_batches_row_splits1_array =
+  Array1<int32_t> &arc_batches_row_splits1_array =
       entering_arc_batches.RowSplits(1);
   Array1<int32_t> cpu_state_idx0xx =
```
danpovey, line 1095: I would call this like arc_batches_row_splits12_cpu

danpovey: ... I see this is pre-existing code...

k2/python/csrc/torch/fsa.cu
```diff
@@ -111,11 +122,109 @@ static void PybindFsaUtil(py::module &m) {
...
+        py::arg("fsas"), py::arg("state_batches"));
+}
+
+template <typename T>
+static void PybindGetForwardScores(py::module &m, const char *name) {
+  // Return a std::pair
+  //   - forward_scores, a torch::Tensor of dtype torch.float32 or torch.float64
+  //   (depending on T) containing the scores
+  //
+  //   - best_path_arc_indexes (optional)
```
danpovey, line 171: mm.. this was called entering_arcs.  I don't like the name best_path_arc_indexes; this is traceback information.

k2/python/csrc/torch/fsa.cu
```diff
@@ -124,21 +233,23 @@ static void PybindDenseFsaVec(py::module &m) {
...
-        dense_fsa->shape = RaggedShape2(&_row_splits, nullptr, -1);
-        dense_fsa->scores = FromTensor<float>(scores, Array2Tag{});
-
-        K2_CHECK(IsCompatible(dense_fsa->shape, dense_fsa->scores));
-
-        return dense_fsa;  // Python takes the ownership
-      }));
+  pyclass.def(py::init([](torch::Tensor scores,
+                          torch::Tensor row_splits) -> DenseFsaVec * {
+                // remove the contiguous check once the following comment
```
danpovey, line 238: I have added a constructor recently, I think.

k2/python/csrc/torch/torch_util.h
```diff
@@ -47,6 +47,7 @@ struct ToScalarType;
 
 // TODO(fangjun): add other types if needed
 TO_SCALAR_TYPE(float, torch::kFloat);
```
danpovey, line 49: The comments around ToScalarType do not really explain what it is or what it's needed for.

csukuangfj: It is used to convert a c type to an instance of `torch::ScalarType`. For example

```cpp
static_assert(torch::kFloat == ToScalarType<float>::value>, "");
static_assert(torch::kInt == ToScalarType<int32_t>::value>, "");

```

csukuangfj: Its usages are like the following:

https://github.com/k2-fsa/k2/blob/4efa9d21acb3798e33bb9b53efa743cff9b8f3f7/k2/python/csrc/torch/torch_util.cu#L77

k2/python/k2/fsa.py
```diff
@@ -109,9 +109,50 @@ def __init__(self,
...
+        #           returned by :func:`_k2._get_state_batches`
+        #  - `dest_states`:
+        #           returned by :func:`_k2._get_dest_states`
+        #  - `incoming_arcs`:
+        #           returned by :func:`_k2._get_incoming_arcs`
+        #  - `entering_arc_batches`:
+        #           returned by :func:`_k2._get_entering_arc_index_batches`
+        #  - `leaving_arc_batches`:
+        #           returned by :func:`_k2._get_leaving_arc_index_batches`
+        #  - `forward_scores_tropical_semiring`:
```
danpovey, line 141: these names are getting long, maybe remove _semiring

k2/python/csrc/torch/fsa.cu
csukuangfj: `entering_arcs` contains arc indexes for every state in the FsaVec, 
while `best_path_arc_indexes` contains arc indexes ONLY for the states that are on the best path.

csukuangfj: Will change it.

csukuangfj: Fixed.

k2/csrc/fsa_utils.cu
csukuangfj: Renamed. This file is auto formatted by `clang-format`.

k2/python/csrc/torch/fsa.cu
csukuangfj: Updated.

k2/python/k2/fsa.py
csukuangfj: Fixed.

csukuangfj:
Renamed to `_grad_cache`.

k2/python/csrc/torch/torch_util.h
csukuangfj: I think its meaning is clear from its definition:
https://github.com/k2-fsa/k2/blob/b118afd2be4dc8eded911d25c90e858dec63e8b7/k2/python/csrc/torch/torch_util.h#L43-L46

k2/csrc/fsa_algo.h
```diff
@@ -230,23 +230,18 @@ FsaVec LinearFsas(Ragged<int32_t> &symbols);
...
-
-   @param [out,optional]
-                    If non-NULL, it will contain array of total scores,
-                    of dimension fsas.Dim0(), which will contain the scores
-                    in the final-states of `forward_scores`, or -infinity for
-                    FSAs that had no states.
+   @param [in] entering_arcs  An array indexed by state-index (idx01 into
+                              fsas), i.e.
+                              `entering_arcs.Dim()==fsas.TotSize(1)`,
+                              containing forward scores. (these will be zero
```
qindazhu, line 236: it should be `containing the best arc entering it (state_idx01)`, not `forward scores`

k2/csrc/fsa_algo_test.cu
```diff
@@ -340,7 +340,19 @@ TEST(FsaAlgo, ShortestPath) {
...
-    Ragged<int32_t> best_path_arc_indexes = ShortestPath(fsa_vec);
+    Ragged<int32_t> state_batches = GetStateBatches(fsa_vec, true);
+    Array1<int32_t> dest_states = GetDestStates(fsa_vec, true);
+    Ragged<int32_t> incoming_arcs = GetIncomingArcs(fsa_vec, dest_states);
+    Ragged<int32_t> entering_arc_batches =
+        GetEnteringArcIndexBatches(fsa_vec, incoming_arcs, state_batches);
+
+    bool log_semiring = false;
+    Array1<int32_t> entering_arcs;
+    GetForwardScores<float>(fsa_vec, state_batches, entering_arc_batches,
```
qindazhu, line 351: should we create a wrapper function to get `forward_scores` and `entering_arcs`? I just think maybe we never use above values (e.g. `state_batches`, `dest_states`...) directly in most of calling place, they are just used to call `GetForwardScores`

k2/python/tests/autograd_test.py
```diff
@@ -0,0 +1,294 @@
...
+            1 2 3 0.3
+            1 3 4 0.4
+            2 3 5 0.5
+            3 4 -1 0
+            4
+        '''
+        fsa = k2.Fsa.from_str(s)
+        fsa.scores.requires_grad_(True)
+        fsa_vec = k2.create_fsa_vec([fsa])
+        log_like = k2.get_forward_log_like(fsa_vec,
```
danpovey, line 183: It's great that you have done the autograd tests!
I'm not sure that I like the name `get_forward_log_like`.  I think I'd prefer `get_tot_scores` which is more consistent with the C++ code.  I particularly don't like the 'forward' part of the name, because the log_like is just a property of the FSA (+semiring), and we could just as easily have chosen to implement it via backward.
In fact, it might make sense to do this (implement via backward_loglikes), because that code is likely faster than the forward code.  [We'd have to add a tot_scores arg to the forward-pass to make this work, and also a function to get the tot-scores from the start states, so for now leave it as-is.]

k2/csrc/fsa_algo_test.cu
csukuangfj: The advantage of calling them separately is that `state_batches` can be reused
across different calls, e..g, `GetForwardScores`,  `GetEnteringArcIndexBatches`, etc.

k2/csrc/fsa_algo.h
csukuangfj: thanks, fixed.

k2/python/tests/autograd_test.py
csukuangfj: Renamed to `get_tot_scores`.

csukuangfj:
Ready to merge if there are no other comments.

k2/csrc/fsa_algo_test.cu
qindazhu: I didn't mean to put them in `GetForwardScores` or `ShortestPath`, I just mean that we can define a helper function to wrap those things, then we can just call this helper function in the case where we just want to get `forward_scores` or `entering_arcs`;

k2/python/csrc/torch/fsa.cu
```diff
@@ -124,21 +233,21 @@ static void PybindDenseFsaVec(py::module &m) {
   // We do not need to access its members in Python
 
   // TODO(fangjun): add docstring for this funciton
-  pyclass.def(py::init(
-      [](torch::Tensor scores, torch::Tensor row_splits) -> DenseFsaVec * {
+  pyclass.def(
+      py::init([](torch::Tensor scores,
+                  torch::Tensor row_splits) -> std::unique_ptr<DenseFsaVec> {
```
qindazhu, line 238: just a question, why is `row_splits` here instead of `RaggedShape`?

csukuangfj: `row_splits`, `sizes`, and `shape` are all fine here.
`RaggedShape` has not been wrapped to Python since I don't know when and where we need `RaggedShape` in Python, so `row_splits` is used here. `sizes` requires an extra call to `ExclusiveSum`.

danpovey:
OK for me to merge.

k2/python/tests/autograd_test.py
```diff
@@ -0,0 +1,294 @@
...
+            3
+        '''
+
+        fsa1 = k2.Fsa.from_str(s1)
+        fsa2 = k2.Fsa.from_str(s2)
+        fsa3 = k2.Fsa.from_str(s3)
+
+        fsa1.scores.requires_grad_(True)
+        fsa2.scores.requires_grad_(True)
+        fsa3.scores.requires_grad_(True)
```
qindazhu, line 116: seems we need to define a function to do something like `FsaVec.required_grad_(True)`?

danpovey: The scores are the only differentiable part of it, so if anything it would just be a wrapper doing fsa1.scores.requires_grad_(True)

csukuangfj: So do we need `requires_grad_` ?

danpovey: I think it's normal for that option to be supported.  I don't think we'll really be saving any work by having requires_grad_ be false, but also we need to propagate it appropriately as later operations may need it.  So: yes, we need it.

k2/python/tests/autograd_test.py
```diff
@@ -0,0 +1,294 @@
...
+        '''
+        fsa = k2.Fsa.from_str(s)
+        fsa.scores.requires_grad_(True)
+        fsa_vec = k2.create_fsa_vec([fsa])
+        log_like = k2.get_tot_scores(fsa_vec,
+                                     log_semiring=True,
+                                     use_float_scores=True)
+        assert log_like.dtype == torch.float32
+        # The expected_log_like is computed using gtn.
+        # See https://bit.ly/3oUiRx9
```
qindazhu, line 188: what not just using the host version of `GetForwardScores` to compute this? (we have wrapped it to `csrc/host_shim.h`). BTW, for grad stuff, we may need to create a helper function with `GetTotScore()` (a host version should be fine) for (automatic) test purpose.

danpovey: If you are just talking about the  test code, I think we can just write extra tests later.

csukuangfj: I just wanted to compare that `k2` computes the same gradient values as `gtn` when
they are given the same input FSA.

k2/python/csrc/torch/fsa.cu
qindazhu: OK, thanks for the explanation. Just noted that when we pass `row_splits`, we may do a Device2Host copy in this line of this wraper function for `cached_tot_size`
https://github.com/k2-fsa/k2/blob/24afc76e23c2db4bd089f74d241ac5f82fe2fc5d/k2/python/csrc/torch/fsa.cu#L245

k2/python/tests/autograd_test.py
qindazhu: > If you are just talking about the test code, I think we can just write extra tests later.

Yeah, sure, what I mean is also not doing this now.

qindazhu: > I just wanted to compare that `k2` computes the same gradient values as `gtn` when
> they are given the same input FSA.

What I thought is, if we use gtn for test purpose (not experiment purpose) in a low level, it may make people feel that the correctness of  k2 is relying on gtn. And I'm not even sure if gtn handles those `inifity` stuff correctly with fast-math.

But, anyway, it's OK for now to leave as-is, since this is just a simple test case.

qindazhu:
+2, @csukuangfj can merge if you think it's fine.

k2/python/tests/autograd_test.py
csukuangfj: Added.

csukuangfj:
Merging.

k2/python/csrc/torch/fsa.cu
csukuangfj: If we pass a RaggedShape, there would also be a copy, but it happens in Python.

PR #305, opened 2020-11-03
Add extra test case to intersect_test.cu; bug fix.

danpovey:
Fixes to intersection, more tests

### DIFF
--- a/k2/csrc/host/fsa_equivalent.cc
+++ b/k2/csrc/host/fsa_equivalent.cc
@@ -244,23 +244,20 @@ bool IsRandEquivalent(const Fsa &a, const Fsa &b,
        (*(labels_difference.begin())) != kEpsilon))
     return false;
 
-  double loglike_cutoff_a, loglike_cutoff_b;
-  if (beam != kFloatInfinity) {
-    loglike_cutoff_a = ShortestDistance<Type>(valid_a) - beam;
-    loglike_cutoff_b = ShortestDistance<Type>(valid_b) - beam;
-    if (Type == kMaxWeight &&
-        !DoubleApproxEqual(loglike_cutoff_a, loglike_cutoff_b))
-      return false;
-  } else {
-    loglike_cutoff_a = kDoubleNegativeInfinity;
-    loglike_cutoff_b = kDoubleNegativeInfinity;
-  }
+  double dist_a = ShortestDistance<Type>(valid_a),
+      dist_b = ShortestDistance<Type>(valid_b),
+      loglike_cutoff_a = dist_a - beam,
+      loglike_cutoff_b = dist_b - beam;
+  if (abs(dist_a - dist_b) > delta)
+    return false;
 
   std::random_device rd;
   std::mt19937 gen(rd());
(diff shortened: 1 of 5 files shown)


k2/csrc/host/fsa_equivalent.cc
```diff
@@ -244,23 +244,20 @@ bool IsRandEquivalent(const Fsa &a, const Fsa &b,
...
 
   std::random_device rd;
   std::mt19937 gen(rd());
   std::bernoulli_distribution coin(0.5);
-  std::size_t n = 0;
-  while (n < npath) {
+
+  for (std::size_t n = 0; n < npath; ++n) {
+    // Note: for now we `continue` when paths are not above the beam,
+    // which may mean that for some FSAs
```
qindazhu, line 260: seems incomplete comments?

k2/csrc/host/fsa_equivalent.cc
```diff
@@ -288,8 +285,6 @@ bool IsRandEquivalent(const Fsa &a, const Fsa &b,
     if (cost_a < loglike_cutoff_a && cost_b < loglike_cutoff_b) continue;
```
qindazhu, line 285: As we allow epsilon arc in path, I believe `RandomPath(fsa, false, &valid_path_storage)` will never fail. So is that better to add a value `tries` here for `if (cost_a < loglike_cutoff_a && cost_b < loglike_cutoff_b) continue` and return true with a warning LOG when, e.g. `tries > 50`. I just feel the default `npath = 100` now will not be a good default value for most of large size Fsas.

k2/csrc/intersect_test.cu
```diff
@@ -30,8 +46,8 @@ TEST(Intersect, Simple) {
   // clang-format off
   DenseFsaVec dfsavec {
     RaggedShape("[ [ x x x ] ]"),
-    Array2<float>("[ [ -Inf 0.1 0.2 0.3 ] [ -Inf 0.04 0.05 0.06 ] [ 1.0 -Inf -Inf -Inf]]")  // NOLINT
-  };
+        Array2<float>("[ [ -Inf 0.1 0.2 0.3 ] [ -Inf 0.04 0.05 0.06 ] [ 1.0 -Inf -Inf -Inf]]")  // NOLINT
```
csukuangfj, line 49: Should the arc entering the final state have `score == 0` instead of 1.0 ?

danpovey: There isn't a special constraint here (it would have that value, though, if it came from a normal log-like matrix; I wanted to test something though).

k2/csrc/host/fsa_equivalent.cc
danpovey: Mm the right fix here, I think, is to use a rand-path algorithm that does not generate paths outside the beam.  I dont want changes that will make this super slow, thats why I did it like this.

k2/csrc/host/fsa_equivalent.cc
danpovey: OK, will fix it later.

PR #306, opened 2020-11-04
Write and test function to add epsilon self-loops

danpovey:


danpovey:
test code is not great, just based on printing, but I've tested the cases I was worried about...

PR #316, opened 2020-11-08
Replace Renumber() with more general Index() function

danpovey:


danpovey:
Also fix a bug or two, add move constructors; add a few new functions in Array2.

k2/csrc/ragged_ops.h
```diff
@@ -662,6 +665,64 @@ bool Equal(Ragged<T> &a, Ragged<T> &b) {
...
+              `ans.NumAxes() == src.NumAxes()`
+              and `ans.Dim0() == indexes.Dim()`.
+
+*/
+template <typename T>
+Ragged<T> Index(Ragged<T> &src, const Array1<int32_t> &indexes,
+                Array1<int32_t> *value_indexes_out = nullptr) {
+  Array1<int32_t> value_indexes;
+  RaggedShape ans_shape = Index(src.shape, indexes, &value_indexes);
+  return Ragged<T>(ans_shape,
```
qindazhu, line 718: `return` should be the last line.

csukuangfj: This has been fixed in my opening pullrequest #318

k2/csrc/ragged_ops.h
```diff
@@ -662,6 +665,64 @@ bool Equal(Ragged<T> &a, Ragged<T> &b) {
...
+                   src.values[value_indexes]);
+  if (value_indexes_out)
+    *value_indexes_out = std::move(value_indexes);
```
csukuangfj, line 721: There is no move assignment operator for `Array1<T>`, so this actually invokes the assignment operator. That is, nothing is moved here. Fixed in #318

PR #322, opened 2020-11-09
Add cuda support to GitHub actions

### LINKED ISSUE #319: build pip package automatically for dev purpose?
@csukuangfj Can you kindly build a latest pip package with master branch? And wondering if there's any way we can create pip package automatically every day (or when master is updated), for dev purpose for now?

csukuangfj:
Closes #319

Wheels are automatically built for every pullrequest and are saved on GitHub for downloading for 90 days.

The wheel packages for this pullrequest can be found at https://github.com/k2-fsa/k2/actions/runs/355073976

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,9 +22,18 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        # os: [ubuntu-16.04, ubuntu-18.04]
         os: [ubuntu-16.04]
-        gcc: [5]
+        cuda: ["10.0", "10.1", "10.2"]
+        gcc: ["5"]
+        # torch: ["1.6.0", "1.7.0"]
+        torch: ["1.6.0"]
         python-version: [3.6, 3.7, 3.8]
+        exclude:
+          - cuda: "10.0" # No wheels for torch==1.6.0+cu100
+            torch: "1.6.0"
+          - cuda: "10.0" # No wheels for torch==1.7.0+cu100
+            torch: "1.7.0"
 
     steps:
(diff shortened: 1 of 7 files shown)


danpovey:
Impressive!
How did you debug this?  Is there some kind of shell you can log into to debug the environment that's used for this build?
And are these builds uploaded to PyPi, or if not, how are they discovered when you do `pip install`?

csukuangfj:
> How did you debug this? Is there some kind of shell you can log into to debug the environment that's used for this build?

The debug was done by fixing every error reported by GitHub Actions until there were no errors. I think it is not possible to
log into these machines.

These builds are not uploaded to PyPI, they are saved on GitHub instead. But it is possible to upload them to PyPI. We can create another action that is triggered whenever a new version is released and upload wheel files in this action.

To use `pip install`, we have to first download the artifact, unzip it to get the `*.whl` file, and then use `pip install` to install it.

I've downloaded one of the artifacts and tested it in this [colab notebook][1]. Everything works fine. So I think there is nothing wrong in the pip packages built by GitHub actions.

[1]: https://colab.research.google.com/drive/1YYjLEkphmPY1l5WoTqT86zpfVA3lLf4v?usp=sharing

csukuangfj:
One drawback is that the file size of `libcontext.so` built by GitHub actions is about 3 times larger than the one built locally.

danpovey:
Ah.
What do you think about renaming context to k2context?  The name is generic
enough that it can be confusable whether it comes from k2 or externally.

csukuangfj:
I agree. Renamed.

csukuangfj:
Ready to merge if there are no other problems/comments.

PR #324, opened 2020-11-10
fix a bug in Fsa.invert_.

csukuangfj:
By default, `fsa.labels` returns a reference.
We need to get a copy.


qindazhu:
OK, thanks!

PR #327, opened 2020-11-10
Implement k2.index_select

### LINKED ISSUE #307: k2.index
I am thinking at the Python level we could add a function k2.index1 that does 1-d indexing with int32_t's, with special treatment of -1.  So like what PyTorch does with a[b], except it treats -1 separately and returns 0 in that case.

At the python level it could be something like the following:
```
   def index1(a, b):
     """ Return a[b] which is the elements of a indexed by tensor b.  
         Args:
           a: a torch.Tensor, of a fundamental type such as int32, int64, float, double.
               (We can later extend this to deal with other types as needed).  The number
               of axes should be 1 or 2 currently; and if 2, then a.stride[1] should equal the
               size of the type.
           b: a torch.Tensor of type torch.int32, and one axis.  Its elements must be
               either 0 <= i < a.shape[0], or -1.  If -1, then the corresponding element of
              the result will be 0. 
          Return:
              Returns a torch.Tensor whose first axis has dimension b.shape[0], and whose
              remaining axes (if any) are the same as a. 
      """
    pass
```
However the internal implementation would have to be done with a torch.Function, so we can implement the backward pass.

csukuangfj:
TODOs

~~- [ ] Support 2-D tensors~~
- [x] Support non-contiguous tensors

Closes #307  and #343

### DIFF
--- a/k2/python/csrc/torch/index_add.cu
+++ b/k2/python/csrc/torch/index_add.cu
@@ -1,0 +1,111 @@
+/**
+ * @brief index_add for k2.
+ *
+ * It has identical semantics as torch.Tensor.index_add_
+ * except that it requires the dtype of the input index
+ * to be torch.int32, whereas PyTorch expects the dtype to be
+ * torch.int64.
+ *
+ * @copyright
+ * Copyright (c)  2020  Mobvoi Inc.        (authors: Fangjun Kuang)
+ *
+ * @copyright
+ * See LICENSE for clarification regarding multiple authors
+ */
+
+#include "k2/csrc/context.h"
+#include "k2/python/csrc/torch/index_add.h"
+#include "k2/python/csrc/torch/torch_util.h"
+#include "torch/extension.h"
+
+namespace k2 {
+
+// It implements  (*in_out)[index[i]] += value[i]
+// for i in range(index.numel())
+void IndexAdd(torch::Tensor index, torch::Tensor value, torch::Tensor *in_out) {
+  // We support only 1-D tensors at present
+  K2_CHECK_EQ(index.scalar_type(), ToScalarType<int32_t>::value);
+  K2_CHECK_EQ(index.dim(), 1);
+
+  K2_CHECK_EQ(value.scalar_type(), ToScalarType<float>::value);
+  K2_CHECK_EQ(value.dim(), 1);
+
+  K2_CHECK_EQ(index.numel(), value.numel());
+
+  K2_CHECK_EQ(in_out->scalar_type(), ToScalarType<float>::value);
+  K2_CHECK_EQ(in_out->dim(), 1);
+
+  const int32_t *index_data = index.data_ptr<int32_t>();
+  const float *value_data = value.data_ptr<float>();
+  float *in_out_data = in_out->data_ptr<float>();
+
+  ContextPtr context;
+  if (index.device().type() == torch::kCPU) {
+    context = GetCpuContext();
+  } else {
+    K2_CHECK(index.is_cuda());
+    context = GetCudaContext(index.device().index());
+  }
+
+  int32_t index_numel = index.numel();
+  int32_t in_out_numel = in_out->numel();
+
+  if (index.is_contiguous() && value.is_contiguous() &&
+      in_out->is_contiguous()) {
+    if (context->GetDeviceType() == kCpu) {
+      for (int32_t i = 0; i != index_numel; ++i) {
+        int32_t in_out_i = index_data[i];
+        K2_DCHECK_GE(in_out_i, 0);
+        K2_DCHECK_LT(in_out_i, in_out_numel);
+
+        // for host thread, we assume single thread at present
+        in_out_data[in_out_i] += value_data[i];
+      }
+    } else {
+      // for cuda
+      auto lambda_contiguous = [=] __device__(int32_t i) -> void {
+        int32_t in_out_i = index_data[i];
+        K2_DCHECK_GE(in_out_i, 0);
+        K2_DCHECK_LT(in_out_i, in_out_numel);
+
+        atomicAdd(in_out_data + in_out_i, value_data[i]);
+      };
+      EvalDevice(context, index_numel, lambda_contiguous);
+    }
+    return;
+  }
+
+  // for non-contiguous tensors
+  int64_t index_stride = index.strides()[0];
+  int64_t value_stride = value.strides()[0];
+  int64_t in_out_stride = in_out->strides()[0];
+
+  if (context->GetDeviceType() == kCpu) {
+    for (int32_t i = 0; i != index_numel; ++i) {
+      int32_t in_out_i = index_data[i * index_stride];
+      K2_DCHECK_GE(in_out_i, 0);
+      K2_DCHECK_LT(in_out_i, in_out_numel);
+
+      // for host thread, we assume single thread at present
+      in_out_data[in_out_i * in_out_stride] += value_data[i * value_stride];
+    }
+  } else {
+    // for cuda
+    auto lambda_noncontiguous = [=] __device__(int32_t i) -> void {
+      int32_t in_out_i = index_data[i * index_stride];
+      K2_DCHECK_GE(in_out_i, 0);
+      K2_DCHECK_LT(in_out_i, in_out_numel);
+
+      atomicAdd(in_out_data + in_out_i * in_out_stride,
+                value_data[i * value_stride]);
+    };
+    EvalDevice(context, index_numel, lambda_noncontiguous);
+  }
+}
+
+}  // namespace k2
+
+void PybindIndexAdd(py::module &m) {
+  m.def("index_add_", &k2::IndexAdd, py::arg("index"), py::arg("value"),
+        py::arg("in_out"));
+}
--- a/k2/python/csrc/torch/index_add.cu
+++ b/k2/python/csrc/torch/index_add.cu
@@ -4,7 +4,7 @@
  * It has identical semantics as torch.Tensor.index_add_
  * except that it requires the dtype of the input index
  * to be torch.int32, whereas PyTorch expects the dtype to be
- * torch.int64.
+ * torch.int64. Furthermore, it ignores index[i] == -1.
  *
  * @copyright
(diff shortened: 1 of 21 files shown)


k2/python/csrc/torch/index_select.cu
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  const T *src_data = src.data_ptr<T>();
+  int32_t src_numel = src.numel();
+  const int32_t *index_data = index.data_ptr<int32_t>();
+
+  torch::Tensor ans = torch::empty(index.sizes(), src.options());
+  T *ans_data = ans.data_ptr<T>();
+
+  auto lambda = [=] __host__ __device__(int32_t i) {
```
danpovey, line 46: I tend to prefer to implement the CPU version with a loop rather than a lambda for things like this, because the current implementation of lambdas from nvcc has to follow a function pointer each time the lambda is invoked, i.e. it's not inline.  So you could do:
```
   if (context->GetDeviceType() == kCpu) {
        // use a loop
   } else {
       EvalDevice(context, index.numel(), lambda);
  }
```
I introduced EvalDevice() fairly recently so you won't see it in a ton of code yet.  It just means, we won't call this for CPU.
```

File: k2/python/csrc/torch/index_select.cu [hunk 2]
```diff
@@ -0,0 +1,143 @@
...
+      ans_data[i] = src_data[src_i];
+    } else {
+      ans_data[i] = 0;
+    }
+  };
+  Eval(context, index.numel(), lambda);
+  return ans;
+}
+
+static torch::Tensor IndexSelect1DBackward(torch::Tensor src,
```
danpovey, line 61: Incidentally, the core part of this function is the same as torch's index_add_, and we might eventually want to make that callable independently.  Just something to think about; no change required.

k2/python/csrc/torch/index_select.cu
```diff
@@ -0,0 +1,143 @@
...
+torch::Tensor IndexSelectWrapper(torch::Tensor src, torch::Tensor index) {
+  auto scalar_type = src.scalar_type();
+  if (src.dim() == 1) {
+    switch (scalar_type) {
+      case ToScalarType<int32_t>::value:
+        return IndexSelect1D<int32_t>(src, index);
+      case ToScalarType<float>::value:
+        return IndexSelect1D<float>(src, index);
+      default:
+        K2_LOG(FATAL) << "Unknown scalar type: " << scalar_type;
```
danpovey, line 115: perhaps "unsupported scalar type" would be more accurate.

k2/python/k2/autograd.py
```diff
@@ -186,6 +182,40 @@ def backward(
...
+        returned tensor is 0.
+
+        Caution:
+          `index.dtype == torch.int32` and `index.ndim == 1`.
+
+        Args:
+          src:
+            The input tensor. Either 1-D or 2-D with dtype torch.int32 or torch.float32.
+          index:
+            1-D tensor of dtype torch.int32 containing the indexes to index. If an entry
```
danpovey, line 202: Just say "containing the indexes", not "indexes to index".  And you can add the sentence at the end: "The values in `index` should be in the range [-1..src.shape[0]-1].

k2/python/k2/autograd.py
```diff
@@ -186,6 +182,40 @@ def backward(
...
+            is -1, the corresponding entry in the returned value is 0.
+
+        Returns:
+          The output tensor.
```
danpovey, line 206: Saying it returns "the output tensor" is not informative at all.  You could say
 ```
    A tensor with shape index.shape+src.shape[1:] and dtype the same as src,
    e.g. if src.ndim == 1 ans.shape would be (index.shape[0]), if src.ndim==2 ans.shape
    would be (index.shape[0],src.shape[1]).    Will satisfy ans[i] == src[index[i]] if 
   src.ndim == 1, or ans[i,j] == src[index[i],j] if src.ndim == 2, except for entries 
   where index[i] == -1 which will be zero.
```
   Notice how what I say is unambiguous because i'm using code.

File: k2/python/csrc/torch/index_select.cu (hunk 1 above)
Comment (csukuangfj) [reply k2/python/csrc/torch/index_select.cu]: Fixed.

File: k2/python/csrc/torch/index_select.cu (hunk 2 above)
Comment (csukuangfj) [reply k2/python/csrc/torch/index_select.cu:109]: Added `index_add_`.

File: k2/python/csrc/torch/index_select.cu (hunk 3 above)
Comment (csukuangfj) [reply k2/python/csrc/torch/index_select.cu]: Fixed.

File: k2/python/k2/autograd.py (hunk 4 above)
Comment (csukuangfj) [reply k2/python/k2/autograd.py]: Fixed.

File: k2/python/k2/autograd.py (hunk 5 above)
Comment (csukuangfj) [reply k2/python/k2/autograd.py]: Fixed.

**Comment by csukuangfj** [comment]
The support of 2-D tensors is missing. Otherwise, it is ready to review.

There are currently no use cases of 2-D tensors that require `index_add` and `index_select` in k2.

File: k2/python/csrc/torch/index_select.cu (hunk 2 above)
Comment (danpovey) [reply k2/python/csrc/torch/index_select.cu:109]: Do this neeed to be different from index_add, i.e. can't you just call index_add?

I don't think index_add_ is the right name if it's not a member function (?)   Should be index_add?

File: k2/python/csrc/torch/index_select.cu [hunk 6]
```diff
@@ -0,0 +1,221 @@
...
+
+#include "k2/csrc/context.h"
+#include "k2/python/csrc/torch/index_select.h"
+#include "k2/python/csrc/torch/torch_util.h"
+#include "torch/extension.h"
+
+namespace k2 {
+
+template <typename T>
+static torch::Tensor IndexSelect1D(torch::Tensor src, torch::Tensor index) {
```
danpovey, line 22: The functions in this file still need documentation!  Including the wrappers!   Does n't have to be very long/detailed b ut should exist.

k2/python/csrc/torch/index_select.cu
csukuangfj: > Do this need to be different from index_add, i.e. can't you just call index_add?

Yes, we still need it. 

`index_add_` does not handle the case where an index may be -1.
If we do not need to consider that indexes may be -1, then `IndexSelect1DBackward` can be
actually removed. We can call `index_add_` in Python for the backpropagation, which is
the case for `intersect_dense_pruned`.

danpovey:
I thought we were talking about a k2 version of index_add that would treat
-1's specially.
I'm sure we will have some kind of use for that.

k2/python/csrc/torch/index_select.cu
csukuangfj: Fixed by removing `IndexSelect1DBackward`.

k2/python/csrc/torch/index_select.cu
csukuangfj: Fixed.

csukuangfj:
> I thought we were talking about a k2 version of index_add that would treat
-1's specially.

Fixed.

danpovey:
Thanks!
Anyway going to bed in  a minute, will be in touch tomorrow.

k2/python/csrc/torch/index_add.cu
```diff
@@ -0,0 +1,119 @@
...
+    } else {
+      // for cuda
+      auto lambda_contiguous = [=] __device__(int32_t i) -> void {
+        int32_t in_out_i = index_data[i];
+        if (in_out_i == -1) return;
+
+        K2_DCHECK_GE(in_out_i, 0);
+        K2_DCHECK_LT(in_out_i, in_out_numel);
+
+        atomicAdd(in_out_data + in_out_i, value_data[i]);
```
qindazhu, line 75: Just a question, @danpovey , do you think we should avoid `atomicAdd` here by using an approach like `gather(with index) then SumPerSublist`, I'm just not sure which is faster as the latter approach will have multiple kernels.

csukuangfj:
I guess the barrier exists only for multiple accesses to the same memory address.



Sent from myMail for iOS


Sunday, 15 November 2020, 12:38 +0800 from notifications@github.com  <notifications@github.com>:

qindazhu:
> I guess the barrier exists only for multiple accesses to the same memory address

that's true, that's why I said I'm not sure which is faster as I'm not sure what the index and data distribution in our usages.

danpovey:
We can commit for now with indexAdd(), and figure this out later if the speed becomes a problem.
I guess for the typical cases where the mapping is really one-to-one, indexAdd() will be much faster, but
it scales linearly with a poor constant if there are lots of writes to a single memory location.

In this case it would definitely make sense to have a reduction-based approach.  
Two points:
   (i) it requires some data rearrangement.  I believe GetTransposeReordering() can be used here.  
Internally it uses mergesort from moderngpu, which probably isn't the fastest or best way to do it, but
we can try to optimize it later.  
  (ii) Since the reduction-based one will be substantially slower for "nice" data, we will need to 
   figure out when it makes sense to use the reduction one, e.g. via some kind of option or a differently
    named interface.  One possibility is to force the user to do this.

BTW there is a branch conflict.

qindazhu:
Sure，then let's add a TODO here for notice.

qindazhu:
RE reduction-based approach, I was thinking we can do:
1. Historgram/Count on index and exclusive to get row_splits;
2. Copy data with index as row_ids
3. Do SumPerSublist

Not sure if it's faster than GetTransposeReordering.

danpovey:
... I don't think there is a way, using Histogram/Count on indexes, to figure out the reordering that is needed.
It would be necessary to use some kind of sorting, I think (even if one of those reduction/count-based sorting methods
that internally uses the same code).

danpovey:
Don't forget to fix the conflict!

csukuangfj:
Unfortunately, I don't have a laptop at hand currently. Will fix it later.



Sent from myMail for iOS


Sunday, 15 November 2020, 13:52 +0800 from notifications@github.com  <notifications@github.com>:

csukuangfj:
Conflicts have been resolved.

csukuangfj:
Merging as I need `index_select` for implementing backward propagation of `union`.

If there are any problems, I will fix them later.

Support for 2-D tensors will be added when it is needed and when I have time.

PR #330, opened 2020-11-11
Fix some bugs

qindazhu:
Noted most changes were from @danpovey (except change aux_labels to -1 in OpenFstTransducerFromStream)

### DIFF
--- a/k2/csrc/array.h
+++ b/k2/csrc/array.h
@@ -176,7 +176,6 @@ class Array1 {
   */
   Array1 To(ContextPtr ctx) const {
     if (ctx->IsCompatible(*Context())) return *this;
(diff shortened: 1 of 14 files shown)


danpovey:
Note, I changed how the properties are dealt with.  you now access fsa.properties or fsa.properties_str.  Currently, the properties are always computed when the FSA is constructed.  However we'll need to recompute them when we do fsa._invert().  We should make a call _recompute_properties() whenever we do something that changes the 'arcs' field that might affect the properties.  @qindazhu can you look at that?  In fact, _init_properties() will do the job just fine.

qindazhu:
OK, will do.

PR #332, opened 2020-11-12
[WIP] fix bugs

qindazhu:
Will continue working on this to fix other bugs.

### DIFF
--- a/k2/python/csrc/torch/torch_util.cu
+++ b/k2/python/csrc/torch/torch_util.cu
@@ -80,6 +80,7 @@ torch::Tensor ToTensor(Array1<Arc> &array) {
   std::vector<int64_t> sizes = {array.Dim(), 4};  // [num_rows, num_cols]
   std::vector<int64_t> strides = {4, 1};          // in number of elements
   auto options = torch::device(device).dtype(scalar_type);
+  if (array.Dim() == 0) return torch::empty({0, 4}, options);
 
   // NOTE: we keep a copy of `Region` inside the lambda
   // so that the returned tensor outlives the input array.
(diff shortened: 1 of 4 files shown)


k2/python/csrc/torch/torch_util.cu
```diff
@@ -80,6 +80,7 @@ torch::Tensor ToTensor(Array1<Arc> &array) {
   std::vector<int64_t> sizes = {array.Dim(), 4};  // [num_rows, num_cols]
   std::vector<int64_t> strides = {4, 1};          // in number of elements
   auto options = torch::device(device).dtype(scalar_type);
+  if (array.Dim() == 0) return torch::empty((0, 4), options);
```
csukuangfj, line 83: `(0, 4)` should be changed to `{0, 4}`.

csukuangfj: `(0, 4)` is a comma expression in C++.
So `torch::empty((0, 4), options);` is equivalent to `torch::empty(4, options);`

danpovey:
Thanks!!

k2/python/csrc/torch/torch_util.cu
qindazhu: Oh, yes, and @csukuangfj can you kindly help to check other code of those convertion between k2 and torch, I think we should add tests for empty cases

csukuangfj: Sure, I'd like to do it.

qindazhu: Updated.

PR #336, opened 2020-11-13
fix bugs..

qindazhu:


k2/csrc/ragged_ops.cu
```diff
@@ -1010,10 +997,11 @@ static Array1<int32_t> GetTransposeReorderingThreeAxesCuda(Ragged<int32_t> &src,
 
   int32_t n = src.values.Dim();
   Array1<int32_t> ans = Range(context, n, 0);
-  K2_CUDA_SAFE_CALL(mgpu::segmented_sort(ans.Data(),          // keys
-                                         ans.Dim(),           // count
-                                         segments.Data(),     // segments
-                                         segments.Dim() - 1,  // num_segments
```
qindazhu, line 1016: @csukuangfj not sure why you `-1` here, but all tests get passed by moving this.

k2/csrc/ragged_ops.cu
```diff
@@ -1038,6 +1026,7 @@ Array1<int32_t> GetTransposeReordering(Ragged<int32_t> &src, int32_t num_cols) {
   const int32_t *value_data = src.values.Data();
   int32_t n = src.values.Dim();
   Array1<int32_t> ans = Range(context, n, 0);
+  if (n == 0) return ans;
```
qindazhu, line 1029: mgpu will throw cuda error if ans is emtpy, @csukuangfj it would be great if you can add random tests (especially empty case) for all our mgpu usages. (IIRC, we use it for `arc_sort` and `TransposeReorder`?)

qindazhu:
Note now the trainning scripts can run but we get `inf` for some Fsas

k2/csrc/ragged_ops.cu
csukuangfj: I tried both of them and both worked. I don't remember why I chose `-1`.

k2/csrc/ragged_ops.cu
csukuangfj: Will add it.

csukuangfj:
> Note now the trainning scripts can run but we get inf for some Fsas

For what kind of fsas?

k2/csrc/fsa_utils.cu
```diff
@@ -1026,7 +1028,8 @@ Array1<FloatType> GetForwardScores(FsaVec &fsas, Ragged<int32_t> &state_batches,
           num_arcs = fsas.TotSize(2);
   int32_t num_batches = state_batches.Dim0();
   // just using DCHECK below to save time in production code
-  K2_DCHECK_EQ((state_batches.TotSize(1) / num_batches), num_fsas);
+  K2_DCHECK(num_batches == 0 ||
```
danpovey, line 1031: We chould just do: `K2_DCHECK_EQ(state_batches.TotSize(1), num_batches * num_fsas);`...

csukuangfj: yes, multiplication is faster than division.

qindazhu:
> For what kind of fsas?

Not sure, This is what I'm doing now

k2/csrc/fsa_utils.cu
qindazhu: sure, will update

PR #338, opened 2020-11-13
wrap ConvertDenseToFsaVec for testing

qindazhu:


k2/python/k2/dense_fsa_vec.py
```diff
@@ -92,3 +93,17 @@ def dim0(self) -> int:
 
     def __str__(self) -> str:
         return self.dense_fsa_vec.to_str()
+
+
+def convert_dense_to_fsa_vec(dense_fsa_vec: DenseFsaVec) -> Fsa:
+    '''Convert a DenseFsaVec to an FsaVec. Intended for use in testing code.
```
csukuangfj, line 99: I would suggest putting a warning/caution here:

Caution:

    Intended for use in testing/debug mode only. This operation is NOT differentiable.

danpovey:
Merge when you guys think it's OK...

k2/python/k2/dense_fsa_vec.py
qindazhu: Updated

PR #342, opened 2020-11-14
Fix computing bprop of GetTotScores.

### LINKED ISSUE #341: Backward of get_tot_scores ignores gradient
This is not right:
https://github.com/k2-fsa/k2/blob/59e7757c23e360acc5f655977683b2dfd8a3ce02/k2/python/k2/autograd.py#L55
the backward of get_tot_scores needs to multiply by the gradient at the output.
In our training, this is causing us to get the sign of the update wrong!!!

csukuangfj:
Closes #341 

TODOs:

- [x] Fix log-semiring

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,24 +22,17 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        # os: [ubuntu-16.04, ubuntu-18.04]
-        os: [ubuntu-16.04]
-        cuda: ["10.0", "10.1", "10.2"]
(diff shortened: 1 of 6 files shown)


csukuangfj:
Ready for review.

qindazhu:
+2, the tot_objf now is decreasing, the only critical issue we have now is the out-of-memory issue.

danpovey:
Great!  Let me merge this...

danpovey:
The way I would try to debug this is first to add code in the tests of the individual things we backprop (tot-scores, etc.) to check that the CUDA memory used decreases to zero after all the objects are made unreachable.

PR #346, opened 2020-11-15
WIP: implement union.

csukuangfj:
TODOs

- [ ] support arc_map
- [ ] compute `row_splits` of the output fsa in the lambda

### DIFF
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -540,4 +540,93 @@ void AddEpsilonSelfLoops(FsaOrVec &src, FsaOrVec *dest,
   }
 }
 
+// TODO(fangjun): compute arc_map
+Fsa Union(FsaVec &fsas, Array1<int32_t> *arc_map /*= nullptr*/) {
+  K2_CHECK_EQ(fsas.NumAxes(), 3);
+
+  ContextPtr &context = fsas.Context();
+  const int32_t *fsas_row_splits1_data = fsas.RowSplits(1).Data();
+  const int32_t *fsas_row_splits2_data = fsas.RowSplits(2).Data();
+  const int32_t *fsas_row_ids1_data = fsas.RowIds(1).Data();
+  const int32_t *fsas_row_ids2_data = fsas.RowIds(2).Data();
+  const Arc *arcs_data = fsas.values.Data();
+
+  int32_t num_fsas = fsas.Dim0();
+  int32_t num_states = fsas.TotSize(1);
+  int32_t num_arcs = fsas.TotSize(2);
+
+  // A new start state and a new final state are added (+2).
+  // The final state of each fsa is removed (-num_fsas)
+  int32_t num_out_states = num_states + 2 - num_fsas;
+  int32_t out_final_state = num_out_states - 1;
+
+  // For every fsa, a new arc is added from the new start state
+  // to its original start state (+num_fsas)
+  int32_t num_out_arcs = num_arcs + num_fsas;
+
+  Array1<int32_t> out_row_splits(context, num_out_states + 1);
+  Array1<int32_t> out_row_ids(context, num_out_arcs);
+  Array1<Arc> out_arcs(context, num_out_arcs);
+
+  int32_t *out_row_splits_data = out_row_splits.Data();
+  int32_t *out_row_ids_data = out_row_ids.Data();
+  Arc *out_arcs_data = out_arcs.Data();
+
+  auto lambda_set_out = [=] __host__ __device__(int32_t fsas_arc_idx012) {
+    int32_t fsas_state_idx01 = fsas_row_ids2_data[fsas_arc_idx012];
+    int32_t fsas_idx0 = fsas_row_ids1_data[fsas_state_idx01];
+    int32_t this_fsa_final_state_idx01 =
+        fsas_row_splits1_data[fsas_idx0 + 1] - 1;
+
+    K2_DCHECK_GE(this_fsa_final_state_idx01, fsas_state_idx01);
+
+    int32_t fsas_state_idx0x = fsas_row_splits1_data[fsas_idx0];
+    int32_t fsas_state_idx1 = fsas_state_idx01 - fsas_state_idx0x;
+    int32_t this_fsa_final_state_idx1 =
+        this_fsa_final_state_idx01 - fsas_state_idx0x;
+
+    int32_t fsas_arc_idx01x = fsas_row_splits2_data[fsas_state_idx01];
+    int32_t fsas_arc_idx2 = fsas_arc_idx012 - fsas_arc_idx01x;
+
+    // fsa0: +1 (a new start state)
+    // fsa1: +0 (the final state of fsa0 is removed)
+    // fsa2: -1 (the final state of fsa1 is removed)
+    // fsa3: -2 (the final state of fsa2 is removed)
+    int32_t state_offset = 1 - fsas_idx0;
+    int32_t out_state_idx0 = fsas_state_idx01 + state_offset;
+
+    int32_t out_arc_idx01 = fsas_arc_idx012 + num_fsas;
+    out_row_ids_data[out_arc_idx01] = out_state_idx0;
+    Arc arc = arcs_data[fsas_arc_idx012];
+
+    K2_DCHECK_EQ(arc.src_state, fsas_state_idx1);
+
+    if (arc.dest_state == this_fsa_final_state_idx1)
+      arc.dest_state = out_final_state;
+    else
+      arc.dest_state = arc.dest_state - arc.src_state + out_state_idx0;
+
+    arc.src_state = out_state_idx0;
+    out_arcs_data[out_arc_idx01] = arc;
+
+    if (fsas_state_idx1 == 0 && fsas_arc_idx2 == 0) {
+      // add a new arc from the new start state to this state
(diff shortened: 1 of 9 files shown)


csukuangfj:
Since k2 supports visualizing only a single FSA,  one possible use of `union` is  convert a FsaVec to a single FSA and then
visualize the converted FSA.

k2/csrc/fsa_algo.h
```diff
@@ -239,6 +239,19 @@ FsaVec LinearFsas(Ragged<int32_t> &symbols);
 Ragged<int32_t> ShortestPath(FsaVec &fsas,
                              const Array1<int32_t> &entering_arcs);
 
+/* Compute the union of all fsas in a FsaVec.
+
```
danpovey, line 243: OK... this is not the most general interface that we might need, but could be OK for now.
We might eventually need to do a union (FsaVec, FsaVec -> FsaVec), or (FsaVec, Fsa -> FsaVec);
or in fact (FsaVec, [Ragged<Int> that shows how the Fsas in the FsaVec are to be grouped]) -> FsaVec.

But we can add other version later or extend the interface.

k2/csrc/fsa_algo.cu
```diff
@@ -540,4 +540,93 @@ void AddEpsilonSelfLoops(FsaOrVec &src, FsaOrVec *dest,
...
+  K2_LOG(INFO) << out_arcs;
+  K2_LOG(INFO) << out_row_ids;
+
+  // TODO(fangjun): figure out out_row_splits in the lambda
+  //
+  // RaggedShape shape = RaggedShape2(&out_row_splits, &out_row_ids,
+  // num_out_arcs);
+  RaggedShape shape = RaggedShape2(nullptr, &out_row_ids, num_out_arcs);
+  Fsa ans = Ragged<Arc>(shape, out_arcs);
+  return ans;
```
danpovey, line 629: If you are not going to support the arc-map for this PR, this should probably be documented and you should assert that it's nullptr.

csukuangfj: I am going to support arc-map in this PR. Not finished yet.

k2/csrc/fsa_algo_test.cu
```diff
@@ -376,4 +374,26 @@ TEST(FsaAlgo, ShortestPath) {
...
+  )";
+
+  for (auto &context : {GetCpuContext(), GetCudaContext()}) {
+    Fsa fsa1 = FsaFromString(s1);
+    Fsa fsa2 = FsaFromString(s2);
+    Fsa *fsa_array[] = {&fsa1, &fsa2};
+    FsaVec fsa_vec = CreateFsaVec(2, &fsa_array[0]);
+    fsa_vec = fsa_vec.To(context);
+    Fsa fsa = Union(fsa_vec);
+    K2_LOG(INFO) << FsaToString(fsa);
```
qindazhu, line 395: Again, I **strongly** suggest we add random tests (especially empty case) for any algorithms, it really will take our too much time to dubug in the final python training scripts if there's any issue. You can use RandomFsa/RandomFsaVec to create Fsas.

k2/csrc/fsa_algo.cu
```diff
@@ -540,4 +540,93 @@ void AddEpsilonSelfLoops(FsaOrVec &src, FsaOrVec *dest,
...
+  const int32_t *fsas_row_ids2_data = fsas.RowIds(2).Data();
+  const Arc *arcs_data = fsas.values.Data();
+
+  int32_t num_fsas = fsas.Dim0();
+  int32_t num_states = fsas.TotSize(1);
+  int32_t num_arcs = fsas.TotSize(2);
+
+  // A new start state and a new final state are added (+2).
+  // The final state of each fsa is removed (-num_fsas)
+  int32_t num_out_states = num_states + 2 - num_fsas;
```
qindazhu, line 560: what happened here if one of input Fsa is empty?

k2/csrc/fsa_algo.cu
```diff
@@ -540,4 +540,93 @@ void AddEpsilonSelfLoops(FsaOrVec &src, FsaOrVec *dest,
...
+  int32_t out_final_state = num_out_states - 1;
+
+  // For every fsa, a new arc is added from the new start state
+  // to its original start state (+num_fsas)
+  int32_t num_out_arcs = num_arcs + num_fsas;
```
qindazhu, line 565: and here, if one of the input fsa is empty. That's why I strongly ask you add random tests and I think I have asked this too many times..

danpovey:
Mm, search for lambda_set_fsa_nonempty in fsa_algo.cu for an example of
code that deals with the empty FSA issue.

danpovey:
For dealing with empty FSAs, we might be able to solve it in a general way,
like:

  /*
     Remove empty FSAs from `src`, setting `dest` to an FsaVec containing
FSAs in the
     same order but with empty FSAs removed.
        ...
  */
  RemoveEmptyFsas(FsaVec &src, Renumbering *renumbering, FsaVec *dest);

That would be sufficient in this case.

qindazhu:
> RemoveEmptyFsas(FsaVec &src, Renumbering *renumbering, FsaVec *dest);

I like this approach to deal with empty case in a general way, but I just doubt it's not efficient in some cases, e.g the original algorithm is relatively simple or even has only one kernel in it, call `RemoveEmptyFsas` will add 2 kernel calls before it.

For simple algorithms, I would like to call `RemoveEmptyFsas` before we feed those fsas into a chain of operations, e,g,
```
RemoveEmptyFsas();
union()
rmepsilon()
determinize()
...
```

danpovey:
Mm.  Most algorithms are slow enough that we wouldn't notice the
difference.  I'm concerned about adding extra assumptions that are not
quickly
checkable.  Unless we make the properties an extra argument to these
function calls.  Since we are currently computing the properties whenever
we construct an Fsa, this might be feasible.

In the majority of cases, there won't be empty FSAs; so we could have a
renumbering inside each algorithm that only happens if
there were empty FSAs, and that wouldn't really impact the speed.  So
regardless, I don't see a good reason to add "nonempty" as a
requirement for algorithms (but we can consider making the input properties
an extra arg to functions).

qindazhu:
> I don't see a good reason to add "nonempty" as a
requirement for algorithms (but we can consider making the input properties
an extra arg to functions).

Agree

k2/csrc/fsa_algo.cu
csukuangfj: It will abort the program.

It indeed checks for the empty case
https://github.com/k2-fsa/k2/blob/c1e16fc0ea65b5f7be322291b6852d465b44637b/k2/csrc/fsa_algo.cu#L581


The empty case is not supported as the precondition is documented clearly
in the comment  of this function
https://github.com/k2-fsa/k2/blob/c1e16fc0ea65b5f7be322291b6852d465b44637b/k2/csrc/fsa_algo.h#L244-L245

k2/csrc/fsa_algo.cu
```diff
@@ -540,4 +540,93 @@ void AddEpsilonSelfLoops(FsaOrVec &src, FsaOrVec *dest,
...
+
+    if (arc.dest_state == this_fsa_final_state_idx1)
+      arc.dest_state = out_final_state;
+    else
+      arc.dest_state = arc.dest_state - arc.src_state + out_state_idx0;
+
+    arc.src_state = out_state_idx0;
+    out_arcs_data[out_arc_idx01] = arc;
+
+    if (fsas_state_idx1 == 0 && fsas_arc_idx2 == 0) {
```
qindazhu, line 612: it's start state of the origianl Fsa, right?  we can just use `fsas_state_idx0x == fsa_state_idx01`

k2/csrc/fsa_algo.cu
```diff
@@ -540,4 +540,93 @@ void AddEpsilonSelfLoops(FsaOrVec &src, FsaOrVec *dest,
...
+      // add a new arc from the new start state to this state
```
qindazhu, line 613: and may be it's better to change `this_state` to something like `start state of this fsa`? If I understand correctly

k2/csrc/fsa_algo.cu
qindazhu: Oh, seems I miss your documentation here, sorry. But as @danpovey said below, I now also think we should not non=empty requirements on such basic algorithms, we should handle such cases in it.

danpovey:
I'm OK with the current implementation if it checks for nonempty and also
documents it.
We can fix this later, possibly along with making properties an arg to
functions.
Let's just get this merged quickly so Yiming (@freewym) can make progress.

qindazhu:
OK

csukuangfj:
Will support arc map later today 



Sent from myMail for iOS


Sunday, 15 November 2020, 12:19 +0800 from notifications@github.com  <notifications@github.com>:

qindazhu:
> Will support arc map later today 

As @freewym does not need arc_map stuff for now, I think you can add arc_map support (with row_splits) in another PR.

danpovey:
OK I'll merge this now.

csukuangfj:
Not wrong but it's the 0th arc of the start state.



Sent from myMail for iOS


Sunday, 15 November 2020, 12:11 +0800 from notifications@github.com  <notifications@github.com>:

k2/csrc/fsa_algo_test.cu
csukuangfj: Will add it.

PR #348, opened 2020-11-15
WIP: add arc_map support to union

### LINKED ISSUE #350: MemoryCopy CPU only
If this code requires a GPU to be initialized for CPU->CPU transfers then we should change it.

https://github.com/k2-fsa/k2/blob/4c3c2a9428014ed24bd7abdca2e43ea50ac4c27e/k2/csrc/context.h#L176

csukuangfj:
TODOs

- [x] compute row_splits in the lamda
- [ ] use random generated fsas for testing

Closes #350


danpovey:
Merging this for the memcpy thing.

PR #355, opened 2020-11-17
support FsaVec printing in python

qindazhu:


k2/python/k2/fsa.py
```diff
@@ -120,7 +120,17 @@ def __str__(self) -> str:
             aux_labels = self.aux_labels.to(torch.int32)
         else:
             aux_labels = None
-        ans = "k2.Fsa: " + _fsa_to_str(self.arcs, False, aux_labels)
+        if self.arcs.num_axes() == 2:
+            ans = "k2.Fsa: " + _fsa_to_str(self.arcs, False, aux_labels)
+        else:
+            ans = "k2.FsaVec: \n"
+            for i in range(self.shape[0]):
```
danpovey, line 127: I believe using "+=" is very slow.  Better to do ' '.join[ some_expr for something in something ], or construct a list with a loop and then ' '.join(list).

qindazhu: Actually, according to some experiments, + or `+=` may be the fastest way, and they even recommend this way as it's more readable than other approach. BTW, usually We don't print Fsa for a very large Fsa so the difference would be small.
https://blog.ganssle.io/articles/2019/11/string-concat.html 
https://stackoverflow.com/questions/12169839/which-is-the-preferred-way-to-concatenate-a-string-in-python

Anyway, I'm open with this.

PR #361, opened 2020-11-18
Fix segfaults in __host__ __device__ lambdas.

danpovey:
The segfaults seem to be due to compiler bugs.
Fixed them by converting all one or two-line __host__ __device__ lambdas into __device__ lambdas, dealing with the host case separately via simple loop.


qindazhu:
+N!!!!!

qindazhu:
@danpovey , just a question, so more than 3 lines should be fine?

danpovey:
It appears to be fine.  Of course I don't know the exact limit.  Relates to
compiler inlining decisions, I think.

PR #364, opened 2020-11-19
wrap host rmepsilon and determinize

qindazhu:


k2/csrc/fsa_algo.h
```diff
@@ -195,6 +195,41 @@ bool Intersect(FsaOrVec &a_fsas, FsaOrVec &b_fsas,
...
+/*
+    Remove epsilons (symbol zero) in the input Fsas, it works for both
+    Fsa and FsaVec.
+    @param [in] src   Source Fsa or FsaVec. Must be top-sorted as we will
+                      compute forward and backward scores on it.
+    @param [out] dest Destination; at exit will be equivalent to `src` but will
+                      be epsilon-free, i.e. there's no arc in dest with
+                      arc.label==0, its Properties() will contain
+                      kFsaPropertiesEpsilonFree.
+    We don't output arc_map here as this version will not support auto grad; we
```
danpovey, line 207: autograd one word

danpovey: ... also, equivalent in what semiring?  There were 2 versions of the host function.

k2/csrc/fsa_algo.h
```diff
@@ -195,6 +195,41 @@ bool Intersect(FsaOrVec &a_fsas, FsaOrVec &b_fsas,
...
+void RemoveEpsilon(FsaOrVec &src, FsaOrVec *dest);
+
+/*
+    Determinize the input Fsas, it works for both Fsa and FsaVec.
+    @param [in] src   Source Fsa or FsaVec. Must be top-sorted as we will
+                      compute forward and backward scores on it.
+                      Expected to be epsilon free, but this is not checked;
+                      in any case, epsilon will be treated as a normal
+                      symbol. We also assume src is connected.
+    @param [out] dest Destination; at exit will be equivalent to `src` but will
```
danpovey, line 221: Equivalent under what semiring?  I believe there were 2 versions of the underlying host code.

danpovey (commented):
some comments..

k2/csrc/fsa_algo.h
qindazhu: updated autogrid

k2/csrc/fsa_algo.h
qindazhu: As this version doesn't support pruning, I was thinking it's equivalent to src under both log and tropical smearing. Correct me if I'm wrong

danpovey: No, because it needs to know how weights are summed.  Consider that there are 2 arcs from state 0 to state 1 with the same symbol, and cost=0.  Should the cost after determinization be 0, or log(2.0)?

qindazhu: Oh, yes, we did them in `TraceBack` stuff. 
 For now I implement with tropical, Is that OK for our decoding usage? If it's OK, will add documentation here to say `under tropical semiring`  (as we will abandon this version after getting GPU version done I was thinking we don't need both log and semiring version). Or should I change to log semiring implementation?

danpovey:
Just saying it's under tropical is OK for now.

qindazhu:
Updated, thanks!

csukuangfj:
I am adding Python wrappers from them.

qindazhu:
mm, wait, I was just doing this. Will make a PR now and add tests in another PR.

csukuangfj:
ok, then I will work on nvtx.

PR #366, opened 2020-11-19
wrap rmepsion and determinize to Python

qindazhu:
Just merge to avoid future conflicts, will add tests in another PR.

### DIFF
--- a/k2/python/csrc/torch/fsa.cu
+++ b/k2/python/csrc/torch/fsa.cu
@@ -61,6 +61,20 @@ static void PybindFsaBasicProperties(py::module &m) {
                kFsaPropertiesMaybeCoaccessible;
       },
       py::arg("properties"));
+  m.def(
+      "is_epsilon_free",
+      [](int32_t properties) -> bool {
+        return (properties & kFsaPropertiesEpsilonFree) ==
+               kFsaPropertiesEpsilonFree;
+      },
+      py::arg("properties"));
+  m.def(
+      "is_arc_sorted_and_deterministic",
+      [](int32_t properties) -> bool {
+        return (properties & kFsaPropertiesArcSortedAndDeterministic) ==
+               kFsaPropertiesArcSortedAndDeterministic;
+      },
+      py::arg("properties"));
 }
 
 static void PybindFsaUtil(py::module &m) {
--- a/k2/python/k2/fsa_algo.py
+++ b/k2/python/k2/fsa_algo.py
@@ -1,4 +1,5 @@
 # Copyright (c)  2020  Mobvoi Inc.        (authors: Fangjun Kuang)
+#                      Xiaomi Corporation (authors: Haowen Qiu)
 #
 # See ../../../LICENSE for clarification regarding multiple authors
 
@@ -283,6 +284,8 @@ def remove_epsilon(fsa: Fsa) -> Fsa:
 
     ragged_arc = _k2.remove_epsilon(fsa.arcs)
     out_fsa = Fsa.from_ragged_arc(ragged_arc)
+    for name, value in fsa.named_non_tensor_attr():
+        setattr(out_fsa, name, value)
     return out_fsa
 
 
@@ -312,4 +315,6 @@ def determinize(fsa: Fsa) -> Fsa:
 
     ragged_arc = _k2.determinize(fsa.arcs)
     out_fsa = Fsa.from_ragged_arc(ragged_arc)
+    for name, value in fsa.named_non_tensor_attr():
+        setattr(out_fsa, name, value)
     return out_fsa
--- a/k2/python/k2/fsa_algo.py
+++ b/k2/python/k2/fsa_algo.py
@@ -8,12 +8,13 @@ from typing import List
 import torch
 import _k2
 
-
 from .autograd import index_select
 from .fsa import Fsa
 from .fsa_properties import is_accessible
 from .fsa_properties import is_arc_sorted
 from .fsa_properties import is_coaccessible
+from .fsa_properties import is_arc_sorted_and_deterministic
+from .fsa_properties import is_epsilon_free
 
 
 def linear_fsa(symbols: Union[List[int], List[List[int]]]) -> Fsa:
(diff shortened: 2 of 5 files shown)


qindazhu:
@danpovey , there's no change to fsa.py

k2/python/csrc/torch/fsa.cu
```diff
@@ -61,6 +61,20 @@ static void PybindFsaBasicProperties(py::module &m) {
                kFsaPropertiesMaybeCoaccessible;
       },
       py::arg("properties"));
+  m.def(
```
csukuangfj, line 64: Please also update the copyright info of this file and the remaining files.

k2/python/k2/fsa_algo.py
```diff
@@ -257,3 +258,58 @@ def add_epsilon_self_loops(fsa: Fsa) -> Fsa:
...
+        It will be the same as the input ``fsa`` if the input
+        ``fsa`` is epsilon-free. Otherwise, a new epsilon-free fsa
+        is returned and the input ``fsa`` is NOT modified.
+    '''
+    properties = getattr(fsa, 'properties', None)
+    if properties is not None and is_epsilon_free(properties):
+        return fsa
+
+    ragged_arc = _k2.remove_epsilon(fsa.arcs)
+    out_fsa = Fsa.from_ragged_arc(ragged_arc)
```
qindazhu, line 285: @danpovey, I just realized we may need arc_map here, if not, how can we keep aux_labels after rm-epsilon? we need aux_labels (e.g. words) in decoding, right? I will do this in another PR.

csukuangfj, line 285: The non-tensor attributes can be copied without an arc map.

danpovey:
MM yeah but an arc map will always be needed, for aux_labels, tensor
attributes and for the backprop of the scores.

k2/python/k2/fsa_properties.py
```diff
@@ -80,3 +80,36 @@ def is_coaccessible(p: int) -> bool:
       False otherwise.
     '''
     return _k2.is_coaccessible(p)
+
+
+def is_epsilon_free(p: int) -> bool:
+    '''Determine whether the given properties imply an FSA
```
csukuangfj, line 86: imply an FSA that is epsilon free

or

imply an epsilon-free FSA.

k2/python/k2/fsa_properties.py
```diff
@@ -80,3 +80,36 @@ def is_coaccessible(p: int) -> bool:
       False otherwise.
     '''
     return _k2.is_coaccessible(p)
+
+
+def is_epsilon_free(p: int) -> bool:
+    '''Determine whether the given properties imply an FSA 
+       is epsilon free, i.e. no symbol zero (epsilon) 
+       present in the Fsa.
```
csukuangfj, line 88: is present

danpovey:
... for the backprop of scores, it would be helpful if we create a "phantom
assignment" operator, where for instance we could set:

phantom_assign(ans_fsa.scores, this_fsa.scores[arc_map])

and
phantom_assign(a, b)
would be equivalent to doing
a[:] = b
for backprop purposes, but in fact it would not do the assignment (its
forward method would do nothing), it would only do anything
in the backprop

k2/python/k2/fsa_algo.py
qindazhu: Just a question, @csukuangfj , How do we copy aux_labels? it's a non-tensor or tensor, I was thinking it's a tensor attribute, but didn't find code in fsa.py which shows it's a tensor attribute

danpovey:
Forget that, I am replacing that code.

danpovey:
aux_labels is a tensor attribute.  It doesn't need to be treated specially,
it behaves like other attributes.

k2/python/csrc/torch/fsa.cu
qindazhu: Added, thanks.

k2/python/k2/fsa_properties.py
qindazhu: Fixed

k2/python/k2/fsa_properties.py
qindazhu: Sure

k2/python/k2/fsa_algo.py
csukuangfj: https://github.com/k2-fsa/k2/blob/6013e15598ed673d4eec34dcf2f59b8bc47717a7/k2/python/k2/fsa.py#L113-L114

`self.aux_labels` will invoke 
https://github.com/k2-fsa/k2/blob/6013e15598ed673d4eec34dcf2f59b8bc47717a7/k2/python/k2/fsa.py#L206

https://github.com/k2-fsa/k2/blob/6013e15598ed673d4eec34dcf2f59b8bc47717a7/k2/python/k2/fsa.py#L215-L222

`aux_labels` is just like other attributes. If it is of type `torch.Tensor`, it is saved in `self._tensor_attr`.

qindazhu:
Will add arc_map and tests tomorrow

qindazhu:
> aux_labels is a tensor attribute

Yeah, I think so, but I just didn't find the code that does something like `self._tensor_attr['aux_labes'] = aux_labesl`, I was thinking it was put into `self._tensor_attr`. Can you show me the code related? I'm just confused how aux_labels is copied from input Fsa to output Fsa.

danpovey:
You just do
  fsa.aux_labels = some_tensor
and __setattr__ handles it.
This kind of code is the calling code:

        *for* name, a_value *in* a_fsas.named_tensor_attr():

            *if* name == 'scores':

                *continue*

            value = _k2.index_select(a_value, arc_map_a)

            *setattr*(out_fsa[0], name, value)


        *for* name, a_value *in* a_fsas.named_non_tensor_attr():

            *setattr*(out_fsa[0], name, a_value)

k2/python/k2/fsa_algo.py
qindazhu: Ha, many thanks!

qindazhu:
See that as the code Fangjun shows, many thanks!

qindazhu:
Copied non-tensor attribute, will add arc_map in another PR. Ready to merge if there's no other issue.

csukuangfj:
+2

PR #369, opened 2020-11-19
add nvtx

csukuangfj:


k2/csrc/nvtx.cu
```diff
@@ -0,0 +1,25 @@
...
+ * See LICENSE for clarification regarding multiple authors
+ */
+
+#include "k2/csrc/nvtx.h"
+#include "nvToolsExt.h"
+
+namespace k2 {
+
+#ifdef K2_ENABLE_NVTX
+NvtxRange::NvtxRange(const char *name) { nvtxRangePushA(name); }
```
danpovey, line 18: I'd have put these in the header, since they are so lightweight-- no reason having an extra function call in the mix.

csukuangfj: Fixed.

danpovey:
Fantastic! 
Next step is to sprinkle these NVTX_RANGE macros everywhere.
We can insert them fairly liberally.

PR #376, opened 2020-11-20
Support running tests without GPU.

### LINKED ISSUE #359: Make tests run if no GPU available
Would be nice to have it so tests can be run (and not crash) when there is no GPU.

(In the longer run we should aim to compile without CUDA as well.)

### LINKED ISSUE #381: Class Timer should support CPU as well as GPU
Could make the constructor accept the ContextPtr, and work it out from that.

csukuangfj:
~~WIP~~

- Closes #359 
- Fixes #381


k2/csrc/pytorch_context.cu
```diff
@@ -112,12 +114,19 @@ class PytorchCudaContext : public Context {
...
+#endif
 
 ContextPtr GetCpuContext() { return std::make_shared<PytorchCpuContext>(); }
 
 ContextPtr GetCudaContext(int32_t gpu_id /*= -1*/) {
+#ifdef K2_USE_CUDA
   if (gpu_id < 0) gpu_id = c10::cuda::current_device();
   return std::make_shared<PytorchCudaContext>(gpu_id);
+#else
+  K2_LOG(FATAL) << "\nk2 is compiled without CUDA!"
```
qindazhu, line 126: How about we print a warning here and return a CpuContext so that we don't need to write those `#if K2_USE_CUDA` in testing code?

danpovey:
Yeah it seems to me your current solution doesn't actually address the
problem, which was runtime test failures.

csukuangfj:
All tests are passed locally without using a GPU. 
I have enabled running tests with GitHub actions. 
Let's see whether they will work on GitHub actions: https://github.com/k2-fsa/k2/actions/runs/375604015

danpovey:
Cool!

k2/csrc/pytorch_context.cu
csukuangfj: Fixed.

csukuangfj:
Ready for review.

csukuangfj:
You can run the tests with CPU via
```
export CUDA_VISIBLE_DEVICES=
cd build
ctest --verbose
```

k2/csrc/timer.h
```diff
@@ -12,37 +12,64 @@
...
 
 namespace k2 {
+
 class Timer {
  public:
   Timer() {
-    K2_CUDA_SAFE_CALL(cudaEventCreate(&time_start_));
-    K2_CUDA_SAFE_CALL(cudaEventCreate(&time_end_));
-    Reset();
+    static std::once_flag has_cuda_init_flag;
```
danpovey, line 24: Why doesn't class Timer work for CPU? 
It might make sense for it to take a cudaStream_t, which can be kCudaStreamInvalid if we want it to work on CPU.

k2/python/k2/fsa.py
```diff
@@ -226,14 +224,14 @@ def __setattr__(self, name: str, value: Any) -> None:
...
-        if name in ('_tensor_attr', '_non_tensor_attr', 'arcs',
-                    '_cache'):
+            # For attribute names that have a getter/setter function
+            # (like 'property' or which exist as members of the object's
+            # __dict__ (like 'arcs', '_tensor_attr', '_non_tensor_attr'),
+            # do the member lookup in the normal way as if __setattr__ was
+            # not defined.
+            object.__setattr__(name, value)
+
+        if name in ('_tensor_attr', '_non_tensor_attr', 'arcs', '_cache'):
```
danpovey, line 234: To avoid having to do this we can, in __init__ _init_internal, have it do things like, instead of setting self.arcs, self._tensor_attr, etc., self.__dict__['arcs'] = arcs and so on.

k2/csrc/timer.h
csukuangfj: `cudaEventCreate` returns an error status and `K2_CUDA_SAFE_CALL` will fail.

danpovey: Yeah but what I meant is, there's no reason why that class should be for CPU only. 
It should probably take the Context and behave appropriately.
But I guess these changes are OK as a temporary thing, if we plan to fix it later.

csukuangfj: > It might make sense for it to take a cudaStream_t, which can be kCudaStreamInvalid if we want it to work on CPU.

How is the `kCudaStreamInvalid` argument going to be used by cpu code?
The previous Cuda code did not take any arguments.

danpovey:
I was thinking it could take the stream as a constructor arg.
But maybe more future-proof to take the context as constructor arg.
OK to do later.  I'll create an issue.

k2/csrc/timer.h
qindazhu: > Why doesn't class Timer work for CPU?

Back to the time we created time.h, we actually have not stable context stuff and we only use it to test GPU matrix transpose (it's actually an experiment code that time), that's why we don't write CPU version that time.

Maybe we can just copy code from Kaldi for CPU version https://github.com/kaldi-asr/kaldi/blob/master/src/base/timer.h

csukuangfj: thanks, I am porting it.

qindazhu:
BTW, I prefer to create a new class (e.g. CpuTimer, copy from Kaldi) for CPU timer, it will make code simpler and clearer

csukuangfj:
yes, 

I will create four classes: `BaseTimer`, `CpuTimer`, `GpuTimer`, and `Timer`.

Only `Timer` is visible outside. It will contain a pointer to `BaseTimer`.

k2/csrc/pytorch_context.cu
```diff
@@ -116,8 +117,21 @@ class PytorchCudaContext : public Context {
 ContextPtr GetCpuContext() { return std::make_shared<PytorchCpuContext>(); }
 
 ContextPtr GetCudaContext(int32_t gpu_id /*= -1*/) {
-  if (gpu_id < 0) gpu_id = c10::cuda::current_device();
-  return std::make_shared<PytorchCudaContext>(gpu_id);
+  static std::once_flag has_cuda_init_flag;
```
qindazhu, line 120: how about we put those `call_once` code in to a function and return `has_cuda`, then there will be no duplicate code, another advantage is we really print only once the warning if we call this in more than 1 function (e.g. context.cu and time.cu

csukuangfj: I think timer.cu will not need it since Timer will accept a ContextPtr,  which cannot be a cuda context when GetCudaContext() returns a cpu context.

qindazhu: Yeah, I just take the current code as an example to show that we will print warning twice if we do the same thing in different functions. Anyway, let's remove the duplicate code if possible.

csukuangfj: Fixed.

danpovey:
Great!  You can  merge whenever you want.

k2/python/k2/fsa.py
```diff
@@ -252,36 +250,35 @@ def __setattr__(self, name: str, value: Any) -> None:
                 self.arcs.values()[:, -1] = _as_int(value.detach())
         else:
             assert name != 'properties'  # should be set by getter/setter or by
```
freewym, line 252: assertion fails here

k2/python/k2/fsa.py
```diff
@@ -226,14 +224,14 @@ def __setattr__(self, name: str, value: Any) -> None:
...
-            object.__setattr__(key, value)
-
-        if name in ('_tensor_attr', '_non_tensor_attr', 'arcs',
-                    '_cache'):
+            # For attribute names that have a getter/setter function
+            # (like 'property' or which exist as members of the object's
+            # __dict__ (like 'arcs', '_tensor_attr', '_non_tensor_attr'),
+            # do the member lookup in the normal way as if __setattr__ was
+            # not defined.
+            object.__setattr__(name, value)
```
freewym, line 232: object.__setattr__(self, name, value)

k2/csrc/pytorch_context.cu
danpovey: RE the naming, since it's kind of an `Impl` paradigm, I wonder whether it would make sense to call BaseTimer TimerImpl, and have CpuTimerImpl and CudaTimerImpl.  I have been avoiding the word `Gpu` in these names because one day we may want to support AMD Gpu's or other GPUs.

k2/python/k2/fsa.py
danpovey: @freewym, are you sure you're not using an older version of k2?  I fixed this by changing:
```
@property
 def properties(self, value):
```
to:
```
@properties.setter
 def properties(self, value):
```

k2/python/k2/fsa.py
danpovey: mm, you're right, this code was doubly wrong.  Also it should have  had a `return` statement afterwards, according to the original intention of the code.  But I now realize that the whole thing is unnecessary, and better to delete it.

danpovey: Let me rework this slightly.

danpovey: Fixing this via separate PR

k2/csrc/pytorch_context.cu
csukuangfj: Fixed. Renamed to TimerImpl.

k2/python/k2/fsa.py
csukuangfj: Mark as resolved. This pullrequest has been rebased to the latest master.

csukuangfj:
Merging.

PR #383, opened 2020-11-21
Implement extra indexing operations (+declare some tensor ones not implemented yet)

danpovey:


k2/csrc/array_ops.h
```diff
@@ -380,6 +406,40 @@ Array2<T> ToContiguous(const Array2<T> &src);
 template <typename T>
 bool Equal(const Array2<T> &a, const Array2<T> &b);
 
+/*
+  Index `src` with `indexes`, as in src[indexes].
+     @param [in] src   Array whose elements are to be read
+     @param [in] indexes  Indexes into `src`; must satisfy
+                       `0 <= indexes[i] < indexes.Dim()` if
+                       `allow_minus_one == true`,
```
qindazhu, line 414: if `allow_minus_one == false`?

k2/csrc/array_ops.h
```diff
@@ -380,6 +406,40 @@ Array2<T> ToContiguous(const Array2<T> &src);
 template <typename T>
 bool Equal(const Array2<T> &a, const Array2<T> &b);
 
+/*
+  Index `src` with `indexes`, as in src[indexes].
+     @param [in] src   Array whose elements are to be read
+     @param [in] indexes  Indexes into `src`; must satisfy
+                       `0 <= indexes[i] < indexes.Dim()` if
```
qindazhu, line 413: `0 <= indexes[i] < src.Dim()`?

k2/csrc/array_ops.h
```diff
@@ -380,6 +406,40 @@ Array2<T> ToContiguous(const Array2<T> &src);
...
+  Index `src` with `indexes`, as in src[indexes].
+     @param [in] src   Array whose elements are to be read
+     @param [in] indexes  Indexes into `src`; must satisfy
+                       `0 <= indexes[i] < indexes.Dim0()` if
+                       `allow_minus_one == true`,
+                       else -1 is also allowed and the corresponding
+                       output element will be zero.
+     @return  Returns an `Array2<T>` of shape (indexes.Dim(), src.Dim1()),
+               with `ans[i,j] = src[indexes[i,j]]` (or zero if
+               `allow_minus_one == true` and `indexes[i] == 0`).
```
qindazhu, line 436: and `indexes[i] == -1`? Add in the implementation seems we just check `indexes[i] < 0` instead of `indexes[i] == -1`

k2/csrc/array_ops.h
```diff
@@ -380,6 +406,40 @@ Array2<T> ToContiguous(const Array2<T> &src);
...
+template <typename T>
+Array1<T> Index(const Array1<T> &src, const Array1<int32_t> &indexes,
+                bool allow_minus_one);
+
+
+/*
+  Index `src` with `indexes`, as in src[indexes].
+     @param [in] src   Array whose elements are to be read
+     @param [in] indexes  Indexes into `src`; must satisfy
+                       `0 <= indexes[i] < indexes.Dim0()` if
```
qindazhu, line 430: `src.Dim0()`?

k2/csrc/array_ops.h
```diff
@@ -380,6 +406,40 @@ Array2<T> ToContiguous(const Array2<T> &src);
...
+                       `allow_minus_one == true`,
```
qindazhu, line 431: `allow_minux_one == false`?

k2/csrc/array_ops.h
```diff
@@ -380,6 +406,40 @@ Array2<T> ToContiguous(const Array2<T> &src);
...
+                       else -1 is also allowed and the corresponding
+                       output element will be zero.
+     @return  Returns an `Array2<T>` of shape (indexes.Dim(), src.Dim1()),
+               with `ans[i,j] = src[indexes[i,j]]` (or zero if
```
qindazhu, line 435: `ans[i,j] = src[indexes[i], j]`? seems this function is to do a  row-index/select? maybe it's better to add documentation at the first line, or use another function name (e.g. IndexRow or something else, I'm not sure)?

k2/csrc/array_ops_test.cu
```diff
@@ -1302,6 +1302,84 @@ void TestMonotonicLowerBound() {
...
+
+
+    Array1<int32_t> indexes_no_minus_one = RandUniformArray1<int32_t>(
+        c, ans_dim, 0, src_dim - 1),
+                 indexes_minus_one = RandUniformArray1<int32_t>(
+                     c, ans_dim, -1, src_dim - 1);
+    Array1<T> ans_no_minus_one = Index(src, indexes_no_minus_one, false),
+        ans_no_minus_one_check = src[indexes_no_minus_one],
+       ans_no_minus_one_check2 = Index(src, indexes_no_minus_one, true);
+    ASSERT_EQ(Equal(ans_no_minus_one, ans_no_minus_one_check), true);
```
qindazhu, line 1325: just metion it, you can use `ASSERT_TRUE` or `ASSERT_FALSE` in such case.

k2/csrc/array_ops.h
danpovey: MM.. I just called it Index because it's the same thing as what [] does in Python.  But yes, IndexRow() makes sense. 
Can you change that and fix the other typos?
Regarding testing <0 instead of -1: it saves one instruction, it's an optimization because the only -ve number allowed is -1.

danpovey: IndexRows() would be fine as a name.

qindazhu: OK, will do.

PR #384, opened 2020-11-22
Bug fixes

danpovey:


danpovey:
I made the properties dicts rather than OrderedDicts because I couldn't see the reason to make them ordered dicts... LMK about objections, we can change it back or solve it a different way.

PR #387, opened 2020-11-22
Fix some ctest fails

jimbozhang:
The following unit tests failed on my environment (where the torch version is `1.7.0+cu101`):
```
         32 - get_tot_scores_test_py (Failed)
         36 - fsa_test_py (Failed)
         38 - index_add_test_py (Failed)
         46 - union_test_py (Failed)
```

This PR fixed the test `fsa_test_py` and `index_add_test_py`. But the `get_tot_scores_test_py` and `union_test_py` still fails with the error:

```
  File "/home/tj-storage07/zhangjunbo/src/k2/k2/python/tests/union_test.py", line 107, in test_autograd
    fsa0 = k2.Fsa.from_str(s0).to(device).requires_grad_(True)
  File "/home/tj-storage07/zhangjunbo/src/k2/k2/python/k2/fsa.py", line 677, in to
    self.arcs = self.arcs.to(device)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

I haven't found the solution. Could @csukuangfj have a look?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,7 @@ GSYMS
 GPATH
 tags
 TAGS
+.tags
 
 # python build files
 **/__pycache__
--- a/k2/python/tests/fsa_test.py
+++ b/k2/python/tests/fsa_test.py
@@ -176,7 +176,7 @@ class TestFsa(unittest.TestCase):
     def test_transducer_from_tensor(self):
         devices = [torch.device('cpu')]
         if torch.cuda.is_available():
-            devices.append('cuda', 0)
+            devices.append(torch.device('cuda'))
 
         for device in devices:
             fsa_tensor = torch.tensor(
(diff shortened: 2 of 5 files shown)


csukuangfj:
@jimbozhang 
Thanks.
It turns out all of the following assignments with `to` are equivalent. That might be the reason that all tests got passed on my computer.

```python
a = torch.tensor([1])
b = a.to(torch.device('cuda', 0))
c = a.to('cuda')
d = a.to(0)
```

csukuangfj:
@jimbozhang 
It seems that you are.not.using the latest master branch.
https://github.com/k2-fsa/k2/blob/adf717276b05608f1cec7305bac68f6da0b22da9/k2/python/k2/fsa.py#L667

Could you try the latest k2 on the master branch and see if  the test failures remain.

csukuangfj:
@jimbozhang 
Could you change `torch.device('cuda')` to `torch.device('cuda', 0)`? This should fix the problem. I will make a change
to handle this case when it is called without the extra `0`.

Sorry, I only run all the tests without GPU.

jimbozhang:
Thanks @csukuangfj , all unit test passes now.

csukuangfj:
thanks, merging.

PR #393, opened 2020-11-23
Support virtualenv by upgrading pybind11 from v2.5 to v2.6

jimbozhang:
`PYTHON_EXECUTABLE` is set as `/usr/bin/python` by pybind11 under version 2.5, which should be `<venv-path>/bin/python` for the virtual environment.
The version 2.6 pybind11 works well with virtualenv.

I have tested the pybind11 v2.6, and all the unit tests passed. So I think it is safe to upgrade pybind11 from v2.5 to v2.6.

### DIFF
--- a/cmake/pybind11.cmake
+++ b/cmake/pybind11.cmake
@@ -8,8 +8,8 @@ function(download_pybind11)
 
   include(FetchContent)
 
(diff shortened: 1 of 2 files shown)


csukuangfj:
+2
thanks, merging

PR #395, opened 2020-11-23
Support Valgrind for unit tests

jimbozhang:
If `valgrind` was installed, it allows to run `ctest -R <NAME> -D ExperimentalMemCheck` to check heap corruptions and memory leaks.


csukuangfj:
+1

Are there any memory leaks found by valgrind?

cmake/Modules/FindValgrind.cmake
```diff
@@ -0,0 +1,44 @@
...
+find_program(Valgrind_EXECUTABLE NAMES valgrind PATH ${Valgrind_BINARY_PATH})
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(Valgrind DEFAULT_MSG Valgrind_INCLUDE_DIR Valgrind_EXECUTABLE)
+
+mark_as_advanced(Valgrind_INCLUDE_DIR Valgrind_EXECUTABLE)
+
+if(NOT Valgrind_FOUND)
+  if(Valgrind_FIND_REQUIRED)
+     message(FATAL_ERROR "Valgrind required but it seems has not be installed. ${ERR_MSG}")
```
csukuangfj, line 39: Where is `ERR_MSG` defined?

csukuangfj, line 39: it seems **it** has not

CMakeLists.txt
```diff
@@ -106,6 +106,14 @@ foreach(COMPUTE_ARCH IN LISTS K2_COMPUTE_ARCHS)
 endforeach()
 # ========= Settings for CUB end=========
 
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/Modules)
+find_package(Valgrind)
+if(Valgrind_FOUND)
+  find_program(MEMORYCHECK_COMMAND ${Valgrind_EXECUTABLE})
```
csukuangfj, line 112: I think `Valgrind_EXECUTABLE` is a path, but `find_program`
https://github.com/k2-fsa/k2/blob/3dc0d7cb108bd9dbef522b45f7baeb7ebaa3f33c/cmake/Modules/FindValgrind.cmake#L30
requires a `name` or `NAMES` argument.

https://cmake.org/cmake/help/latest/command/find_program.html

jimbozhang:
> Are there any memory leaks found by valgrind?

Yes, many memory leaks. But most of them occurs inside `libcuda.so` and `libtorch_cuda.so`. You can run `ctest -D ExperimentalMemCheck` to generate the reports.

csukuangfj:
> Yes, many memory leaks. But most of them occurs inside libcuda.so and libtorch_cuda.so

I think there is little we can do about the warnings from `libcuda.so` and `libtorch_cuda.so`. Can we suppress these warnings?

danpovey:
Perhaps you could mantion at the end of INSTALL how to run the valgrind checks.

jimbozhang:
Updated. Suppressed Valgrind messages of cuda libraries, and added some instructions in `INSTALL`.
@csukuangfj @danpovey

jimbozhang:
For the valgrind checking is very slow, I just checked some tests, and no mem leak reported, with this suppression.

danpovey:
Fantastic.  Guys, pls merge if you think it's OK...

PR #396, opened 2020-11-24
implement simple version index with Ragged for determinize and rm-epsilon

qindazhu:
Will add tests in another PR.


danpovey:
Cool!

PR #397, opened 2020-11-24
remove template on device type for tests

qindazhu:
And changed some functions' context parameter from `ContextPtr& c` to `ContextPtr c` to accept lvalue (It also could be fine if we change them to `const ContextPtr& c`, but as `ContextPtr` is a shared_ptr, I think it is very cheap to pass by value).


csukuangfj:
+1

PR #398, opened 2020-11-24
Add Closure operation

### LINKED ISSUE #390: possibility to provide fstclosure?
I realize I need an operation equivalent to fstclosure to  make an fsa accepting "(a|b|c)*" from an existing fsa accepting "(a|b|c)". Would it be possible to provide such function? Thanks!

csukuangfj:
Closes #390 

- [x] add arc_map
- [x] add more tests with random fsas
- [x] add Python interface and support autograd
- [x] handle aux_labels in Python

### DIFF
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -782,8 +782,8 @@ Fsa Closure(Fsa &fsa, Array1<int32_t> *arc_map /* = nullptr*/) {
     int32_t fsa_state_idx0 = fsa_row_ids_data[fsa_arc_idx01];
     int32_t fsa_arc_idx0x = fsa_row_splits_data[fsa_state_idx0];
     int32_t fsa_arc_idx1 = fsa_arc_idx01 - fsa_arc_idx0x;
-    int32_t this_state_num_arcs = fsa_row_splits_data[fsa_state_idx0 + 1] -
-                                  fsa_row_splits_data[fsa_state_idx0];
+    int32_t this_state_num_arcs =
+        fsa_row_splits_data[fsa_state_idx0 + 1] - fsa_arc_idx0x;
 
     Arc arc = fsa_arcs_data[fsa_arc_idx01];
     if (arc.dest_state == fsa_final_state) {
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -775,6 +775,9 @@ Fsa Closure(Fsa &fsa, Array1<int32_t> *arc_map /* = nullptr*/) {
   Array1<Arc> out_arcs(c, num_out_arcs);
   Arc *out_arcs_data = out_arcs.Data();
 
+  Array1<int32_t> tmp_arc_map(c, num_out_arcs, -1);
+  int32_t *tmp_arc_map_data = tmp_arc_map.Data();
+
   auto lambda_set_arcs = [=] __host__ __device__(int32_t fsa_arc_idx01) {
     int32_t fsa_state_idx0 = fsa_row_ids_data[fsa_arc_idx01];
     int32_t fsa_arc_idx0x = fsa_row_splits_data[fsa_state_idx0];
@@ -815,11 +818,15 @@ Fsa Closure(Fsa &fsa, Array1<int32_t> *arc_map /* = nullptr*/) {
       out_row_ids_data[0] = 0;
     }
 
+    tmp_arc_map_data[out_arc_idx01] = fsa_arc_idx01;
+
     out_arcs_data[out_arc_idx01] = arc;
     out_row_ids_data[out_arc_idx01] = arc.src_state;
   };
 
   Eval(c, fsa.values.Dim(), lambda_set_arcs);
+
+  if (arc_map != nullptr) *arc_map = std::move(tmp_arc_map);
   Array1<int32_t> out_row_splits(c, num_out_states + 1);
   RowIdsToRowSplits(out_row_ids, &out_row_splits);
   RaggedShape shape = RaggedShape2(&out_row_splits, &out_row_ids, num_out_arcs);
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -788,6 +788,7 @@ Fsa Closure(Fsa &fsa, Array1<int32_t> *arc_map /* = nullptr*/) {
       //   - dest_state == 0
       //   - label == 0
       arc.dest_state = 0;
+      K2_DCHECK_EQ(arc.label, -1);
       arc.label = 0;
     }
 
@@ -799,15 +800,17 @@ Fsa Closure(Fsa &fsa, Array1<int32_t> *arc_map /* = nullptr*/) {
     } else {
       out_arc_idx01 = fsa_arc_idx01;
       if (fsa_arc_idx1 == this_state_num_arcs - 1) {
-        Arc new_arc(0, fsa_final_state, 0, 0.0f);
+        // This is the last arc of the original start state,
+        // so we add a new arc just after it.
+        Arc new_arc(0, fsa_final_state, -1, 0.0f);
         out_arcs_data[out_arc_idx01 + 1] = new_arc;
         out_row_ids_data[out_arc_idx01 + 1] = 0;
       }
     }
 
-    // it may happen that the start state have no leaving arcs
+    // it may happen that the start state has no leaving arcs
     if (fsa_row_splits_data[1] == 0) {
-      Arc new_arc(0, fsa_final_state, 0, 0.0f);
+      Arc new_arc(0, fsa_final_state, -1, 0.0f);
       out_arcs_data[0] = new_arc;
       out_row_ids_data[0] = 0;
     }
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -509,7 +509,7 @@ void AddEpsilonSelfLoops(FsaOrVec &src, FsaOrVec *dest,
     if (num_states < 2) {
       K2_CHECK_EQ(num_states, 0);
       *dest = src;
-      if (arc_map) *arc_map = Array1<int32_t>(c, 0);
+      if (arc_map != nullptr) *arc_map = Array1<int32_t>(c, 0);
       return;
     }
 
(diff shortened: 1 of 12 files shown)


danpovey:
Cool!

k2/csrc/fsa_algo.cu
```diff
@@ -733,4 +733,91 @@ Fsa Union(FsaVec &fsas, Array1<int32_t> *arc_map /*= nullptr*/) {
...
+
+  Array1<int32_t> tmp_arc_map(c, num_out_arcs, -1);
+  int32_t *tmp_arc_map_data = tmp_arc_map.Data();
+
+  auto lambda_set_arcs = [=] __host__ __device__(int32_t fsa_arc_idx01) {
+    int32_t fsa_state_idx0 = fsa_row_ids_data[fsa_arc_idx01];
+    int32_t fsa_arc_idx0x = fsa_row_splits_data[fsa_state_idx0];
+    int32_t fsa_arc_idx1 = fsa_arc_idx01 - fsa_arc_idx0x;
+    int32_t this_state_num_arcs = fsa_row_splits_data[fsa_state_idx0 + 1] -
+                                  fsa_row_splits_data[fsa_state_idx0];
```
freewym, line 772: may be better to use `fsa_arc_idx0x`

danpovey: I was having difficulty with what to call such things.  Technically this is a difference of two arc_idx0x's, which I think makes it an idx1 (i.e. a count of arcs), but it's a special kind of idx1.

csukuangfj: > may be better to use fsa_arc_idx0x

thanks, will change it. I guess the compiler will make some optimizations to reuse `fsa_arc_idx0x` in line 769.

danpovey:
You could *maybe* call it num_arcs_idx0x.  But I have actually been
avoiding calling such things idx0x because they are not really usable as
an idx0x, it behaves more like an idx1.   Maybe num_arcs_idx1

qindazhu:
I guess what @freewym means is to reuse `fsa_arc_idx0x` in line 772? BTW, I prefer name `this_stat_num_arcs` or `num_arcs_this_state`

k2/csrc/fsa_algo.h
```diff
@@ -298,6 +298,38 @@ Ragged<int32_t> ShortestPath(FsaVec &fsas,
...
+   Note: the output will be arc-sorted if the input was arc-sorted and
+   epsilon-free (actually we only need for the start-state to not have epsilon
+   arcs leaving it). The output will not be epsilon-free though, so if an
+   epsilon-free output is desired, which it generally will be, the caller will
+   need to determinize afterward.
+
+   Caution: The caller will have to modify any extra labels (like aux_labels) to
+   deal with -1's correctly.
+
+   @param in]   fsa        The input FSA.
```
qindazhu, line 323: maybe better to add `Must have NumAxes() == 2` or `it must be a single FSA`

freewym:
> I guess what @freewym means is to reuse `fsa_arc_idx0x` in line 772? BTW, I prefer name `this_stat_num_arcs` or `num_arcs_this_state`

Yes I meant that. For me.  The name `this_state_num_arcs` helped me understand the code and guess what `fsa_row_splits` means before seeing  its definition

k2/csrc/fsa_algo.h
```diff
@@ -298,6 +298,38 @@ Ragged<int32_t> ShortestPath(FsaVec &fsas,
...
+   - For all arcs entering the final-state, change the destination-state from
+   the final-state to state 0 (the start state), and the label from -1 to 0.
+
+   - Add an arc from the start-state to the final-state (with label -1, and
+   score 0, of course); this can be the last numbered arc of the start state of
+   the output. The implementation wouldn't be in-place, we'd create a new FSA.
+   Assume that the original start state has two arcs, then the arc-map would be
+   [ 0 1 -1 2 3 ... ].
+
+   Note: the output will be arc-sorted if the input was arc-sorted and
```
qindazhu, line 314: How does the current implementation make sure the ouput arc-sorted if input arc-sorted? As we changed label from -1 to 0 for those final-arcs, accoding to our definition of arc-sort (compared labels with type `uint32_t` instead of `int32_t`), the output would not be arc-sorted as arc with label `0` now is the last arc leaving a state.

k2/csrc/fsa_algo.h
```diff
@@ -298,6 +298,38 @@ Ragged<int32_t> ShortestPath(FsaVec &fsas,
...
+   epsilon-free (actually we only need for the start-state to not have epsilon
+   arcs leaving it). The output will not be epsilon-free though, so if an
+   epsilon-free output is desired, which it generally will be, the caller will
+   need to determinize afterward.
```
qindazhu, line 318: `the caller should call rmepsilon afterward`?

k2/csrc/fsa_algo.h
```diff
@@ -298,6 +298,38 @@ Ragged<int32_t> ShortestPath(FsaVec &fsas,
...
+   The closure is implemented in the following way:
+
+   - For all arcs entering the final-state, change the destination-state from
+   the final-state to state 0 (the start state), and the label from -1 to 0.
+
+   - Add an arc from the start-state to the final-state (with label -1, and
+   score 0, of course); this can be the last numbered arc of the start state of
+   the output. The implementation wouldn't be in-place, we'd create a new FSA.
+   Assume that the original start state has two arcs, then the arc-map would be
+   [ 0 1 -1 2 3 ... ].
```
qindazhu, line 312: I just thought if we can implement it with such approach:
- add another super final_state, and let the original fina_state has an arc with label -1 to the super final state and an arc with label 0 to the start state
- change all final_arcs' labels from -1 to 0, still has the original final_state as the dest_state.

This approach may be simpler as all row_splits and row_ids in the original Fsa can be copied to the out FSA (as we don't insert an arc for start_state), all we need to do this to add one super final_state and two arcs for it. The logic in the lambda would be much simpler. Correct me if I'm wrong.

k2/csrc/fsa_algo.h
csukuangfj: Some states of the returned FSA  may have more than two leaving arcs with label == 0 even if
the input FSA is determinized.

k2/csrc/fsa_algo.h
csukuangfj: I agree. The downside is we have one more state.

k2/csrc/fsa_algo.h
csukuangfj: The output fsa will not be arc sorted even if the input fsa was arc sorted. I am updating the documentation.

k2/csrc/fsa_algo.h
csukuangfj: If @danpovey  agrees with this approach, I would like to change the code.

k2/csrc/fsa_algo.h
qindazhu: > The output will not be epsilon-free though, so if an epsilon-free output is desired, which it generally will be, the caller will need to determinize afterward.

The point here is even we call determinize, the output will still have epsilon-arcs (even though the input is epsilon-free), as we have changed label from -1 to 0 for those final-arcs. so `if an epsilon-free output is desired` , we should call rm-epsilon instead of determinize.

BTW,  there may be two arcs with lable -1 from start_state to final_state in current approach.

k2/csrc/fsa_algo.h
qindazhu: Another advantage of approach I metioned above is: if the input is top-sorted, then the ouput will be top-sorted.

k2/csrc/fsa_algo.cu
csukuangfj: Fixed.

k2/csrc/fsa_algo.h
csukuangfj: thanks, added.

k2/csrc/fsa_algo.h
freewym: > I just thought maybe we can implement it with such approach:
> 
> * add another super final_state, and let the original fina_state has an arc with label -1 to the super final state and an arc with label 0 to the start state
> * change all final_arcs' labels from -1 to 0, still has the original final_state as the dest_state.
> 
> This approach may be simpler as all row_splits and row_ids in the original Fsa can be copied to the out FSA (as we don't insert an arc for start_state), all we need to do this to add one super final_state and two arcs for it. The logic in the lambda would be much simpler. Correct me if I'm wrong.

May still need to add an epsilon arc from the start state to the super final state to make the new fsa accept empty string

k2/csrc/fsa_algo.h
csukuangfj: Yes, you're right. Will change it.

k2/csrc/fsa_algo.h
qindazhu: Another advantage of the approach I metioned above is: we can make sure the output arc-sorted easily as the row_splits of the original FSA is not changed.  We can do `count` to know how many epsilon-arcs for each state, and then re-arrange the input arcs accoding to the number we just get. Just mention it, we may not need to implement arc-sort in this function.

qindazhu: > May still need to add an epsilon arc from the start state to the original final state to make the new fsa accept empty string

Ah, OK, your requirement is A* instead of A+. Then we have to insert an arc for start state and can't use the approach I metioned above.

k2/csrc/fsa_algo.cu
```diff
@@ -733,4 +733,91 @@ Fsa Union(FsaVec &fsas, Array1<int32_t> *arc_map /*= nullptr*/) {
...
+
+    out_arcs_data[out_arc_idx01] = arc;
+    out_row_ids_data[out_arc_idx01] = arc.src_state;
+  };
+
+  Eval(c, fsa.values.Dim(), lambda_set_arcs);
+
+  if (arc_map != nullptr) *arc_map = std::move(tmp_arc_map);
+  Array1<int32_t> out_row_splits(c, num_out_states + 1);
+  RowIdsToRowSplits(out_row_ids, &out_row_splits);
```
qindazhu, line 817: However, we can still get row_splits by doing something like Plus(input_fsa_row_splits, 1)?

k2/csrc/fsa_algo.h
qindazhu: Just metion it, actually we can do this by:
1. do `count` on the input Fsa to get how many epsilon-arcs we have for each state
2. for a specific arc, if it's label == 0, then just copy to the corresponding position in the output Fsa; if lable != 0, then copy to the corresponding position +1. 
3. Add an arc from this state to start state, with label == 0 at position count.

k2/csrc/fsa_algo.cu
csukuangfj: Assume the input fsa_row_splits is `[0, 2, 3, 7, 7]` and every state has at least one leaving arc,
then output fsa_row_splits will be `[0, 3, 4, 8, 8]`

We have to handle the first element carefully.

danpovey:
I think it may be bug-prone to make the algorithm too complicated.  We
don't need this to be super-efficient; better keep it simple IMO.

csukuangfj:
ok, I will leave the current implementation as it is.

k2/csrc/fsa_algo.cu
qindazhu: yeah, then we should not call Plus, instead, do this in a kernel and handle 0th element when idx==0

qindazhu: I believe this is much efficient than calling `RowIdsToRowSplits`

csukuangfj: Will change it.

csukuangfj: Now I remember that I have tried to compute row splits in the lambda while implementing Union.
It is quite error-prone. We have to consider edges cases like some states may have no leaving arcs at all.

Computing row splits inside the lambda will make the code quite complicated.

qindazhu: what's the complexity here? Are you doing this in a separate lambda? e.g.
```
auto lambda_set_row_splits = [=]  __device__(int32_t i){
  if(i = 0) ans_row_splits_data[i] = 0;
  else  ans_row_splits_data[i] = src_row_splits_data[i] + 1;
};
EvalDevice(c, row_splits.Dim(), lambda_set_row_splits);
```

csukuangfj: thanks, fixed.

k2/csrc/fsa_algo.cu
```diff
@@ -747,4 +747,100 @@ Fsa Union(FsaVec &fsas, Array1<int32_t> *arc_map /*= nullptr*/) {
...
+  // An arc from the start state to the final state with label == 0 is added.
+  int32_t num_out_arcs = fsa.values.Dim() + 1;
+
+  Array1<int32_t> out_row_ids(c, num_out_arcs);
+  int32_t *out_row_ids_data = out_row_ids.Data();
+
+  Array1<Arc> out_arcs(c, num_out_arcs);
+  Arc *out_arcs_data = out_arcs.Data();
+
+  Array1<int32_t> tmp_arc_map(c, num_out_arcs, -1);
```
qindazhu, line 778: it may not need to init with -1 if you set arc map data for the added arc from start_state to final_state

csukuangfj: yes, you are right.

k2/csrc/fsa_algo.cu
```diff
@@ -747,4 +747,100 @@ Fsa Union(FsaVec &fsas, Array1<int32_t> *arc_map /*= nullptr*/) {
...
+    out_row_ids_data[out_arc_idx01] = arc.src_state;
+  };
+
+  Eval(c, fsa.values.Dim(), lambda_set_arcs);
+
+  if (arc_map != nullptr) *arc_map = std::move(tmp_arc_map);
+
+  Array1<int32_t> out_row_splits(c, num_out_states + 1);
+  int32_t *out_row_splits_data = out_row_splits.Data();
+  auto lambda_set_row_splits = [=] __host__ __device__(int32_t i) {
```
qindazhu, line 833: @csukuangfj , you may need to do this
```
if(c->GetDeviceType() == kCpu)
  do this in a for-loop
else
  call EvalDevice with this lambda.
```
Otherwise we may get crash in Release build as the lambda is too short.

csukuangfj: Fixed. Thanks

qindazhu:
+1

k2/csrc/fsa_algo_test.cu
```diff
@@ -585,4 +585,119 @@ TEST(FsaAlgo, Determinize) {
...
+    Fsa ans = Closure(fsa, &arc_map);
+
+    arc_map = arc_map.To(GetCpuContext());
+    CheckArrayData(arc_map, std::vector<int32_t>{0, -1, 1, 2});
+
+    ans = ans.To(GetCpuContext());
+    EXPECT_EQ(ans.TotSize(0), 4);  // number of states
+    EXPECT_EQ(ans.TotSize(1), 4);  // number of arcs
+
+    EXPECT_EQ((ans[{0, 0}]), (Arc{0, 1, 1, 0.1f}));   // state 0, arc 0
```
qindazhu, line 608: Just metion it (we don't need to do this now), the current test approach depends on the implementation, we may need to find a implementation independent way to test `Closure`. One of a way I think of is, we construct a src FSA with every leaving arc for a state has different label, then for any path `A` in src, check if we can find AA in ans, or check if we can find A in ans and the last state has an arc entering the start_state with label 0, etc. Not sure if @danpovey has any better idea.

danpovey:
What I'd do is get a path from src, then check that (that path times a
random number of repeats [0,1,2,3] for instance, is in the closure.
Of course that doesn't check that the closure doesn't contain *extra*
paths.  Could do that by, say, checking that: Union(Append(src, closure),
[Fsa that only accepts empty sequence]) is equivalent to closure.

But I don't think we need to spend *too* much time on this.

PR #399, opened 2020-11-24
Intersection simplification, interface changes

danpovey:


danpovey:
caution: this will break snowfall, will commit fix separately.

PR #406, opened 2020-11-25
Add Sort for Array1.

### LINKED ISSUE #405: Sort() needed.
It would be nice if we had a Sort method that would work with Array1<T>.
I'm open minded on the interface.  Right now I just need it to sort int32_s, but I want the `argsort` output, i.e.
I need the order.  Perhaps someone can look at our underlying libraries' sort capabilities (e.g. moderngpu
and/or cub) and write something?
Right now my main concern is that it be done fast, I am not too picky on the details or the speed.

csukuangfj:
Closes #405

### DIFF
--- a/k2/csrc/array_ops_test.cu
+++ b/k2/csrc/array_ops_test.cu
@@ -1414,7 +1414,7 @@ TEST(OpsTest, Array2IndexTest) {
 }
 
 template <typename T>
-static void Array1SortTest() {
+static void Array1SortTestSimple() {
   std::vector<T> data = {3, 2, 5, 1};
   for (auto &context : {GetCpuContext(), GetCudaContext()}) {
     {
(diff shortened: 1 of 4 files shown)


k2/csrc/array_ops_test.cu
```diff
@@ -1295,4 +1296,31 @@ TEST(OpsTest, Array2IndexTest) {
   }
 }
 
+template <typename T>
+static void Array1SortTest() {
+  std::vector<T> data = {3, 2, 5, 1};
```
qindazhu, line 1301: it's better if we can random tests for this (especially the empty case, I'm not convinced by what mgpu does with empty case, IIRC, there was a bug about this before in `GetTransposeReordering`)

csukuangfj: Will add tests for random cases and empty cases.

csukuangfj: Fixed.

csukuangfj:
Ready for review.

k2/csrc/array_ops_test.cu
```diff
@@ -1295,4 +1296,81 @@ TEST(OpsTest, Array2IndexTest) {
...
+}
+
+template <typename T>
+static void Array1SortTestRandom() {
+  int32_t num = RandIntGenerator()(0, 10000);
+  std::vector<T> data(num);
+  std::iota(data.begin(), data.end(), 0);
+  std::vector<T> saved_data = data;
+
+  std::random_shuffle(data.begin(), data.end());
```
qindazhu, line 1339: just call `RandUniformArray1` would be fine., no `iota`, no `shuffle`

qindazhu: And you can check ans with `is_sorted`

qindazhu: BTW, iota approach has a disadvantage that it never generates duplicate values

k2/csrc/array_ops_test.cu
```diff
@@ -1295,4 +1296,81 @@ TEST(OpsTest, Array2IndexTest) {
...
+      Array1<T> array(context, data);
+      Sort(&array);
+
+      CheckArrayData(array, saved_data);
+    }
+  }
+}
+
+TEST(OpsTest, Array1Sort) {
+  Array1SortTestSimple<int32_t>();
```
qindazhu, line 1366: it doesn't matter, but just wondering why you don't put those template test cases into one function, e.g.
```
for(context....){
{ // empty case..}
{// simple case ...}
}
```
it may save some duplicate code. Anyway, it doesn't matter, it's up to you to change it or not

k2/csrc/macros.h
```diff
@@ -21,4 +21,4 @@
 #define K2_FUNC __func__
 #endif
 
-#endif  // K2_MACROS_H_
+#endif  // K2_CSRC_MACROS_H_
```
qindazhu, line 24: mm, Sorry, seems we have conflicts here as I changed it as well in https://github.com/k2-fsa/k2/pull/415. I'll merge that PR and you may need to revert your changes, many thanks.

k2/csrc/array_ops_test.cu
csukuangfj: Just want to make the test code simple and modular.
I prefer to test each case in a single function.

k2/csrc/macros.h
csukuangfj: Conflicts have been resolved.

k2/csrc/array_ops_test.cu
csukuangfj: Was not aware of `std::is_sort`,  thanks.

Issue #353, opened 2020-11-16
Crashes/segfaults when compiled in release mode


danpovey:
Haowen and I have been debugging this today.
It can be fixed by setting optimization level to 0; it still happens at optimization level=1.
How to set that optimization is complicated:
```
git diff CMakeLists.txt  
diff --git a/CMakeLists.txt b/CMakeLists.txt
index e261522..65fd97f 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,10 @@ endif()
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0")
 set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -O0")
 
+set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -g -O0")
+set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -g -O0")
+
+
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 option(BUILD_SHARED_LIBS "Whether to build shared or static lib" ON)
@@ -91,6 +95,9 @@ else()
     set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --expt-extended-lambda -gencode arch=compute_${COMPUTE_ARCH},code=sm_${COMPUTE_ARCH}")
   endforeach()
 endif()
+set(CMAKE_CUDA_FLAGS_RELEASE "-DNDEBUG -O1 -g")
+
+
 
 enable_testing()
 
```

danpovey:
Fangjun: we haven't made any progress on the crashes in release mode.  I suspect it may be a bug in CUDA 10.1 that can't be worked around other than by using  -O0.  What version of the CUDA toolkit are you compiling with at your work?

csukuangfj:
I am also using cuda 10.1. I can try it with cuda 10.2

qindazhu:
The simplest code I found until now which can repro this bug on my local machine (cuda 10.1) is:
```
ContextPtr context = GetCudaContext(1);
 Array1<int32_t> a(context, 1, 0);
```

it will crash at this line https://github.com/k2-fsa/k2/blob/1bd2652e692ff5d8264b449b8c4cd37bc5560f9f/k2/csrc/array.h#L283-L284

[...]

PR #407, opened 2020-11-26
Update INSTALL.md

csukuangfj:
Users do not need to install CUDA toolkit to run k2.

@qindazhu Please have a review.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -13,27 +13,31 @@ you install is compatible with the CUDA toolkit you are using.
 For example, if your local CUDA toolkit version is 10.1, and
 you want to install `torch 1.6.0`, you should install it with:
 
-```
+```bash
 pip install torch==1.6.0+cu101 -f https://download.pytorch.org/whl/torch_stable.html
 ```
 
-If you use conda to manage packages, you can install those 
+If you use conda to manage packages, you can install those
 dependencies with:
 
-```
+```bash
 conda create -n k2 python=3.7
 conda activate k2
 conda install pytorch==1.6.0 cudatoolkit=10.1 -c pytorch
 ```
 
-To install other versions of torch, please check below URLs:
+To install other versions of torch, please check the following URLs:
 
-https://pytorch.org/get-started/locally/
+- https://pytorch.org/get-started/locally/
 
-https://pytorch.org/get-started/previous-versions/
+- https://pytorch.org/get-started/previous-versions/
+
+**HINT**: If you install k2 from pre-built wheel packages, neither do you need
+a GPU nor need to install CUDA toolkit. k2 is ready to run with a CPU. However,
+you still need to install PyTorch with CUDA support.
 
 
-## Install k2 from Pre-built wheel packages
+## Install k2 from pre-built wheel packages
 
 There are two ways to install k2 from pre-built wheel packages.
 
@@ -91,10 +95,10 @@ Before compiling k2, some preparation work has to be done:
   CUDA toolkit you are using.
 
 (Note we need NVCC to build k2, if you use conda to install CUDA toolkit,
-you may need to install nvcc_linux-64 or cudatoolkit-dev as well since the
+you may need to install `nvcc_linux-64` or `cudatoolkit-dev` as well since the
 default installation of CUDA toolkit in conda did not include NVCC.
-However, nvcc_linux-64 or cudatoolkit-dev may not work well on all platforms,
-so it's better if can install CUDA toolkit using a normal way instead of 
+However, `nvcc_linux-64` or `cudatoolkit-dev` may not work well on all platforms,
+so it's better if you can install CUDA toolkit using a normal way instead of
 using conda if you want to build k2 from source.)
 
 After setting up the environment, we are ready to build k2:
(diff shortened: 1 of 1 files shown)


qindazhu:
+1, merge

danpovey:
Have we tested this?  Wouldn't it try to dynamically load libraries from
the CUDA toolkit?

csukuangfj:
Yes, I have verified it locally on a virtual machine with Ubuntu 18.04. There are no GPUs or CUDA toolkit on the VM. I only need to install torch==1.6.0+cu101 and install k2 downloaded from GitHub actions. All  Python tests get passed.

GetCudaContext will not be called in this case since all Python tests will not use GPU if torch.cuda.is_available() returns False.




Thursday, 26 November 2020, 12:06 +0800 from notifications@github.com  <notifications@github.com>:

danpovey:
And all CPU tests?  Surely it would try to load the CUDA library to see if
a GPU is available?

csukuangfj:
Yes, all CPU tests got passed.

You can see from the following output that there is a function `_lazy_init`, which is not called if we do
not access the GPU.

```
$ python3
Python 3.6.9 (default, Oct  8 2020, 12:12:24) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
>>> torch.cuda.is_available()
False
>>> a = torch.arange(3)
>>> a.cuda()
  File "<stdin>", line 1, in <module>
  File "/home/fangjun/sdb1/py36/lib/python3.6/site-packages/torch/cuda/__init__.py", line 186, in _lazy_init
    _check_driver()
  File "/home/fangjun/sdb1/py36/lib/python3.6/site-packages/torch/cuda/__init__.py", line 68, in _check_driver
    http://www.nvidia.com/Download/index.aspx""")
AssertionError: 
Found no NVIDIA driver on your system. Please check that you
have an NVIDIA GPU and installed a driver from
http://www.nvidia.com/Download/index.aspx
>>> import k2
>>> s = '''0 1 -1 0\n1'''
>>> fsa = k2.Fsa.from_str(s)
>>> fsa = fsa.to(torch.device('cuda', 0))
[W] /home/runner/work/k2/k2/k2/csrc/pytorch_context.cu:operator():126 CUDA is not available. Return a CPU context.
  File "<stdin>", line 1, in <module>
  File "/home/fangjun/sdb1/py36/lib/python3.6/site-packages/k2/fsa.py", line 688, in to
    phantom_set_scores_to(ans, self.scores.to(device))
  File "/home/fangjun/sdb1/py36/lib/python3.6/site-packages/torch/cuda/__init__.py", line 186, in _lazy_init
    _check_driver()
  File "/home/fangjun/sdb1/py36/lib/python3.6/site-packages/torch/cuda/__init__.py", line 68, in _check_driver
    http://www.nvidia.com/Download/index.aspx""")
AssertionError: 
Found no NVIDIA driver on your system. Please check that you
have an NVIDIA GPU and installed a driver from
http://www.nvidia.com/Download/index.aspx
```

csukuangfj:
RE CPU tests, i.e., C++ tests, they are all passed without GPU as can be seen from
https://github.com/k2-fsa/k2/actions?query=workflow%3Arun-tests

However, the **Release** build will segfault on GitHub actions, which I cannot reproduce locally.
Add C++ tests are able to run without the need of a GPU.

To run C++ tests, they have to be compiled first, which means CUDA toolkit has to be installed.

danpovey:
can you show the url where the C++ tests are failing in release?
I can try to repro on a remote machine I have.

csukuangfj:
I commented on it at https://github.com/k2-fsa/k2/issues/353#issuecomment-732478499

One of the URLs is https://github.com/k2-fsa/k2/runs/1440725780?check_suite_focus=true#step:11:276

> /home/runner/work/_temp/24ef5e13-3c98-46fb-9c31-b30c4a1e294d.sh: line 4: 16032 Segmentation fault      (core dumped) ./bin/cu_fsa_algo_test

danpovey:
Hm.  I wonder if we could modify the scripts so it will run with valgrind
on failure, to get a stack trace?
I wasn't able to install CUDNN to make the installation work on that
machine, NVidia makes it too hard to download.

csukuangfj:
@danpovey 
Here is the stack trace using the latest master

https://github.com/csukuangfj/k2/runs/1457937727?check_suite_focus=true#step:11:312

danpovey:
ChangeSublistSize() and Transpose() (in ragged_ops.cu) both contain very short lambdas.
I think it's that compiler bug we previously encountered.  I think those should be replaced with
an if-statement where one branch uses a loop on CPU and the other calls EvalDevice()
using a __device__ macro.

Incidentally, it would be good if we could make minimal repo where we demonstrate this bug, so NVidia
can perhaps fix it (or confirm that it's already fixed, in later versions of their toolkit).
I wonder whether @jimbozhang  could try to do that?

csukuangfj:
> I think those should be replaced with
an if-statement where one branch uses a loop on CPU and the other calls EvalDevice()
using a __device__ lambda.

I have updated the code and am waiting to see whether GitHub actions will segfault or not.

csukuangfj:
Replacing short lambdas with `if + for`  is not able to prevent segfault in GitHub actions for `Release` build.
After some trial-and-error attempts, I managed to avoid segfault by replacing GCC 5 with GCC 6 and GCC 7
without changing a signle line of C++ code. Will create a pullrequest.

PR #408, opened 2020-11-26
Enable NVTX_RANGE for more functions.

csukuangfj:
@danpovey @qindazhu 
Currently, we are using `__func__` for `NVTX_RANGE`. Should we replace it with `__PRETTY_FUNCTION__`?

`__PRETTY_FUNCTION__`, as can be seen from its name, is more readable than `__func__`, though it is an extension
of GCC and is supported by GCC and Clang, not supported by MSVC.

We can define a macro `K2_PRETTY_FUNCTION`, which is equivalent to `__PRETTY_FUNCTION__` when
the compiler is GCC or Clang; otherwise, it is defined as `__func__`.

---

The following code
```cpp
#include <iostream>

template<typename T, int N>
int Foo(int i){
    std::cout << __PRETTY_FUNCTION__ << "\n";
    std::cout << __func__ << "\n";
    return 0;
}
int main()
{
    Foo<float, 10>(2);
    Foo<int, 20>(5);
    return 0;
}
```

prints
```
int Foo(int) [T = float, N = 10]
Foo
int Foo(int) [T = int, N = 20]
Foo
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,7 +87,7 @@ jobs:
           pwd
           mkdir build
           cd build
-          cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE ..
+          cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DK2_ENABLE_NVTX=OFF ..
           cat k2/csrc/version.h
 
       - name: Build _k2
@@ -95,10 +95,7 @@ jobs:
         run: |
           echo "number of cores: $(nproc)"
           cd build
-          make VERBOSE=1
-          ls -lh lib/*
-          rm lib/*
-          make VERBOSE=1 _k2
+          make VERBOSE=1 -j _k2
           ls -lh lib/*
 
       - name: Build pip packages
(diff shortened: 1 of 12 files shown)


.github/workflows/build.yml
```diff
@@ -87,18 +87,15 @@ jobs:
           pwd
           mkdir build
           cd build
-          cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE ..
+          cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DK2_ENABLE_NVTX=OFF ..
```
danpovey, line 90: Does this turn it off?  I don't think nvtx adds much cost so I'd rather have it on at the moment while debugging is our main priority.

danpovey:
Sure, I think your idea is good.
I'd name it __K2_FUNC__ which is shorter to write.

.github/workflows/build.yml
csukuangfj: Yes, it disables NVTX for building wheel packages. I thought wheel files do not need NVTX.
Will revert it.

PR #410, opened 2020-11-26
construct Ragged with RaggedShape and Tensor in python

qindazhu:


k2/python/csrc/torch/ragged.cu
```diff
@@ -18,12 +19,17 @@
...
 namespace k2 {
-
 template <typename T>
 static void PybindRaggedTpl(py::module &m, const char *name) {
   using PyClass = Ragged<T>;
   py::class_<PyClass> pyclass(m, name);
 
+  pyclass.def(py::init([](const RaggedShape &shape, torch::Tensor values) {
+                K2_CHECK_EQ(shape.NumElements(), values.sizes()[0]);
+                return std::unique_ptr<PyClass>(
```
csukuangfj, line 29: I would suggest to use
```cpp
return std::make_unique<PyClass>(shape, FromTensor<T>(values));
```
and to add a return type hint for the lambda.

qindazhu: Yeah, as we required at least C++14,  `std::make_unique` is available. Fixed, thanks!

csukuangfj:
+2

qindazhu:
Merging.

PR #417, opened 2020-11-26
Intersect returns an FsaVec if either input is an FsaVec.

### LINKED ISSUE #414: return type of k2.intersect()
if a and b are both FsaVec but actually contain only one fsa respectively, `k2.intersect(a, b)` will return an Fsa rather than FsaVec.

I think it should still return FsaVec in this case

csukuangfj:
Closes #414


qindazhu:
+1, merging

PR #422, opened 2020-11-27
add num_work_items in ParallelRunner

qindazhu:


k2/csrc/context.h
```diff
@@ -468,7 +476,7 @@ class ParallelRunnerDummy {
   cudaStream_t stream_;
 };
 
-using ParallelRunner = ParallelRunnerDummy;
+using ParallelRunner = ParallelRunnerActive;
```
csukuangfj, line 479: Why is it switched back? `NewStream()` always returns the default stream in the current code.

qindazhu: As the comments showed above, we decided that if `num_work_items < a threshould, e.g. 10k` we don't create a new stream (as Dan found it would be heavy if we create new stream every time (recalling the event stuff related to this), so he got an idea that we only create new stream when >10k)

csukuangfj: The point is that the current code always uses the default argument 0.

qindazhu: Yes, so it's now still equivalent to ParallelRunnerDummy, as you asked in another quesiton.

csukuangfj:
Actually they are not.equivalent.  The dummy  runner returns stream_, but the active runner returns 0.

This pullrequest essentially undoes what Dan has done as number of work items is always 0





Friday, 11 December 2020, 12:33 +0800 from notifications@github.com  <notifications@github.com>:

qindazhu:
They are both default stream (at the beginning of program), the point we used `0x0` here is that `context->GetCudaStream()` may be overrided by `With` if we use `ParallelRunnerActive` instead of `ParallelRunnerDummy` (as user cannot override the default stream in `ParallelRunnerDummy` but they can do this with `ParallelRunnerActive` when num_work_items > 10k with `With`）

csukuangfj:
> They are both default stream (at the beginning of program)

https://github.com/k2-fsa/k2/blob/1d328494367efd3f8341f6f300d66c799ef1894c/k2/csrc/context.h#L509-L510

The constructor of `ParallelRunnerDummy` sets `stream_` to `c->GetCudaStream()`. Is `c->GetCudaStream()` equal
to the default stream at the beginning of the program?

qindazhu:
I think so, at least for default_context, I'm not quite sure about pytorch stuff.

csukuangfj:
> I think so, at least for default_context

This is how `stream_` been created in the default context:

https://github.com/k2-fsa/k2/blob/1d328494367efd3f8341f6f300d66c799ef1894c/k2/csrc/default_context.cu#L59

Is it the same as the default stream, i.e., 0x0 ?

qindazhu:
Oh, we have created a new stream in context, then you're right, we should use `c->GetCudaStream()` here (I mean, the very begining stream of that context, not one overrrided by the following calls)

I think we need to print stream id here to see what happen in a multiple thread case.

PR #423, opened 2020-11-27
Declarations of Prefix, etc

danpovey:


danpovey:
Would be nice if someone can implement this.  This is a couple of functions I'll need for the faster intersection code that I am writing.
Actually I'll only need the 2nd one; the 1st one clarifies it and makes testing easier.

qindazhu:
Taking, will do this at weekend

danpovey:
Thanks!!

PR #424, opened 2020-11-27
reset stream_override_ to zero when stack in CudaStreamOverride is empty

qindazhu:


csukuangfj:
+2

k2/csrc/context.cu
```diff
@@ -26,6 +26,7 @@ void CudaStreamOverride::Pop(cudaStream_t stream) {
   K2_DCHECK(!stack_.empty());
   K2_DCHECK_EQ(stack_.back(), stream);
   stack_.pop_back();
+  if (stack_.empty()) stream_override_ = static_cast<cudaStream_t>(0x0);
```
csukuangfj, line 29: what should be assigned to `stream_override_` if the stack is not empty?
When a stream is pushed to the stack, `stream_override_`  points to the top
of the stack. When an element is popped out, should `stream_override_` points
to the new top?

PR #426, opened 2020-11-28
Implemented Prefix and GetPrefixes for RaggedShape

qindazhu:


k2/csrc/ragged_ops.cu
```diff
@@ -1176,6 +1176,56 @@ RaggedShape ChangeSublistSize(RaggedShape &src, int32_t size_delta) {
...
+  int32_t ans_size = static_cast<int32_t>(sizes.size());
+  std::vector<RaggedShape> ans(ans_size);
+  for (int32_t i = 0; i != ans_size; ++i) {
+    std::vector<RaggedShapeDim> axes_out(axes_in.size());
+    int32_t row_end = sizes[i];
+    K2_CHECK(row_end >= 0 && row_end <= dim0);
+    for (int32_t axis = 0; axis < num_axes - 1; ++axis) {
+      axes_out[axis].row_splits =
+          axes_in[axis].row_splits.Arange(0, row_end + 1);
+      row_end = axes_in_cpu[axis].row_splits[row_end];
```
danpovey, line 1220: What I actually had in mind with this is to avoid any kernels being called in a loop.  Here, `row_splits[row_end]` involves a GPU->CPU data transfer, which is slow.
It would probably involve creating an Array1 containing the sizes, use indexing operations to get Array1s containing the sizes at each level (on device), then transfer those to CPU in one operation.  (If they can be made part of a longer array to have just one GPU->CPU transfer that would be nice).  Also when you construct RaggedShape, if you add `check=false` it will  help the speed a lot (in debug mode)... after it's debugged, of course..

qindazhu: Updated, thanks

danpovey:
Cool!  Merging.

k2/csrc/array.h
```diff
@@ -312,12 +312,24 @@ class Array1 {
    */
   Array1 operator[](const Array1<int32_t> &indexes) const {
     NVTX_RANGE(K2_FUNC);
-    const ContextPtr &c = Context();
-    K2_CHECK(c->IsCompatible(*indexes.Context()));
+    const ContextPtr &c = GetContext(*this, indexes);
```
csukuangfj, line 315: We cannot use a reference here. `GetContext` returns a variable that lives on the stack.

PR #427, opened 2020-11-29
New intersect

danpovey:
Un-pruned intersection that should be much faster for problems that are not too large.
Not finished- putting it up in case anyone wants to make progress before I start work to-morrow.
(The remaining parts should be reasonably obvious if you read the code).. but no hurry, I can do it tomorrow myself.

### DIFF
--- a/k2/csrc/algorithms.h
+++ b/k2/csrc/algorithms.h
@@ -26,6 +26,9 @@ class Renumbering {
   Renumbering(const Renumbering &src) = default;
   // move constructor
   Renumbering(Renumbering &&src) = default;
+  // move assignment
+  Renumbering &operator=(Renumbering &&) = default;
+
   Renumbering(ContextPtr c, int32_t num_old_elems) { Init(c, num_old_elems); }
 
   void Init(ContextPtr c, int32_t num_old_elems) {
--- a/k2/csrc/algorithms.h
+++ b/k2/csrc/algorithms.h
@@ -21,6 +21,10 @@ namespace k2 {
 class Renumbering {
  public:
   Renumbering() = default;
+  // copy constructor
+  Renumbering(const Renumbering &src) = default;
+  // move constructor
+  Renumbering(Renumbering &&src) = default;
   Renumbering(ContextPtr c, int32_t num_old_elems) { Init(c, num_old_elems); }
 
   void Init(ContextPtr c, int32_t num_old_elems) {
@@ -91,6 +95,8 @@ class Renumbering {
                            // old2new_ is created.
   Array1<int32_t> new2old_;
 };
+
+
 
 }  // namespace k2
 
--- a/k2/csrc/array_ops.h
+++ b/k2/csrc/array_ops.h
@@ -558,6 +558,16 @@ Array2<T> IndexRows(const Array2<T> &src, const Array1<int32_t> &indexes,
 template <typename T, typename Compare = LessThan<T>>
 void Sort(Array1<T> *array, Array1<int32_t> *index_map = nullptr);
 
+
+/*
+  Assign elements from `src` to `dest`; they must have the same shape.  For now
+  this only supports cross-device copy if the data is contiguous.
+ */
+template <typename T>
+void Assign(Array2<T> &src, Array2<T> *dest);
+
+
+
 }  // namespace k2
 
 #define IS_IN_K2_CSRC_ARRAY_OPS_H_
--- a/k2/csrc/array_ops.h
+++ b/k2/csrc/array_ops.h
@@ -330,7 +330,7 @@ bool IsMonotonic(const Array1<T> &a);
                       strictly) and whose last element is zero, e.g.
                       [ 5 5 4 2 0 ]
      @return          Returns an array such with ans.Dim() == src[0] + 1,
-                      sich that ans[i] = min(j >= 0 : src[i] <= j).
+                      sich that ans[i] = min(j >= 0 : src[j] <= i).
                       In this case the result would be:
                       [ 4 4 3 3 2 0 ].
 
@@ -338,7 +338,7 @@ bool IsMonotonic(const Array1<T> &a);
                       will always equal x if x satisfies the preconditions.
 
    Implementation notes: allocate ans as zeros; run lambda { if
-   (src[i+1] < src[i]) ans[src[i] - 1] = i + 1 } -> ans = [ 0 4 0 3 2 0 ]
+   (i+1 == dim || src[i+1] < src[i]) ans[src[i] - 1] = i + 1 } -> ans = [ 0 4 0 3 2 0 ]
    in this example; call MonotonicDecreasingUpperBound(ans, &ans).
  */
 Array1<int32_t> InvertMonotonicDecreasing(const Array1<int32_t> &src);
--- a/k2/csrc/array_ops.h
+++ b/k2/csrc/array_ops.h
@@ -403,6 +403,16 @@ bool IsMonotonic(const Array1<T> &a);
  */
 Array1<int32_t> InvertMonotonicDecreasing(const Array1<int32_t> &src);
 
+
+/*
+  Assuming `src` is a permutation of Range(0, src.Dim()), returns the inverse of
+  that permutation, such that ans[src[i]] = i.  It is an error, and may cause a
+  segfault or undefined results, if `src` was not a permutation of Range(0,
+  src.Dim()).
+ */
+Array1<int32_t> InvertPermutation(const Array1<int32_t> &src);
+
+
 /*
    Validate a row_ids vector; this just makes sure its elements are nonnegative
    and non-decreasing.
--- a/k2/csrc/array_ops.h
+++ b/k2/csrc/array_ops.h
@@ -383,8 +383,8 @@ bool IsMonotonic(const Array1<T> &a);
   monotonically decreasing.
 
      @param [in] src   Array which is monotonically decreasing (not necessarily
-                      strictly) and whose last element is zero, e.g.
-                      [ 5 5 4 2 0 ]
+                      strictly) and whose elements are all positive, e.g.
+                      [ 5 5 4 2 1 ]
      @return          Returns an array such with ans.Dim() == src[0] + 1,
                       sich that ans[i] = min(j >= 0 : src[j] <= i).
      @return          Returns an array such with ans.Dim() == src[0],
--- a/k2/csrc/array_ops.h
+++ b/k2/csrc/array_ops.h
@@ -342,6 +342,18 @@ void RowSplitsToRowIds(const Array1<int32_t> &row_splits,
                        Array1<int32_t> *row_ids);
 
 /*
+  Given a vector of row_splits, return a vector of sizes.
+
+     @param [in] row_splits   Row-splits array, should be non-decreasing,
+                            and have Dim() >= 1
+     @return   Returns array of sizes, satisfying `ans.Dim() ==
+              row_splits.Dim() - 1` and
+              `ans[i] = row_splits[i+1] - row_splits[i]`.
+ */
+Array1<int32_t> RowSplitsToSizes(const Array1<int32_t> &row_splits);
+
+
+/*
   This is a convenience wrapper for the function of the same name in utils.h.
    @param [in] row_ids     Input row_ids vector, of dimension num_elems
    @param [out] row_splits  row_splits vector to whose data we will write,
@@ -377,6 +389,13 @@ bool Equal(const Array1<T> &a, const Array1<T> &b);
  */
 template <typename T>
 bool IsMonotonic(const Array1<T> &a);
+/*
+  Return true if array `a` is monotonically decreasing, i.e.
+  a[i+1] >= a[i].
+ */
+template <typename T>
+bool IsMonotonicDecreasing(const Array1<T> &a);
+
 
 /*
   Generalized function inverse for an array viewed as a function which is
@@ -549,11 +568,12 @@ Array2<T> IndexRows(const Array2<T> &src, const Array1<int32_t> &indexes,
 /* Sort an array **in-place**.
 
    @param [inout]   array        The array to be sorted.
-   @param [out]     index_map    If non-null, it maps the index
-                                 of the returned array to the original
-                                 unsorted array. That is,
-                                 out[i] = unsorted[index_map[i]]
-                                 for i in [0, array->Dim()).
+   @param [out]     index_map    If non-null, it will be set to
+                          a new array that maps the index of the returned array
+                          to the original unsorted array. That is, out[i] =
+                          unsorted[index_map[i]] for i in [0, array->Dim()) if
+                          `unsorted` was the value of `array` at input and `out`
+                          is the value after the function call.
  */
 template <typename T, typename Compare = LessThan<T>>
 void Sort(Array1<T> *array, Array1<int32_t> *index_map = nullptr);
--- a/k2/csrc/array_ops.h
+++ b/k2/csrc/array_ops.h
@@ -385,14 +385,12 @@ bool IsMonotonic(const Array1<T> &a);
      @param [in] src   Array which is monotonically decreasing (not necessarily
                       strictly) and whose elements are all positive, e.g.
                       [ 5 5 4 2 1 ]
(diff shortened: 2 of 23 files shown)


k2/csrc/array_ops.h
```diff
@@ -383,8 +383,10 @@ bool IsMonotonic(const Array1<T> &a);
   monotonically decreasing.
 
      @param [in] src   Array which is monotonically decreasing (not necessarily
-                      strictly) and all elements are positive, e.g.
+                      strictly) and whose elements are all positive, e.g.
                       [ 5 5 4 2 1 ]
+     @return          Returns an array such with ans.Dim() == src[0] + 1,
```
qindazhu, line 388: These two lines should be removed, seems issue because of confilict resolving

k2/csrc/ragged_ops.h
```diff
@@ -172,6 +172,34 @@ RaggedShape Stack(int32_t axis, int32_t src_size, RaggedShape **src);
...
+        @param [in] src    Source RaggedShape
+        @param [in] sizes  Lengths of desired prefixes; all elements
+                           will satisfy 0 <= sizes[i] < src.Dim0().
+        @return   Returns vector of prefixes of a RaggedShape;
+                  ans[i] will be equal to Prefix(src, sizes[i]).
+                  We provide this interface because individual
+                  calls to `Prefix` would otherwise require multiple
+                  GPU->CPU memory transfers.
+ */
+std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
```
qindazhu, line 200: I thought you have merged these two functions' declarations before.

k2/csrc/ragged_ops.cu
```diff
@@ -223,6 +223,122 @@ RaggedShape RaggedShape3(Array1<int32_t> *row_splits1,
...
 }
 
+
+RaggedShape RaggedShape4(Array1<int32_t> *row_splits1,
+                         Array1<int32_t> *row_ids1, int32_t cached_tot_size1,
+                         Array1<int32_t> *row_splits2,
+                         Array1<int32_t> *row_ids2, int32_t cached_tot_size2,
+                         Array1<int32_t> *row_splits3,
+                         Array1<int32_t> *row_ids3, int32_t cached_tot_size3) {
+  NVTX_RANGE(__func__);
```
qindazhu, line 233: Just a question: can we implement `RaggedShape4` with `RaggedShape3` and `RaggedShape2` and then `ComposeShape`, it may save a lot of code as `ComposeShape` is very cheap.

k2/csrc/algorithms.h
```diff
@@ -22,6 +22,10 @@ namespace k2 {
 class Renumbering {
  public:
   Renumbering() = default;
+  // copy constructor
+  Renumbering(const Renumbering &src) = default;
+  // move constructor
+  Renumbering(Renumbering &&src) = default;
```
csukuangfj, line 28: Move assignment operator is also needed if move constructor is present.

```cpp
Renumbering &operator=(Renumbering &&) = default;

File: k2/csrc/array_ops.h [hunk 5]
```diff
@@ -383,8 +383,10 @@ bool IsMonotonic(const Array1<T> &a);
...
+                      sich that ans[i] = min(j >= 0 : src[j] <= i).
```
csukuangfj, line 389: typo: such that

k2/csrc/intersect.cu
```diff
@@ -90,950 +69,777 @@ using namespace intersect_internal;  // NOLINT
...
+
+    { // check that b_fsas are in order of decreasing length.
+      Array1<int32_t> r = b_fsas.shape.RowSplits(1).To(GetCpuContext());
+      int32_t *r_data = r.Data();
+      int32_t prev_t = r_data[1] - r_data[0];
+      for (int32_t i = 1; i + 1 < r.Dim(); i++) {
+        int32_t this_t = r_data[i+1] - r_data[i];
+        if (this_t < prev_t)
+          K2_LOG(FATAL) << "Sequences (DenseFsaVec) must be in sorted from greatest to least.";
+        prev_t = this_t;
```
qindazhu, line 151: wondering if this is a check for only debug mode? why we do it on CPU?

k2/csrc/intersect.cu
```diff
@@ -90,950 +69,777 @@ using namespace intersect_internal;  // NOLINT
...
+    int32_t ans_tot_num_states = new2old.Dim();
+
+    // t_per_fsa will be set below to the number of time-steps that each FSA has
+    // states active on; if each FSA i has scores for 0 <= t < T_i, then
+    // t_per_fsa[i] will be T_i + 1, because there is also a copy of the state
+    // at time T_i.
+    Array1<int32_t> t_per_fsa(c_, num_fsas_ + 1);
+    int32_t *t_per_fsa_data = t_per_fsa.Data();
+    auto lambda_set_t_per_fsa_etc = [=] __host__ __device__ (int32_t i) -> void {
+      t_per_fsa_data[i] = fsa_info_data[i].T + 1;
```
qindazhu, line 484: This is a very short lambda, maybe we should change it to `if CPU else lambda` approach.

k2/csrc/array_ops.h
qindazhu: I believe these two lines should be deleted

k2/csrc/ragged_ops.cu
```diff
@@ -223,6 +223,122 @@ RaggedShape RaggedShape3(Array1<int32_t> *row_splits1,
...
+  K2_CHECK(row_splits1 != nullptr || row_ids1 != nullptr)
+      << "At least one of row_splits1 and row_ids1 must be defined";
+  K2_CHECK(row_splits2 != nullptr || row_ids2 != nullptr)
+      << "At least one of row_splits2 and row_ids2 must be defined";
+  K2_CHECK(row_splits3 != nullptr || row_ids3 != nullptr)
+      << "At least one of row_splits3 and row_ids3 must be defined";
+
+
+  ContextPtr ctx = ::GetContext(row_splits1, row_ids1);
```
csukuangfj, line 242: `GetContext` is a variadic function.

```cpp
ContextPtr ctx = ::GetContext(row_splits1, row_ids1, row_splits2, row_ids2, row_splits3, row_ids3);
```
is simpler.

qindazhu: mm, `::GetContext` is an internal function https://github.com/k2-fsa/k2/blob/ddd30a16d0c0fe28d82773737a939afca0b4e285/k2/csrc/ragged_ops.cu#L34-L35 which will check if the input is null or not, this is different from `k2::GetContext`

So I think there's no need to change.

csukuangfj: ok

k2/csrc/ragged_ops.h
danpovey: github is getting this wrong, this is not the diff.  there are some small changes that I'm reverting though.

k2/csrc/ragged_ops.cu
danpovey: MM yeah, let me change it in this PR, I'll change RaggedShape3() as well.

k2/csrc/algorithms.h
danpovey: ok..

k2/csrc/intersect.cu
danpovey: I figured that for typical number of FSAs, the overhead of transfer will probably exceed the time from doing it on CPU..

k2/csrc/intersect.cu
danpovey: So we're still getting that problem with short lambdas?  We didn't somehow resolve it another way?

qindazhu: I supposed we just try our best to avoid short lambdas nowdays.

k2/csrc/intersect.cu
```diff
@@ -90,950 +69,840 @@ using namespace intersect_internal;  // NOLINT
...
   DenseFsaVec &b_fsas_;
-  float search_beam_;
+
+  // num_fsas_ equals b_fsas_.shape.Dim0() == a_fsas_.Dim0().
+  int32_t num_fsas_;
+
+  // This is just a copy of a_fsas_.arcs, with a couple extra pieces of information.
+  struct CompressedArc {
+    // src_state of Arc, as uint16 (an idx1)
+    uint16_t src_state;
```
csukuangfj, line 725: Will `uint16_t` limit the size of the input FsaVec？
It cannot contain more than (1<<16) == 64 k states.

k2/csrc/intersect.cu
```diff
@@ -90,950 +69,840 @@ using namespace intersect_internal;  // NOLINT
...
   /**
-     Pruned intersection (a.k.a. composition) that corresponds to decoding for
+     Intersection (a.k.a. composition) that corresponds to decoding for
      speech recognition-type tasks
 
        @param [in] a_fsas  The decoding graphs, one per sequence.  E.g. might
                            just be a linear sequence of phones, or might be
                            something more complicated.  Must have either the
-                           same Dim0() as b_fsas, or Size0()==1 in which
+                           same Dim0() as b_fsas, or Dim0() == 1 in which
```
csukuangfj, line 93: `Dim0() == 1` can be removed. The implementation assumes that a_fsas.Dim0() == b_fsas_.shape.Dim0().

k2/csrc/intersect.cu
```diff
@@ -90,950 +69,840 @@ using namespace intersect_internal;  // NOLINT
...
-        dynamic_beams_(a_fsas.Context(), b_fsas.shape.Dim0(), search_beam) {
-    NVTX_RANGE(K2_FUNC);
+        output_beam_(output_beam) {
+    NVTX_RANGE(__func__);
     c_ = GetContext(a_fsas.shape, b_fsas.shape);
+
+    K2_CHECK(a_fsas_.Dim0() == b_fsas_.shape.Dim0());
+    num_fsas_ = a_fsas_.Dim0();
+    K2_CHECK_GT(num_fsas_, 0);
     K2_CHECK(b_fsas.scores.IsContiguous());
```
csukuangfj, line 116: I think the constraint `IsContiguous()` is not necessary any more.

k2/csrc/intersect.cu
```diff
@@ -90,950 +69,840 @@ using namespace intersect_internal;  // NOLINT
...
+  RaggedShape backward_then_forward_shape_;
+
+  // Temporary array used in combined forward+backward computation, of dimension
+  // a_fsas_.TotSize(2) * 2.  Used indirectly through fb_arc_scores_, which
+  // share this data.
+  Array1<float> arc_scores_;
+
+  struct FsaInfo {
+    // T is the number of frames in b_fsas_.scores that we have for this FSA, i.e.
+    // `b_fsas_.scores.RowSplits(1)[i+1] -  b_fsas_.scores.RowSplits(1)[i].`
```
csukuangfj, line 769: typo: `b_fsas.scores_.RowSplits` --> `b_fsas_.shape.RowSplits`

k2/csrc/intersect.cu
```diff
@@ -90,950 +69,840 @@ using namespace intersect_internal;  // NOLINT
...
-      // idx01; this relies on the fact that sinfo.abs_state_id == arc.src_state
-      // + a_fsas_fsa_idx0x.
-      ai.u.dest_a_fsas_state_idx01 =
-          sinfo.a_fsas_state_idx01 + arc.dest_state - arc.src_state;
-      ai_data[ai_arc_idx012] = ai;
+  void InitSteps() {
+    NVTX_RANGE(__func__);
+    // This vector, of length num_fsas_, tells us how many copies of (the states
+    // of the i'th decoding graph) we have.  It equals (the length of the sequence
+    // of log-likes in n_fsas_) + 1.  It is monotonically decreasing (thanks to
```
csukuangfj, line 499: typo: `n_fsas_` --> `b_fsas_`

k2/csrc/intersect.cu
```diff
@@ -90,950 +69,840 @@ using namespace intersect_internal;  // NOLINT
...
-                           be active on any given frame for any given
-                           intersection/composition task. This is advisory,
-                           in that it will try not to have fewer than this
-                           number active.
-       @param [in] max_active  Maximum number of FSA states that are allowed to
-                           be active on any given frame for any given
-                           intersection/composition task. This is advisory,
-                           in that it will try not to exceed that but may not
-                           always succeed.
+                           (in general) different length.  MUST BE SORTED BY
```
csukuangfj, line 97: > MUST BE SORTED BY DECREASING LENGTH, or it is an error.

We need to change the Python code of constructing DenseFsaVec.
The current implementation does not satisfy this constraint.

danpovey: You're right, needs change.  Should be easy though.

k2/csrc/intersect.cu
danpovey: It does limit it.  
This version of the code is not suitable for intersection with large FSAs though, because there is no pruning.
So in practice it's not much loss.

k2/csrc/intersect.cu
csukuangfj: Should the sort be performed by the user or k2?
If it is done by the user, do we need to check that the input supervision segments are sorted?

danpovey:
By the user is the easiest way.   Sort the segments by length when we
create them, that's easy.

danpovey:
OK I merged this- please resolve any merge problems!
Also the Python binding needs to be done for this new intersection code, then we can test it in snowfall.

k2/csrc/intersect.cu
danpovey: Mm, you are right I think... should remove any mention of that from the documentation too.

danpovey: sorry I missed some of these comments from long ago (as you pointed out in another thread).. since I am away for a few days, is there any chance you could clean up these issues?
BTW, for `intersect`, it's definitely possible to improve its behavior memory-wise but this would be at the expense of time (part of what we do for speed is to make the forward and backward passes simultaneous, which is a factor of 2 in memory since otherwise we could share memory for the alphas and betas).   Let's wait until it becomes an issue and then optimize, maybe?  If it's an issue already, LMK.

One possibility is to store just every *n* frames permanently, in the forward and backward (still done simultaneously), and use temporary memory for the in-between ones; the missing ones would then be computed as needed in a loop over n.  [n might be 2 or 4 or so.]   Not a *completely* trivial change, of course.

csukuangfj: >  is there any chance you could clean up these issues?

working on it.

k2/csrc/intersect_pruned.cu
```diff
@@ -0,0 +1,1044 @@
...
+     FloatToOrderedInt/OrderedIntToFloat so we can use atomic max.  It
+     represents a Viterbi-style 'forward probability'.  (Viterbi, meaning: we
+     use max not log-sum).  You can take the pruned lattice and rescore it if
+     you want log-sum.  */
+  int32_t forward_loglike;
+
+  /* Note: this `backward_loglike` is the best score of any path from here to
+     the end, minus the best path in the overall FSA, i.e. it's the backward
+     score you get if, at the final-state, you set backward_loglike ==
+     forward_loglike. So backward_loglike + OrderedIntToFloat(forward_loglike)
```
csukuangfj, line 41: Is it assigned to **negative** `forward_loglike`?

danpovey: yes, will fix.

k2/csrc/intersect_pruned.cu
```diff
@@ -0,0 +1,1044 @@
...
+      : a_fsas_(a_fsas),
+        b_fsas_(b_fsas),
+        search_beam_(search_beam),
+        output_beam_(output_beam),
+        min_active_(min_active),
+        max_active_(max_active),
+        dynamic_beams_(a_fsas.Context(), b_fsas.shape.Dim0(), search_beam) {
+    NVTX_RANGE(__func__);
+    c_ = GetContext(a_fsas.shape, b_fsas.shape);
+    K2_CHECK(b_fsas.scores.IsContiguous());
```
csukuangfj, line 138: Does this constraint still hold? I find that the current code
is using `Accessor` to access its element.

PR #436, opened 2020-12-02
create Fsa from vector of arcs (maybe cyclic)

qindazhu:
The original `CreateFsa` (host version) requires the input arcs are acyclic, I add a version by removing this requirement
 I need this in un-pruned version of `determinize` as usually the output is cyclic, e.g. for L*G. The original version is renamed to `CreateTopSortedFsa`


csukuangfj:
+1

danpovey:
cool...

PR #442, opened 2020-12-02
avoid repeated check for IsArcSorted in intersect()

### LINKED ISSUE #441: Intersect() checks
Guys, it turns out the reason why CPU intersection is slow is that it's repeatedly checking IsArcSorted() each time we intersect (and for each individual FSA).
Remember I previously proposed that we pass in the FSA properties to these kinds of calls.
I suggest to do the following:

 For the Intersect() function in host/, add a 'bool check_properties' argument, after treat_epsilons_specially.
(We may later want to introduce a config class for these kinds of things, to avoid args getting out of control).
If false, it will just assume the properties are true and not do the checks.

To Intersect() in csrc/, change
  `FsaOrVec &a_fsas, FsaOrVec &b_fsas,`
to:
`  FsaOrVec &a_fsas, int32_t properties_a, FsaOrVec &b_fsas, int32_t properties_b`
where the user can pass -1 if they haven't computed the properties, and it will compute them 
internally.  We can pass in the properties from the Python code (the Fsa object caches them).
Any checks that are normally done in host/ Intersect(), we can put instead at the outer level, in the
wrapping code in csrc/, and use the properties for the check.

freewym:
resolves #441 

@danpovey I just had a try. Let me know how it looks.

### DIFF
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -108,8 +108,8 @@ bool HostTopSort(Fsa &src, Fsa *dest, Array1<int32_t> *arc_map /*=nullptr*/) {
   return ans;
 }
 
-bool Intersect(FsaOrVec &a_fsas, FsaOrVec &b_fsas,
-               bool treat_epsilons_specially, FsaVec *out,
+bool Intersect(FsaOrVec &a_fsas, int32_t properties_a, FsaOrVec &b_fsas,
+               int32_t properties_b, bool treat_epsilons_specially, FsaVec *out,
                Array1<int32_t> *arc_map_a, Array1<int32_t> *arc_map_b) {
   NVTX_RANGE(K2_FUNC);
   K2_CHECK(a_fsas.NumAxes() >= 2 && a_fsas.NumAxes() <= 3);
@@ -118,13 +118,13 @@ bool Intersect(FsaOrVec &a_fsas, FsaOrVec &b_fsas,
   K2_CHECK_EQ(c->GetDeviceType(), kCpu);
   if (a_fsas.NumAxes() == 2) {
     FsaVec a_fsas_vec = FsaToFsaVec(a_fsas);
-    return Intersect(a_fsas_vec, b_fsas, treat_epsilons_specially, out,
-                     arc_map_a, arc_map_b);
+    return Intersect(a_fsas_vec, properties_a, b_fsas, properties_b,
+                     treat_epsilons_specially, out, arc_map_a, arc_map_b);
   }
   if (b_fsas.NumAxes() == 2) {
     FsaVec b_fsas_vec = FsaToFsaVec(b_fsas);
-    return Intersect(a_fsas, b_fsas_vec, treat_epsilons_specially, out,
-                     arc_map_a, arc_map_b);
+    return Intersect(a_fsas, properties_a, b_fsas_vec, properties_b,
+                     treat_epsilons_specially, out, arc_map_a, arc_map_b);
   }
 
   int32_t num_fsas_a = a_fsas.Dim0(), num_fsas_b = b_fsas.Dim0();
@@ -141,6 +141,16 @@ bool Intersect(FsaOrVec &a_fsas, FsaOrVec &b_fsas,
     }
     // the check on the previous line will fail.
   }
+  if (properties_a < 0) {
+    properties_a = GetFsaBasicProperties(a_fsas);
+  }
+  if (properties_b < 0) {
+    properties_b = GetFsaBasicProperties(b_fsas);
+  }
+  bool check_properties = true;
+  if ((properties_a & kFsaPropertiesArcSorted) && (properties_b & kFsaPropertiesArcSorted)) {
+    check_properties = false;
+  }
   int32_t num_fsas = std::max(num_fsas_a, num_fsas_b);
 
   std::vector<std::unique_ptr<k2host::Intersection>> intersections(num_fsas);
@@ -149,7 +159,7 @@ bool Intersect(FsaOrVec &a_fsas, FsaOrVec &b_fsas,
     k2host::Fsa host_fsa_a = FsaVecToHostFsa(a_fsas, i * stride_a),
                 host_fsa_b = FsaVecToHostFsa(b_fsas, i * stride_b);
     intersections[i] = std::make_unique<k2host::Intersection>(
-        host_fsa_a, host_fsa_b, treat_epsilons_specially);
+        host_fsa_a, host_fsa_b, treat_epsilons_specially, check_properties);
     intersections[i]->GetSizes(&(sizes[i]));
   }
   FsaVecCreator creator(sizes);
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -148,7 +148,8 @@ bool Intersect(FsaOrVec &a_fsas, int32_t properties_a, FsaOrVec &b_fsas,
     properties_b = GetFsaBasicProperties(b_fsas);
   }
   bool check_properties = true;
-  if ((properties_a & kFsaPropertiesArcSorted) && (properties_b & kFsaPropertiesArcSorted)) {
+  if ((properties_a & kFsaPropertiesArcSorted) &&
+      (properties_b & kFsaPropertiesArcSorted)) {
     check_properties = false;
   }
   int32_t num_fsas = std::max(num_fsas_a, num_fsas_b);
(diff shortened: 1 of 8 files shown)


k2/csrc/fsa_algo.h
```diff
@@ -225,8 +225,8 @@ void IntersectDense(FsaVec &a_fsas, DenseFsaVec &b_fsas,
                   (requires input FSAs to be arc-sorted and at least one of
                   them to be epsilon free).
  */
-bool Intersect(FsaOrVec &a_fsas, FsaOrVec &b_fsas,
-               bool treat_epsilons_specially, FsaVec *out,
+bool Intersect(FsaOrVec &a_fsas, int32_t properties_a, FsaOrVec &b_fsas,
```
csukuangfj, line 228: Would you mind also updating the document about `properties_a` and `properties_b`?

k2/python/csrc/torch/fsa_algo.cu
```diff
@@ -99,15 +99,17 @@ static void PybindLinearFsa(py::module &m) {
 static void PybindIntersect(py::module &m) {
   m.def(
       "intersect",  // works only on CPU
-      [](FsaOrVec &a_fsas, FsaOrVec &b_fsas,
+      [](FsaOrVec &a_fsas, int32_t properties_a,
```
csukuangfj, line 102: Please also update line 128 in this file to add arguments for `properties_a` and `properties_b`. I guess the current code won't compile.

freewym: sorry I forgot to push my local changes to that

k2/csrc/host/intersect.cc
```diff
@@ -52,7 +52,7 @@ void Intersection::GetSizes(Array2Size<int32_t> *fsa_size) {
   arc_map_a_.clear();
   arc_map_b_.clear();
 
-  if (IsEmpty(a_) || IsEmpty(b_)) return;
+  if (IsEmpty(a_) || IsEmpty(b_) || !check_properties_) return;
   if (!IsArcSorted(a_) || !IsArcSorted(b_)) {
```
csukuangfj, line 56: I would suggest

```cpp
if (check_properties_ && (!IsArcSorted(a_) || !IsArcSorted(b_)) {
  ...
}
```
The main computation of intersect is done in this function (`GetSizes`). If `check_properties_` is false,
we just skip `IsArcSorted` and go ahead.

k2/csrc/fsa_algo.cu
```diff
@@ -141,6 +141,16 @@ bool Intersect(FsaOrVec &a_fsas, FsaOrVec &b_fsas,
...
+  if (properties_a < 0) {
+    properties_a = GetFsaBasicProperties(a_fsas);
+  }
+  if (properties_b < 0) {
+    properties_b = GetFsaBasicProperties(b_fsas);
+  }
+  bool check_properties = true;
+  if ((properties_a & kFsaPropertiesArcSorted) && (properties_b & kFsaPropertiesArcSorted)) {
+    check_properties = false;
+  }
```
csukuangfj, line 153: @danpovey 
should we return early here if one of the input FSAs is not arc sorted?
If an FSA is not arc sorted, there is no need to call k2host::Intersect
as we know it will return an empty FSA.

k2/csrc/host/intersect.cc
```diff
@@ -53,7 +53,7 @@ void Intersection::GetSizes(Array2Size<int32_t> *fsa_size) {
   arc_map_b_.clear();
 
   if (IsEmpty(a_) || IsEmpty(b_)) return;
-  if (!IsArcSorted(a_) || !IsArcSorted(b_)) {
+  if (!check_properties_ || !IsArcSorted(a_) || !IsArcSorted(b_)) {
```
csukuangfj, line 56: Not a request for changing code.

I feel negative conditions are hard to read.

k2/csrc/host/intersect.h
```diff
@@ -36,9 +36,11 @@ class Intersection {
                       symbol.
   */
   Intersection(const Fsa &a, const Fsa &b,
-               bool treat_epsilons_specially = true) :
+               bool treat_epsilons_specially = true,
+               bool check_properties = true) :
```
csukuangfj, line 40: And the documentation for `check_properties`, please.

csukuangfj:
+1

k2/csrc/host/intersect.cc
freewym: I think it should be 

`if (check_properties_ && (!IsArcSorted(a_) || !IsArcSorted(b_)))`

k2/csrc/host/intersect.cc
```diff
@@ -53,7 +53,7 @@ void Intersection::GetSizes(Array2Size<int32_t> *fsa_size) {
   arc_map_b_.clear();
 
   if (IsEmpty(a_) || IsEmpty(b_)) return;
-  if (!IsArcSorted(a_) || !IsArcSorted(b_)) {
+  if (check_properties_ && (!IsArcSorted(a_) || !IsArcSorted(b_))) {
     K2_LOG(WARNING) << "One of the inputs is not arc-sorted";
```
csukuangfj, line 57: This is for the original code. I think we can do better here to prevent unnecessary invoking of `IsArcSorted`.

```cpp
if (check_properties_) {
  if (!IsArcSorted(a_) {
    K2_LOG(INFO) << "Not arc-sorted: " << a_;
    status_ = false;
    return;
  }

  if (!IsArcSorted(b_) {
    K2_LOG(INFO) << "Not arc-sorted: " << b_;
    status_ = false;
    return;
  }
}
```

csukuangfj: And it is less error prone.

k2/csrc/fsa_algo.cu
danpovey: I was thinking we would always call it with check_properties = false (from here).
Since we have called GetFsaBasicProperties() we know the properties.
It will be an error if they are not arc-sorted, i.e. we will die with error message.

k2/python/k2/fsa_algo.py
```diff
@@ -87,9 +87,12 @@ def intersect(a_fsa: Fsa, b_fsa: Fsa) -> Fsa:
       if and only if the two input FSAs are single FSAs;
       otherwise, len(out_fsa.shape) is 3.
     '''
+    properties_a = a_fsa.properties if a_fsa.properties is not None else -1
```
danpovey, line 90: Actually I think a_fsa.properties can never be None.  There is a getter that will compute it.

danpovey:
Perhaps someone in our time zone can address those small comments?

k2/csrc/fsa_algo.cu
freewym: you mean, remove this if block and always pass check_properties = false?

k2/python/k2/fsa_algo.py
freewym: ok

csukuangfj:
> Perhaps someone in our time zone can address those small comments?


Let's merge this and I am going to resolve the comments.

csukuangfj:
@freewym 
you can resolve the comments if you have time.

k2/csrc/fsa_algo.cu
freewym: or use ASSERT_EQ for the properties?

csukuangfj: I guess this is what Dan meant:
```cpp
K2_CHECK_EQ(check_properties, false) << "some log message";
```

freewym: OK, done

csukuangfj:
+2

k2/csrc/fsa_algo.cu
freewym: @csukuangfj this check failed in two tests. Not familiar with the error message. Not sure if it is because of some un-arc-sorted input fsa...

k2/csrc/fsa_algo.cu
```diff
@@ -141,6 +141,19 @@ bool Intersect(FsaOrVec &a_fsas, FsaOrVec &b_fsas,
     }
     // the check on the previous line will fail.
   }
+  if (properties_a < 0) {
+    properties_a = GetFsaBasicProperties(a_fsas);
+  }
+  if (properties_b < 0) {
+    properties_b = GetFsaBasicProperties(b_fsas);
```
qindazhu, line 148: Noted we'll get mutiple testing failure here as we are calling `GetFsaBasicProperties` for a FsaVec, it always return 0. I have fixed it locally, will push together with my other changes later

qindazhu: I mean, we should call `GetFsaVecBasicProperties` for FsaVec.

qindazhu: Fixed in https://github.com/k2-fsa/k2/pull/446.

PR #444, opened 2020-12-03
Add support for 2-d tensors to index_select and index_add.

### LINKED ISSUE #307: k2.index

csukuangfj:
TODOs
- [x] Handle non-contiguous cases
- [x] add 2-d support to index_add

Closes #307

### DIFF
--- a/k2/python/csrc/torch/index_add.cu
+++ b/k2/python/csrc/torch/index_add.cu
@@ -14,6 +14,8 @@
  */
 
 #include "k2/csrc/context.h"
+#include "k2/csrc/macros.h"
+#include "k2/csrc/nvtx.h"
 #include "k2/python/csrc/torch/index_add.h"
 #include "k2/python/csrc/torch/torch_util.h"
 #include "torch/extension.h"
@@ -22,12 +24,15 @@ namespace k2 {
 
 // It implements  (*in_out)[index[i]] += value[i]
 // for i in range(index.numel())
-void IndexAdd(torch::Tensor index, torch::Tensor value, torch::Tensor *in_out) {
-  // We support only 1-D tensors at present
+static void IndexAdd1D(torch::Tensor index, torch::Tensor value,
+                       torch::Tensor *in_out) {
+  NVTX_RANGE(K2_FUNC);
   K2_CHECK_EQ(index.scalar_type(), ToScalarType<int32_t>::value);
   K2_CHECK_EQ(index.dim(), 1);
 
-  K2_CHECK_EQ(value.scalar_type(), ToScalarType<float>::value);
+  K2_CHECK_EQ(value.scalar_type(), ToScalarType<float>::value)
+      << "index_add is supposed to be used in back propagation, "
+         "which means the value type should be of float";
   K2_CHECK_EQ(value.dim(), 1);
 
   K2_CHECK_EQ(index.numel(), value.numel());
@@ -52,30 +57,21 @@ void IndexAdd(torch::Tensor index, torch::Tensor value, torch::Tensor *in_out) {
 
   if (index.is_contiguous() && value.is_contiguous() &&
       in_out->is_contiguous()) {
-    if (context->GetDeviceType() == kCpu) {
-      for (int32_t i = 0; i != index_numel; ++i) {
-        int32_t in_out_i = index_data[i];
-        if (in_out_i == -1) continue;
+    K2_EVAL(
+        context, index_numel, lambda_contiguous, (int32_t i)->void {
+          int32_t in_out_i = index_data[i];
+          if (in_out_i == -1) return;
 
-        K2_DCHECK_GE(in_out_i, 0);
-        K2_DCHECK_LT(in_out_i, in_out_numel);
+          K2_DCHECK_GE(in_out_i, 0);
+          K2_DCHECK_LT(in_out_i, in_out_numel);
 
-        // for host thread, we assume single thread at present
-        in_out_data[in_out_i] += value_data[i];
-      }
-    } else {
-      // for cuda
-      auto lambda_contiguous = [=] __device__(int32_t i) -> void {
-        int32_t in_out_i = index_data[i];
-        if (in_out_i == -1) return;
-
-        K2_DCHECK_GE(in_out_i, 0);
-        K2_DCHECK_LT(in_out_i, in_out_numel);
-
-        atomicAdd(in_out_data + in_out_i, value_data[i]);
-      };
-      EvalDevice(context, index_numel, lambda_contiguous);
-    }
+#ifdef __CUDA_ARCH__
+          atomicAdd(in_out_data + in_out_i, value_data[i]);
+#else
+          // for host thread, we assume single thread at present
+          in_out_data[in_out_i] += value_data[i];
+#endif
+        });
     return;
   }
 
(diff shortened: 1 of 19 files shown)


danpovey:
OK with me. 
Note: there is a function IndexRows() in array_ops.h that indexes Array2 with Array1, but I'm not saying you should necessarily use it here.

k2/python/k2/autograd.py
```diff
@@ -322,11 +322,14 @@ def forward(ctx, src: torch.Tensor, index: torch.Tensor) -> torch.Tensor:
...
         src, index = ctx.saved_tensors
 
-        ans = torch.zeros(src.size(0),
+        ans = torch.zeros(src.shape,
                           dtype=torch.float32,
                           device=src.device,
                           requires_grad=False)
-        _k2.index_add(index, out_grad, ans)
+        if out_grad.stride()[0] == 0:
+            # TODO(fangjun): why is the stride of out_grad 0 ??
```
csukuangfj, line 330: I tried to figure out why `out_grad.stride() == (0, 0)` but failed.
@danpovey @qindazhu could you give some comments about it?

k2/python/csrc/torch/index_add.cu
```diff
@@ -52,68 +57,149 @@ void IndexAdd(torch::Tensor index, torch::Tensor value, torch::Tensor *in_out) {
...
+    K2_EVAL(
+        context, index_numel, lambda_contiguous, (int32_t i)->void {
+          int32_t in_out_i = index_data[i];
+          if (in_out_i == -1) return;
+
+          K2_DCHECK_GE(in_out_i, 0);
+          K2_DCHECK_LT(in_out_i, in_out_num_rows);
+          float *cur_in_out_data = in_out_data + in_out_i * in_out_num_cols;
+          const float *cur_value_data = value_data + i * in_out_num_cols;
+          for (int32_t j = 0; j != in_out_num_cols; ++j) {
```
qindazhu, line 143: seems there's a loop in kernel, can't we add `j` in the parameter of lambda and use `Eval2`? BTW, I think we need also add a `TODO` here as we definitely need to find a way to avoid `atomicAdd` in future.

k2/python/csrc/torch/index_select.cu
```diff
@@ -45,70 +50,116 @@ static torch::Tensor IndexSelect1D(torch::Tensor src, torch::Tensor index) {
...
+  if (src.is_contiguous() && index.is_contiguous()) {
+    K2_EVAL(
+        context, index_numel, lambda, (int32_t i)->void {
+          int32_t src_i = index_data[i];
+          const T *src_cur_row = src_data + src_i * src_num_cols;
+          T *ans_cur_row = ans_data + i * src_num_cols;
+          if (src_i != -1) {
+            K2_DCHECK_GE(src_i, 0);
+            K2_DCHECK_LT(src_i, src_num_rows);
+            for (int32_t j = 0; j != src_num_cols; ++j)
```
qindazhu, line 125: add also move `j` here to the parameter of lambda and use Eval2

k2/python/csrc/torch/index_add.cu
csukuangfj: The number of loops is determined by the number of entries in the arc map, which is a 1-D tensor, so `Eval` is used here. 

I didn't find a way to avoid the loop here.

qindazhu: we can just do like below
```
 K2_EVAL2(
        context, index_numel, *in_out_num_cols*,  lambda_contiguous, (int32_t i, *int32_t j*)->void {
          int32_t in_out_i = index_data[i];
          if (in_out_i == -1) return;

          K2_DCHECK_GE(in_out_i, 0);
          K2_DCHECK_LT(in_out_i, in_out_num_rows);
          float *cur_in_out_data = in_out_data + in_out_i * in_out_num_cols;
          const float *cur_value_data = value_data + i * in_out_num_cols;
#ifdef __CUDA_ARCH__
            atomicAdd(cur_in_out_data + j, cur_value_data[j]);
#else
            // for host thread, we assume single thread at present
            cur_in_out_data[j] += cur_value_data[j];
#endif
```

k2/python/k2/autograd.py
qindazhu: have you ever printed some logs at C++ level of `index_select`?

csukuangfj: Yes, I have a check inside c++ to ensure that the stride for the column is 1. But the actual value is 0. That is
how I discovered that the `out_grad.stride() == (0, 0)`.

`out_grad` is generated by autograd from Python.

danpovey: What is the dimension of out_grad?

k2/python/csrc/torch/index_add.cu
danpovey: We have a __host__ version of atomicAdd so you don't need the ifdef.

danpovey:
Guys, can you get out of the habit of doing force-push?  It makes it hard to view commits.
Better to merge and not rebase while you are working on a branch; I can do 'squash and merge' when merging from github.

k2/python/csrc/torch/index_add.cu
```diff
@@ -22,12 +24,15 @@ namespace k2 {
...
-  // We support only 1-D tensors at present
+static void IndexAdd1D(torch::Tensor index, torch::Tensor value,
+                       torch::Tensor *in_out) {
+  NVTX_RANGE(K2_FUNC);
   K2_CHECK_EQ(index.scalar_type(), ToScalarType<int32_t>::value);
   K2_CHECK_EQ(index.dim(), 1);
 
-  K2_CHECK_EQ(value.scalar_type(), ToScalarType<float>::value);
+  K2_CHECK_EQ(value.scalar_type(), ToScalarType<float>::value)
+      << "index_add is supposed to be used in back propagation, "
```
danpovey, line 34: I'd rather just say something to the effect that only float is supported right now.  It's a general purpose function.

k2/python/csrc/torch/index_add.cu
```diff
@@ -52,68 +57,149 @@ void IndexAdd(torch::Tensor index, torch::Tensor value, torch::Tensor *in_out) {
...
+        atomicAdd(in_out_data + in_out_i * in_out_stride,
+                  value_data[i * value_stride]);
+#else
+      // for host thread, we assume single thread at present
+      in_out_data[in_out_i * in_out_stride] += value_data[i * value_stride];
+#endif
+      });
+}
+
+static void IndexAdd2D(torch::Tensor index, torch::Tensor value,
```
danpovey, line 101: There is no documentation here!!
You should be writing the documentation first!

danpovey: I can't review this without documentation- how can I tell if it is doing what is intended when that is not stated?

danpovey:
The structure of this is problematic.  It's in torch/.  Remember, the code in torch/ is intended to be a fairly thin wrapper for our library.  The actual functionality should probably be implemented in tensor_ops.cu, and declared in tensor_ops.h.

k2/python/k2/autograd.py
qindazhu: > I have a check inside c++ to ensure that the stride for the column is 1. But the actual value is 0

I suppose you add a check locally(sorry, I didn't find it with current code), so you mean even in C++, the stride is 0? when we create tensor with something like `src.options()`? what if we add `stride` manually?

danpovey:
If the dimension is 1 it might be OK that the stride is 0.  In that case we
could treat the stride as being whatever value we like.

danpovey:
... in fact, I have already declared functions IndexSelect() and IndexAdd() in tensor.h, just not implemented them yet.
The idea is that *where possible* these functions will call corresponding functions declared for Array1 and Array2.
There may be cases where, due to strides etc., this is not possible.

Note the functgions Index() and IndexRows() in array_ops.h.

We don't want the same code implemented in different parts of the software-- better to have one place where each operation is implemented, for easier optimization!

csukuangfj:
> If the dimension is 1 it might be OK that the stride is 0.  In that case we
could treat the stride as being whatever value we like.

The problem is that `out_grad` is a 2-d tensor and `out_grad.stride() == (0, 0)`, 
whereas `out_grad.clone().stride() == (some_nonzero_value, 1)`. I cannot figure out the reason.

----

> Guys, can you get out of the habit of doing force-push?

Sorry, no force-push from now on.

---

> The structure of this is problematic.

Will refactor it as suggested.

danpovey:
Thanks!!

danpovey:
So was the shape of `out_grad` in this case (1, 1)?  In that case we can ignore the stride, or treat it as whatever we want.

danpovey:
So was the shape of `out_grad` in this case (1, 1)?  In that case we can ignore the stride, or treat it as whatever we want.

csukuangfj:
> So was the shape of out_grad in this case (1, 1)

The `out_grad.shape` contains two very large positive numbers that are generated by `torch.randint`.
It's strange that its stride is `(0, 0)`. I have never encountered this problem before.

danpovey:
I wonder whether torch could be doing something where it lazily allocates
things and doesn't decide the stride unless you attempt to access the data?
When I was designing a tensor toolkit previously I was planning to do
something like that
(although in that case the strides would have been defined).

k2/python/k2/autograd.py
csukuangfj: > I suppose you add a check locally(sorry, I didn't find it with current code),

Here is the check in the code:

https://github.com/k2-fsa/k2/blob/da3c30e0d277a69013610bf0838592f25e003861/k2/python/csrc/torch/index_add.cu#L167

where `value` is actually `out_grad` from Python:
https://github.com/k2-fsa/k2/blob/da3c30e0d277a69013610bf0838592f25e003861/k2/python/k2/autograd.py#L329-L332

danpovey:
I see now why the stride is zero.
If Torch was backpropping the result of you summing all the elements of a tensor, then the gradient would be the same for all elements, so the easiest way for Torch to do the backprop for that would be take a scalar Tensor containing 1 and just view it as having shape (x,y), with zero stride.
It's actually quite nice that it does that.

k2/python/k2/autograd.py
csukuangfj: fixed.

k2/python/csrc/torch/index_add.cu
csukuangfj: Fixed.

There was no `__host__ `version of atomicAdd.  A host version is added in this pullrequest.

k2/python/csrc/torch/index_select.cu
csukuangfj: fixed.

k2/python/csrc/torch/index_add.cu
csukuangfj: Fixed by removal.

k2/python/csrc/torch/index_add.cu
csukuangfj: Removed.

csukuangfj:
Ready for review.

danpovey:
OK, I have one point of discussion...

I notice you are implementing the operations directly in class Tensor.
It would be possible, if the strides are suitable, to use the existing implementations involving Array1 and Array2,
e.g. Index(), IndexRows().  [We can consider changing these names for consistency.]

The advantage of using the existing implementations is (1) slight kernel simplification due to
knowing some of the strides are 1, and (2) Having a smaller number of locations to optimize.

I may be overthinking this though.  We could perhaps merge as-is but just make a note for the future to address this later.

csukuangfj:
> It would be possible, if the strides are suitable, to use the existing implementations involving Array1 and Array2,
e.g. Index(), IndexRows()

Yes, I am using Index() and IndexRows() from array_ops when the stride is suitable:
https://github.com/k2-fsa/k2/blob/fc9e7154a11f3cd3a4d3ab953f37c20423287ee6/k2/python/csrc/torch/index_select.cu#L55-L60

https://github.com/k2-fsa/k2/blob/fc9e7154a11f3cd3a4d3ab953f37c20423287ee6/k2/python/csrc/torch/index_select.cu#L95-L99

There is no `IndexAdd` in array_ops but there is `IndexAdd` declared in tensor_ops, so I
implemented IndexAdd in tensor_ops and wrapped it to Python.

danpovey:
OK cool, thanks!  Merging.

qindazhu:
Seems there's a build failure? 
```
/home/runner/work/k2/k2/k2/csrc/utils.h(518): error: no instance of overloaded function "atomicAdd" matches the argument list
            argument types are: (double *, double)
          detected during:
            instantiation of "void k2::AtomicAdd(T *, T) [with T=double]" 
/home/runner/work/k2/k2/k2/csrc/tensor_ops.cu(298): here
            instantiation of "void k2::IndexAdd1DImpl(k2::ContextPtr, const T *, int32_t, int32_t, const int32_t *, __nv_bool, int32_t, int32_t, T *) [with T=double]" 
/home/runner/work/k2/k2/k2/csrc/tensor_ops.cu(373): here
```

csukuangfj:
> Seems there's a build failure?


Fixed in #454

danpovey:
Oops.  That may relate to different CUDA versions, possibly?  Looks like atomicAdd may only be implemented from CUDA version 6.0.

csukuangfj:
> there is a way, it seems, to implement it for earlier CUDA versions. We can maybe have an #ifdef on the number of __CUDA_ARCH__ ?

I just removed `double`.

danpovey:
OK I can merge your change for now to fix the build problem, but can you
please either make an issue or make a PR to fix it differently?

csukuangfj:
> OK I can merge your change for now to fix the build problem, but can you
please either make an issue or make a PR to fix it differently?

Fixed in #455

PR #447, opened 2020-12-03
Note for epsilon removal on GPU

danpovey:


danpovey:
it's OK to take individual pieces out of this to implement.
 @freewym  if you would like to contribute more to this project, AddSuffixToRagged() and AddPrefixToRagged() are not that difficult to implement.  The shapes can be modified using ChangeSublistSize( ..., 1), and you'll need a lambda to copy the actual data.

freewym:
OK. Let me take a look.

qindazhu:
Merging, as I need those declarations, thanks!

PR #448, opened 2020-12-04
fix some build issues of rm_epsilon

qindazhu:
Just push some fixes of build issues to avoid future conflicts between us.

### DIFF
--- a/k2/csrc/ragged_ops.h
+++ b/k2/csrc/ragged_ops.h
@@ -186,35 +186,6 @@ RaggedShape ChangeSublistSize(RaggedShape &src, int32_t size_delta);
  */
 RaggedShape ChangeSublistSizePinned(RaggedShape &src, int32_t size_delta);
 
-
-/*
-  Return a sub-part of a RaggedShape containing indexes 0 through n-1 of
-  its 1st axis.
-      @param [in] src  Source RaggedShape
-      @param [in] n    Number of (leading) indexes to keep; result will
-                       satisfy ans.Dim0() == n.  Must have 0 <= n <= src.Dim0().
-
-      @return          Returns RaggedShape containing a prefix of `src`.
-                       It will share memory with `src`.
- */
-RaggedShape Prefix(RaggedShape &src, int32_t n);
-
-
-/*
-  Return a vector of RaggedShapes containing various prefixes of `src`.
-
-        @param [in] src    Source RaggedShape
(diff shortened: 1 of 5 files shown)


k2/csrc/ragged_ops.h
```diff
@@ -216,31 +213,34 @@ std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
...
 
 /*
-  Return a sub-part of a RaggedShape containing indexes 0 through n-1 of
-  its 1st axis.
-      @param [in] src  Source RaggedShape
-      @param [in] n    Number of (leading) indexes to keep; result will
-                       satisfy ans.Dim0() == n; Must have 0 <= n <= src.Dim0().
-      @return          Returns RaggedShape containing a prefix of `src`.
-                       It will share memory with `src`.
+  Append a single element to each sub-array of a ragged matrix (we consider
```
qindazhu, line 216: We have duplicate declarations for `Prefix` and `GetPrefixes`, seems something got wrong when github did merging before.

k2/csrc/rm_epsilon.cu
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    RaggedShape epsilon_closure_combined_with_fool =
+        ComposeRaggedShapes(epsilon_closure_mapped.shape, foll_shape);
+    RaggedShape foll_fsa_shape =
+        RemoveAxis(epsilon_closure_combined_with_fool, 2);
+
+    combined_foll = FsaVec(foll_fsa_shape, arcs);
+
+    Ragged<int32_t> epsilon_closure_mapped_arc_map_foll =
```
qindazhu, line 155: @danpovey , as some of Ragged Ops require rvalue instead of lvalue (e.g. `AddSuffixToRagged`, `RemoveAxis`), I created some intermediate variables here. I just name them with some LONG BAD name for now to pass build ASAP, you may want to change them.

k2/csrc/ragged_ops.h
danpovey: ah OK, please fix when convenient..

k2/csrc/rm_epsilon.cu
danpovey: Adding the vars is, I wasn't sure if that would compile, you can change as you see fit.

qindazhu:
OK, merging, in case we have any conflicts about this in the following PR, we can fix any other issues later.

PR #453, opened 2020-12-05
implement AddPrefixToRagged and AddSuffixToRagged

freewym:
@danpovey Just want to make sure I am in the right direction before writing some test code...

### DIFF
--- a/k2/csrc/ragged_ops.cu
+++ b/k2/csrc/ragged_ops.cu
@@ -1257,8 +1257,9 @@ Ragged<int32_t> AddSuffixToRagged(Ragged<int32_t> &src,
   ContextPtr &c = src.Context();
   Array1<int32_t> dst_values(c, src.NumElements() + suffix.Dim());
   int32_t *dst_values_data = dst_values.Data();
-  // "row_splits1" and "row_ids1" below are actually on the last axis. We name them with "1" so that we can use
-  // "idx01" and "idx0" for those indexes in lambda, following the naming convention explained in k2/csrc/utils.h
+  // "row_splits1" and "row_ids1" below are actually on the last axis. We name
+  // them with "1" so that we can use "idx01" and "idx0" for those indexes in
+  // lambda, following the naming convention explained in k2/csrc/utils.h
   const int32_t *src_row_splits1_data = src.RowSplits(num_axes - 1).Data(),
                 *src_row_ids1_data = src.RowIds(num_axes - 1).Data(),
                 *src_values_data = src.values.Data(),
@@ -1288,8 +1289,9 @@ Ragged<int32_t> AddPrefixToRagged(Ragged<int32_t> &src,
   ContextPtr &c = src.Context();
   Array1<int32_t> dst_values(c, src.NumElements() + prefix.Dim());
   int32_t *dst_values_data = dst_values.Data();
-  // "row_splits1" and "row_ids1" below are actually on the last axis. We name them with "1" so that we can use
-  // "idx01" and "idx0" for those indexes in lambda, following the naming convention explained in k2/csrc/utils.h
+  // "row_splits1" and "row_ids1" below are actually on the last axis. We name
+  // them with "1" so that we can use "idx01" and "idx0" for those indexes in
+  // lambda, following the naming convention explained in k2/csrc/utils.h
   const int32_t *src_row_splits1_data = src.RowSplits(num_axes - 1).Data(),
                 *src_row_ids1_data = src.RowIds(num_axes - 1).Data(),
                 *src_values_data = src.values.Data(),
--- a/k2/csrc/ragged_ops.cu
+++ b/k2/csrc/ragged_ops.cu
@@ -5,6 +5,7 @@
  * @copyright
  * Copyright (c)  2020  Xiaomi Corporation (authors: Daniel Povey, Haowen Qiu)
  *                      Mobvoi Inc.        (authors: Fangjun Kuang)
+ *                      Microsoft Corporation (authors: Yiming Wang)
  *
  * @copyright
  * See LICENSE for clarification regarding multiple authors
@@ -1249,14 +1250,62 @@ std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
 
 Ragged<int32_t> AddSuffixToRagged(Ragged<int32_t> &src,
                                   const Array1<int32_t> &suffix) {
-  K2_LOG(FATAL) << "Not Implemented!";
-  return Ragged<int32_t>();
+  NVTX_RANGE(K2_FUNC);
+  K2_CHECK_EQ(suffix.Dim(), src.TotSize(src.NumAxes() - 2));
+  ContextPtr &c = src.Context();
+  Array1<int32_t> dst_values(c, src.NumElements() + suffix.Dim());
+  int32_t *dst_values_data = dst_values.Data();
+  int32_t num_axes = src.NumAxes();
+  const int32_t *src_row_splits_lastm1_data =
+      src.RowSplits(num_axes - 1).Data(),
+                *src_row_ids_lastm1_data = src.RowIds(num_axes - 1).Data(),
+                *src_values_data = src.values.Data(),
+                *suffix_data = suffix.Data();
+
+  K2_EVAL(
+      c, src.NumElements(), lambda_copy_src, (int32_t idx)->void {
+        int32_t dst_idx = idx + src_row_ids_lastm1_data[idx];
+        dst_values_data[dst_idx] = src_values_data[idx];
+      });
+  K2_EVAL(
+       c, suffix.Dim(), lambda_copy_suffix, (int32_t idx)->void {
+         int32_t row_id = src_row_ids_lastm1_data[idx];
+         int32_t dst_idx = src_row_splits_lastm1_data[row_id + 1] - 1;
+         dst_values_data[dst_idx] = suffix_data[idx];
+     });
+  RaggedShape dst_shape = ChangeSublistSize(src.shape, 1);
+
+  return Ragged<int32_t>(dst_shape, dst_values);
 }
 
 Ragged<int32_t> AddPrefixToRagged(Ragged<int32_t> &src,
                                   const Array1<int32_t> &prefix) {
-  K2_LOG(FATAL) << "Not Implemented!";
-  return Ragged<int32_t>();
+  NVTX_RANGE(K2_FUNC);
+  K2_CHECK_EQ(prefix.Dim(), src.TotSize(src.NumAxes() - 2));
+  ContextPtr &c = src.Context();
+  Array1<int32_t> dst_values(c, src.NumElements() + prefix.Dim());
+  int32_t *dst_values_data = dst_values.Data();
+  int32_t num_axes = src.NumAxes();
+  const int32_t *src_row_splits_lastm1_data =
+      src.RowSplits(num_axes - 1).Data(),
+                *src_row_ids_lastm1_data = src.RowIds(num_axes - 1).Data(),
+                *src_values_data = src.values.Data(),
+                *prefix_data = prefix.Data();
+
+  K2_EVAL(
+      c, src.NumElements(), lambda_copy_src, (int32_t idx)->void {
+        int32_t dst_idx = idx + src_row_ids_lastm1_data[idx] + 1;
+        dst_values_data[dst_idx] = src_values_data[idx];
+      });
+  K2_EVAL(
+       c, prefix.Dim(), lambda_copy_prefix, (int32_t idx)->void {
+         int32_t row_id = src_row_ids_lastm1_data[idx];
+         int32_t dst_idx = src_row_splits_lastm1_data[row_id];
+         dst_values_data[dst_idx] = prefix_data[idx];
+     });
+  RaggedShape dst_shape = ChangeSublistSize(src.shape, 1);
+
+  return Ragged<int32_t>(dst_shape, dst_values);
 }
 
 RaggedShape SubsampleRaggedShape(RaggedShape &src, Renumbering &renumbering) {
--- a/k2/csrc/ragged_ops.cu
+++ b/k2/csrc/ragged_ops.cu
@@ -1251,27 +1251,28 @@ std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
 Ragged<int32_t> AddSuffixToRagged(Ragged<int32_t> &src,
                                   const Array1<int32_t> &suffix) {
   NVTX_RANGE(K2_FUNC);
-  K2_CHECK_EQ(suffix.Dim(), src.TotSize(src.NumAxes() - 2));
+  int32_t num_axes = src.NumAxes();
+  K2_CHECK_GE(num_axes, 2);
+  K2_CHECK_EQ(suffix.Dim(), src.TotSize(num_axes - 2));
   ContextPtr &c = src.Context();
   Array1<int32_t> dst_values(c, src.NumElements() + suffix.Dim());
   int32_t *dst_values_data = dst_values.Data();
-  int32_t num_axes = src.NumAxes();
-  const int32_t *src_row_splits_lastm1_data =
-      src.RowSplits(num_axes - 1).Data(),
-                *src_row_ids_lastm1_data = src.RowIds(num_axes - 1).Data(),
+  // "row_splits1" and "row_ids1" below are actually on the last axis. We name them with "1" so that we can use
+  // "idx01" and "idx0" for those indexes in lambda, following the naming convention explained in k2/csrc/utils.h
+  const int32_t *src_row_splits1_data = src.RowSplits(num_axes - 1).Data(),
+                *src_row_ids1_data = src.RowIds(num_axes - 1).Data(),
                 *src_values_data = src.values.Data(),
                 *suffix_data = suffix.Data();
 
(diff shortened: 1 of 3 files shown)


csukuangfj:
@freewym 
FYI: Dan does NOT recommend us to use force push.

k2/csrc/ragged_ops.cu
```diff
@@ -1249,14 +1250,62 @@ std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
 
 Ragged<int32_t> AddSuffixToRagged(Ragged<int32_t> &src,
                                   const Array1<int32_t> &suffix) {
-  K2_LOG(FATAL) << "Not Implemented!";
-  return Ragged<int32_t>();
+  NVTX_RANGE(K2_FUNC);
+  K2_CHECK_EQ(suffix.Dim(), src.TotSize(src.NumAxes() - 2));
```
csukuangfj, line 1254: The last axis is `NumAxes() - 1`. 

I think there is a typo in the documentation of the interface.

qindazhu:
@freewym many thanks! The implementation is right except that you can merge those two kenerls into 1 kenerl, something like this:
```
K2_EVAL(
      c, *dest_values()*, lambda_copy_value, (int32_t idx01)->void {
        int32_t idx0 = dest_row_id1_data[idx01];  // noted dest_row_splits1_data is the row splits on last axis of dest
       int32_t idx0x_this_row = dest_row_split1_data[idx0]; // first element index in this row
        int32_t idx0x_next_row = dest_row_split1_data[idx0 + 1]; // first element index in the next row
       K2_CHECK_GT(idx0x_next_row, idx0x_this_row); // as we have inserted an extra element for each row in src, so there's no empty rows
       if(idx01 == indx0x_next_row -1 ){
         // the last element of this row in dest, should copy from suffix data
       }
      else{
         // not the last element of this row in dest, should copy for src data
       }
        
```
Noted I added `idx0x_this_row` and `K2_CHECK` stuff just for showing the process clearly, you may delete those lines in the final code. Also noted in `Prefix` version, you will use `idx0x_this_row` to check if current element is the first element in this row of dest.

k2/csrc/ragged_ops.cu
```diff
@@ -1249,14 +1250,62 @@ std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
 
 Ragged<int32_t> AddSuffixToRagged(Ragged<int32_t> &src,
                                   const Array1<int32_t> &suffix) {
-  K2_LOG(FATAL) << "Not Implemented!";
-  return Ragged<int32_t>();
+  NVTX_RANGE(K2_FUNC);
```
csukuangfj, line 1253: Please also add a check that `src` has at least two axes.

k2/csrc/ragged_ops.cu
```diff
@@ -1249,14 +1250,62 @@ std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
...
+        int32_t dst_idx = idx + src_row_ids_lastm1_data[idx];
+        dst_values_data[dst_idx] = src_values_data[idx];
+      });
+  K2_EVAL(
+       c, suffix.Dim(), lambda_copy_suffix, (int32_t idx)->void {
+         int32_t row_id = src_row_ids_lastm1_data[idx];
+         int32_t dst_idx = src_row_splits_lastm1_data[row_id + 1] - 1;
+         dst_values_data[dst_idx] = suffix_data[idx];
+     });
+  RaggedShape dst_shape = ChangeSublistSize(src.shape, 1);
```
qindazhu, line 1276: As I suggested above, you can call this before lambda to get `row_splits` and `row_ids` of dst

k2/csrc/ragged_ops.cu
```diff
@@ -1249,14 +1250,62 @@ std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
...
                                   const Array1<int32_t> &suffix) {
-  K2_LOG(FATAL) << "Not Implemented!";
-  return Ragged<int32_t>();
+  NVTX_RANGE(K2_FUNC);
+  K2_CHECK_EQ(suffix.Dim(), src.TotSize(src.NumAxes() - 2));
+  ContextPtr &c = src.Context();
+  Array1<int32_t> dst_values(c, src.NumElements() + suffix.Dim());
+  int32_t *dst_values_data = dst_values.Data();
+  int32_t num_axes = src.NumAxes();
+  const int32_t *src_row_splits_lastm1_data =
```
qindazhu, line 1259: I aslo suggest that you rename this `src/dst_row_split1_data` and `src/dst_row_ids1_data`, and add documentation to say that they are actually row_splits and row_ids on the last axis.
The advantage of naming `row_split1` and `row_ids1` is that we can use `idx01` and `idx0` to name those indexes in lambda, which will make code easy to read according to our naming convention

qindazhu:
BTW, regarding testing, you need to add testing with RandomShapes, e.g. https://github.com/k2-fsa/k2/blob/7e084e9cfd0982c9d2db3db8ff8a2fd6d97e3b6c/k2/csrc/ragged_test.cu#L1994-L1999,

k2/csrc/ragged_ops.cu
```diff
@@ -1249,14 +1250,62 @@ std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
...
+
+  K2_EVAL(
+      c, src.NumElements(), lambda_copy_src, (int32_t idx)->void {
+        int32_t dst_idx = idx + src_row_ids_lastm1_data[idx];
+        dst_values_data[dst_idx] = src_values_data[idx];
+      });
+  K2_EVAL(
+       c, suffix.Dim(), lambda_copy_suffix, (int32_t idx)->void {
+         int32_t row_id = src_row_ids_lastm1_data[idx];
+         int32_t dst_idx = src_row_splits_lastm1_data[row_id + 1] - 1;
```
csukuangfj, line 1273: I think 
```cpp
int32_t dst_idx = src_row_splits_lastm1_data[idx + 1] + idx;
```

k2/csrc/ragged_ops.cu
```diff
@@ -1249,14 +1250,62 @@ std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
...
-  return Ragged<int32_t>();
+  NVTX_RANGE(K2_FUNC);
+  K2_CHECK_EQ(suffix.Dim(), src.TotSize(src.NumAxes() - 2));
+  ContextPtr &c = src.Context();
+  Array1<int32_t> dst_values(c, src.NumElements() + suffix.Dim());
+  int32_t *dst_values_data = dst_values.Data();
+  int32_t num_axes = src.NumAxes();
+  const int32_t *src_row_splits_lastm1_data =
+      src.RowSplits(num_axes - 1).Data(),
+                *src_row_ids_lastm1_data = src.RowIds(num_axes - 1).Data(),
```
csukuangfj, line 1261: What is the meaning of `m1` in `lastm1`?

k2/csrc/ragged_ops.cu
```diff
@@ -1249,14 +1250,62 @@ std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
...
+
+  K2_EVAL(
+      c, src.NumElements(), lambda_copy_src, (int32_t idx)->void {
+        int32_t dst_idx = idx + src_row_ids_lastm1_data[idx] + 1;
+        dst_values_data[dst_idx] = src_values_data[idx];
+      });
+  K2_EVAL(
+       c, prefix.Dim(), lambda_copy_prefix, (int32_t idx)->void {
+         int32_t row_id = src_row_ids_lastm1_data[idx];
+         int32_t dst_idx = src_row_splits_lastm1_data[row_id];
```
csukuangfj, line 1303: I think this line is not correct and can be debugged by unit tests.

freewym: Yes, I realized..

k2/csrc/ragged_ops.cu
freewym: m1 means "minus 1"

k2/csrc/ragged_ops.cu
qindazhu: mm, yeah, as @freewym use idx as row id (i.e. idx0) here. Anyway, I suggest we do this in one kenerl indexes with dest_values.Dim() instead of two kenerls

csukuangfj:
> @freewym many thanks! The implementation is right except that you can merge those two kenerls into 1 kenerl, something like this:

I prefer @freewym 's approach:
- It is simpler and less error-prone. Not sure whether it is more efficient since it has no branches in the kernel.
- It handles the case that some `idx0` may have no elements at all.

k2/csrc/ragged_ops.cu
csukuangfj: I think `src_row_ids_data` is fine here as there are no other `row_ids` here
and there exist no confusions.

k2/csrc/ragged_ops.cu
csukuangfj: Here idx means idx0, so it is incorrect to use it to index `src_row_ids_lastm1_data`.
I would suggest following the naming convention established by Dan.

k2/csrc/ragged_ops.cu
csukuangfj: I think his approach is fine and easy to understand.

k2/csrc/ragged_ops.cu
freewym: I am not familiar with the naming convention, so I may be wrong: I think  I was using `idx` to index idx012 if `src` is 3D, so `row_id` is idx01. I should have written `dst_idx` as
```c++
int32_t dst_idx = src_row_splits_lastm1_data[row_id + 1] - 1 + row_id;
```

where ` src_row_splits_lastm1_data[row_id + 1] - 1` is the position of the last element of row `row_id` in `src`

csukuangfj: Please refer to https://github.com/k2-fsa/k2/blob/7e084e9cfd0982c9d2db3db8ff8a2fd6d97e3b6c/k2/csrc/utils.h#L107

k2/csrc/ragged_test.cu
```diff
@@ -2028,4 +2029,66 @@ TEST(RaggedShapeOpsTest, GetPrefixesTest) {
...
+        int32_t num_axes = src.NumAxes();
+        Array1<int32_t> prefix =
+            RandUniformArray1<int32_t>(context, src.TotSize(num_axes - 2),
+                                       0, 100);
+        Ragged<int32_t> dst = AddPrefixToRagged(src, prefix);
+        EXPECT_EQ(dst.NumAxes(), num_axes);
+        EXPECT_EQ(dst.NumElements(), src.NumElements() + prefix.Dim());
+        Ragged<int32_t> src_cpu = src.To(GetCpuContext());
+        Ragged<int32_t> dst_cpu = dst.To(GetCpuContext());
+        for (RaggedShapeIndexIterator src_iter = src_cpu.shape.Iterator();
```
qindazhu, line 2077: I like this testing approach with Iterator!!!!

k2/csrc/ragged_ops.cu
```diff
@@ -1249,14 +1250,66 @@ std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
...
+  int32_t *dst_values_data = dst_values.Data();
+  // "row_splits1" and "row_ids1" below are actually on the last axis. We name
+  // them with "1" so that we can use "idx01" and "idx0" for those indexes in
+  // lambda, following the naming convention explained in k2/csrc/utils.h
+  const int32_t *src_row_splits1_data = src.RowSplits(num_axes - 1).Data(),
+                *src_row_ids1_data = src.RowIds(num_axes - 1).Data(),
+                *src_values_data = src.values.Data(),
+                *prefix_data = prefix.Data();
+
+  K2_EVAL(
```
qindazhu, line 1300: @danpovey do we need to wrap those two kernels with `ParallelRunner` even if `ParallelRunner` does nothing currently because of cuda record event issue? or merge them into one as I suggested before https://github.com/k2-fsa/k2/pull/453#issuecomment-739109372

k2/csrc/ragged_test.cu
```diff
@@ -2028,4 +2029,74 @@ TEST(RaggedShapeOpsTest, GetPrefixesTest) {
...
+            RandUniformArray1<int32_t>(context, src.TotSize(num_axes - 2),
+                                       0, 100);
+        Ragged<int32_t> dst = AddPrefixToRagged(src, prefix);
+        EXPECT_EQ(dst.NumAxes(), num_axes);
+        EXPECT_EQ(dst.NumElements(), src.NumElements() + prefix.Dim());
+        Ragged<int32_t> src_cpu = src.To(GetCpuContext());
+        Ragged<int32_t> dst_cpu = dst.To(GetCpuContext());
+        Array1<int32_t> prefix_cpu = prefix.To(GetCpuContext());
+        Array1<int32_t> dst_row_splits = dst_cpu.RowSplits(num_axes - 1);
+        for (RaggedShapeIndexIterator dst_iter = dst_cpu.shape.Iterator();
```
qindazhu, line 2084: Ah, @freewym , I thought your previous testing approach (i.e. testing src first with indexes then testing prefix) is much more clear than this (as we don't need to write compact/very-high-performace code in testing, it's more important to make it clear.

Anyway, it's amazing that you use `RaggedShapeIndexIterator` to test this, I definitely like the testing approach with iterator, I just realized that I have not used it for a long time (the last time I used it is to test Transpose, Append, Stack stuff..)

freewym: The test for prefix/suffix in the previous one is basically duplicate what we do in the functions; in this one I use the triplet to used to index all elements in dst including those from prefix/suffix, which I think is closer to the description of the functions

qindazhu: (duplicate is fine as long as our testing depend on the interface instead of implementation). Anyway, this version is also fine with me,  it's up to you to change it or not. (and again, I like `Iterator` here). Will make this issue resloved

k2/csrc/ragged_ops.cu
danpovey: Mm.  I think it would be better to use a single kernel like you suggested.
The time CUDA would take to deal with these events, or to have sequential kernels run, would likely be much greater than the one or two instructions of extra latency that we'd get in your kernel that has an if-statement.

k2/csrc/ragged_ops.cu
```diff
@@ -1249,14 +1250,70 @@ std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
...
+                *src_values_data = src.values.Data(),
+                *suffix_data = suffix.Data();
+  int32_t *dst_values_data = dst_values.Data();
+
+  K2_EVAL(
+      c, dst_shape.NumElements(), lambda_copy_values, (int32_t idx01)->void {
+        int32_t idx0 = dst_row_ids1_data[idx01];
+        if (idx01 == dst_row_splits1_data[idx0 + 1] - 1) {
+          // idx01 points to the last element of this row; copy from suffix
+          dst_values_data[idx01] = suffix_data[idx0];
```
csukuangfj, line 1274: Some elements in `suffix_data` will never be accessed
if some `idx0` has no corresponding idx01.

For example:
```
row splits: [ 0 2 2 3 3]
values: [x x y]
```
You will never get `idx0 == 1` inside the kernel in the current implementation.

csukuangfj: Not sure if your tests cover this kind of case.

freewym: idx01 is an index into dst, so I think it's always has the corresponding idx0

csukuangfj: > idx01 is an index into dst, so I think it always has the corresponding idx0


Yes, the above statement is true. But the reverse is not.

csukuangfj: You can run `cu_ragged_test` locally and you will find that the random case will not pass.

freewym: Sorry I don't quite catch the point. there are no empty sublists in dst, so each idx0 of dst will be obtained.

freewym: > You can run `cu_ragged_test` locally and you will find that the random case will not pass.

For some reason (library conflicts after cmake), I am currently unable to build targets for tests.

danpovey:
Try removing the build directory and starting again.

k2/csrc/ragged_ops.cu
csukuangfj: > there are no empty sublists in dst, so each idx0 of dst will be obtained.

You are correct if there are no empty sublists. But the current implementation does not handle
sublists that are empty.  The code does not check for empty sublists and `RandomRagged`
may return empty sublists. That is why all your passed tests contain non-empty lists.

qindazhu: mm, I didn't see there's any problem of @freewym 's implementation, as there are no empty rows in `dest` ( we have called ChangedSublist to add 1 for each row/sublist).

csukuangfj: sorry, you're correct. The current code has been updated.

danpovey:
Thanks!  Just asking a question offline about copyright (see email), after that should be OK to merge.

qindazhu:
Thanks, merging.

PR #454, opened 2020-12-05
Fix build errors for GitHub actions.

csukuangfj:
Remove `double` in `atomicAdd` as it is not always supported.

See https://stackoverflow.com/questions/37566987/cuda-atomicadd-for-doubles-definition-error

### DIFF
--- a/k2/csrc/dtype.h
+++ b/k2/csrc/dtype.h
@@ -173,16 +173,11 @@ struct DtypeOf<uint64_t> {
     }                                                                    \
   } while (0)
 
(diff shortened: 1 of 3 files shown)


danpovey:
I think a better solution might be to have an ifdef where we provide the appropriate definition of atomicAdd for double.
like in the article I linked in a previous comment.

Bear in mind that atomicAdd won't scale well to situations where many of the indexes are the same.  In that case we would need to solve it differently, by sorting pairs of indexes and using a reduction-type kernel.

PR #457, opened 2020-12-05
Ragged utils etc

danpovey:


danpovey:
This is some lower-level utilities that will be needed in an Append() for ragged tensors that can work on any axis.  It will be used in the epsilon removal algorithm.

k2/csrc/ragged_ops.cu
```diff
@@ -1379,4 +1380,14 @@ Array1<int32_t> GetDecreasingSizeOrder(RaggedShape &shape) {
   return index_map;
 }
 
+RaggedShape GetLayer(RaggedShape &src, int32_t layer) {
+  K2_CHECK_GE(layer, 0);
```
qindazhu, line 1384: also check `src.NumAxes() >=2`?

k2/csrc/ragged_ops.h
```diff
@@ -292,6 +292,18 @@ std::vector<RaggedShape> UnsqueezeParallel(int32_t num_srcs, RaggedShape **src,
...
+/*
+  Returns a `sub-shape` of `src` consisting of one of its RaggedShapeLayer
+  elements, i.e. one of the levels of its shape.  This returned shape
+  will have NumAxes() == 2, but it is the minimal case of a RaggedShape.
+
+    @param [in] src   Source RaggedShape
+    @param [in] layer  Layer that is desired, from 0 .. src.NumAxes() - 2.
+                      View this as an index into its Layers() vector.
+ */
+RaggedShape GetLayer(RaggedShape &src, int32_t layer);
```
qindazhu, line 304: not sure, but if it's possible, it's better if we can change the parameter to `const RaggedShape& src` in case we'd call this function with lvalue.

k2/csrc/ragged_ops.h
```diff
@@ -388,10 +400,17 @@ Ragged<T> Transpose(Ragged<T> &src,
                            same value
       @param [in] num_srcs Number of source shapes to append; require
                            num_srcs > 0.
+      @param [out] merge_map  If not nullptr, will be set to the merge-map
+                          that tells us for each 0 <= i < ans.NumElements(),
+                          which element of `src` it came from (available
+                          as `ans[i]i % num_srcs`) and its element-index within
```
qindazhu, line 406: `merge_map[i] % num_srcs`? I'm confused with this documentation.

k2/csrc/ragged_ops.h
```diff
@@ -388,10 +400,17 @@ Ragged<T> Transpose(Ragged<T> &src,
                            same value
       @param [in] num_srcs Number of source shapes to append; require
                            num_srcs > 0.
+      @param [out] merge_map  If not nullptr, will be set to the merge-map
```
qindazhu, line 403: it's better if we can say `meger_map.Dim() == ans.Dim()`,  even if user can figure this out with `0 <= i < ans.NumElements()`

k2/csrc/ragged_ops.h
```diff
@@ -388,10 +400,17 @@ Ragged<T> Transpose(Ragged<T> &src,
...
+                          that tells us for each 0 <= i < ans.NumElements(),
+                          which element of `src` it came from (available
+                          as `ans[i]i % num_srcs`) and its element-index within
+                          `src[i]` (available as `ans[i] / num_srcs`.
+
       @param [in] src      Array of sources to append
```
qindazhu, line 409: this line should be in front of the documentation of `merge_map` according to the parameter order.

k2/csrc/ragged_utils_inl.h
```diff
@@ -0,0 +1,34 @@
...
+ * @brief
+ * ragged_utils_inl
+ *
+ * @note
+ * This is to be included only from ragged_utils.h.
+ *
+ * @copyright
+ * Copyright (c)  2020  Xiaomi Corporation (authors: Daniel Povey
+ *                                                   Haowen Qiu)
+ *                      Mobvoi Inc.        (authors: Fangjun Kuang)
```
qindazhu, line 11: Ah, I think we can add Fangjun's copyright under `Xiaomi Corporation` now.

k2/csrc/array_ops_test.cu
```diff
@@ -1663,4 +1663,15 @@ TEST(OpsTest, Array2Assign) {
   }
 }
 
+TEST(OpsTest, SizesToMergeMapTest) {
+  for (int loop = 0; loop < 10; loop++) {
```
qindazhu, line 1667: Seems below test is only a simple testing, we don't need to `loop 10` here. I guess you plan to add random tests here. 
Anyway, at least we can add a TODO (for random case, as kenerls definitely need to be tested with random case)

k2/csrc/array_ops.cu
```diff
@@ -373,4 +374,77 @@ Array1<int32_t> RowSplitsToSizes(const Array1<int32_t> &row_splits) {
...
+  int32_t num_srcs = sizes.size();
+
+  ContextPtr cpu_context = GetCpuContext();
+  Array1<int32_t> row_splits_cpu(cpu_context, num_srcs + 1);
+  int32_t *row_splits_cpu_data = row_splits_cpu.Data();
+  int32_t tot_size = 0;
+  row_splits_cpu_data[0] = 0;
+  for (int32_t i = 0; i < num_srcs; i++) {
+    tot_size += sizes[i];
+    row_splits_cpu_data[i+1] = tot_size;
```
qindazhu, line 413: just a question, seems we are doing `exlcusiveSum` here? Can we call `ExclusiveSum` here ( it works for CPU as well), I mean, something like below:
```
Array1<int32_t> array_sizes(cpu, sizes);
Exclusive(array_sizes, &row_splits_cpu); // note this is the version that dst.Dim() == row_splits_cpu.Dim() = src.Dim() +1
```

k2/csrc/array_ops.cu
```diff
@@ -373,4 +374,77 @@ Array1<int32_t> RowSplitsToSizes(const Array1<int32_t> &row_splits) {
...
+
+  if (c->GetDeviceType() == kCpu) {
+    int32_t *ans_data = ans.Data();
+    int32_t cur = 0;
+    for (int32_t src = 0; src < num_srcs; src++) {
+      int32_t begin = cur,  // i.e. the previous end.
+                end = row_splits_cpu_data[src+1];
+      for (; cur != end; ++cur) {
+        // the 'src' says which source this item came from, and (cur - begin) *
+        // num_srcs is the position within that source.
```
qindazhu, line 425: `cur - begin` instead of `(cur - begin) * num_srcs` is the position within that source

k2/csrc/array_ops.cu
```diff
@@ -373,4 +374,77 @@ Array1<int32_t> RowSplitsToSizes(const Array1<int32_t> &row_splits) {
...
+        ans_data[cur] = src + (cur - begin) * num_srcs;
+      }
+    }
+    return ans;
+  }
+
```
qindazhu, line 431: add check `c->GetDeviceType() == kCuda` here.

k2/csrc/array_ops.cu
```diff
@@ -373,4 +374,77 @@ Array1<int32_t> RowSplitsToSizes(const Array1<int32_t> &row_splits) {
...
+  Array1<int32_t> row_splits = row_splits_cpu.To(c);
+  int32_t *row_splits_data = row_splits.Data();
+  int32_t *merge_map_data = ans.Data();
+  int32_t avg_elems_per_row = (tot_size + num_srcs - 1) / num_srcs,
+            threads_per_row = RoundUpToNearestPowerOfTwo(avg_elems_per_row),
+              tot_threads = num_srcs * threads_per_row;
+  int32_t block_size = 256;
+  int32_t grid_size = NumBlocks(tot_threads, block_size);
+
+  K2_CUDA_SAFE_CALL(SizesToMergeMapKernel<<<grid_size, block_size, 0,
```
qindazhu, line 441: @danpovey , wondering if we can do this in a lambda (together with CPU version) instead of in a kenerl, saying,
we construct a RaggedShape2 with row_splits, then `idx0 + idx1 * num_srcs` will be the value we want for index idx01. The code is simpler but the disadvantage is that we needs populate row_ids first. Not sure we should care about the performace here or code simplity (I thought those two approach will not make a very big difference?)

k2/csrc/ragged_utils.h
```diff
@@ -0,0 +1,172 @@
...
+// ragged implementation, and not user-facing-- for example, operations that are
+// limited to a RaggedShape with NumAxes() == 2 (i.e. a single ragged axis).
+namespace k2 {
+
+/*
+  This checks that the RaggedShapeLayer indexed `axis`, i.e. src.Axes()[axis],
+  is the same for all sources (i.e. represents the same ragged shape).
+
+  This function may do a quick or more thorough check depending on the debug
+  level k2 was compiled with it is intended for situations where it's required
```
qindazhu, line 36: `compiled with.`, there's a `.` or `,` after `with`?

k2/csrc/ragged_utils.cu
```diff
@@ -0,0 +1,182 @@
...
+  K2_CHECK(axis >= 0 && axis + 1 < src[0]->NumAxes());
+  std::vector<int32_t*> row_splits_data_vec;
+  row_splits_data_vec.reserve(num_srcs);
+  int32_t row_splits_dim, row_ids_dim;
+  for (int32_t s = 0; s < num_srcs; s++) {
+    // RowSplits(1) .. is the lowest numbered row-splits...
+    int32_t *data = src[s]->RowSplits(axis + 1).Data();
+    if (s == 0 || data != row_splits_data_vec[0])
+      row_splits_data_vec.push_back(data);
+    if (s == 0) {
```
qindazhu, line 39: is that possible we put code related with those two `if (s==0)` outside of the loop ( and `s` starts from 1 of course)? I was thinking that would be faster and clearer?

k2/csrc/ragged_utils.cu
```diff
@@ -0,0 +1,182 @@
...
+      row_splits_dim = src[s]->RowSplits(axis + 1).Dim();
+      row_ids_dim = src[s]->RowIds(axis + 1).Dim();
+    } else {
+      K2_CHECK_EQ(row_splits_dim, src[s]->RowSplits(axis + 1).Dim());
+      K2_CHECK_EQ(row_ids_dim, src[s]->RowIds(axis + 1).Dim());
+    }
+  }
+  if (row_splits_data_vec.size() <= 1)
```
qindazhu, line 47: maybe it's better to add documentation to say those srcs are exactly same (same address of row_splits) on that aixs before this line?

k2/csrc/ragged_utils.cu
```diff
@@ -0,0 +1,182 @@
...
+  int32_t **row_splits_ptrs_data = row_splits_ptrs.Data();
+  int32_t *is_bad_data = is_bad.Data();
+  K2_EVAL2(c, row_splits_ptrs.Dim() - 1,
+           row_splits_dim, lambda_check_row_splits,
+           (int32_t i, int32_t j) -> void {
+             if (row_splits_ptrs_data[i+1][j] !=
+                 row_splits_ptrs_data[0][j])
+               is_bad_data[0] = 1;
+           });
+  if (!(is_bad[0] == 0)) {
```
qindazhu, line 62: `if(is_bad[0] == 1)`?

k2/csrc/ragged_utils.h
```diff
@@ -0,0 +1,172 @@
...
+                                   Array1<int32_t> *merge_map = nullptr);
+
+/*
+  Merge a ragged axis given a 'merge_map' obtained from a previous operation on an
+  axis.
+
+     @param [in] num_srcs  Number of RaggedShapes being merged
+     @param [in] axis      Axis to operate on, viewed as index into src[i]->Axes(),
+                           with 0 <= axis < src[i]->NumAxes() - 1 (for 0 <= i < num_srcs).
+     @param [in] src       src   Array of pointers to RaggedShapes to be merged;
```
qindazhu, line 143: do we need them (i.e. `src`) to be on the same device?

k2/csrc/ragged_utils.h
```diff
@@ -0,0 +1,172 @@
...
+ */
+RaggedShape AppendRaggedAxisBefore(int32_t num_srcs,
+                                   int32_t axis,
+                                   RaggedShape **src,
+                                   Array1<int32_t> *merge_map = nullptr);
+
+
+
+/*
+   Intersperses rows from a single layer of an array of RaggedShape sources,
```
qindazhu, line 94: @danpovey, just a question, this function will be called in `MergeRaggedAxis`? Anyway, it's glad to see your these declarations, I was thinking `Append` on other axis would be very hard.

k2/csrc/ragged_utils.cu
```diff
@@ -0,0 +1,182 @@
...
+          num_rows = src[0]->TotSize(layer),
+       tot_elems = 0;
+  for (int32_t i = 0; i < num_srcs; i++) {
+    if (i > 0) {
+      K2_CHECK_EQ(src[i]->NumAxes(), num_axes);
+      K2_CHECK_EQ(src[i]->TotSize(layer), num_rows);
+    }
+    Array1<int32_t> &row_ids = src[i]->RowIds(layer + 1),
+                 &row_splits = src[i]->RowSplits(layer + 1);
+    tot_elems += (row_ids_dims_vec[i] = row_ids.Dim());
```
qindazhu, line 100: sorry, where did `row_ids_dims_vec[i]` get value set? I just didn't find the code..

k2/csrc/ragged_utils.cu
```diff
@@ -0,0 +1,182 @@
...
+                  pos = j / num_srcs;
+          int32_t this_row_split = row_splits_ptrs_data[src][pos];
+          sum += this_row_split;
+        }
+        row_splits_data[i] = sum;
+      };
+      EvalDevice(c, new_num_rows + 1, lambda_get_row_splits);
+    } else {
+      // Set the row_splits initially to the sizes, then do exclusive-sum.
+      auto lambda_get_sizes = [=] __device__ (int32_t i) -> void {
```
qindazhu, line 146: not sure if it's worthy or not, but this operation can be viewed `splice` row_splits of those srcs on axis 1,  it may be worth putting this along with `SpliceRowSplits` (can be viewed `splice` on axis 0?)
 https://github.com/k2-fsa/k2/blob/b85b9dbe2cab9a0d7d6926f1e7fb1ae2b036aeb0/k2/csrc/array_ops.h#L169

Anyway, just metion it..

k2/csrc/ragged_ops.cu
danpovey: Mm, I don't always check this because such a Ragged objecct would not be valid, and for instance probably could not be constructed from Python.

k2/csrc/ragged_ops.h
danpovey: ok..

k2/csrc/ragged_utils_inl.h
danpovey: mm ok, lets have him do that when he modifies files, I am trying to keep these to just the list of those who contributed to that file as a matter of general principle.
if it turns out we don't need this file I will delete it.

k2/csrc/array_ops_test.cu
danpovey: will reduce to 2

k2/csrc/array_ops.cu
danpovey: mm yeah in principle, but would have unnecessary copy... would prefer to leave it.

k2/csrc/array_ops.cu
danpovey: the whole point of this function is to avoid having to call 2 kernels (this kind of operation might become the limiting factor in some casess).

k2/csrc/ragged_utils.h
danpovey: I'm not saying this every time; it's assumed that in general things must be on the same device.

k2/csrc/ragged_utils.h
danpovey: It is a primitive to be used in Append() and Stack() operations.

k2/csrc/array_ops.cu
qindazhu: Ah, got it, thanks!

k2/csrc/ragged_utils.cu
danpovey: Right there: (row_ids_dims_vec[i] = row_ids.Dim()), assignment as expression.
It turns out it is not being used, I will delete.

k2/csrc/ragged_utils.cu
danpovey: mm, will think about this..

k2/csrc/ragged_utils.cu
qindazhu: Ah, I read it as `==` before, many thanks!

k2/csrc/ragged_ops.cu
qindazhu: OK

PR #460, opened 2020-12-06
Ragged utils 2

danpovey:
This makes Append() and Stack() use the new ragged utils, shoule be faster as well as more flexible.
Not finished test after latest bug fix .

### DIFF
--- a/k2/csrc/array_ops_inl.h
+++ b/k2/csrc/array_ops_inl.h
@@ -504,6 +504,11 @@ Array1<T> Range(ContextPtr c, int32_t dim, T first_value, T inc /*=1*/) {
   return ans;
 }
 
+template<typename T>
+Array1<T> Arange(ContextPtr c, T begin, T end, T inc) {
+  return Range<T>(c, (end + inc - 1 - begin) / inc, begin, inc);
+}
+
 template <typename T>
 Array2<T> ToContiguous(const Array2<T> &src) {
   NVTX_RANGE(K2_FUNC);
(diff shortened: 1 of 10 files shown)


k2/csrc/array_ops_inl.h
```diff
@@ -877,6 +882,32 @@ void Assign(Array2<T> &src, Array2<T> *dest) {
...
 
+template <typename T>
+Array1<T> MergeWithMap(const Array1<uint32_t> &merge_map,
+                       int32_t num_srcs, const Array1<T> **src) {
+  int32_t dim = merge_map.Dim();
+  ContextPtr &c = merge_map.Context();
+  std::vector<const T*> src_ptrs_vec(num_srcs);
+  int32_t src_tot_dim = 0;
+  for (int32_t i = 0; i < num_srcs; i++) {
+    src_tot_dim += src[i]->Dim();
```
qindazhu, line 893: it's better to add `K2_CHECK(IsCompatible(c, src->Context())`?

k2/csrc/array_ops_inl.h
```diff
@@ -877,6 +882,32 @@ void Assign(Array2<T> &src, Array2<T> *dest) {
...
+    src_ptrs_vec[i] = src[i]->Data();
+  }
+  Array1<const T*> src_ptrs(c, src_ptrs_vec);
```
qindazhu, line 896: I believe you want to check src_tot_dim == dim here, as you have computed src_tot_dim above but did not use it.

PR #464, opened 2020-12-07
Add pinned context that allocates pinned memory.

### LINKED ISSUE #451: Pinned memory allocator
This is a change that could significantly reduce the latency of some operations.

In addition to the contexts returned by GetCpuContext() and GetCudaContext(), we need a 3rd type of context which would return pinned (page-locked) memory.  Call this function GetPinnedContext().
(We need to remove the member-function of the same name from class Context).

Internally, in the PyTorch case this could probably use `at::cuda::getPinnedMemoryAllocator();`
In the default case (not compiling with Torch), for now we can just have the call to GetPinnedContext() return the CPU context, unless there is an easy way to allocate pinned memory (and cache it), IDK about that moderngpu allocator we use.

These pinned context virtual classes  would return kCpu as device type, but we could add a virtual function IsPinned() to class Context that will return true if the allocator returns pinned memory (i.e. only for the pinned context type).

We will have to change the function MemoryCopy().  Currently it takes only one Context object as an arg.  That appears to be the destination context, although this was not documented, which is not good.  We can give it the src and dest Contexts.
In the case of host to device transfers, if the source is pinned memory we can do async copy.  (For device to host, for now we'll make the transfer always synchronous).

It might be easier if GetMemoryCopyKind() were to accept ContextPtr instead of Context (avoid the hassle of dereferencing); and MemoryCopy could also accept ContextPtr.  That will make the code a bit easier to read and more consistent.

csukuangfj:
Closes #451 

We need to change `MemoryCopy`:
https://github.com/k2-fsa/k2/blob/bd4495ac7a665282e99c4c7fef059bc3eebc55a3/k2/csrc/context.h#L174-L180

If we are copying pinned memory to the device, we can use `cudaMemcpyAsync`. But in this case, we
have to set up a CUDA event to prevent the pinned memory from being reused in PyTorch while the copying
is in progress.

---

I suggest adding a new method in `class Context`:

```cpp
virtual void CopyData(ContextPtr src_context, const void *src, void *dest, size_t num_bytes) = 0;
```

If `src_context` is not a pinned context, it becomes a wrapper around the current `MemoryCopy`.


csukuangfj:
The following is the benchmark results from our server:

```
$ ./bin/cu_array_ops_test --gtest_filter=PinnedContext.*
Running main() from /root/fangjun/open-source/k2/build/_deps/googletest-src/googletest/src/gtest_main.cc
Note: Google Test filter = PinnedContext.*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from PinnedContext
[ RUN      ] PinnedContext.SpeedTest
non-pinned memory copy (host->device): : 3.21 GB/s
pinned memory copy (host->device): : 9.47 GB/s
[       OK ] PinnedContext.SpeedTest (13592 ms)
[----------] 1 test from PinnedContext (13592 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (13592 ms total)
[  PASSED  ] 1 test.
```

k2/csrc/context.h
```diff
@@ -329,6 +321,10 @@ ContextPtr GetCpuContext();
 // from PyTorch
 ContextPtr GetCudaContext(int32_t gpu_id = -1);
 
+// Returns a (CPU) context that will allocate pinned memory.  (This is CPU
+// memory that's pinned for faster GPU memory transfers).  May or may not
+// return the same value as ::k2::GetCpuContext()... this is so, for instance,
+// if you have a GPU PyTorch context you can get a CPU PyTorch context.
 // Returns a (CPU) context that will allocate pinned memory.
 ContextPtr GetPinnedContext();
```
danpovey, line 329: I'm thinking we could rename this to `GetHostContext()`, and say: if `this` is a CPU context it will just return a pointer to `this`, but if `this` is a GPU context it will return a pinned context from which we can do asynchronous transfers.

danpovey:
Your suggestion about the CopyData() think is reasonable, IMO.
I am thinking it could be named CopyDataFrom(), for greater clarity.

One possibility is that the CopyDataFrom() if copying from CPU non-pinned to GPU pinned, could allocate
pinned memory and use it as a temporary buffer (copy to it), then do asynchronous transfer from the pinned 
memory.

k2/csrc/context.h
csukuangfj: `GetPinnedContext ` is not a member method so there is no `this` pointer. The returned context's GetDeviceType()
always returns `kCPU` and it is compatible with contexts returned from `GetCpuContext()`.

Do we still need to rename it to `GetHostContext()`?  I think there are confusions between `GetCpuContext()`
and `GetHostContext()`. 

From #451 
> In addition to the contexts returned by GetCpuContext() and GetCudaContext(), we need a 3rd type of context which would return pinned (page-locked) memory. Call this function GetPinnedContext().
(We need to remove the member-function of the same name from class Context).

danpovey: Mm.  If it's non-member, GetPinnedContext makes the most sense as a function name.
We'll need a not-too-wordy mechanism to do the logic of:
If the input is CUDA, give a pinned context, else give a CPU context.
which will commonly be needed.

k2/csrc/pytorch_context.cu
```diff
@@ -149,18 +192,53 @@ class PytorchPinnedContext : public Context {
...
 
 ContextPtr GetCpuContext() { return std::make_shared<PytorchCpuContext>(); }
 
-ContextPtr GetPinnedContext() {
-  std::call_once(has_cuda_init_flag, InitHasCuda);
-
-  if (has_cuda) return std::make_shared<PytorchPinnedContext>();
-
-  return GetCpuContext();
+ContextPtr GetPinnedContext(DeviceType device_type) {
```
danpovey, line 229: I would rather have a function GetPinnedContext() that returns a pinned context, and a function
GetContextForTransfer(DeviceType) that returns the appropriate one.

k2/csrc/pytorch_context.cu
```diff
@@ -149,18 +192,53 @@ class PytorchPinnedContext : public Context {
...
+                  void *dst) override {
+    DeviceType device_type = dst_context->GetDeviceType();
+    switch (device_type) {
+      case kCpu:
+        // we assume that src and dst do not overlap
+        memcpy(dst, src, num_bytes);
+        break;
+      case kCuda: {
+        // from host to device
+        c10::cuda::CUDAStream stream = c10::cuda::getCurrentCUDAStream();
```
qindazhu, line 205: just a question, @danpovey , which stream should we use here, the default stream or `dest->GetCudaStream()`?

k2/csrc/pytorch_context.cu
```diff
@@ -59,6 +60,26 @@ class PytorchCpuContext : public Context {
     return other.GetDeviceType() == kCpu;
   }
 
+  void CopyDataTo(size_t num_bytes, const void *src, ContextPtr dst_context,
+                  void *dst) override {
```
danpovey, line 64: This doesn't make sense to me, I thought we were taling about a member function of class Context.
How do we know where the source data is located?

danpovey: oh I see.. its in the class.

danpovey:
> *@qindazhu* commented on this pull request.
> ------------------------------
>
> In k2/csrc/pytorch_context.cu
> <https://github.com/k2-fsa/k2/pull/464#discussion_r537322308>:
>
> > @@ -149,18 +192,53 @@ class PytorchPinnedContext : public Context {
>      return other.GetDeviceType() == kCpu;
>    }
>
> +  void CopyDataTo(size_t num_bytes, const void *src, ContextPtr dst_context,
> +                  void *dst) override {
> +    DeviceType device_type = dst_context->GetDeviceType();
> +    switch (device_type) {
> +      case kCpu:
> +        // we assume that src and dst do not overlap
> +        memcpy(dst, src, num_bytes);
> +        break;
> +      case kCuda: {
> +        // from host to device
> +        c10::cuda::CUDAStream stream = c10::cuda::getCurrentCUDAStream();
>
> just a question, @danpovey <https://github.com/danpovey> , which stream
> should we use here, the default stream or dest->GetCudaStream()?
>

Should be using dest->GetCudaStream().  Otherwise it won't behave correctly
if we're using `child streams'.

k2/csrc/context.h
```diff
@@ -321,12 +347,17 @@ ContextPtr GetCpuContext();
...
-ContextPtr GetPinnedContext();
+/* Returns a (CPU) context that will allocate pinned memory.  (This is CPU
+   memory that's pinned for faster GPU memory transfers).  May or may not
+   return the same value as ::k2::GetCpuContext()... this is so, for instance,
+   if you have a GPU PyTorch context you can get a CPU PyTorch context.
+   Returns a (CPU) context that will allocate pinned memory.
+
+   @param [in] device_type  If device_type is kCpu, it is equivalent
+                            to `GetCpuContext()`. If device_type is kCuda,
+                            it returns a context that allocates pinned memory.
```
qindazhu, line 358: if `has_cuda` return false, it will return CPU context? maybe it's better to add documentation about this.

k2/csrc/pytorch_context.cu
csukuangfj: How are child streams created? I cannot find any code about child stream creation.
There are only several places in the documentation mentioning child streams.

danpovey:
When ParallelRunner is used and somone calls NewStream().
There is no notion of child stream in CUDA, but the way we sync them in k2,
they have a child/parent relationship.

k2/csrc/pytorch_context.cu
csukuangfj: If we are going to use `dest->GetCudaStream()`, then we have to modify the interface `GetCudaStream()`.

https://github.com/k2-fsa/k2/blob/6854bbb8ff85c5da8b4aced629344299917489d2/k2/csrc/pytorch_context.cu#L205-L213

`THCCachingHostAllocator_recordEvent` requires an argument of type `CUDAStream`, which is a wrapper
around `cudaStream_t` with some other information, e.g., the priority of the stream and the device
associated with the stream. 

`CUDAStream` can be converted to `cudaStream_t` implictly, but the reverse is impossible.

I would propose to change `cudaStream_t GetCudaStream() const` to

```cpp
cudaStream_t GetCudaStream(void *pstream) const override {
   *(c10::cuda::CUDAStream*)pstream = c10::cuda::getCurrentCUDAStream();
   return *(c10::cuda::CUDAStream*)pstream;
}
```
The default value of `pstream` is nullptr, so the existing code will not be affected.

k2/csrc/pytorch_context.cu
csukuangfj: `GetContextForTransfer` is added but I don't know where it will be called.
I think we can always use `GetPinnedContext()`.

k2/csrc/pytorch_context.cu
csukuangfj: > just a question, @danpovey , which stream should we use here, the default stream or dest->GetCudaStream()?


Fixed.

k2/csrc/context.h
csukuangfj: Added.

k2/csrc/pytorch_context.cu
danpovey: I don't like this design because we call GetCudaStream a lot, and it can't be optimized to be inline because it's virtual.  I'd prefer to have a separate function return this extra info (e.g. GetTorchCudaStream() maybe.. we don't have to have this be defined for most instances of class Context.

k2/csrc/pytorch_context.cu
danpovey: No, we don't want to use pinned memory if we are doing CPU-only computations.  It's bad for the computer (stops it from swapping).. can make the system unstable if you use too much.  It's only for when you want to copy something to the device.

k2/csrc/pytorch_context.cu
csukuangfj: > I don't like this design because we call GetCudaStream a lot, and it can't be optimized to be inline because it's virtual.

The existing `GetCudaStream()` is virtual. If we add `GetTorchCudaStream()`, both of them
will be virtual and cannot be inlined.

I will change it to `GetTorchCudaStream()`.

danpovey: Yes it will still be virtual, but the point is, we don't want to add overhead to all the *other* calls to GetCudaStream().

There is still a slight problem here that we can only return the base Torch stream in this way, not any stream that might have been created by class ParallelRunner.

danpovey: Incidentally, if all this function call GetCudaStream() is doing is calling that Torch function, and since it can only be invoked from the Torch pinned-context object, there is no point using it at all-- better to just invoke that Torch function directly that gets the stream.

k2/csrc/pytorch_context.cu
csukuangfj: I see, thanks.

k2/csrc/pytorch_context.cu
danpovey: You could perhaps find out whether there is a way to make Torch creaor provide additional streams when we need them, so we have this info.
Alternatively, we could ensure correctness by syncing the streams by means of events.  (Adds a bit of overhead, though).

csukuangfj: https://github.com/k2-fsa/k2/blob/0b055f07329cdca82d3f91c0b938803ba590644e/k2/csrc/pytorch_context.cu#L82-L85

The return value is from `g_stream_override.OverrideStream()`, which is its `stream_` that can be different from the one
returned from PyTorch.

danpovey: Yes, that is my concern, that for the streams which we created there is no way to convert them into Torch's notion of a stream.
It might be easier to always use our own allocator for pinned allocation, i.e. for the pinned allocators in the Torch and default cases to use the same underlying cached-allocator code (we would have to handle syncing by means of events).

csukuangfj:
ok, I have read the implementation of pinned allocator from PyTorch. Will write our own using PyTorch as a reference.






Monday, 7 December 2020, 20:22 +0800 from notifications@github.com  <notifications@github.com>:

k2/csrc/pytorch_context.cu
csukuangfj: > It might be easier to always use our own allocator for pinned allocation

A PinnedContext of our own is added (not tested yet).

k2/csrc/pytorch_context.cu
```diff
@@ -20,15 +20,22 @@
 
 namespace k2 {
 
+static std::once_flag has_cuda_init_flag;
+static bool has_cuda = false;
+static void InitHasCuda() {
```
qindazhu, line 25: It may be much more compact if we can put `has_cuda` and `has_cuda_init_flag` into `(Init)HasCuda` (and return bool, i.e. static `has_cuda`, in this function). Then we can just do `if (HasCuda())`. And it's better to add `inline` here?

csukuangfj: `HasCuda()` involves an extra function call. `inline` is just a suggestion to the compiler.
I remeber that we added `K2_LOG` inside `CudaStreamOverride::Push` and the function
was no longer inlined.

k2/csrc/pinned_context.cu
```diff
@@ -0,0 +1,343 @@
...
+  PinnedAllocator *allocator_;  // NOT owned here
+};
+
+ContextPtr GetPinnedContext() {
+  static std::once_flag has_cuda_init_flag;
+  static bool has_cuda = false;
+
+  std::call_once(has_cuda_init_flag, []() {
+    int32_t count = 0;
+    cudaError_t err = cudaGetDeviceCount(&count);
```
qindazhu, line 313: So what does it do in `torch::cuda::is_available` below, I just feel it may be duplicate with below `InitHasCuda` stuff?

csukuangfj: pinned_context does not depend on PyTorch. The implementation is similar to `torch::cuda::is_available`.

csukuangfj:
Ready for review.

k2/csrc/context.h
csukuangfj: Added `GetContextForTransfer`

k2/csrc/pytorch_context.cu
csukuangfj: Fixed by adding `GetContextForTransfer`

k2/csrc/pytorch_context.cu
csukuangfj: mark as resolved.

k2/csrc/context.cu
```diff
@@ -77,7 +77,7 @@ void ParallelRunnerActive::Finish() {
     for (std::size_t i = 0; i != streams_.size(); ++i) {
       // create and record event on `stream_[i]`, and wait on c_->GetCudaStream
       cudaEvent_t event;
-      auto ret = cudaEventCreate(&event);
+      auto ret = cudaEventCreateWithFlags(&event, cudaEventDisableTiming);
```
danpovey, line 80: can we have a comment to explain why we are using this flag?

csukuangfj: From https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EVENT.html#group__CUDART__EVENT_1g7b317e07ff385d85aa656204b971a042

> cudaEventDisableTiming: Specifies that the created event does not need to record timing data. Events created with this flag specified and the cudaEventBlockingSync flag not specified will provide the **best performance** when used with cudaStreamWaitEvent() and cudaEventQuery().

I think the documentation of `cudaEventCreateWithFlags` explains the flag clearly.

k2/csrc/pinned_context.cu
```diff
@@ -0,0 +1,343 @@
...
+                                the allocated memory.
+
+     @return Return cudaSuccess on success. Return a CUDA error code on failure.
+   */
+  cudaError_t Malloc(size_t size, void **ptr) {
+    NVTX_RANGE(K2_FUNC);
+    K2_CHECK_NE(ptr, nullptr);
+    std::lock_guard<std::mutex> lock(mutex_);
+
+    // ProcessEvents may free blocks
```
danpovey, line 85: I think it's best to delay processing these events, so that we don't repeatedly attempt to wait on events that are pending, which would be wasteful.  My suggestion:  put this code inside:
 ``` 
  for (int iter = 0; iter < 2; ++iter) {
      if (events_.size() > 100 || iter > 1) {
        cudaError_t err = ProcessEvents();
         if (err != cudaSuccess) return err;
     }
     // then search for smallest block..
  }
```

Comment (csukuangfj) [reply k2/csrc/pinned_context.cu:85]: Will update it.

Comment (csukuangfj) [reply k2/csrc/pinned_context.cu]: Fixed.

File: k2/csrc/pinned_context.cu [hunk 10]
```diff
@@ -0,0 +1,357 @@
...
+    // than 100, we will invoke `ProcessEvents()`, which
+    // may make some blocks available for reuse.
+    //
+    // If the number of cuda_events_ is small,
+    // we first try to find a block from the pool. If it
+    // cannot find one, we then invoke `ProcessEvents`.
+    //
+    // The purpose is to reduce the time of waiting for
+    // the pending events.
+    for (int32_t iter = 0; iter <= 2; ++iter) {
```
danpovey, line 95: Should be iter < 2, and "iter > 0" below....  as it is, iters 0 and 1 would be identical.

csukuangfj: thanks, fixed.

danpovey:
Before I merge, can we clarify the plan to use this?
We are going to change calls of MemoryCopy to calls of CopyTo() from the context object?
And instead of using GetCpuContext() when we need to transfer some data from CPU, we'll normally 
call GetContextForTransfer(c->GetDeviceType()) and use that as the CPU context, right?

csukuangfj:
I think pinned memory is useful for transferring data from host to device.

For snowfall, we need to move two kinds of things to GPU:

- log_prob, the output of the neural network, i.e., the DenseFsaVec. I think log_prob is already on GPU if
the model is running on GPU, so I think we don't need PinnedContext in this case

- the decoding graph, which is constructed from a list of LinearFsas. Initially, the information is on the host
and we need to move it to the device to construct LienarFsas. If the graph is large, I think we can use PinnedContext
in this case.

I cannot think of other usages of PinnedContext in k2.

danpovey:
No, there are lots of places where we transfer small-ish vectors of metadata.  Look for places where we either call GetCpuContext() and create an Array1 with it, and later do .To(c) to transfer i to a device context, e.g. in fsa.cu:
`  num_states_array = num_states_array.To(GetCpuContext());`
 or (more common) initialize an Array1 from std::vector.  This constructor should do the transfer via pinned memory if it's to GPU.

danpovey:
.. we'll capture most of these cases by modifying Array1's constructor from std::vector.

csukuangfj:
https://discuss.pytorch.org/t/when-to-set-pin-memory-to-true/19723

The above post says it is not always good to use pinned memory for data transfer. I don't have
much experience in CUDA and don't know under which circumstances it is correct.

danpovey:
Also to make this functional I think you'd have to replace all instances of Memcpy with CopyDataTo().
I think it would be a good idea to actually do this, and run the tests.  
But let me merge this now so we reduce the chance of conflicts.

csukuangfj:
> Also to make this functional I think you'd have to replace all instances of Memcpy with CopyDataTo().


Will do it.

danpovey:
We can test it, but I doubt there will be the problem, these are probably
PyTorch-specific issues.
My main concern is we shouldn't use pinned memory for large quantities of
data that won't be transferred very often to GPU (e.g. an entire training
set).

k2/csrc/pinned_context.cu
```diff
@@ -0,0 +1,357 @@
...
+    K2_CHECK_NE(ptr, nullptr);
+    std::lock_guard<std::mutex> lock(mutex_);
+
+    // If the number of outstanding cuda events is larger
+    // than 100, we will invoke `ProcessEvents()`, which
+    // may make some blocks available for reuse.
+    //
+    // If the number of cuda_events_ is small,
+    // we first try to find a block from the pool. If it
+    // cannot find one, we then invoke `ProcessEvents`.
```
qindazhu, line 91: `If it cannot find one, we then invoke ProcessEvents or create a new one?` seems there's a mismatch between code and documentation?

k2/csrc/pinned_context.cu
```diff
@@ -0,0 +1,357 @@
...
+  std::mutex mutex_;
+};
+
+static PinnedAllocator *GetPinnedAllocator() {
+  static std::once_flag init_flag;
+  static PinnedAllocator *allocator = nullptr;
+
+  std::call_once(init_flag, []() {
+    // it is never freed.
+    allocator = new PinnedAllocator;
```
qindazhu, line 264: also add documentation in the section of `valgrind` of INSTALL.md about this?

qindazhu:
Except `Array1(const std::vector<int32_t> &value)` and all `To` method (including `Array`, `Ragged`, `Tensor`）, I just think of some other places we may need pinned memory:

- `Array.CopyFrom` is dst is on GPU
- `Array1(ContextPtr ctx, int32_t size, T elem)` and of course `Array1.operator==(T t)`
- istream code for Array? (but it's fine as it may be used just in test code)

BTW, I just thought maybe we can even cache a gloabl GPU array for some usages? e.g. we always need `Array1<int32_t>(c, 1)` when doing some property/status check, it may be much more efficient if we can get an address from the global GPU memory cache instead of creating a new one for such usages? @danpovey what do you think about this?

k2/csrc/pinned_context.cu
csukuangfj: The comment is for the `for` loop. It does not create a new block in the `for` loop.
The `for` loop will try to reuse an existing block. But if it cannot find one, we just exit the `for` loop.
Statements after the for loop will allocate a new block.

csukuangfj:
> we always need Array1<int32_t>(c, 1)

I would suggest `Array1<int32_t>(GetPinnedContext(), 1)` in this case.

danpovey:
@qindazhu : regarding `Array1(ContextPtr ctx, int32_t size, T elem)` and `Array1.operator==(T t)`, we don't actually need pinned memory here because the T is a function argument so the asynchronous processing already happens by default.

Regarding `Array1<int32_t>(c, 1)` for status checks: I'm not quite sure what you are saying here.  Firstly, if you call `Array1<int32_t>(c, 1)` it will of course be using the caching allocator (not big overhead).  @csukuangfj I don't think you're right about using GetPinnedContext() in these kinds of situations either... firstly, we'd want the address on GPU if it's on GPU.   Secondly, in situations where we request a CPU context, we don't want to just convert that to a request pinned context because it would require to initialize the CUDA library which is wasteful if the context of the data we're processing was not on GPU in the first place: so we want to be calling GetContextForTransfer(device_type) or GetHostContext(device_type) or whatever it was we called it.

k2/csrc/pinned_context.cu
csukuangfj: I am wondering about the value of `valgrind` in k2. 

It shows lots of warnings from `libcuda.so` and we can do nothing about them.
`valgrind` only shows that there are some memory being leaked.
 
What should we document in INSTALL.md?

PR #469, opened 2020-12-08
Now Fsa.as_dict() saves tensor attributes as well as non-tensor attributes

csukuangfj:
Ready for review.

### DIFF
--- a/k2/python/k2/fsa.py
+++ b/k2/python/k2/fsa.py
@@ -83,7 +83,7 @@ class Fsa(object):
       and ``aux_labels`` are instances of ``torch.Tensor``.
 
       Implementation note: most of this class's attributes are not
-      real attributes in the objcet's dict; the real attributes are
+      real attributes in the object's dict; the real attributes are
       'arcs', '_non_tensor_attr', '_tensor_attr', '_properties',
       '_cache'.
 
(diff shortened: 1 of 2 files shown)


k2/python/k2/fsa.py
```diff
@@ -193,25 +193,34 @@ def __init__(self,
...
-        ans += "properties_str = " + _k2.fsa_properties_as_str(
-            self._properties) + "."
+        ans += 'properties_str = ' + _k2.fsa_properties_as_str(
+            self._properties) + '.'
+        for name, value in self.named_tensor_attr(include_scores=False):
+            sep = '\n'
+            ans += f'{sep}{name}: {value}'
+        for name, value in self.named_non_tensor_attr():
+            sep = '\n'
+            ans += f'{sep}{name}: {value}'
```
danpovey, line 222: mm I'm concerned if we us symbol tables a lot, this could be too verbose.

danpovey:
Looks OK to me.

k2/python/k2/fsa.py
csukuangfj: Ok, I will remove the symbol table from the output.

csukuangfj:
Merging.

PR #476, opened 2020-12-08
support creating LinearFsa(s) on a given device.

csukuangfj:


k2/python/tests/linear_fsa_test.py
```diff
@@ -35,32 +42,40 @@ def test_fsa_vec(self):
...
-        assert torch.allclose(
-            fsa.scores,
-            torch.zeros(num_symbols + len(symbols), dtype=torch.float32))
+
+        devices = [torch.device('cpu')]
+        if torch.cuda.is_available():
+            devices.append(torch.device('cuda', 0))
+
+        for device in devices:
+            fsa = k2.linear_fsa(symbols, device)
```
csukuangfj, line 51: @pzelasko
It is possible to convert a list of list of symbols to LinearFsa**s** on GPU,
which I think is faster than creating them one by one on CPU.

pzelasko: got it - so you're suggesting to convert the list of strings into an FsaVec, and then iterate over the FsaVec to cache each Fsa independently?

csukuangfj: I don't know what role LRU cache plays here. At least, the graphs can be created and cached on GPU.

Not sure whether `k2.linear_fsa` is faster than `k2.create_fsa_vec`. If `k2.linear_fsa` is faster,
we may not need the cache.

danpovey:
The slow part is intersection, because it has to be done on CPU.

pzelasko:
If it has to be done on CPU, then is there a point to creating the linear FSAs on the GPU?

qindazhu:
> If it has to be done on CPU, then is there a point to creating the linear FSAs on the GPU?

I doubt if this approach would be faster or not, if we create tons of linear Fsa on the GPU, we are actually doing tons of HOST to DEVICE memory transfer? I was thinking create them on CPU and construct an FsaVec, then transfering the FsaVec to GPU at once may be faster. Of course the benfit here depends on how many epoch we run, but I think at least we should run some benchmark on this?

csukuangfj:
> if we create tons of linear Fsa on the GPU, we are actually doing tons of HOST to DEVICE memory transfer?

Actually, there is only a single transfer per FsaVec.  
The more FSAs in a FsaVec, the less memory transfer is needed.
Constructions of LinearFsas on GPU can be done in parallel.

qindazhu:
Oh, I see, there is a version of  `LinearFsa` which accepts `Ragged` as input parameter (list of list of symbols).

qindazhu:
I mean, we are still construct the Ragged<int32_t> from `list of list of symbols` on CPU (and then tranfer to GPU), so I would be fine.

danpovey:
There is actually no point in transferring the data to GPU until after the intersection, because it will have to be transferred back.
In fact,  Intersect() will just crash if the data was on GPU.

qindazhu:
> There is actually no point in transferring the data to GPU until after the intersection, because it will have to be transferred back.
> In fact, Intersect() will just crash if the data was on GPU.

Can't agree anymore.

>  if we create tons of linear Fsa on the GPU, we are actually doing tons of HOST to DEVICE memory transfer? I was thinking create them on CPU and construct an FsaVec, then transfering the FsaVec to GPU at once may be faster.  Of course the benfit here depends on how many epoch we run but I think at least we should run some benchmark on this?

however, my comments above are about `create_fsa_vec`  in below code 


`decoding_graphs = k2.create_fsa_vec([self.compile_one_and_cache(text) for text in texts])` 

where `[self.compile_one_and_cache(text) for text in texts]` is cached decoding graph (on CPU, have been intersected with L) for each text, my point is we can just keep them as-is (on CPU) instead of keep them on GPU,  and after `k2.create_fsa_vec` we transfer the resulted FsaVec to GPU at once. (But I'm not sure, if we run many epochs, caching each decoding graph on GPU before training may get some gain as we don't need do trasfering to GPU every time after `create_fsa_vec`?)

danpovey:
Mm I see your point, doing individual transfers to the GPU will be slower than single large transfers.  So having compile_one_and_cache(text) return something on CPU and then having the caller transfer the FsaVec to GPU would possibly be faster.   (And also more robust when datasets get large).

csukuangfj:
> There is actually no point in transferring the data to GPU until after the intersection, because it will have to be transferred back.

Can it happen that the time of constructing  LinearFsas on GPU is less than the sum of the following time:
- host->device
- device->host
- construct the LinearFsas on CPU
- k2.create_fsa_vec

danpovey:
Very doubtful; host/device transfer is very slow, much slower than a simple
copy operation on CPU.

qindazhu:
BTW, we cannot actually create `LinearFsas` for a batch of `texts` and cache the result `LinearFsas` (even if we don't consider intersection we need to do later is on CPU), we can only create `Linear` for each `text` on them, because we may do shuffle between epoch.

csukuangfj:
Will close this pull-request since LinearFsas will be constructed on CPU.

PR #477, opened 2020-12-09
Replace MemoryCopy with CopyDataTo.

csukuangfj:
~~WIP~~


csukuangfj:
CopyDataTo and MemoryCopy are equivalent in the following cases:
- host -> host
- device -> device
- device -> host

For `host -> device`, `CopyDataTo`  will use pinned-memory as an intermediate buffer.

PR #480, opened 2020-12-09
support torch.device in string form.

### LINKED ISSUE #479: torch devices vs. strings
This line of code:
`  best_paths = best_paths.to('cpu') `
in snowfall is failing with:
```
  File "/ceph-dan/.local/lib/python3.8/site-packages/k2/fsa.py", line 694, in to
    assert device.type in ('cpu', 'cuda')
AttributeError: 'str' object has no attribute 'type'
```
Wouldn't it be better if, when we expect a torch.device, we construct a torch.device from the thing, as in
torch.device(device), so that it still works if a string is provided?  This seems to me more consistent with the behavior of torch.

csukuangfj:
Closes #479


danpovey:
thanks!

PR #482, opened 2020-12-09
Implemented ComputeEpsilonClosure

qindazhu:


danpovey:
Great work!!!
I'll merge this, but don't forget to add random tests!

danpovey:
Also you might want to test with simple examples that it does what we intended about combining epsilons with either the following or preceding arcs depending what would be more efficient.

qindazhu:
Definitely yes, I have 3 functions needed to be add random tests in rm_epsilon, `MapFsaVecStates`, `ComputeEpsilonClosure(Iter)`, will do later). Thanks

qindazhu:
> we intended about combining epsilons with either the following or preceding arcs depending what would be more efficient.

Yeah, but actually I have not run `RemoveEpsilonsIterativeTropical`.. will do later.

qindazhu:
Merging..

PR #483, opened 2020-12-09
import torch in __init__.py

qindazhu:


k2/python/k2/__init__.py
```diff
@@ -1,3 +1,4 @@
+import torch
```
csukuangfj, line 1:
```python
import torch  # noqa
```
to make `flake8` happy.

qindazhu: updated, thanks

csukuangfj:
+2

PR #484, opened 2020-12-09
Add benchmark for k2 APIs.

csukuangfj:
TODOs

- [x] Support running specified benchmarks
- [x] Print system information (e.g., CPU & GPU info) and date
~~- [ ] Support pretty output~~


k2/csrc/array_ops_benchmark.cu
```diff
@@ -0,0 +1,36 @@
+/**
+ * @brief Benchmarks for k2 APIs.
```
qindazhu, line 2: not sure if we'd like put those files under `csrc`, maybe a folder like `benchmark`. I just feel compared to `**_test.cu`, those benchmark files would not be read/used by normal users.

k2/csrc/array_ops_benchmark.cu
```diff
@@ -0,0 +1,36 @@
...
+    timer.Reset();
+    Array1<T> dst = ExclusiveSum(src);
+    state.SetIterationTime(timer.Elapsed());  // in seconds
+    num_bytes_processed += dim * sizeof(T);
+  }
+  state.SetBytesProcessed(num_bytes_processed);
+}
+
+BENCHMARK_TEMPLATE(BM_ExclusiveSum, int32_t)
+    ->Range(8, 8 << 10)
```
qindazhu, line 33: Compared to use those `marco` (or `function`?) from google benchmark, I prefer to declare problem sizes manually here, so the reader can easily know what problem sizes we'll run.

(I have to check the documentation of google benchmark to know what this line means: `By default the arguments in the range are generated in multiples of eight and the command above selects [ 8, 64, 512, 4k, 8k ]`)

qindazhu:
Just a question, with output below, how many CPUs do one benchmark (e.g. `ExclusiveSum<int32_t>/8k`) use? I guess the benchmark library takes very few CPU time, right?
```
Running ./bin/array_ops_benchmark
Run on (72 X 2600 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x36)
  L1 Instruction 32 KiB (x36)
  L2 Unified 1024 KiB (x36)
  L3 Unified 25344 KiB (x2)
Load Average: 31.42, 28.01, 27.58
```

k2/csrc/array_ops_benchmark.cu
csukuangfj: You can find the problem sizes at the output:
```
BM_ExclusiveSum<int32_t>/8/manual_time 
BM_ExclusiveSum<int32_t>/64/manual_time 
BM_ExclusiveSum<int32_t>/512/manual_time 
BM_ExclusiveSum<int32_t>/4096/manual_time 
BM_ExclusiveSum<int32_t>/8192/manual_time
```

`8->64->512->4096->8192`.

csukuangfj:
> Just a question, with output below, how many CPUs do one benchmark 

By default, it is single-threaded, I think.

To use multiple threads, you have to specify explicitly the number of threads you need

k2/csrc/array_ops_benchmark.cu
qindazhu: of course we can as you have pasted the output here. But what I mean is: reader should be able to figure out the sizes by just reading code instead of running code...

csukuangfj: `Range` is to free us specifying the sizes exhaustively. What do you think are the sutiable sizes that should be written down manually?

qindazhu: I'd like to add `100, 500,1000, 2000, 5000, 10000` etc here, maybe +m whre m is a small random number which can be negative or positive. We need to run multiple times for a particular size and get the average time (e.g. with the random number `m`, for `size=1000`, we run 5 times, we get the result of size = 999, size == 1002, size = 1001, size= 997, size = 1001, and we compute the average time)

Another point here is: we should not just run benchmark on those 2^n sizes (that is you did here with `Range`), they may be faster than real normal case (non 2^n sizes). 

 @danpovey as he may have some more insights.

qindazhu: BTW, for kernels performace tests, we usually add a `warm-up` before the real benchmark starting in case the first kernel lanuch takes longer time. e.g. 
```
mykenerl(); // just a warm-up
Time t;
for(int32_t i = 0; i != 10; ++i){
  mykenerl();
}
t.Elapsed()
```

But I'm not sure if we need to do this for our op testings. Maybe @danpovey can give some comments here.

danpovey:
Sounds like a good idea (the first-launch thing).

k2/csrc/array_ops_benchmark.cu
csukuangfj: >  But what I mean is: reader should be able to figure out the sizes by just reading code instead of running code

Users that have used `google/benchmark` before should be able to figure out the sizes from `Range`.
For users who have never used `google/bechmark`, I recommend them to refer to the documentation
and code of `google/benchmark`.

qindazhu: mm, you're just trying to ask our code readers learn more nonessential tools....Anyway, it's up to you, but I believe those 2^n sizes you set are definitely not right for a typical benchmark running (especially for cuda kenerls)

danpovey: @csukuangfj I don't like to assume people read the documentation of projects we require because:
  - in most cases they won't have
  - we're probably not providing a pointer
  - there may be a lot of documentation and it's too much reading to be able to figure out what's needed for our stuff.
I'd rather provide some brief comments in our code to help people understand what's going on.

csukuangfj: > I'd like to add 100, 500,1000, 2000, 5000, 10000 etc here, maybe +m whre m is a small random number which can be negative or positive

If random numbers come  into scope, will it be difficult to compare benchmark results in different settings, i.e.,
release vs debug?

> We need to run multiple times for a particular size and get the average time (e.g. with the random number m, for size=1000, we run 5 times, we get the result of size = 999, size == 1002, size = 1001, size= 997, size = 1001, and we compute the average time

You can see from the output posted above
```
----------------------------------------------------------------------------------------------------
Benchmark                                          Time             CPU   Iterations
----------------------------------------------------------------------------------------------------
BM_ExclusiveSum<int32_t>/8/manual_time         15451 ns        38989 ns        43121
```

It is the average of `43121` iterations. The number of iterations needed is computed automatically
by google/benchmark.


> BTW, for kernels performace tests, we usually add a warm-up before the real benchmark starting in case the first kernel lanuch takes longer time. e.g.

The results are the average of thousands of iterations. Anyway, warm up is added as you suggested.

csukuangfj: >  but I believe those 2^n sizes you set are definitely not right for a typical benchmark running (especially for cuda kenerls)

Can you explain why the sizes of power of two are.NOT.right for benchmarks? I think it can show how the algorithm
scales with the increasing number of items in logarithmic scale.

csukuangfj: > I don't like to assume people read the documentation of projects we require because:

I agree, but the point is the relevant code is about implementation details. If the user/developer is
really interested in it, he/she can read the API documentation.

We are using `cub::DeviceScan::ExclusiveSum` in k2, but I think we only add documentation
to `k2::ExclusiveSum`. If someone is interested in `cub::DeviceScan::ExclusiveSum`, he/she
should try to find its doc provided by cub.

qindazhu: > > but I believe those 2^n sizes you set are definitely not right for a typical benchmark running (especially for cuda kenerls)
> 
> Can you explain why the sizes of power of two are.NOT.right for benchmarks? I think it can show how the algorithm
> scales with the increasing number of items in logarithmic scale.

Well, Cuda does computation with `warps`  instead of a single thread, normally a warp of thread is 16 or 32 (depends on CUDA ARCH, IIRC). If we are using 2^n sizes, actually, there may be no idle thread, this is not real for most normal cases. Besides, we always to start block with size of 2^m (e.g. 128, 256) to leverage cuda arch (IIRC, apply for such block will be faster than non 2^n sizes, according to CUDA arch), testing with 2^n problems mean we may get the highest performace, this is not reasonal for a typical benchmark running.

csukuangfj: I see, thanks.

Then the problem was that I used 8 and its default multiplier `8` in `Range(8, 8 << 10)` .
`Range` would be fine here if we change its default multiplier.

qindazhu: BTW, IIRC ,block size should be always a round multiple of the warp size (e.g. 32); otherwise, we are wasting some computation resources as cuda always make it to be a multiple of 32.

qindazhu: > I see, thanks.
> 
> Then the problem was that I used 8 and its default multiplier `8` in `Range(8, 8 << 10)` .
> `Range` would be fine here if we change its default multiplier.

mm, of course we can, but I still prefer to add some random numbers here, the disadvantage of `Range` stuff here is: no matter what `multiplier` we use, we are still testing with a particular distribution of data (thinking about `size % 32==warp_size`). But as you said, it's hard to do this in google benchmark. so I'm not sure about this (if google benchmark cannot, should we still use it for benchmark or not? or we just don't use some fixed problem sizes) @danpovey any suggestions?

danpovey:
Perhaps there's an easy way, like: 10, 10 << 10  (?)

qindazhu:
OK, so we just use some fixed sizes for benchmark running,right? (no random numbers, e,g. to test size=1000, we run tons of iteration with 997, 998, 1002, 1001, etc. and get the average time)

danpovey:
Sure, that sounds good, it would help to detect kernel problems.
But only if it doesn't conflict with some aspect of the benchmark library
that I don't understand, e.g. regarding finding the number of repeats,
or would make the code too ugly.

csukuangfj:
The number of iterations is now set to a fixed number `10000` and a rand number in the range (-2, 2) is added
to the number of elements to avoid power of two. Hope that the randomness will not affect the results when we have
tens of thousands of elements.

The output looks like the following:

```
------------------------------------------------------------------------------------
Benchmark                                          Time             CPU   Iterations
------------------------------------------------------------------------------------
ExclusiveSum_int32/10/iterations:10000         0.040 ms        0.040 ms        10000 3
ExclusiveSum_int32/100/iterations:10000        0.040 ms        0.040 ms        10000 2
ExclusiveSum_int32/1000/iterations:10000       0.065 ms        0.064 ms        10000 1
ExclusiveSum_int32/10000/iterations:10000      0.321 ms        0.319 ms        10000 -2
ExclusiveSum_int32/10240/iterations:10000      0.324 ms        0.323 ms        10000 1
ExclusiveSum_float/10/iterations:10000         0.038 ms        0.038 ms        10000 -2
ExclusiveSum_float/100/iterations:10000        0.040 ms        0.040 ms        10000 -2
ExclusiveSum_float/1000/iterations:10000       0.048 ms        0.048 ms        10000 -1
ExclusiveSum_float/10000/iterations:10000      0.158 ms        0.158 ms        10000 3
ExclusiveSum_float/10240/iterations:10000      0.166 ms        0.165 ms        10000 3
```

The last column is the random number added to the number of elements in benchmarking.

danpovey:
I don't think it's a good idea to fix the number of iterations; this will end up being a hassle to maintain as some of them will get too slow and we'll have to do work to find out which ones they are and fix the tests.  There's no way to make it take a fixed time instead?

Simplifying things a bit (e.g. getting rid of something you previously changed) might be OK.

qindazhu:
Amazing, +2!

 just a suggestion, 10000 iterations may be too large ( I guess we may run too long time for some complex OPs), I was thinking 100 iteration is good enough for our purpose.

qindazhu:
> There's no way to make it take a fixed time instead

@danpovey, I didn't get your idea here, what's the meaning of running a benchmark with a fixed time? I was thinking different Ops will take different time to run one iteration, so I thought we can just run (e.g. 10-100) iterations for each ops and get the average time.

danpovey:
I think 10ms would be a better cutoff than a fixed num iterations.

Also: you have to be careful of calling timer.Elapsed() every iteration.  That's a system call, I believe, it can take a significant amount of time.

danpovey:
@qindazhu I'm concerned that if we just have fixed num-iters:
  - for fast things the time increment may be to small to be accurately measured
  - for slow things it will be too slow.

If we don't want to break after a certain amount of time, it might be better to do something like:
  num_iters = min(1000, 10000000 / problem_size)
(ignore the specific numbers , that might not be well thought out)

csukuangfj:
> I don't think it's a good idea to fix the number of iterations

>  I was thinking 100 iteration is good enough for our purpose.

If we don't set the number of iterations, google/bechmark will compute the number of iterations given
how time-consuming the OP is within a timing budget. The faster the OP executes, the more
the number of iterations it will be run by google/bechmark.

I am fine to change it to a fixed number or let google/benchmark computes it.

csukuangfj:
I made some changes so that the benchmark will run for at least 1 second.

```
----------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations
----------------------------------------------------------------------------------
ExclusiveSum_int32/10/min_time:1.000         0.039 ms        0.039 ms        35312 2
ExclusiveSum_int32/100/min_time:1.000        0.039 ms        0.039 ms        36182 2
ExclusiveSum_int32/1000/min_time:1.000       0.063 ms        0.063 ms        22370 1
ExclusiveSum_int32/10000/min_time:1.000      0.321 ms        0.319 ms         4496 1
ExclusiveSum_int32/10240/min_time:1.000      0.321 ms        0.320 ms         4268 3
ExclusiveSum_float/10/min_time:1.000         0.039 ms        0.038 ms        38305 2
ExclusiveSum_float/100/min_time:1.000        0.037 ms        0.037 ms        38064 3
ExclusiveSum_float/1000/min_time:1.000       0.047 ms        0.047 ms        29997 -1
ExclusiveSum_float/10000/min_time:1.000      0.161 ms        0.159 ms         8747 3
ExclusiveSum_float/10240/min_time:1.000      0.161 ms        0.159 ms         8676 3
```

You can see that the larger the problem size is, the fewer iterations it will run.

qindazhu:
>  you have to be careful of calling timer.Elapsed() every iteration

yeah, I was thinking the main issue here is calling `Elapsed` every time,  I prefer to call it after all iterations and compute the average time as I did here (though it is a very rough code as that time I was learning cuda programming)

https://github.com/k2-fsa/k2/blob/1d328494367efd3f8341f6f300d66c799ef1894c/k2/csrc/array_ops_test.cu#L78-L82

> for fast things the time increment may be to small to be accurately measured
> for slow things it will be too slow.

`Transpose` (Array not ragged) is a fast algorithm, IIRC, each time I run is relatively stable. For slow Op, I agree with your formulation of  `num_iters = min(1000, 10000000 / problem_size)` (except I thought 100 iterations would be fine for most algorithms)

k2/csrc/array_ops_benchmark.cu
csukuangfj: I am fine with where to put these files.

danpovey:
I think 1 second is way too long, we want these tests to finish fast so we can easily compare (there will be hundreds).  I'd go for 10 ms to 100ms.
I am a little concerned about the effects of using that ElapsedTime() inside the loop:
  - For CPU this is redundant as I think google benchmark does this itself, and getting the time can be slow.
  - For GPU, I think the call to ElapsedTime() will probably wait on whatever the GPU is doing, so for throughput-oriented tests this is not good; we'll mostly be measuring the time to sync the GPU, rather than the operation itself.  If we wanted to use the gpu's ElapsedTime(), the only sensible way to do it would be to call it only after the loop was done.

csukuangfj:
From https://github.com/google/benchmark/blob/bf585a2789e30585b4e3ce6baf11ef2750b54677/include/benchmark/benchmark.h#L548

```
  // REQUIRES: called exactly once per iteration of the benchmarking loop.
  // Set the manually measured time for this benchmark iteration, which
  // is used instead of automatically measured time if UseManualTime() was
  // specified.
  //
  // For threaded benchmarks the final value will be set to the largest
  // reported values.
  void SetIterationTime(double seconds);
```

`SetIterationTime` is expected to be called per iteration.

https://github.com/google/benchmark/blob/bf585a2789e30585b4e3ce6baf11ef2750b54677/include/benchmark/benchmark.h#L903

```
  // If a benchmark must measure time manually (e.g. if GPU execution time is
  // being
  // measured), call this method. If called, each benchmark iteration should
  // call
  // SetIterationTime(seconds) to report the measured time, which will be used
  // to control how many iterations are run, and in the printing of items/second
  // or MB/second values.
  Benchmark* UseManualTime();
```

I guess we need to set the number of iterations if we call `SetIterationTime` after the loop.
Need to dig into the code to learn more.

danpovey:
Here's an idea: we don't use ElapsedTime(), but we synchronize the GPU every n iterations, where n is chosen so that the GPU-syncing does not dominate the time taken.  For example, synchronize every n iterations where n is the minimum of (500,  K / num_items)   where e.g. K = 1000000 or something, and  num_items is the problem size (e.g. num elements in the array).
Perhaps could add a call context->Synchronize() ?  Or just use whatever code we use now to synchronize GPU.

qindazhu:
I guess for benchmark code, we always starts from the default stream, so I suggest we just keep it simper here, i.e. call `time.ElapsedTime` only once after the loop (which will wait on the default stream), as this example shows https://developer.nvidia.com/blog/how-implement-performance-metrics-cuda-cc/

csukuangfj:
I feel that it will be easier if we write our own benchmark framework and output in a format that can be
recognized by google/benchmark.

k2/csrc/array_ops_benchmark.cu
```diff
@@ -0,0 +1,94 @@
...
+  K2_CHECK_GT(num_iter, 0);
+
+  cudaEvent_t start;
+  cudaEvent_t stop;
+  cudaEventCreate(&start);
+  cudaEventCreate(&stop);
+  cudaStream_t stream = context->GetCudaStream();
+  cudaEventRecord(start, stream);
+  for (int32_t i = 0; i != num_iter; ++i) {
+    DoNotOptimize(ExclusiveSum(src));
```
csukuangfj, line 67: I feel the pattern is somewhat fixed and we can make it a template argument
or a function argument.

k2/csrc/array_ops_benchmark.cu
```diff
@@ -0,0 +1,98 @@
...
+// ------- from google/benchmark  ---- end ----
+
+/* Return the number of milliseconds per iteration on average.
+
+   @param  [in]  dim  Number of elements in Array1<T> for testing.
+   @return Return number of ms per iteration on avarage.
+ */
+
+template <typename T>
+using UnaryOp = Array1<T> (*)(const Array1<T> &);
```
csukuangfj, line 50: We can make it a variadic template that accepts an arbitrary number of arguments.

danpovey: Bear in mind that not all the functions we might want to call output their data via return values.
I'm not objecting, just pointing it out.

k2/csrc/array_ops_benchmark.cu
```diff
@@ -0,0 +1,98 @@
...
+  cudaEvent_t stop;
+  cudaEventCreate(&start);
+  cudaEventCreate(&stop);
+  cudaStream_t stream = context->GetCudaStream();
+  cudaEventRecord(start, stream);
+  for (int32_t i = 0; i != num_iter; ++i) {
+    DoNotOptimize(op(src));
+  }
+  cudaEventRecord(stop, stream);
+  cudaEventSynchronize(stop);
```
qindazhu, line 74: seems these code are exactly what our `Timer` class does?

danpovey: If so I'd prefer to use an instance of the CuTimer class instead of this code, it would be more readable.

danpovey: @csukuangfj I think using our Timer clas would be better thant his code, i.e. around line 64 do:
`Timer t(context);`
and at the end do
`return t.Elapsed();`
(and maybe change the calling code to accept seconds?)

csukuangfj: Sure, I am updating it.

k2/csrc/array_ops_benchmark.cu
```diff
@@ -0,0 +1,98 @@
...
+ * See LICENSE for clarification regarding multiple authors
+ */
+
+#include "k2/csrc/array_ops.h"
+#include "k2/csrc/log.h"
+#include "k2/csrc/timer.h"
+
+namespace k2 {
+
+// ------- from google/benchmark  ---- begin ----
```
danpovey, line 17: We should probably put this generic stuff in something like csrc/benchmark/benchmark.h, and I think it would be good to put the benchmarks in csrc/benchmark/array_ops_benchmark.cu.  To separate it from the other stuff.

And I think there should be a function to print out relevant info about the CPU and GPU, and the date and time, and maybe certain key compiler options (at least debug/release mode)  at the beginning of running the program- so we can keep track of this stuff over time.  What do you think?
I know we'll mostly be comparing this on the same machine before and after a change, but we may also want to make other kinds of comparisons and do various kinds of data analytics later on, e.g. to find problematic hardware.

csukuangfj: Yes, I agree.

csukuangfj: > And I think there should be a function to print out relevant info about the CPU and GPU, and the date and time,

I've printed some info about GPU. Not sure what kind of GPU information is needed.

k2/csrc/benchmark/array_ops_benchmark.cu
```diff
@@ -0,0 +1,64 @@
...
+#include <cstdlib>
+
+#include "k2/csrc/array_ops.h"
+#include "k2/csrc/benchmark/benchmark.h"
+
+namespace k2 {
+
+template <typename T>
+static BenchmarkStat BenchmarkExclusiveSum(int32_t dim) {
+  ContextPtr context = GetCudaContext();
```
danpovey, line 20: Don't we want to do the test for both Cuda and CPU?
Of course we could easily do this in separate processes (e.g. separate invocations of the benchmark program) which would help speed... dunno how easy it would be to set up and what the others think...

csukuangfj: Support for CPU is also added.

The current output format is CSV, which I think is very easy to parse. We can also output a format
that is compatible with google/benchmark. Don't know if you guys have some suggestions here.

csukuangfj:
Here is the output of this pullrequest:

```
k2/build$ ./bin/array_ops_benchmark
Mon Dec 14 13:01:50 2020

CUDA device name: Tesla V100-PCIE-32GB
Compute capability: 7.0
GPU clock freq: 1.38 GHz
Driver version: 10.2
Runtime version: 10.1
Warp size: 32
L2 cache size: 6 MB
Total global memory: 31.7485 GB
Total const memory: 64 KB
Total shared memory per block: 48 KB
Total shared memory per multiprocessor: 96 KB
ECC enabled: 1
Number of multiprocessors: 80
Number of CUDA cores: 64

ExclusiveSum_int32_kCpu_100,500,0.000000
ExclusiveSum_int32_kCpu_500,500,0.000001
ExclusiveSum_int32_kCpu_1000,500,0.000001
ExclusiveSum_int32_kCpu_2000,500,0.000001
ExclusiveSum_int32_kCpu_5000,200,0.000003
ExclusiveSum_int32_kCpu_10000,100,0.000007
ExclusiveSum_int32_kCpu_100000,10,0.000059
ExclusiveSum_int32_kCuda_100,500,0.000016
ExclusiveSum_int32_kCuda_500,500,0.000016
ExclusiveSum_int32_kCuda_1000,500,0.000016
ExclusiveSum_int32_kCuda_2000,500,0.000017
ExclusiveSum_int32_kCuda_5000,200,0.000014
ExclusiveSum_int32_kCuda_10000,100,0.000012
ExclusiveSum_int32_kCuda_100000,10,0.000011
```

For smaller problem sizes, CPU is not slower than CUDA.

danpovey:
Mm.  Could you print out some kind of legend so we can understand what the fields mean?
E.g. in 100,500,0.000000, what do the numbers represent?
The number of iterations being used seems to be too low (so we can't tell what the speed is), you may have to tweak some of the constants in the code.

danpovey:
Could you have all lines that aren't in csv format start with a special character like '#', so they can easily be parsed out?
Is it possible to have things like ExclusiveSum_int32_kCuda_1000  be printed out with commas instead of _ so they can be parsed as csv too?  Or are they function names?

danpovey:
... perhaps the issue with the time is that you are only printing out 6 significant digits (and those are times per iteration).
You could perhaps print out the time in milliseconds instead of seconds.

danpovey:
... or even microseconds.

csukuangfj:
Updated output:

```
# Mon Dec 14 13:58:26 2020

# CUDA device name: Tesla V100-PCIE-32GB
# Compute capability: 7.0
# GPU clock freq: 1.38 GHz
# Driver version: 10.2
# Runtime version: 10.1
# Warp size: 32
# L2 cache size: 6 MB
# Total global memory: 31.7485 GB
# Total const memory: 64 KB
# Total shared memory per block: 48 KB
# Total shared memory per multiprocessor: 96 KB
# ECC enabled: 1
# Number of multiprocessors: 80
# Number of CUDA cores: 64

op_name,dtype,device,problem_size,number_of_iterations,elapsed_us_per_iteration
ExclusiveSum,int32,kCpu,100,500,0.482082
ExclusiveSum,int32,kCpu,500,500,0.646114
ExclusiveSum,int32,kCpu,1000,500,0.892162
ExclusiveSum,int32,kCpu,2000,500,1.447678
ExclusiveSum,int32,kCpu,5000,200,2.850294
ExclusiveSum,int32,kCpu,10000,100,5.290508
ExclusiveSum,int32,kCpu,100000,10,51.784515
ExclusiveSum,int32,kCuda,100,500,12.491071
ExclusiveSum,int32,kCuda,500,500,12.378112
ExclusiveSum,int32,kCuda,1000,500,12.280128
ExclusiveSum,int32,kCuda,2000,500,12.514560
ExclusiveSum,int32,kCuda,5000,200,12.456161
ExclusiveSum,int32,kCuda,10000,100,12.446720
ExclusiveSum,int32,kCuda,100000,10,13.334400
```

danpovey:
Cool.  I'd be OK to merge at this point.

csukuangfj:
`RowSplitsToRowIds` is also added as a demonstration for how to add other operations.

```
op_name,dtype,device,problem_size,number_of_iterations,elapsed_us_per_iteration
ExclusiveSum,int32,kCpu,100,500,0.430107
ExclusiveSum,int32,kCpu,500,500,0.652313
ExclusiveSum,int32,kCpu,1000,500,0.862122
ExclusiveSum,int32,kCpu,2000,500,1.428127
ExclusiveSum,int32,kCpu,5000,200,2.830029
ExclusiveSum,int32,kCpu,10000,100,5.679131
ExclusiveSum,int32,kCpu,100000,10,57.196617
ExclusiveSum,int32,kCuda,100,500,15.522688
ExclusiveSum,int32,kCuda,500,500,13.503168
ExclusiveSum,int32,kCuda,1000,500,12.769920
ExclusiveSum,int32,kCuda,2000,500,12.796735
ExclusiveSum,int32,kCuda,5000,200,13.019680
ExclusiveSum,int32,kCuda,10000,100,10.883519
ExclusiveSum,int32,kCuda,100000,10,13.872001
RowSplitsToRowIds,int32,kCpu,100,500,6.326199
RowSplitsToRowIds,int32,kCpu,500,500,39.864063
RowSplitsToRowIds,int32,kCpu,1000,500,107.650284
RowSplitsToRowIds,int32,kCpu,2000,500,185.015686
RowSplitsToRowIds,int32,kCpu,5000,200,536.605164
RowSplitsToRowIds,int32,kCpu,10000,100,1650.300049
RowSplitsToRowIds,int32,kCpu,100000,10,51252.414062
RowSplitsToRowIds,int32,kCuda,100,500,27.510401
RowSplitsToRowIds,int32,kCuda,500,500,47.054462
RowSplitsToRowIds,int32,kCuda,1000,500,121.602242
RowSplitsToRowIds,int32,kCuda,2000,500,115.474808
RowSplitsToRowIds,int32,kCuda,5000,200,240.739502
RowSplitsToRowIds,int32,kCuda,10000,100,453.319977
RowSplitsToRowIds,int32,kCuda,100000,10,4515.548828
```

csukuangfj:
Merging if there are no other comments.

danpovey:
Great!  RowSplitsToRowIds() should not be anywhere near that slow.  It
could easily be implemented using a simple scan-type operator like
ExclusiveSum() is.  (It's likely this slow because I don't understand GPU
hardware well enough to write that kernel; actually the problem is easier
than
the exclusive-sum problem in typical cases.)

danpovey:
.. the way it could be implemented using a scan-type operator would be:
initialize the data using -1's, then have a kernel write the 1st element of
each row, and -1's for the rest; and then scan using max as the scan
operator.  It would require 3 kernels which may be overkill.

csukuangfj:
> .. the way it could be implemented using a scan-type operator would be

Trying to implement it and then to benchmark it.

PR #486, opened 2020-12-10
Implement some indexing-with-ragged operations

danpovey:


danpovey:
@qindazhu  I realized that ComposeArcMaps() is really a more general operation that it would be better to just call Index().
Could you replace that code with a call to this?

Also I added a function RemoveValuesLeq(ragged, cutoff) which will be useful for removing 0's and -1's from ragged tensors.

qindazhu:
Sure, merging (will update `ComposeArcMaps()` locally, I'm now working on converting those your comments in `RemoveEpsilonsIterativeTropical` to real code).

PR #488, opened 2020-12-11
Implemented left lambdas and code in RemoveEpsilonIterative, tested with simple case

qindazhu:
Those `ragged*` files are just style fixes, you can just ignore them when doing review.


k2/csrc/rm_epsilon.cu
```diff
@@ -586,7 +586,11 @@ void ComputeEpsilonClosureOneIter(FsaVec &epsilon_fsa, FsaVec *closure_fsa,
 
 void RemoveEpsilonsIterativeTropical(FsaVec &src_fsa, FsaVec *dest_fsa,
```
qindazhu, line 587: @danpovey , I just fell we do too many things in this function and the code may be hard to read for others. If you think so, I'll split it into a few functions, e.g. one to compute `epsilon_closure_mapped`, one to compute `epsilon_num_foll_arcs` and `epsilon_pre_renumbering`, one to compute `combined_foll`, one to compute `combined_prec`, one to get the result, etc. Of course not so urgent, I'll try to add random tests first.

k2/csrc/rm_epsilon_test.cu
```diff
@@ -318,4 +320,59 @@ TEST(RmEpsilon, ComputeEpsilonClosureSimple) {
...
+    int32_t p;
+    GetFsaVecBasicProperties(dest, &properties, &p);
+    EXPECT_EQ(p & kFsaPropertiesEpsilonFree, kFsaPropertiesEpsilonFree);
+    bool log_semiring = false;
+    float beam = std::numeric_limits<float>::infinity();
+    fsa_vec = fsa_vec.To(GetCpuContext());
+    dest = dest.To(GetCpuContext());
+    // as both fsa_vec and dest would be top-sorted, we can call
+    // IsRandEquivalent to check the result, but for now top-sorted Fsas, how
+    // can we check the result?
```
qindazhu, line 369: And here, any suggestion? our current implementation `IsRandEquivalent` requires the inputs are top-sorted as it will compute forward/backward weights in it. For non-top-sorted tests, I think we can call ` IsRandEquivalentUnweighted` to check the sequence, but have no idea now about checking the weights.

k2/csrc/rm_epsilon.cu
danpovey: ok with me...

k2/csrc/rm_epsilon_test.cu
danpovey: Mm, theres no fundamental reason why IsRandEquivalent should not be able to operate on non-top-sorted input I believe (?).  Would have to check.   Perhaps we chould do tests on top-sorted input to RmEpsilon, it will stay top-sorted.

danpovey:
conflicts..

danpovey:
Don't forget to fix the merge conflicts!

qindazhu:
Will do now

qindazhu:
wait,  seems there's some format issue, will update

k2/csrc/rm_epsilon_test.cu
qindazhu: as I said above, it's becasue currently we call `shortestDistance` in `IsRandEquivalent` to get the score of the sequence, which calls `computeForwardWeights` internally and then it requires the input top-sorted.

danpovey: I think I just merged a change so IsRandEquivalent will work on non-top-sortable input as long as log_semiring==false.

qindazhu: sorry, I just tried to find but seems there's no such one，can you paste the link? BTW, so if there's a cyclic, how do you compute the forward weights (even for tropical), using relax variable stuff?

danpovey: It's a change to the existing algorithm.
It uses a queue; you re-add something to the queue if the score changes.

PR #489, opened 2020-12-11
Add hashing code

danpovey:


danpovey:
still working on this, test fails because it's not finished.

qindazhu:
Dan, just be curious where it would be used in a short time, determinize?

danpovey:
OK, this works, I'm going to merge it.
The next stage is that I modify our IntersectDensePruned code so it can use hashes, to avoid the out-of-memory error that I previously encountered when trying to decode.  I should finish that to-morrow.

@qindazhu when that is finished, assuming you have been able to finish the things we discussed regarding letting the aux_labels be sequences, we should be able to decode.  
It would be nice if we can get the epsilon removal stuff finished and tested fairly soon because the decoding will be more correct once we remove epsilons; but we can still decode without that.

k2/csrc/hash.h
```diff
@@ -0,0 +1,281 @@
...
+
+/*
+  How Hash32 works:
+    - It's a map from uint32_t to int32_t.  (All keys except
+      ~0 == -1 == UINT32_MAX are allowed).
+    - Each bucket has a struct { uint32_t key; int32_t value; }
+    - If there is no value present, we set the key to ~0 == -1.
+    - The number of buckets is a power of 2 provided by the user.
+    - When accessing hash[key], we use bucket_index == key % num_buckets,
+      leftover_index = 1 | ((key * 2) / num_buckets).  This is
```
csukuangfj, line 54: The documentation does not match the implementation.

There is no `2` in the implementation.

k2/csrc/hash.h
```diff
@@ -0,0 +1,281 @@
...
+      leftover part of the index times 2, plus 1.
+    - If the bucket at `bucket_index` is occupied, we look in locations
+      `bucket_index + n * leftover_index` for n = 1, 2, ....;  this choice
```
csukuangfj, line 57: Looks like it is not multiplied by `n` in the code.

k2/csrc/hash.h
```diff
@@ -0,0 +1,281 @@
...
+    } p;
+  };
+
+  class Accessor {
+   public:
+    Accessor(uint64_t *data,
+             uint32_t num_buckets_mask,
+             int32_t bucket_num_bitsm1) :
+        data_(data), num_buckets_mask_(num_buckets_mask),
+        bucket_num_bitsm1_(bucket_num_bitsm1) { }
```
csukuangfj, line 106: Do we need to check that

```cpp
K2_CHECK_EQ(num_buckets_mask_ + 1,  1 << (bucket_num_bitsm1 + 1));
```

k2/csrc/hash.h
```diff
@@ -0,0 +1,281 @@
...
+                    hash (or set by another thread in this kernel), i.e. only if
+                    this function returns false.
+
+       @return  Returns true if this (key,value) pair was inserted, false otherwise.
+
+      Note: the const is with respect to the metadata only; it is required, to
+      avoid compilation errors.
+   */
+    __forceinline__ __host__ __device__ bool Insert(uint32_t key, int32_t value,
+                                                    int32_t *old_value = nullptr) const {
```
csukuangfj: `old_value` is never used.

k2/csrc/hash.h
danpovey: the index is incremented by leftover_index each time.

k2/csrc/hash.h
danpovey: I checked this in the constructor of Hash32.

PR #493, opened 2020-12-12
Update documentation.

csukuangfj:
Documentation can be written in either `*.md` or in `*.rst` format.
The `installation.md` is copied from INSTALLATION.md


csukuangfj:
You can find what the generated documentation of this pullrequest looks like by visiting

https://k3.readthedocs.io/en/latest/

k2/docs/source/_k2.py
```diff
@@ -0,0 +1,17 @@
+# this file is for generating documentation
+
```
danpovey, line 2: So do the inclusion of these things make it document them?   Or not document them?

k2/docs/source/conf.py
```diff
@@ -12,6 +12,8 @@
 #
```
danpovey, line 12: How about we move the docs to (top-level)/docs/, and remove the old docs?

k2/docs/source/_k2.py
csukuangfj: No, the purpose is to make `import k2` working as `k2` depends on `_k2`. The code is not going to be executed.
Everything that is going to be generated is specified explicitly in
https://github.com/k2-fsa/k2/blob/master/k2/docs/source/python_api/index.rst

```rst
Python API Reference
====================

.. toctree::
   :maxdepth: 2
   :caption: Python API

   fsa
   fsa_algo
   symbol_table
```

If we want to document some of the functions from `_k2`, we can document them here, add a file `python_api/_k2.rst`, and add it to `python_api/index.rst`.

k2/docs/source/conf.py
csukuangfj: Sure, will move it in the next pullrequest.

PR #497, opened 2020-12-13
Add easier way to visualize FSAs

danpovey:


k2/python/k2/fsa.py
```diff
@@ -13,6 +13,10 @@
 from . import fsa_properties
 from .autograd_utils import phantom_set_scores_to
 
+
+import k2.utils
```
csukuangfj, line 17: I would suggest
```python
from .utils import to_dot
```

I think `import k2.utils` requires that we add `the/path/to/k2` to `PYTHONPATH`.

k2/python/k2/utils.py
```diff
@@ -8,15 +8,14 @@
 
 import torch
 
-from .fsa import Fsa
```
csukuangfj, line 11: Is there a reason to remove it?

danpovey: avoid circular import

csukuangfj: Then how about

```python
def save_image(......):
  from .utils import to_dot
  # use to_dot in below
```

That is, we do import inside the function.

k2/python/k2/fsa.py
```diff
@@ -232,6 +235,35 @@ def __str__(self) -> str:
 
         return ans
 
+
+    def save_image(self, filename: str, title: Optional[str] = None):
```
pzelasko, line 239: If this function returns the graphviz object I think it can be easily rendered inside a Jupyter notebook

csukuangfj:
Yes, you are right. I've used it in colab notebook and it is rendered correctly.

Sunday, 13 December 2020, 21:13 +0800 from notifications@github.com  <notifications@github.com>:

k2/python/k2/utils.py
danpovey: Mm it seems to be a bit wrong, that we are making the type-checking become an obstacle to writing the code itself in the way we want...

csukuangfj: I've tried it locally that doing import inside the function will not cause a circular import error.

How about moving `save_image` to `utils.py` and making it an independent function instead of a method of Fsa?

danpovey:
My objection is that you are "letting the tail wag the dog" here.  The
design of the library itself should be the  number 1 priority.
The type-hint stuff is just an extra feature.  Having the type-hints as
strings instead of the typenames is harmless, I don't see  how it detracts
from anything.

k2/python/k2/fsa.py
danpovey: You mean return that in addition to saving the image?  Could be doable.

pzelasko: Yeah. How about naming the method draw(), making it return the graphviz object and having an optional argument for the path where to save the image? Seems like a cleaner API.

k2/python/k2/utils.py
danpovey: I don't want to let the functionality be dictated by the requirements of the type-checking code, would rather design in the best way first.

csukuangfj: `to_dot` and `to_str` were placed in `utils.py` because I think there are just utilities and are useful for debugging
and visualization in jupyter notebook and they are not a core part of `k2`.

It's fine that you put `save_image` as a method of `class Fsa`.

danpovey: @pzelasko when you look at how we're using imports, does it seem to be as it's supposed to be.  The more I look into these issues, the angrier I get (mostly at how stupidly this aspect of Python seems to be designed).

danpovey: ... I do notice that PEP 8 (https://www.python.org/dev/peps/pep-0008/#imports) recommends to mostly use absolute imports, e.g. in utils.py I can get `Fsa` without problems with circular imports by `import k2.fsa` and then using `k2.fsa.Fsa`.  A bit ugly though...
It just seems like a bit of a hack if we have to rely on import statements inside functions.

danpovey: @csukuangfj I think we should make the to_str thing a __repr__ function in class Fsa, it's more canonical.
I wanted to make the 'draw()' method more easily available because it will be used in tutorial material and so on.
We can still keep 'utils', of course, esp. for long functions where the implementation isn't that interesting (and might bore readers of fsa.py).   I want to figure out the correct way of handling problems with circular imports.

csukuangfj: I am fine with the changes. If they are not changed in this pullrequest., I will change them in a new pullrequest when
this one is merged.

k2/python/k2/fsa.py
```diff
@@ -252,6 +251,41 @@ def __str__(self) -> str:
 
         return ans
 
+
+    def draw(self, filename: Optional[str], title: Optional[str] = None) -> 'Digraph':
```
pzelasko, line 255: It should be `filename: Optional[str] = None`, otherwise it will crash when called with no arguments.

k2/python/k2/utils.py
pzelasko: @danpovey the imports overall look OK; the local relative import (`from .utils import to_dot`) is a bit puzzling at first, but then I understood it's about avoiding circular references (I was surprised this works TBH). I am myself not used to having relative imports anywhere besides __init__.py but maybe that's just my experience and other projects also do it like this. 

The circular import problem with typing is also puzzling to me - I'm not sure what is the right way to work around it (or if there is a right way at all). I think using strings for types in `to_dot` makes sense to me and I'd have probably gone that way.

csukuangfj: The relative import stuff is also used by torchaudio. See
https://github.com/pytorch/audio/blob/master/torchaudio/backend/__init__.py

```python
from . import utils
from .utils import (
    list_audio_backends,
    get_audio_backend,
    set_audio_backend,
)
from .sox_backend import (
    save_encinfo,
    sox_signalinfo_t,
    sox_encodinginfo_t,
    get_sox_option_t,
    get_sox_encoding_t,
    get_sox_bool,
)
from .common import (
    SignalInfo,
    EncodingInfo,
)
```

Python even has documentation about relative import. See
https://docs.python.org/3/tutorial/modules.html#intra-package-references

```python
from . import echo
from .. import formats
from ..filters import equalizer
```

pzelasko: Thanks. I am also using relative imports in __init__.py (I think otherwise it would fail most of the time); I just rarely encounter them being used in regular source files.

pzelasko: (I'm actually not sure why and what are pros and cons)

csukuangfj: Here is another example of relative import being used not in `__init__.py` from torchaudio

https://github.com/pytorch/audio/blob/80c97e6a78b6f1ead85d04c40badaa1014d8a24d/torchaudio/backend/sox_backend.py#L11-L12
```python
from . import common
from .common import SignalInfo, EncodingInfo
```
---


> (I'm actually not sure why and what are pros and cons)

It turns out there are two kinds of relative import in Python

- implicit relative import  (no longer supported in Python3)
- explicit relative import

From PEP-8: https://www.python.org/dev/peps/pep-0008/

> Implicit relative imports should never be used and have been removed in Python 3.

> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose

So I think it is ok to use relative import for the existing code in k2, like the one we are using.

pzelasko: Cool, thanks for the info!

PR #500, opened 2020-12-14
[WIP] decoding stuff changes

qindazhu:


k2/python/k2/ops.py
```diff
@@ -68,3 +122,83 @@ def index_add(index: torch.Tensor, value: torch.Tensor,
...
+      Returns a tensor or a ragged tensor (depending on the type of `indexes`)
+    '''
+    if isinstance(indexes, torch.Tensor):
+        return index_select(src, indexes)
+    else:
+        # TODO(haowen): it does not autograd now.
+        return index_tensor_with_ragged_int(src, indexes)
+
+
+def index_attr(src: Union[torch.Tensor, _k2.RaggedInt],
```
danpovey, line 195: can you explain the name of this?

qindazhu: maybe not a good name, I called it `index_attribute(of fsa)`, but that is a general index operation, any suggestion?

qindazhu:
BTW, I move some op to another files to avoid circular import, any better suggestions about this? I have not tried import in a function instead of at the top of file.

qindazhu:
And pickle for ragged tensor only supports RaggedShape2 for now. (not sure if we really need general case?)

danpovey:
I'll merge this, but later we'll change the indexing names so that there is a generic index() function that dispatches the right one, and then a bunch of functions like index_tensor_ragged, index_fsa_tensor and so on, that have the names of the left and right types.

qindazhu:
Sure

k2/python/k2/fsa.py
```diff
@@ -36,58 +37,59 @@ class Fsa(object):
...
-    arcs
-      You will NOT use it directly in Python. It is an instance of
-      ``_k2.RaggedArc`` with only one method ``values()`` which
-      returns a 2-D `torch.Tensor`` of dtype ``torch.int32`` with 4
-      columns. Its number of rows indicates the number of arcs in the
-      FSA. The first column represents the source states, second
-      column the destination states, third column the labels and the
-      fourth column is the score. Note that the score is actually
-      a float number but it is **reinterpreted** as an integer.
+    - ``arcs``: You will NOT use it directly in Python. It is an instance
```
csukuangfj, line 40: Could you use the previous one as it is better rendered in the documentation?

danpovey: Sorry merged so I could debug.. hopefull this can be fixed later...

qindazhu: OK，will do later, seems it's from merging stuff somehow (or local auto format), I never edit those documentation except aux_labels

k2/python/csrc/torch/ragged.cu
```diff
@@ -122,11 +142,28 @@ static void PybindRaggedTpl(py::module &m, const char *name) {
         return std::make_pair(ans, value_indexes_tensor);
       },
       py::arg("src"), py::arg("indexes"), py::arg("need_value_indexes") = true);
+
+  m.def(
+      "index_ragged_with_ragged_int",
```
csukuangfj, line 147: I would suggest renaming `index_ragged_with_ragged_int` and `index_tensor_with_ragged_int` to `index`.

Pybind11 supports function name overloading.

qindazhu: Ah, OK, will do later (but maybe later, I just added the TODO in case I forget this but have something else to do now..)

PR #507, opened 2020-12-14
Ragged fix

danpovey:


danpovey:
Also removing the 'strip' command when building libraries, which hinders debugging

PR #511, opened 2020-12-15
Add quality-of-life improvements to SymbolTable

pzelasko:
I'm proposing a couple of changes for the `SymbolTable`:
- standard Python magic methods (len(st), st[], x in st)
- mutability (it is useful to construct them programatically)
- limited support for generic symbol type (can be anything that is hashable and immutable)
- merging symbol tables (without conflicting ids)

If these are OK to go I'll add the tests.

### DIFF
--- a/k2/python/k2/symbol_table.py
+++ b/k2/python/k2/symbol_table.py
@@ -2,7 +2,7 @@
 #
 # See ../../../LICENSE for clarification regarding multiple authors
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Dict, Optional
 from typing import Generic
 from typing import TypeVar
@@ -22,11 +22,11 @@ class SymbolTable(Generic[Symbol]):
     The SymbolTable can only be read to/written from disk if the
     symbols are strings.
     '''
-    _id2sym: Dict[int, Symbol]
+    _id2sym: Dict[int, Symbol] = field(default_factory=dict)
     '''Map an integer to a symbol.
     '''
 
-    _sym2id: Dict[Symbol, int]
+    _sym2id: Dict[Symbol, int] = field(default_factory=dict)
     '''Map a symbol to an integer.
     '''
 
--- a/k2/python/k2/symbol_table.py
+++ b/k2/python/k2/symbol_table.py
@@ -30,6 +30,11 @@ class SymbolTable(Generic[Symbol]):
     '''Map a symbol to an integer.
     '''
 
+    _next_available_id: int = 1
+    '''A helper internal field that helps adding new symbols 
+    to the table efficiently.
+    '''
+
     eps: Symbol = '<eps>'
     '''Null symbol, always mapped to index 0.
     '''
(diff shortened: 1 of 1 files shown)


csukuangfj:
+1

k2/python/k2/symbol_table.py
```diff
@@ -2,21 +2,38 @@
...
+
+
+@dataclass(repr=False)  # Disable __repr__ otherwise it could freeze e.g. Jupyter.
+class SymbolTable(Generic[Symbol]):
+    '''SymbolTable that maps symbol IDs, found on the FSA arcs to
+    actual objects. These objects can be arbitrary Python objects
+    that can serve as keys in a dictionary (i.e. they need to be
+    hashable and immutable).
+
+    The SymbolTable can only be read to/written from disk if the
```
csukuangfj, line 22: Is this constraint enforced by the implementation?
I find that it uses `print` to convert a symbol to a string.

pzelasko: It won’t know how to convert back to the correct type when reading. If this is needed in the future we could extend it I think.

k2/python/k2/symbol_table.py
```diff
@@ -90,7 +106,55 @@ def from_file(filename: str) -> 'SymbolTable':
...
+            symbol:
+                The symbol to be added.
+            index:
+                Optional int id to which the symbol should be assigned.
+                If it is not available, a ValueError will be raised.
+
+        Returns:
+            The int id to which the symbol has been assigned.
+        '''
+        # Already in the table? Return it's ID.
```
csukuangfj, line 143: typo: its ID

k2/python/k2/symbol_table.py
```diff
@@ -90,7 +106,55 @@ def from_file(filename: str) -> 'SymbolTable':
...
+        if symbol in self._sym2id:
+            return self._sym2id[symbol]
+        # Specific ID not provided - use next available.
+        if index is None:
+            index = len(self)
```
csukuangfj, line 148: It can happen that IDs are not contiguous and the ID returned by `len(self)`
may already be occupied.

Shall we add a field `next_available_id_` that is initialized to the largest ID + 1 ?

k2/python/k2/symbol_table.py
```diff
@@ -104,7 +168,39 @@ def get(self, k: Union[int, str]) -> Union[str, int]:
...
+                A symbol table to merge with ``self``.
+
+        Returns:
+            A new symbol table.
+        '''
+        common_ids = set(self._id2sym).intersection(other._id2sym)
+        assert self.eps == other.eps, f'Mismatched epsilon symbol: ' \
+                                      f'{self.eps} != {other.eps}'
+        for idx in common_ids:
+            assert self[idx] == other[idx], f'ID conflict for id: {idx}, ' \
```
csukuangfj, line 190: This only checks that the same ID maps to the same symbol.

Do we need to add an extra check that the same symbol maps to the same ID?

csukuangfj:
> limited support for generic symbol type (can be anything that is hashable and immutable)

Could you give an example what the generic symbol can be other than `str`?

pzelasko:
> Could you give an example what the generic symbol can be other than `str` and in which situations
> can it be useful?

These could be data structures like tuples/namedtuples that contain multiple pieces of information for each FSA transition. A simple example could be (word, pos_tag).

pzelasko:
I addressed all the remarks - please have a look.

csukuangfj:
Thanks! Merging

PR #516, opened 2020-12-16
add documentation about core concepts in k2.

csukuangfj:
The generated documentation is available for preview at

https://k3.readthedocs.io/en/latest/core_concepts/index.html


csukuangfj:
Looks like that readthedocs cannot render bibtex correctly. Will try to fix it tomorrow.

csukuangfj:
Ready to review.

Any comments would be appreciated.

Doc about ragged arrays will be added in another pullrequrest
as it requires some time to draw figures about them.

docs/source/core_concepts/index.rst
```diff
@@ -0,0 +1,512 @@
...
+  - Arcs that do not enter the final state cannot have -1 as the label
+
+.. HINT::
+
+  Different from other frameworks, FSAs in k2 have only a single final state.
+
+  If you want to convert an FSA from another framework to k2 that contains
+  multiple final states, you can create an extra state and consider it as
+  the super final state. For each final state in the FSA, add an arc to this
+  super final state with label -1 and score 0. The resulting FSA will contain
```
danpovey, line 64: score 0 -> score equal to the final-weight of that final state.   

Note: in general final-states have a final-weight that can be arbitrary.  gtn is unusual in this respect.

docs/source/core_concepts/index.rst
```diff
@@ -0,0 +1,512 @@
...
+
+We summarize the **unique** features of FSA in k2 in below:
+
+  - There is only one start state
+  - The start state is **always** 0
+  - All other states have a state number greater than 0
+  - There is only one final state
+  - The final state **always** has the **largest** state number
+  - Arcs entering the final state **always** have -1 as the label
+  - Arcs that do not enter the final state cannot have -1 as the label
```
danpovey, line 55: We need to mention that we store weights in the positive sense rather than as costs: i.e. we store them as log-probs rather than negative log-probs, and call them "scores" to indicate this.

docs/source/core_concepts/index.rst
```diff
@@ -0,0 +1,512 @@
...
+.. NOTE::
+
+  The visualization code handles the attributes ``aux_labels`` specially.
+  Other than this, ``aux_labels`` is like any other attributes attached
+  to the FSA.
+
+Semirings
+---------
+
+k2 supports two kinds of semirings:
```
danpovey, line 124: Could edit to 

In the FSA literature, generality is achieved through the concept of "semirings".  The two most common are the "tropical semiring" and "log semiring".  The way we will explain these is a little different from the iterature because we are using the opposite sign.  We won't get into the formalism here, but it relates to what happens when you combine scores from multiple alternative paths.  The two common semirings are:
    - tropical semiring: take the maximum score (or minimum cost)
    - log semiring: log-add the scores (or the negatives of the costs).

docs/source/core_concepts/index.rst
```diff
@@ -0,0 +1,512 @@
...
+    :figwidth: 600px
+
+    An FSA with two alternative paths to the final states.
+
+There are two paths from the start state to the final state:
+
+  - Path 0: state 0 -> state 1 -> state 3, with score: 0.1 + 0 = 0.1
+  - Path 1: state 0 -> state 2 -> state 3, with score: 0.2 + 0 = 0.2
+
+So in the tropical semiring, the best score is ``max(0.1, 0.2) == 0.2``.
```
danpovey, line 149: So in the tropical semiring, we would consider that "total score" of this FSA is ``max(0.1, 0.2) == 0.2``.

docs/source/core_concepts/index.rst
danpovey: While k2 only supports these two operations for the core operations, the framework is designed to be flexible through the concept of "attributes" which make it possible to implement the kinds of things that are normally accomplished through exotic semirings such as the Gallic semiring.

danpovey (commented):
Amazing work!!!
A few small comments.

docs/source/core_concepts/index.rst
```diff
@@ -0,0 +1,522 @@
...
+        We store them as log-probs rather than negative log-probs,
+        and call them ``"scores"`` to indicate this.
+
+        They can come directly from the output of a log-softmax layer.
+
+.. HINT::
+
+  Different from other frameworks, FSAs in k2 have only a single final state.
+
+  If you want to convert an FSA from another framework to k2 that contains
```
qindazhu, line 71: Just a suggestion, it's better if we can add the link of the code that how we convert opneFST format to k2 format

docs/source/core_concepts/index.rst
```diff
@@ -0,0 +1,522 @@
...
+
+    .. math::
+
+      \frac{\partial s}{\partial a} = 0
+
+      \frac{\partial s}{\partial b} = \frac{\partial (b + d)}{\partial b} = 1
+
+      \frac{\partial s}{\partial c} = 0
+
+      \frac{\partial s}{\partial b} = \frac{\partial (b + d)}{\partial d} = 1
```
qindazhu, line 340: should be ` \frac{\partial s}{\partial d}` here

docs/source/core_concepts/index.rst
csukuangfj: Fixed.

docs/source/core_concepts/index.rst
csukuangfj: Fixed.

docs/source/core_concepts/index.rst
csukuangfj: Thanks, added.

docs/source/core_concepts/index.rst
csukuangfj: Fixed.

docs/source/core_concepts/index.rst
csukuangfj: Added.

docs/source/core_concepts/index.rst
csukuangfj: Thanks, fixed.

PR #517, opened 2020-12-16
Pruned intersection

danpovey:


danpovey:
This is not the final version, although it will already be more memory efficient than before (though perhaps slower).
I plan to change the backward-pass pruning to run periodically in a background thread (+stream), to both reduce memory usage and make it more concurrent with the forward search.  This will be a bit like what Kaldi does, although Kaldi doesn't use multiple threads.

danpovey:
OK, this is working now.    It will enable us to decode with larger beans and max-states without exhausting memory, as well a being slightly faster.  (Although I'm not saying this will necessarily help WER).

k2/csrc/fsa_utils.cu
```diff
@@ -1789,4 +1789,31 @@ Ragged<int32_t> ComposeArcMaps(Ragged<int32_t> &step1_arc_map,
...
 
+void FixNumStates(FsaVec *fsas) {
+  K2_CHECK_EQ(fsas->NumAxes(), 3);
+  ContextPtr c = fsas->Context();
+  int32_t num_fsas = fsas->Dim0(),
+        num_states = fsas->TotSize(1);
+
+  Array1<int32_t> changed(c, 1, 0);
+  Renumbering renumber_states(c, num_states);
+  renumber_states.Keep() = (char)1;  // by default keep all states..
```
qindazhu, line 1800: why not just set 1 in below lambda? I mean, with `num_states` as the num_items of lambda.

k2/csrc/semaphore.h
```diff
@@ -0,0 +1,48 @@
...
+
+
+// This is intended to implement a subset of the functionality of C++20's
+// counting_semaphore (at the time of writing, we compile with C++14.)
+class counting_semaphore {
+public:
+  counting_semaphore(int count = 0): count_(count) { }
+
+  void release() {  // could also be 'signal'
+    std::unique_lock<std::mutex> lock(mutex_);
```
csukuangfj, line 30: Would `std::lock_guard` be more appropriate here since it does not need
to transfer the ownership of the mutex ?

k2/csrc/semaphore.h
```diff
@@ -0,0 +1,48 @@
+/**
+ * @brief
+ * semaphore
+ *
+ * @copyright
+ * Copyright (c)  2020  Xiaomi Corporation (authors: Daniel Povey)
+ *
+ * @copyright
+ * See LICENSE for clarification regarding multiple authors
+ */
```
csukuangfj, line 10: header guard is missing

k2/csrc/log.h
```diff
@@ -174,6 +174,15 @@ inline bool EnableCudaDeviceSync() {
   return enable_cuda_sync;
 }
```
qindazhu, line 176: I guess user can disable all checks for `RaggedShape` with it, but maybe better with documentation.

k2/csrc/algorithms.h
```diff
@@ -134,7 +137,8 @@ class Renumbering {
   Array1<char> keep_;  // array of elements to keep; dimension is the
                        // `num_old_elems` provided in the constructor but it
                        // was allocated with one extra element.
-  Array1<int32_t> old2new_;  // note: actual dimension is num-old-elems + 1.
+  Array1<int32_t> old2new_;  // note: dimension is num-old-elems + 1.
+  Array1<int32_t> old2new_part_;  // range of old2new_, of dim num-old-elems.
```
csukuangfj, line 141: It is never used.

I guess you were planning to use it in `Old2New(bool extra_element = false)`

k2/csrc/context.cu
```diff
@@ -128,4 +128,50 @@ void GetBlockSizesForLambda2(int32_t m, int32_t n, dim3 *block_dim,
   }
 }
 
+
+void Semaphore:: Signal(ContextPtr c) {
```
csukuangfj, line 132: extra space after the double colons.

k2/csrc/semaphore.h
```diff
@@ -0,0 +1,48 @@
...
+// `k2std` is our replacement for `std` until we require C++20 that has
+// `counting_semaphore` as part of the standard library.
+namespace k2std {
+
+
+// This is intended to implement a subset of the functionality of C++20's
+// counting_semaphore (at the time of writing, we compile with C++14.)
+class counting_semaphore {
+public:
+  counting_semaphore(int count = 0): count_(count) { }
```
csukuangfj, line 27: `explicit` is missing for a constructor with a single parameter.

k2/csrc/fsa_utils.cu
danpovey: mm, might be faster for small data but slower for big.. this is not being used right now and would never be a bottleneck, I don't want to spend time optimizing it...

k2/csrc/semaphore.h
danpovey: will change..

k2/csrc/intersect_pruned.cu
```diff
@@ -1068,16 +1078,299 @@ class MultiGraphDenseIntersectPruned {
...
+        old_states_ptrs_data[i] = cur_frame->states.values.Data();
+
+        // We can't discard any states on t == begin_t because: if it is not t ==
+        // 0, it would be inconvenient to map the dest-states of arcs on t - 1;
+        // and if it is t == 0, this may remove the start-state, which would make
+        // it more complex to avoid invalid FSAs (e.g. with an end-state but no
+        // start-state, or in which we incorrectly interpret a non-start state as
+        // the start state).
+        if (i == 0)  // t == begin_t
+          this_states_keep = (char)1;  // set all elements of the array
```
csukuangfj, line 1173: I have been wondering whether this kind of conversion is needed.

k2/csrc/context.cu
```diff
@@ -128,4 +128,50 @@ void GetBlockSizesForLambda2(int32_t m, int32_t n, dim3 *block_dim,
...
+  if (device_type == kCuda) {
+    cudaEvent_t event;
+    cudaError_t e = cudaEventCreateWithFlags(&event, cudaEventDisableTiming);
+    K2_CHECK_CUDA_ERROR(e) << "Error creating event";
+    // Note: this stream is subject to being overridden by With(stream..), see
+    // class With.
+    cudaStream_t stream = c->GetCudaStream();
+    e = cudaEventRecord(event, stream);
+    K2_CHECK_CUDA_ERROR(e) << "Error recording event.";
+    std::unique_lock<std::mutex> lock(events_mutex_);
```
qindazhu, line 148: I would like to use `std::lock_guard` instead of `unique_lock` as we don't need `defer_lock` feature of `unique_lock`, so using `std::lock_guard` would be simple and intuitive as when it locked in construction and unlocked in destruction (when out of scope)

k2/csrc/intersect_pruned.cu
danpovey: The cast is likely not needed but I feel it helps clarity.

k2/csrc/context.cu
```diff
@@ -128,4 +128,50 @@ void GetBlockSizesForLambda2(int32_t m, int32_t n, dim3 *block_dim,
...
+}
+
+void Semaphore::Wait(ContextPtr c) {
+  DeviceType device_type = c->GetDeviceType();
+  if (device_type_ == kUnk)
+    device_type_ = device_type;
+  else
+    K2_CHECK_EQ(device_type, device_type_)
+        << "Semaphore must always be used with the same device type.";
+  semaphore_.acquire();
```
csukuangfj, line 161: Looks like semaphore_ is used only when the device type is cuda.
Shall we put `acquire` and `release` inside the statement`if (device_type == kCuda) {`
? It can avoid the overhead of using mutex and condition variables when the device type
is cpu, I think.

danpovey: No, the semaphore is needed regardless actually; it's the events that are only needed for CUDA.

danpovey:
Guys, I managed to get to 16% WER via this change.
Actually the key part appears to be a change to atomicMax().  Essentially what I did may be working around a compiler bug; I'll create a separate PR that shows the bug for discussion with the NVidia guys.  It seems it was calling the __host__ atomicMax even in device code, which doesn't make sense as it didn't say __device__.

```
hyp= ['THE', 'RING', 'OF', 'AMETHYST', 'I', 'COULD', 'NOT', 'WEAR', 'HE', 'PLAYED', 'INTO', 'MY', 'SIGHT', 'THEN', 'THAT', 'FIRST', 'KIPPS']
ref= ['I', 'LOVE', 'THEE', 'WITH', 'THE', 'PASSION', 'PUT', 'TO', 'USE', 'IN', 'MY', 'OLD', 'GRIEFS', 'AND', 'WITH', 'MY', "CHILDHOOD'S", 'FAITH']
hyp= ['I', 'LOVE', 'THEE', 'WITH', 'THE', 'PASSION', 'PUT', 'TO', 'USE', 'IN', 'MY', 'OLD', 'GREASIN', 'WITH', 'MY', "CHILDHOOD'S", 'FAITH']
ref= ['DEAREST', 'TEACH', 'ME', 'SO', 'TO', 'POUR', 'OUT', 'GRATITUDE', 'AS', 'THOU', 'DOST', 'GOOD']
hyp= ['DEAREST', 'TEACH', 'ME', 'SO', 'TO', 'POUR', 'OUT', 'GRATITUDE', 'AS', 'THOU', 'DOST', 'GOODT']
ref= ['SLOW', 'TO', 'WORLD', 'GREETINGS', 'QUICK', 'WITH', 'ITS', 'O', 'LIST', 'WHEN', 'THE', 'ANGELS', 'SPEAK']
hyp= ['SLOW', 'TO', 'WORLD', 'GREETINGS', 'QUICK', 'WITH', 'ITS', 'OLD', 'LIST', 'WHEN', 'THE', 'ANGEL', 'SPEAK']
ref= ['I', 'THANK', 'ALL', 'WHO', 'HAVE', 'LOVED', 'ME', 'IN', 'THEIR', 'HEARTS', 'WITH', 'THANKS', 'AND', 'LOVE', 'FROM', 'MINE']
hyp= ['I', 'THINK', 'ALL', 'WHO', 'HAVE', 'LEFT', 'ME', 'IN', 'THEIR', 'HEARTS', 'WITH', 'THANKS', 'AND', 'LOVE', 'FROM', 'MINE']
ref= ['BUT', 'THOU', 'ART', 'NOT', 'SUCH', 'A', 'LOVER', 'MY', 'BELOVED']
hyp= ['BUT', 'THOU', 'ART', 'NOT', 'SUCH', 'A', 'LETTER', 'MY', 'BELOVED']
ref= ['O', 'LOVE', 'O', 'TROTH']
hyp= ['ALL', 'OF', 'A', 'TROUGH']
ref= ['AND', 'LOVE', 'BE', 'FALSE']
hyp= ['AND', 'THE', 'LOVE', 'BE', 'FALSE']
%WER 16.82% [8844 / 52576, 1047 ins, 1033 del, 6764 sub ]
```

csukuangfj:
👍 👍

csukuangfj:
How did you find it?

danpovey:
First I noticed that the pruning wasn't removing as many states as I expected to do it.
But after looking very hard I couldn't see any possible bugs in the code, so just out of general suspicion I put in a check to the code that no paths should be significantly above probability 1.0 (backward+forward > 0.0, actually backward+forward > 2.0 to account for roundoff).

After adding print statements I found that it was apparently a failure in our atomicMax.  I initially suspected IntToOrderedFloat and so on, but after adding print statements to check them I found no problem, so atomicMax was the last likely problem.  I remembered that with atomicAdd it hadn't been possible to just add a __host__ version of the function (I got a compiler error), and I had to add a wrapper called AtomicAdd() that called atomicAdd() instead.  So I tried adding a similar wrapper AtomicMax()- I was suspecting that it somehow wasn't calling the `__device__` version of the function (provided by NVidia).  That turned out to be the problem.

pzelasko:
Congrats!

k2/csrc/ragged_ops.cu
```diff
@@ -1562,4 +1563,125 @@ RaggedShape GetLayer(const RaggedShape &src, int32_t layer) {
...
+  Renumbering r_temp;
+  if (!renumbering_out)
+    renumbering_out = &r_temp;
+
+  ContextPtr c = src_shape.Context();
+  int32_t num_lists = src_shape.Dim0();
+  *renumbering_out = Renumbering(c, num_lists);
+  int32_t *row_splits_data = src_shape.RowSplits(1).Data();
+  char *keep_data = renumbering_out->Keep().Data();
+  K2_EVAL(c, num_lists + 1, lambda_set_keep, (int32_t i) -> void {
```
csukuangfj, line 1628: Should this be `num_lists`? 

Otherwise, when `i` is `num_lists`, `row_splits_data[i+1]` is accessed out of bound.

danpovey: I think you are right.   Would you mind making a PR to fix it?

PR #523, opened 2020-12-18
wrap rm_epsilon_iterative_tropical to python

qindazhu:


qindazhu:
@freewym we have implmented `RemoveEpsilonIterativeTropical` and tested with random cases, here is the python wrapped function and you can try this version now. If there's any issue, please feel free to ping me.

freewym:
Sure, Will work on this and let you know.

qindazhu:
Merigng, so @pzelasko and @freewym  can just pull the code instead of checking out this PR. If there's any issue, just leave your comments and I'll fix later.

freewym:
@qindazhu I need to inverse the returned Fsa of `k2.remove_epsilons_iterative_tropical()`, but its `aux_labels`  is a ragged tensor, and inverse is unsupported for that.

qindazhu:
`invert` an Fsa with `aux_labels` as ragged tensor needs to expand arcs (i.e. we need to convert an arc with two aux_labels to two arcs), we have not implemented it.

Can you show your usage for now? We can try to figure out if it's posibble to do it with current code. If we cannot, I'll add the implementation ( I even thought the host version (no GPU support) I wrote before may work well for your purpose for now)

Also @danpovey here as we talked about `expand`stuff before https://github.com/k2-fsa/k2/pull/481#issuecomment-742219060.

freewym:
I intended to apply `remove_epsilons_iterative_tropical` on H, and then do "H compose L" by `k2.intersect(H.invert_(), L)`

qindazhu:
Ah, OK, I'll add a wrapper of host version `invert` to python today, and will add an GPU implementation later (will talk with Dan offline). 

BTW, bear in mind before you intersect H.invert_ with L, you may need to remove aux_labels of `H.invert_()`, as (IIRC) we don't support both A and B have aux_labels when do `intersect(A,B)`.

qindazhu:
@freewym Host version of `invert` is here https://github.com/k2-fsa/k2/pull/536, noted it may be slow for large graph as it only works for CPU, you may want to cache the result as we did in `snowfall` https://github.com/k2-fsa/snowfall/blob/1a6412259402810f29ef00937b54b6171257b054/egs/librispeech/asr/simple_v1/train.py#L224-L231

PR #533, opened 2020-12-20
Refactor ragged

danpovey:


danpovey:
I am adding a few things here, but also moving some things around for what I see as greater consistency in interface.

danpovey:
Will merge this to avoid blocking snowfall progress.
Can discuss anything later if there are objections.

PR #535, opened 2020-12-21
Change imports etc

danpovey:


k2/python/k2/__init__.py
```diff
@@ -2,6 +2,13 @@
 from _k2 import RaggedInt
 from _k2 import simple_ragged_index_select
 
+from . import autograd
```
csukuangfj, line 5: Will it still work if you remove `from . import xxx` ?

danpovey:
No objections, so will merge..

danpovey:
It doesn't seem to: we get errors like:
```

    k2.autograd_utils.phantom_set_scores_to(ans, self.scores.to(device))

AttributeError: module 'k2' has no attribute 'autograd_utils'
```

PR #539, opened 2020-12-21
Implement ExpandArcs(); various fixes to warnings.

danpovey:


danpovey:
@qindazhu this can be used in the inversion as an alternative, I haven't looked carefully at the other one though so there will likely be some differences.

qindazhu:
OK, thanks, will create a GPU invertion with this.

danpovey:
OK.  Check that the behavior is consistent with CPU e.g. with regard to how
final-arcs are treated.
Since the GPU version supports differentiability it may be easiest just to
use that version even though it will be slower on CPU, since likely this
does not dominate graph construction.

PR #541, opened 2020-12-22
fix a issue of host invertion

qindazhu:
@freewym I have removed the limitation of `no arcs entering start state`. Noted we still have the limitation that aux_labels for final-arcs must be -1. I'll do it tomorrow.


danpovey:
My feeling is that in this case it may not be worth it to maintain a
separate host version of the code in the long run, since it's not likely to
be a performance bottleneck.  But up to you.

PR #543, opened 2020-12-23
add compose

csukuangfj:
I've updated the version from 0.1.1 to 0.1.2 as there are lots of changes after 0.1.1


qindazhu:
+1. You can merge by yourself if you think is OK

PR #544, opened 2020-12-24
fix documentation for intersect dense.

### LINKED ISSUE #542: intersect_dense does not support a_fsas.shape[0] == 1
The doc of `intersect_dense()` says `a_fsas.shape[0]`  can be 1:
https://github.com/k2-fsa/k2/blob/49e824afff691619637d68915ef4934cbec6153b/k2/python/k2/autograd.py#L419

But the underlying c++ seems to not allow it:

https://github.com/k2-fsa/k2/blob/bf274e6855b6e7efddfd3a5a91b5c75e193b4407/k2/csrc/intersect.cu#L108

and I got the check failure for that.

csukuangfj:
Closes #542

### DIFF
--- a/k2/csrc/fsa_algo.h
+++ b/k2/csrc/fsa_algo.h
@@ -179,6 +179,9 @@ void IntersectDensePruned(FsaVec &a_fsas, DenseFsaVec &b_fsas,
 
 /* IntersectDense is a version of IntersectDensePruned that does not
    do pruning in the 1st pass.
+
+   CAUTION: Unlike `IntersectDensePruned`, it requires that
+   a_fsas.Dim0() == b_fsas.shape.Dim0().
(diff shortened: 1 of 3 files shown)


danpovey:
Guys, replying in a random place here, but remember next big thing is
LFMMI/ASG.  Suggest, initially, full bigram LM with weights learned from
numerator, but can also try what GTN guys did, ie learn from num minus den
without normalization.  Motivation for learning from num was compatibility
with LM.  Repeated symbols, as in our CTC impl, may not be are not ideal
for this though.  Could perhaps eventually solve this via a second pass.

csukuangfj:
>  but can also try what GTN guys did, ie learn from num minus den without normalization

I think this is already supported by k2.

PR #545, opened 2020-12-24
implement GPU invert with ExpandArcs

qindazhu:
Have made host version consistent with GPU version (i.e. final-arcs now can have multiple aux_labels), though the state_id of the output Fsa from those two functions may be different, but we can test theire equivalence with `IsRandEquivalent`.


qindazhu:
@freewym, you can now try the GPU version (same interface as before in python), we have only one requirement for final-arcs, that is, aux_labels on final-arcs must have format [X1, X2,...Xn, -1] where Xi != -1, i.e. only the last aux_label can be -1 and must be -1 for final-arcs.

danpovey:
Looks OK to me.  I haven't looked carefully at the compatiblity of the CPU vs GPU versions though.

qindazhu:
They are nearly same except the state-ids of the output Fsa mby be different. Merging, if there's any other issue, just leave your comments and I'll fix them in the next PR.

PR #546, opened 2020-12-24
Implement normalize scores for ragged tensors

### LINKED ISSUE #540: normalize_scores
We might want to implement certain Maximum Likelihood types of things, e.g. suppose in an LF-MMI/ASG-type scheme we want to train the scores of the denominator model to maximize the likelihood of the *numerator* after composition with it, so it's like a genuine language model.

In this case we'd need a function that normalizes a ragged tensor of scores, e.g. a `_k2.RaggedFloat == Ragged<float>`.  In fact, in order to implement the autograd correctly we probably want to break this up, by having the core function 
call LogSumPerSublist from k2/csrc/ragged_ops.h, and then the output would be the input minus the normalizer for the appropriate sub-list.  In the backprop, we'd have
`   input_deriv = output_deriv - exp(output.values) * repeat_out_deriv_sum`
where `repeat_out_deriv_sum` is obtained by doing SumPerSublist on `output_deriv` and then repeating the sum for this sub-list, for each element of the list.

I think all the core C++ code for this already exists, we just need to wrap `Ragged<float>` as `_k2.RaggedFloat` and wrap a few functions.  I recommend to create a file k2/ragged/autograd.py where we can put autograd things that don't pertain directly to FSAs.

The way we'd probably end up training these kinds of things is by having the un-normalized floating-point scores of the FSA be a tensor that lives with the model, and when doing the training forward pass we would compute the actual scores via something like`fsa.scores = k2.ragged.normalize_scores(_k2.Ragged(model.unnorm_scores, fsa.arcs.shape()))`
We could multiply by a small constant e.g. using `model.unnorm_scores * 0.1` if we encountered instability.

csukuangfj:
Closes #540 

---

A Python class `RaggedFloat` is added. 

I suggest creating another class `RaggedInt` if needed or at least moving it from `k2` to `k2.ragged`.

### DIFF
--- a/k2/csrc/ragged_ops_inl.h
+++ b/k2/csrc/ragged_ops_inl.h
@@ -40,7 +40,7 @@ void ApplyOpPerSublist(Ragged<T> &src, T default_value, Array1<T> *dst) {
   K2_CHECK(IsCompatible(src.shape, *dst));
 
   int32_t last_axis = src.NumAxes() - 1;
-  const Array1<int32_t> &row_splits_array = src.shape.RowSplits(last_axis);
+  const Array1<int32_t> &row_splits_array = src.RowSplits(last_axis);
   int32_t num_rows = row_splits_array.Dim() - 1;
   K2_CHECK_EQ(num_rows, dst->Dim());
 
@@ -62,7 +62,7 @@ void ApplyOpPerSublist(Ragged<T> &src, T default_value, Array1<T> *dst) {
       output_data[i] = val;
     }
   } else {
-    K2_CHECK(c->GetDeviceType() == kCuda);
+    K2_CHECK_EQ(c->GetDeviceType(), kCuda);
 
     // This code is based on the example here:
     // https://nvlabs.github.io/cub/structcub_1_1_device_segmented_reduce.html
--- a/k2/csrc/ragged_ops_inl.h
+++ b/k2/csrc/ragged_ops_inl.h
@@ -93,8 +93,7 @@ Ragged<T> NormalizePerSublist(Ragged<T> &src) {
   LogSumPerSublist(src, negative_infinity, &values);
 
   const T *values_data = values.Data();
-  int32_t num_rows = values.Dim();
-  const int32_t *row_splits_data = src.RowSplits(num_axes - 1).Data();
+  const int32_t *row_ids_data = src.RowIds(num_axes - 1).Data();
 
   Array1<T> ans_values(context, src.values.Dim());
   Ragged<T> ans(src.shape, ans_values);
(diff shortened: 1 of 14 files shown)


k2/python/csrc/torch/ragged.cu
```diff
@@ -72,9 +72,7 @@ static void PybindRaggedTpl(py::module &m, const char *name) {
 
   pyclass.def("num_elements", &PyClass::NumElements);
```
csukuangfj, line 74: this file contains only style changes.

csukuangfj:
I tried to wrap sum per sublist and log-sum per sublist to Python but later I removed them
since I think they are not helpful in Python when autograd comes into scope.

csukuangfj:
From https://github.com/k2-fsa/k2/issues/540#issue-772649475

> In this case we'd need a function that normalizes a ragged tensor of scores, e.g. a _k2.RaggedFloat == Ragged<float>. In fact, in order to implement the autograd correctly we probably want to break this up, by having the core function
call LogSumPerSublist from k2/csrc/ragged_ops.h, and then the output would be the **input minus the normalizer for the appropriate sub-list**.

---

> the input minus the normalizer for the appropriate sub-list

The above operation cannot be done efficiently in Python.

k2/csrc/ragged_ops_inl.h
```diff
@@ -71,23 +72,57 @@ void ApplyOpPerSublist(Ragged<T> &src, T default_value, Array1<T> *dst) {
...
+  const T *src_data = src.values.Data();
+
+  K2_EVAL(
+      context, num_rows, lambda_do_normalization, (int32_t i)->void {
+        int32_t begin = row_splits_data[i];
+        int32_t end = row_splits_data[i + 1];
+        T normalizer = values_data[i];
+
+        // it also handles the case when the sublist is empty
+        for (int32_t k = begin; k != end; ++k)
```
qindazhu, line 112: mm, @csukuangfj , I think @danpovey and I have said many times about this: avoid loop in kenel as best as you can...

Please use num_elements as the lambda index and get the index to `normalizer` with row_id.

k2/python/csrc/torch/ragged_ops.cu
```diff
@@ -30,48 +30,111 @@ static void PybindRemoveValuesEq(py::module &m, const char *name) {
...
+          int32_t num_rows = out_grad_sum.Dim();
+
+          if (std::is_same<T, float>::value) {
+            // use `expf` for float
+            K2_EVAL(
+                context, num_rows, lambda_set_ans_grad, (int32_t i)->void {
+                  int32_t begin = row_splits_data[i];
+                  int32_t end = row_splits_data[i + 1];
+                  T scale = out_grad_sum_data[i];
+                  for (int32_t k = begin; k != end; ++k) {
```
qindazhu, line 108: And here, remove the loop

k2/csrc/ragged_ops_inl.h
csukuangfj: thanks, fixed.

qindazhu:
+1, merging.

PR #547, opened 2020-12-25
implemented Arange for ragged tensor

qindazhu:


danpovey:
LGTM.  Could be optimized a bit more, e.g. doing all the transfer to CPU we need in one kernel, but we can leave that for later.

qindazhu:
OK, merging.

RE memory transfer, do you mean use a kenerl to compute those `row_begin` and `row_end` indexes and save to an array (then do one transfer), I suppose we need a loop with `size == ans.NumAxes()` in that kenerl?

danpovey:
yes thats what I meant.  kernel set up is much slower than a single loop
iteration, especially.if.there is cpu syncing.

k2/csrc/ragged_ops.cu
```diff
@@ -1363,6 +1352,47 @@ std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
...
+  int32_t row_begin = begin, row_end = end - 1;
+  for (int32_t cur_axis = axis; cur_axis < num_axes - 1; ++cur_axis) {
+    axes_out[cur_axis - axis].row_splits =
+        axes_in[cur_axis].row_splits.Arange(row_begin, row_end + 1);
+    int32_t row_id = row_begin;
+    // notice here we may do memory copy from GPU to CPU.
+    row_begin = axes_in[cur_axis].row_splits[row_begin];
+    row_end = axes_in[cur_axis].row_splits[row_end];
+    axes_out[cur_axis - axis].row_splits =
+        Minus(axes_out[cur_axis - axis].row_splits, row_begin);
```
csukuangfj, line 1385: When `row_begin` is 0, can the call to `Minus` be avoided?

csukuangfj:
> LGTM. Could be optimized a bit more, e.g. doing all the transfer to CPU we need in one kernel, but we can leave that for later.


The same problem exists for `RaggedShape::Index`:
https://github.com/k2-fsa/k2/blob/d5ecd61c2efd9720d5405f381f6ccc5a741fd634/k2/csrc/ragged.cu#L187

danpovey:
OK.  I'm OK to leave that as a TODO or issue for now, unless someone really wants to fix it now.

csukuangfj:
A TODO has been added in #556. I can fix it when LF-MMI is done.

PR #549, opened 2020-12-25
make treat_epsilons_specially a parameter for intersect/compose.

### LINKED ISSUE #548: Why not make treat_epsilons_specially configurable in python's API intersect
I think it is more convenient to make `treat_epsilons_specially` in k2.intersect() configurable when label 0 means blank. Is there any reason not doing like that?

My current workaround is to add offset +1 to all labels >= 0 before intersect(), and then apply -1 afterwards.

csukuangfj:
Fixes #548


qindazhu:
+1.

PR #551, opened 2020-12-26
rename index_attr to index and dispatch index operation to the right one

qindazhu:


danpovey:
LGTM!

danpovey:
Great work!  I am getting the impression that I am not as necessary to the project as I thought!  But I will keep working when I come back (to-morrow).

qindazhu:
Ah, don't be kidding us @danpovey, apparently we can't work things out without you.. merging, as my next PR based on this (to avoid conflicts).

PR #552, opened 2020-12-26
optimize Arange code to use only one GPU to CPU memory transfer and wrap to python

qindazhu:


k2/csrc/ragged_ops.h
```diff
@@ -238,10 +238,10 @@ std::vector<RaggedShape> GetPrefixes(RaggedShape &src,
       @param [in] axis  The axis we'll get ans, must have
                         0 <= axis < src.NumAxes() - 1.
       @param [in] begin The first element we'll get along axis `axis`.
-                        Must have 0 <= begin <= end <= src.TotSize(axis) + 1.
```
qindazhu, line 241: As `row_splits.dim` is num_rows + 1, here it should be `end <= src.TotSize(axis)`, sorry for my before documentation mistake.

PR #553, opened 2020-12-26
reusing threads from a pool in intersect_pruned.

### LINKED ISSUE #529: Thread pool
Currently, each time IntersectDensePruned() is called (with enough data) it creates a new thread.  Having new threads created instead of using threads from a thread pool is not only inefficient but also makes the nsys profile stats harder to view.  We should create some kind of thread pool.
I am thinking the size of the pool could be based on use: i.e. we create a new thread whenever one isn't already in the pool.
The destructor of the thread-pool object could cause any threads it holds to exit.

csukuangfj:
Fixes #529

### DIFF
--- a/k2/csrc/thread_pool.cu
+++ b/k2/csrc/thread_pool.cu
@@ -28,6 +28,7 @@ ThreadPool::ThreadPool(int32_t num_threads)
 }
 
 ThreadPool::~ThreadPool() {
+  WaitAllTasksFinished();
   {
     std::lock_guard<std::mutex> lock(mutex_);
     keep_running_ = false;
--- a/k2/csrc/thread_pool.cu
+++ b/k2/csrc/thread_pool.cu
@@ -1,0 +1,94 @@
+/**
+ * @brief Thread pool.
+ *
+ *
+ * @copyright
+ * Copyright (c)  2020  Xiaomi Corporation (authors: Fangjun Kuang)
+ *
+ * @copyright
+ * See LICENSE for clarification regarding multiple authors
+ */
+
+#include <utility>
+
+#include "k2/csrc/thread_pool.h"
+
+namespace k2 {
+
+static int32_t GetDefaultNumThreads() {
+  int num_threads = std::thread::hardware_concurrency();
+  if (num_threads == 0) num_threads = 1;
+  return num_threads;
+}
+
+ThreadPool::ThreadPool(int32_t num_threads)
+    : threads_(num_threads > 0 ? num_threads : GetDefaultNumThreads()) {
+  for (auto &thread : threads_)
+    thread = std::thread([this]() { this->ProcessTasks(); });
+}
+
+ThreadPool::~ThreadPool() {
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    keep_running_ = false;
+    not_empty_cond_.notify_all();  // wake up all threads in the pool
+  }
+  for (auto &thread : threads_) thread.join();
+}
+
+void ThreadPool::WaitAllTasksFinished() {
+  std::unique_lock<std::mutex> lock(mutex_);
+  while (!finished_) {
+    // wait for the `empty_cond_` condition.
+    empty_cond_.wait(lock);
+  }
+}
+
+void ThreadPool::ProcessTasks() {
+  std::unique_lock<std::mutex> lock(mutex_);
+  while (keep_running_) {
+    while (tasks_.empty() && keep_running_) {
+      // Wait for the `not_empty_` condition.
+      // RunTask() or the destructor will signal it.
+      not_empty_cond_.wait(lock);
+    }
+
+    if (!keep_running_) break;
+
+    {
+      auto task = std::move(tasks_.front());
+      tasks_.pop();
+      if (!tasks_.empty()) not_empty_cond_.notify_one();
+      lock.unlock();  // let other threads proceed
+      task();
+      // any resource associated with `task` is freed here
+    }
+
+    lock.lock();
+    if (tasks_.empty()) {
+      finished_ = true;
+      // if WaitAllTasksFinished() is waiting, wake it up
+      empty_cond_.notify_one();
+    }
+
+    // the lock is still held here as we assume
+    // that on entering the while() loop, the lock is locked by us
+  }
+}
+
+ThreadPool *GetThreadPool() {
+  static ThreadPool *pool = nullptr;
+  static std::once_flag init_flag;
+
+  std::call_once(init_flag, []() {
+    // TODO(fangjun): how to determine the number
+    // of threads in the pool?
+    //
+    // It is never freed
+    pool = new ThreadPool(2);
+  });
+  K2_CHECK_NE(pool, nullptr);
+  return pool;
+}
+
+}  // namespace k2
--- a/k2/csrc/thread_pool.h
+++ b/k2/csrc/thread_pool.h
@@ -1,0 +1,108 @@
+/**
+ * @brief Thread pool.
+ *
+ *
+ * @copyright
+ * Copyright (c)  2020  Xiaomi Corporation (authors: Fangjun Kuang)
+ *
+ * @copyright
+ * See LICENSE for clarification regarding multiple authors
+ */
+#ifndef K2_CSRC_THREAD_POOL_H_
+#define K2_CSRC_THREAD_POOL_H_
+
+#include <condition_variable>  // NOLINT
+#include <functional>
+#include <mutex>  // NOLINT
+#include <queue>
+#include <thread>  // NOLINT
+#include <vector>
+
+#include "k2/csrc/log.h"
+
+namespace k2 {
+
+class ThreadPool {
+ public:
+  using TaskFunc = std::function<void()>;
+
+  /* Create a pool with given number of threads.
+   *
+   * @param [in] num_threads  The number of threads in the pool. If it is <= 0,
+   *                          the number of threads in the pool is set to
+   *                          `std::thread::hardware_concurrency()`.
+   */
+  explicit ThreadPool(int32_t num_threads);
+
+  /* It ask all threads in the pool to exit after processing what they
+   * are currently executing. If there are still tasks in the queue, they
+   * will not be executed.
+   */
+  ~ThreadPool();
+
+  /* Insert a task into the queue. It is executed immediately
+   * if there are currently idle threads. Otherwise, it is saved
+   * in the queue and waits for threads to be free to pop it from
+   * the queue and to execute it.
+   *
+   * @param [in] task   The task to be run. It should be convertible to
+   *                    `std::function<void()>`.
+   */
+  template <typename Lambda>
+  void RunTask(Lambda task) {
+    K2_CHECK_GT(threads_.size(), 0u);
+    std::lock_guard<std::mutex> lock(mutex_);
+    tasks_.emplace(static_cast<TaskFunc>(task));
+    not_empty_cond_.notify_one();
+    finished_ = false;
+  }
+
+  /* The caller is blocked if `tasks_` is not empty.
+   * If `tasks_` is empty, it returns immediately.
+   */
+  void WaitAllTasksFinished();
+
+ private:
+  /* Remove tasks from `task_` and process them.
+   *
+   * If `keep_running_` is false, the function will return
+   * after processing the current task.
+   *
+   * If the task queue is empty, the caller is blocked.
+   */
+  void ProcessTasks();
+
+ private:
+  // new tasks are added to this queue
+  std::queue<TaskFunc> tasks_;
+
+  std::vector<std::thread> threads_;
+  std::mutex mutex_;
+
+  // When the last task is removed and processed from `tasks_`,
+  // `empty_cond_` is signaled.
+  std::condition_variable empty_cond_;
+
+  // When a new task is added to `tasks_`,
+  // `not_empty_cond_` is signaled.
+  std::condition_variable not_empty_cond_;
+
+  // Set it to false in the destructor to
+  // ask threads to not process tasks in the
+  // waiting queue
+  bool keep_running_ = true;
+
+  // Set it to true when the task queue is empty
+  bool finished_ = true;
+};
+
+/* Get a pointer to global thread pool.
+ *
+ * The returned pointer is NOT owned by the caller and
+ * it must not be freed by the caller.
+ */
+ThreadPool *GetThreadPool();
+
+}  // namespace k2
+
+#endif  // K2_CSRC_THREAD_POOL_H_
--- a/k2/csrc/thread_pool.h
+++ b/k2/csrc/thread_pool.h
@@ -34,22 +34,27 @@ class ThreadPool {
    */
   explicit ThreadPool(int32_t num_threads);
 
-  /* It ask all threads in the pool to exit after processing what they
-   * are currently executing. If there are still tasks in the queue, they
-   * will not be executed.
-   */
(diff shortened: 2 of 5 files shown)


k2/csrc/thread_pool.h
```diff
@@ -0,0 +1,108 @@
...
+  std::condition_variable empty_cond_;
+
+  // When a new task is added to `tasks_`,
+  // `not_empty_cond_` is signaled.
+  std::condition_variable not_empty_cond_;
+
+  // Set it to false in the destructor to
+  // ask threads to not process tasks in the
+  // waiting queue
+  bool keep_running_ = true;
```
qindazhu, line 93: I would call this `stop` or `shutdown` with `false` as the default value.

k2/csrc/thread_pool.h
```diff
@@ -0,0 +1,108 @@
...
+   *                          the number of threads in the pool is set to
+   *                          `std::thread::hardware_concurrency()`.
+   */
+  explicit ThreadPool(int32_t num_threads);
+
+  /* It ask all threads in the pool to exit after processing what they
+   * are currently executing. If there are still tasks in the queue, they
+   * will not be executed.
+   */
+  ~ThreadPool();
```
qindazhu, line 41: I suggest we make assignment operator and copy constructor  deleted, both for lvalue and rvalue (move), since it's dangerous to copy/assgin a thread pool object.

k2/csrc/thread_pool.h
```diff
@@ -0,0 +1,108 @@
...
+
+  /* Create a pool with given number of threads.
+   *
+   * @param [in] num_threads  The number of threads in the pool. If it is <= 0,
+   *                          the number of threads in the pool is set to
+   *                          `std::thread::hardware_concurrency()`.
+   */
+  explicit ThreadPool(int32_t num_threads);
+
+  /* It ask all threads in the pool to exit after processing what they
```
qindazhu, line 37: It asks.

k2/csrc/thread_pool.h
```diff
@@ -0,0 +1,108 @@
...
+   * are currently executing. If there are still tasks in the queue, they
```
qindazhu, line 38: what do you mean here by saying `If there are still tasks in the queue, they will not be executed`? I was thinking the destructor is just join all threads in it.

danpovey:
A good way to test this would be to run the snowfall training (can exit after a few iterations).
You can verify that it's working by either looking at the nsys profile output in nsight systems (visually), or just
by running it gdb, as gdb python3 train.py and (gdb) r, and verifying that it doesn't spawn new threads on every iteration.

k2/csrc/thread_pool.h
```diff
@@ -0,0 +1,108 @@
...
+   * will not be executed.
+   */
+  ~ThreadPool();
+
+  /* Insert a task into the queue. It is executed immediately
+   * if there are currently idle threads. Otherwise, it is saved
+   * in the queue and waits for threads to be free to pop it from
```
qindazhu, line 45: I would say, waits for an idle thread to execute it (i.e. this task)

k2/csrc/thread_pool.h
danpovey: Mm.  Wouldn't it be safer to have the destructor wait for all tasks to be completed?
I have a feeling not doing so would lead to incorrect behavior in some circumstances.

k2/csrc/thread_pool.h
danpovey: I'd say: "The destructor waits for all threads in the pool to exit..."

k2/csrc/thread_pool.h
```diff
@@ -0,0 +1,108 @@
...
+    finished_ = false;
+  }
+
+  /* The caller is blocked if `tasks_` is not empty.
+   * If `tasks_` is empty, it returns immediately.
+   */
+  void WaitAllTasksFinished();
+
+ private:
+  /* Remove tasks from `task_` and process them.
```
qindazhu, line 66: I would say something like, this function keep asking (in an infinite loop) if there's any un-executed task in `task_`, if there's any one, it will remove it from `task_` and process it. If `keep_running_` is false, it will return after processing current task

k2/csrc/thread_pool.cu
```diff
@@ -0,0 +1,94 @@
...
+    thread = std::thread([this]() { this->ProcessTasks(); });
+}
+
+ThreadPool::~ThreadPool() {
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    keep_running_ = false;
+    not_empty_cond_.notify_all();  // wake up all threads in the pool
+  }
+  for (auto &thread : threads_) thread.join();
```
qindazhu, line 36: check `if(thread.joinable()` before join.

k2/csrc/thread_pool.cu
```diff
@@ -0,0 +1,94 @@
...
+  }
+}
+
+void ThreadPool::ProcessTasks() {
+  std::unique_lock<std::mutex> lock(mutex_);
+  while (keep_running_) {
+    while (tasks_.empty() && keep_running_) {
+      // Wait for the `not_empty_` condition.
+      // RunTask() or the destructor will signal it.
+      not_empty_cond_.wait(lock);
```
qindazhu, line 53: @csukuangfj , I think there's some performace issue, you use lock with  `mutex_` for all variables, e.g. not_empty_cond_, task, empty_cond, it's not ineffcient. I suggest here we can add an implemented of `SafeQueue` or class, it contains a separate `mutex` in it, so `task_.empty()`, `front`, `pop` will not lock `mutex_` here.

k2/csrc/thread_pool.cu
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    if (!keep_running_) break;
```
qindazhu, line 56: I think it's not right here as we may leave this task un-executed, but we have wait for it with above mutex, according to the documentation, we should return after execute it.

k2/csrc/thread_pool.cu
```diff
@@ -0,0 +1,94 @@
...
+
+    {
+      auto task = std::move(tasks_.front());
+      tasks_.pop();
+      if (!tasks_.empty()) not_empty_cond_.notify_one();
```
qindazhu, line 61: I think we don't need to notify `not_empty` here, `RunTask` should do this and it has done this.

k2/csrc/thread_pool.cu
```diff
@@ -0,0 +1,94 @@
...
+      lock.unlock();  // let other threads proceed
+      task();
+      // any resource associated with `task` is freed here
+    }
+
+    lock.lock();
+    if (tasks_.empty()) {
+      finished_ = true;
```
qindazhu, line 69: If there's any place set `finished_ = false` again, otherwise what will happly if call `WaitAllTaskFinished` next time?

qindazhu: And I think what `WaitAllTasksFinished` should do is: wait all tasks that has been pushed after the last call of `WaitAllTasksFinished` (except for the first time)

k2/csrc/thread_pool.h
```diff
@@ -0,0 +1,108 @@
...
+  /* Insert a task into the queue. It is executed immediately
+   * if there are currently idle threads. Otherwise, it is saved
+   * in the queue and waits for threads to be free to pop it from
+   * the queue and to execute it.
+   *
+   * @param [in] task   The task to be run. It should be convertible to
+   *                    `std::function<void()>`.
+   */
+  template <typename Lambda>
+  void RunTask(Lambda task) {
```
qindazhu, line 52: Actually

k2/csrc/thread_pool.cu
qindazhu: Oh, I see, it set to false in `RunTask`, but anyway, I think we should clarify the documentation of `WaitAllTasksFinished`

k2/csrc/thread_pool.h
qindazhu, line 52: I would call this function `SubmitTask`, as it just submits task, `ProcessTask` is the real function that `Run` it.

k2/csrc/thread_pool.h
```diff
@@ -0,0 +1,108 @@
...
+   */
+  ~ThreadPool();
+
+  /* Insert a task into the queue. It is executed immediately
+   * if there are currently idle threads. Otherwise, it is saved
+   * in the queue and waits for threads to be free to pop it from
+   * the queue and to execute it.
+   *
+   * @param [in] task   The task to be run. It should be convertible to
+   *                    `std::function<void()>`.
```
qindazhu, line 49: Not sure if we really need to support other function type, but it's doable with parameter pack and `future`, e.g.

```
  template<typename F, typename...Args>
  auto SubmitTask(F&& f, Args&&... args) -> std::future<decltype(f(args...))> {
//  bind function with parameters
    std::function<decltype(f(args...))()> fun= std::bind(std::forward<F>(f), std::forward<Args>(args)...);
// wrap to a shared_ptr so we can copy in below lambda
    auto task_ptr = std::make_shared<std::packaged_task<decltype(f(args...))()>>(fun);
   // convert to `std::function<void()>`
    std::function<void()> wrapper_fun = [task_ptr]() {
      (*task_ptr)(); 
    };
   // push to taks_ and notify
  return task_prt->get_future();
```
Actually this is a pattern to work with C++ threads.

k2/csrc/thread_pool.h
csukuangfj: > since it's dangerous to copy/assgin a thread pool object.

I don't think you are able to copy or assign it.

k2/csrc/thread_pool.cu
csukuangfj: > I think there's some performace issue, you use lock with mutex_ for all variables

All member variables are shared among threads, do you have any proposal modifying them without a lock?

> not_empty_cond_, task, empty_cond, it's not ineffcient

How would you use condition variables without holding a lock?

k2/csrc/thread_pool.cu
csukuangfj: In which case do other threads have a chance to run in case the queue is not empty, but  `RunTask` is not called and this statement is removed?

k2/csrc/thread_pool.h
csukuangfj: Do we need the return value of a task in k2?

csukuangfj:
> and verifying that it doesn't spawn new threads on every iteration.

I have verified it with the following code:

https://github.com/k2-fsa/k2/blob/6df0cd0a53ffc892cf04cd55e4f6be8ace99ff0d/k2/csrc/thread_pool_test.cu#L27-L36

No new threads are created. It is not ideal as the output is checked manually.

k2/csrc/thread_pool.cu
csukuangfj: The behavior is expected.
https://github.com/k2-fsa/k2/blob/6df0cd0a53ffc892cf04cd55e4f6be8ace99ff0d/k2/csrc/thread_pool.h#L37-L41

k2/csrc/thread_pool.cu
qindazhu: I mean, they should not share the same mutex, if they shares, when other thread of `ProcessTask` try to use that `mutex`at the begining,  but it may be used by another thread that is doing `task_.pop()`, it's not reasonable. You should use a separate mutex for `task_`. the pinciple is: don't mix all non-direct-related variables shares a mutex, here `non_empty_cond_` and `task_` are two non-direct-related variables

k2/csrc/thread_pool.cu
csukuangfj: The documentation says:
https://github.com/k2-fsa/k2/blob/6df0cd0a53ffc892cf04cd55e4f6be8ace99ff0d/k2/csrc/thread_pool.h#L60-L61

If you feel the current one is not clear, do you have a better way to make it more understandable ?

k2/csrc/thread_pool.cu
qindazhu: mm, sorry, I didn't understand your question here, the point here is: thread pool is a typical producer-consumer model, so I think it's very clear that `SubmitTask` just push task and notify, the `ProcessTask` just consume the task. I don't what's your point here to `notify` in a consumer, i.e. ProcessTask

k2/csrc/thread_pool.cu
csukuangfj: I think `join` is fine here as there is no interface to detach a thread in the pool.

danpovey:
I think we don't need to spend too much time on this, we can get it working
and move on; I think getting the LF-MMI working is more important than
optimizing this.

danpovey:
i think we can merge when fangjun thinks its ok, later we can have @kkm
check it, he is likely very familiar with these issues.

k2/csrc/thread_pool.cu
csukuangfj: > if they shares, when other thread of ProcessTask try to use that mutex at the begining, but it may be used by another thread that is doing task_.pop(), it's not reasonable

Could you explain why it is not reasonable?  

How would you use a mutex to protect the condition variable `non_empty_cond_` if it needs protection?

Do we need a separate mutex for `keep_running` and a `mutex` for `finished_`?

As the region protected by the mutex is made as small as possible. I doubt it would be a performance bottleneck here.

csukuangfj:
I am fixing:
> Wouldn't it be safer to have the destructor wait for all tasks to be completed?

Since the destructor is never called in our case, it will not be a problem in k2. I will fix it anyway.

k2/csrc/thread_pool.cu
csukuangfj: >  I think it's very clear that SubmitTask just push task and notify, the ProcessTask just consume the task.

Can the consumer always be notified? Is it possible that the notification gets lost?

If all consumers are running, i.e., no consumer is in waiting state, does the `notify` event from the producer
have any effect?

k2/csrc/thread_pool.h
csukuangfj: Thanks, changed.

k2/csrc/thread_pool.h
csukuangfj: Either name would be fine.

k2/csrc/thread_pool.h
csukuangfj: I added some comments explaining why it is not copyable or assignable.

k2/csrc/thread_pool.h
csukuangfj: Thanks, changed.

k2/csrc/thread_pool.h
csukuangfj: Fixed. The destructor will wait until all submitted tasks are finished.

k2/csrc/thread_pool.h
csukuangfj: thanks, changed.

danpovey:
> I am fixing:
> 
> > Wouldn't it be safer to have the destructor wait for all tasks to be completed?
> 
> Since the destructor is never called in our case, it will not be a problem in k2. I will fix it anyway.

The destructor will be called when the main thread is exiting (winding down).

csukuangfj:
> The destructor will be called when the main thread is exiting (winding down).

The object is allocated on the heap by `new ThreadPool`, so I think its destructor is not going to be invoked unless `delete` is called on it.
Anyway,  I have changed it to wait for all tasks being finished before exiting since users may create an object
on the stack or call `delete`.

danpovey:
OK, great.  I'll merge this.

PR #554, opened 2020-12-27
fix a bug in compose when a_fsa has aux_labels and b_fsa does not

freewym:


k2/python/k2/fsa_algo.py
```diff
@@ -188,7 +188,7 @@ def compose(a_fsa: Fsa, b_fsa: Fsa,
         return intersect(a_fsa, b_fsa, treat_epsilons_specially)
 
     if not hasattr(b_fsa, 'aux_labels'):
-        return intersect(a_fsa, b_fsa, treat_epsilons_specially)
+        return invert(intersect(invert(a_fsa), b_fsa, treat_epsilons_specially))
```
csukuangfj, line 191: https://github.com/k2-fsa/k2/blob/d5ce23c44daf814eff40f2a1c28ab210f4c2bee3/k2/python/k2/fsa_algo.py#L148-L150

I was assuming that if either FSA does not have aux_labels, `compose`
is equivalent to `intersect`.

The user has to `invert` the input `a_fsas` himself/herself before
calling `intersect`.

freewym: OK. I see.

PR #557, opened 2020-12-28
support assigning random scores to an FSA.

csukuangfj:
It also demonstrates how to use `torch.optim.SGD` to update the scores of an FSA.

# CAUTION

When an FSA is constructed, `fsa.scores` shares the underlying memory with `fsa.arcs.values`.
But if you do `fsa.scores = something`, then `fsa.scores` shares memory with `something`, but.not.with `fsa.arcs.values`!

Whenever you do `fsa.scores = something`, the code copies `something` to `fsa.arcs.values` and updates `fsa.scores`
to point to `something`.

### DIFF
--- a/k2/python/k2/fsa.py
+++ b/k2/python/k2/fsa.py
@@ -1039,7 +1039,7 @@ class Fsa(object):
           The function name ends with a underline indicating this function
           will change `self` in-place.
         '''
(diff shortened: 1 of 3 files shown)


danpovey:
Great!!

PR #559, opened 2020-12-29
Hash changes

danpovey:


danpovey:
@csukuangfj this is intended to resolve a problem in Snowfall that you encountered where, if the graph size and minibatch size were too large, we got a crash.  
However I'd like, if possible, to verify that it doesn't cause a substantial slow-down.  (It may not be slower at all, but it would be good to check).  We also need to verify that it works.  Can you let us know what snowfall change we'd have to make to test it?

csukuangfj:
https://github.com/k2-fsa/snowfall/blob/09512c3978cd0c614a7580e724767f3d509e2ab9/egs/librispeech/asr/simple_v1/decode.py#L169

```python
    test = K2SpeechRecognitionIterableDataset(cuts_test,
                                              max_frames=30000,
                                              shuffle=False,
                                              concat_cuts=False)
```

`30000` -> `200000`

It was `200000`, which would result in an out of memory error. When I changed it to `30000`, OOM is gone.

csukuangfj:
@zhu-han

I made some comments above. Thanks for your help.

zhu-han:
Testing results are as following:

original version results:

max_frames | memory | time
-- | -- | --
30000 | 5732MiB | 26.033s
100000 | 11248MiB | 20.930s
200000 | 15360MiB | 20.907s

new version results:

max_frames | memory | time
-- | -- | --
30000 | 5002MiB | 26.706s
100000 | 11250MiB | 21.034s
200000 | crashed |

zhu-han:
Testing results are as following:

original version results:

max_frames | memory | time
-- | -- | --
30000 | 5002MiB | 26.706s
100000 | 11250MiB | 21.034s
200000 | crashed |  

new version results:

max_frames | memory | time
-- | -- | --
30000 | 5732MiB | 26.033s
100000 | 11248MiB | 20.930s
200000 | 15360MiB | 20.907s

danpovey:
Fantastic, thanks!!
I'll merge this; can you please merge a PR to snowfall where you change the nframes to 100000?

PR #563, opened 2021-01-01
fix a bug in k2.ragged.normalize_scores.

csukuangfj:
If k2.ragged.RaggedFloat is constructed from a shape and a value,
its scores should be set to the given value to make autograd work.

### DIFF
--- a/k2/python/k2/fsa.py
+++ b/k2/python/k2/fsa.py
@@ -1045,28 +1045,3 @@ class Fsa(object):
 
         # note that `self.scores` also works here, but [:] is more efficient
         self.scores[:] = ragged_scores.scores
-
-    def detach(self) -> 'Fsa':
-        '''Return a new FSA, detached from the current graph.
-
-        Like torch.Tensor.detach(), the returned FSA shares the underlying
-        memory with `self`. The only difference is that the returned FSA's
-        requires_grad is False.
-
-        Caution:
-          The returned FSA shares memory with this FSA.
-
-        Returns:
-          Return an FSA whose `requires_grad` is False.
-        '''
-        # Keep this code in sync with that in to()
-        ans = Fsa(self.arcs, properties=self.properties)
-
-        for name, value in self.named_tensor_attr(include_scores=False):
-            setattr(ans, name, value)
-
-        for name, value in self.named_non_tensor_attr():
-            setattr(ans, name, value)
-
-        ans.scores = self.scores.detach()
-        return ans
--- a/k2/python/k2/fsa.py
+++ b/k2/python/k2/fsa.py
@@ -1030,10 +1030,14 @@ class Fsa(object):
         return Fsa(arcs, aux_labels=aux_labels)
 
     def set_scores_stochastic_(self, scores) -> None:
-        '''Normalize the given `scores` and assign it to self.
+        '''Normalize the given `scores` and assign it to `self.scores`.
 
-        Scores are normalized per state. That is, the sum of the probabilities
-        of all arcs leaving a state equals to 1.
+        Args:
+          scores:
+            Tensor of scores of dtype torch.float32, and shape equal to
+            `self.scores.shape` (one axis). Will be normalized so the
+            sum, after exponentiating, of the scores leaving each state
+            that has at least one arc leaving it is 1.
 
         Caution:
           The function name ends with an underline indicating this function
--- a/k2/python/k2/fsa.py
+++ b/k2/python/k2/fsa.py
@@ -1042,7 +1042,8 @@ class Fsa(object):
         assert scores.ndim == 1
         assert scores.dtype == torch.float32
 
-        ragged_scores = k2.ragged.RaggedFloat(self.arcs.shape(), scores)
+        ragged_scores = k2.ragged.RaggedFloat(
+            self.arcs.shape().to(scores.device), scores)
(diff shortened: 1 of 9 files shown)


csukuangfj:
There already exists `k2.Fsa.detach()`
https://github.com/k2-fsa/k2/blob/75f9378c3bae8c01a280a32cb4c2f6afb938951a/k2/python/k2/fsa.py#L887

so I am removing the one introduced a few days ago.

danpovey:
looks OK to me..

k2/python/k2/ragged/tensor.py
```diff
@@ -42,14 +42,18 @@ def __init__(self,
...
         elif isinstance(ragged, _k2.RaggedShape):
             assert values is not None
             ragged = _k2.RaggedFloat(ragged, values)
 
         assert isinstance(ragged, _k2.RaggedFloat)
 
         self.ragged = ragged
-        self._scores = ragged.values()
+        if values is not None:
+            self._scores = values
```
danpovey, line 54: Wait, why does RaggedFloat have a _scores member?

danpovey: ... it is supposed to be a generic interface, not tied into class Fsa.

csukuangfj: > Wait, why does RaggedFloat have a _scores member?

While doing `normalize_scores`, if we use `_k2.RaggedFloat` directly it is not possible to register the
operation into the autograd graph since there is no `torch.Tensor`. 

`k2.ragged.RaggedFloat` is a wrap around `_k2.RaggedFloat` and the field `_scores` is to make
autograd work.

https://github.com/k2-fsa/k2/blob/75f9378c3bae8c01a280a32cb4c2f6afb938951a/k2/python/k2/ragged/autograd.py#L15-L16

For the autograd to work, we need to pass a `torch.Tensor` to it. One alternative is to let the user pass a tensor
while calling `normalize_scores`.

csukuangfj: > ... it is supposed to be a generic interface, not tied into class Fsa.

The class `RaggedFloat` is not tied to `class Fsa`; it is a wrapper around `_k2.RaggedFloat`.

danpovey: I think the name _values would be better than _scores; we don't want to imply the values have any specific meaning.

csukuangfj: will change it.

danpovey: And please make sure the documentation of class RaggedFloat is as self-contained as possible and explains anything that needs to be explained about autograd...  (it may already be; just a reminder).

csukuangfj: I am adding more documentation for it.

csukuangfj:
Ready for review.

k2/python/k2/fsa.py
```diff
@@ -1029,44 +1029,23 @@ def from_openfst(cls, s: str, acceptor: bool = True) -> 'Fsa':
         arcs, aux_labels = _k2.fsa_from_str(s, acceptor, True)
         return Fsa(arcs, aux_labels=aux_labels)
 
-    def set_scores_stochastic_(self) -> None:
-        '''Set `scores` to random numbers.
+    def set_scores_stochastic_(self, scores) -> None:
+        '''Normalize the given `scores` and assign it to self.
```
danpovey, line 1033: I'd write this as:
   Normalize the given `scores` and assign the result to self.scores.
 Args:
     scores:   Tensor of scores of dtype torch.float32, and shape equal to self.scores.shape
            (one axis).  Will be normalized so the sum, after exponentiating, of the scores 
            leaving each state that has at least one arc leaving it is 1.

k2/python/k2/fsa.py
```diff
@@ -1029,44 +1029,23 @@ def from_openfst(cls, s: str, acceptor: bool = True) -> 'Fsa':
...
-
-        for name, value in self.named_non_tensor_attr():
-            setattr(ans, name, value)
+        ragged_scores = k2.ragged.RaggedFloat(
+            self.arcs.shape().to(scores.device), scores)
+        ragged_scores = k2.ragged.normalize_scores(ragged_scores)
 
-        ans.scores = self.scores.detach()
-        return ans
+        # Note we use `to` here since `scores` and `self.scores` may not
```
danpovey, line 1049: I find it odd that scores and self.scores might not be on the same device, surely we should ensure that they always are?

csukuangfj: It happens in our current ASG training:

- model.P_scores is on CUDA
- P is on CPU since `k2.intersect` works only on CPU

So in this case `model.P_scores` and `P.scores` are on different devices.

k2/python/k2/fsa.py
csukuangfj: thanks, changed.

k2/python/k2/fsa.py
csukuangfj: >  surely we should ensure that they always are?

I assume that log-sum per sub list runs faster on CUDA; otherwise, we can require that `scores.device == self.device`.

csukuangfj:
Merging. If there are any objections, please leave your comments here and I will resolve them.

PR #571, opened 2021-01-05
Implement GetTransposeReodering with MultiSplit

csukuangfj:
[CUDPP][1] has not been updated for more than **4** years. It depends on a very old version of CUB and moderngpu,
which are not compatible with the one used by k2. 

In addition, it uses some deprecated/unsupported CUDA APIs, e.g.,`__shfl` and `__ballot`, which means it cannot be compiled
without modifications.

Since we only want its MultiSplit implementation, I have identified all files related to MultiSplit and placed them into `k2/csrc/cudpp`. Its internals is hard to understand, at least for me. I don't think we need to maintain or modify it, as long as it works.

TODOs:
- [x] Replace `cudaMalloc` with `Context::Allocate`.
- [x] Add benchmark for MultiSplit
- [x] ~Implement IndexAdd with MultiSplit and add benchmark for it~

[1]: https://github.com/cudpp/cudpp.git

---

Currently, `cudpp` is placed into `k2/csrc/cudpp`. I am open about the place where it should be placed.


csukuangfj:
`RandomFsaVec` does not support a fixed seed.

Need to change it so that the results are comparable across different runs.

csukuangfj:
Here is the benchmark result for GetTransposeReordering.

`MultiSplit` is faster for larger FsaVecs, whereas sort from `moderngpu` is faster for smaller FsaVecs.


`GetTransposeReordering_14_265_1576` means:
- there are 14 FSAs
- with 265 states in total
- and with 1576 arcs in total

# With MultiSplit from CUDPP

```
op_name,dtype,device,problem_size,number_of_iterations,elapsed_us_per_iteration
GetTransposeReordering_14_265_1576,int32,kCpu,10,100,97.241402
GetTransposeReordering_51_3181_25958,int32,kCpu,50,100,860.002075
GetTransposeReordering_109_13355_110003,int32,kCpu,100,100,3818.860107
GetTransposeReordering_551_313326_2760114,int32,kCpu,500,20,112707.851562
GetTransposeReordering_11_189_1196,int32,kCuda,10,100,189.125122
GetTransposeReordering_99_6323_50463,int32,kCuda,50,100,317.932495
GetTransposeReordering_186_22261_187643,int32,kCuda,100,100,295.311676
GetTransposeReordering_979_558169_4903852,int32,kCuda,500,20,1556.000000
```

# Via sort for moderngpu
```
op_name,dtype,device,problem_size,number_of_iterations,elapsed_us_per_iteration
GetTransposeReordering_14_265_1576,int32,kCpu,10,100,57.199001
GetTransposeReordering_51_3181_25958,int32,kCpu,50,100,844.709839
GetTransposeReordering_109_13355_110003,int32,kCpu,100,100,3886.411133
GetTransposeReordering_551_313326_2760114,int32,kCpu,500,20,113320.195312
GetTransposeReordering_11_189_1196,int32,kCuda,10,100,60.847996
GetTransposeReordering_99_6323_50463,int32,kCuda,50,100,260.436798
GetTransposeReordering_186_22261_187643,int32,kCuda,100,100,337.603546
GetTransposeReordering_979_558169_4903852,int32,kCuda,500,20,3180.382324
```

k2/csrc/math.h
```diff
@@ -124,6 +124,10 @@ inline std::istream &operator >>(std::istream &is, InputFixer<double> &f) {
   return is;
 }
 
+// Return the seed that can be used for random generators.
+// It reads the environment variable to get the seed.
```
qindazhu, line 128: maybe better to say, it reads the environment variable `K2_SEED` to... or `0` if `K2_SEED` is not set?

danpovey:
I assume it wouldn't be too complex to have an if-statement and choose the
faster one?

danpovey:
I assume it wouldn't be too complex to have an if-statement and choose the faster one?
(re-posting via github as mail has been flaky recently)

qindazhu:
Yeah, it's better if we can print parameters `num_rows` and `num_cols` besides `GetTransposeReordering_14_265_1576`, it may be more intuitive to choose the threshold based on `num_rows` and `num_cols`.

k2/csrc/cudpp/cuda_util.h
```diff
@@ -0,0 +1,52 @@
+// this file is copied/modified from
+// https://github.com/cudpp/cudpp/blob/master/src/cudpp/cuda_util.h
+#ifndef K2_CSRC_CUDPP_CUDA_UTIL_H_
+#define K2_CSRC_CUDPP_CUDA_UTIL_H_
+
+#include <stdio.h>
+#include <stdlib.h>             /* EXIT_FAILURE */
```
qindazhu, line 7: seems we only use `CUDA_SAFE_CALL` in those cudpp files? Is that possible that we delete this file and use `K2_CUDA_SAFE_CALL`

qindazhu: Just want to make sure we can get consistent syncronize behavior with k2, i.e. controlled by DEBUG and K2_SYNC_KERNELS

danpovey:
Agreed on kernel synchronization; better to use our mechanisms as long as
there's no clear incompatibility, so
flags behave the same.

k2/csrc/ragged_ops.cu
```diff
@@ -1144,6 +1145,27 @@ Array1<int32_t> GetTransposeReordering(Ragged<int32_t> &src, int32_t num_cols) {
...
   K2_CHECK_EQ(device_type, kCuda);
 
+#if 1
+  const int32_t *values_data = src.values.Data();
+  int32_t num_buckets = num_cols;
+  int32_t num_elements = src.values.Dim();
+
+  Array1<int32_t> ans = Range(context, num_elements, 0);
+  auto lambda = [values_data, num_elements] __device__(uint32_t i) -> uint32_t {
+    return i < num_elements ? values_data[i] : 0;
```
qindazhu, line 1155: @csukuangfj just a question: why can `i` equal to `num_elements` here? cudpp will pass `i` greater than `num_elements`?

csukuangfj: Yes, you can print what `i` looks like and you will get lots of very large numbers, which result in segfault if you remove the condition statement.

qindazhu: OK, thanks, I guess it may pass some thread Ids to this lambda function instead of element index.

csukuangfj:
The MultiSplit algorithm proposed by [GPU Multisplit: an extended study of a parallel algorithm][1]
is applicable only to `num_buckets <= 32`, at least this is what CUDPP implements.

For `num_buckets > 32`, CUDPP implements MultiSplit with RadixSort from `CUB`. 

Will remove extra code for `num_buckets <= 32` since we will use sort from moderngpu in this case.

[1]: https://arxiv.org/pdf/1701.01189.pdf

qindazhu:
It's a pity that they only implement num_buckets <=32 with histogram approach. But I suggest we don't remove the radixSort approach without benchmark, `MergeSort` maybe faster than `RadixSort` for floating type, I'm not sure if it's still faster for integer type. If we really find `MergeSort` approach is faster, we could just add a if-else branch in our code instaed of changing code in cudpp.

k2/csrc/tensor_ops.cu
```diff
@@ -288,13 +291,96 @@ Tensor Index(Tensor &src, Array1<int32_t> &indexes,
...
+  //
+  // This function will never be called with T==double
+  // at runtime.
+  K2_CHECK_EQ(sizeof(T), sizeof(uint32_t));
+  K2_CHECK_EQ(context->GetDeviceType(), kCuda);
+  int32_t num_buckets = dest_dim;
+  int32_t num_elements = src_dim;
+
+  Array1<int32_t> row_splits(context, num_buckets + 1);
+  Array1<int32_t> histogram = row_splits.Range(0, num_buckets);
```
qindazhu, line 309: As MultiSplit will change both keys and values in place, what I thought originally is that we can re-arrange `index`  use MultiSplit and use the result as the row-ids of the `ragged` below, and gather `ragged.values` with the re-arraged `keys` from `src`, at least for the case where `allow_minus_one=false`

qindazhu: For the case where `allow_minus_one=true`, I think we can get the end position of valid row_ids (note I suppose you put -1 in the last bucket) with only one kenerl, say, the previous value in the re-arranged `index` is not -1 but current value is -1.

csukuangfj: thanks, will try it tomorrow.

csukuangfj:
CUDPP is removed and cub::RadixSort is used to implement MultiSplit.  Now MultiSplit is also faster
for smaller FsaVecs. I would recommend using MutliSplit for GetTransposeReodering. I think the reason for
becoming faster for smaller FsaVecs is that it reduces the number of kernel calls.

## With MultiSplit
```
op_name,dtype,device,problem_size,number_of_iterations,elapsed_us_per_iteration
GetTransposeReordering_14_265_1576,int32,kCpu,10,100,383.939728
GetTransposeReordering_20_667_4231,int32,kCpu,20,100,975.129578
GetTransposeReordering_47_1791_14565,int32,kCpu,30,100,3135.449951
GetTransposeReordering_86_5466_43692,int32,kCpu,50,100,9510.781250
GetTransposeReordering_174_21063_175469,int32,kCpu,100,100,38240.578125
GetTransposeReordering_255_59841_512075,int32,kCpu,200,50,110543.398438
GetTransposeReordering_449_155057_1351205,int32,kCpu,300,33,289998.187500
GetTransposeReordering_680_387983_3406153,int32,kCpu,500,20,731312.812500
GetTransposeReordering_17_289_1816,int32,kCuda,10,100,35.334721
GetTransposeReordering_34_1043_7232,int32,kCuda,20,100,105.194244
GetTransposeReordering_52_2220_16057,int32,kCuda,30,100,116.396797
GetTransposeReordering_74_4634_37552,int32,kCuda,50,100,124.131516
GetTransposeReordering_170_20560_171410,int32,kCuda,100,100,176.891830
GetTransposeReordering_256_60241_514390,int32,kCuda,200,50,225.849625
GetTransposeReordering_594_205468_1787389,int32,kCuda,300,33,474.793671
GetTransposeReordering_957_544306_4793299,int32,kCuda,500,20,1295.924805
```


# Via sort from moderngpu
```
op_name,dtype,device,problem_size,number_of_iterations,elapsed_us_per_iteration
GetTransposeReordering_14_265_1576,int32,kCpu,10,100,356.159210
GetTransposeReordering_20_667_4231,int32,kCpu,20,100,1077.790283
GetTransposeReordering_47_1791_14565,int32,kCpu,30,100,3158.240234
GetTransposeReordering_86_5466_43692,int32,kCpu,50,100,9353.229492
GetTransposeReordering_174_21063_175469,int32,kCpu,100,100,37209.878906
GetTransposeReordering_255_59841_512075,int32,kCpu,200,50,108721.945312
GetTransposeReordering_449_155057_1351205,int32,kCpu,300,33,287676.937500
GetTransposeReordering_680_387983_3406153,int32,kCpu,500,20,727649.750000
GetTransposeReordering_17_289_1816,int32,kCuda,10,100,92.694077
GetTransposeReordering_34_1043_7232,int32,kCuda,20,100,157.120636
GetTransposeReordering_52_2220_16057,int32,kCuda,30,100,192.676163
GetTransposeReordering_74_4634_37552,int32,kCuda,50,100,232.854721
GetTransposeReordering_170_20560_171410,int32,kCuda,100,100,286.725128
GetTransposeReordering_256_60241_514390,int32,kCuda,200,50,501.029785
GetTransposeReordering_594_205468_1787389,int32,kCuda,300,33,1375.094238
GetTransposeReordering_957_544306_4793299,int32,kCuda,500,20,3146.454346
```

csukuangfj:
Will add benchmark for `IndexAdd`.

k2/csrc/cudpp/cuda_util.h
csukuangfj: Fixed.

csukuangfj:
> Yeah, it's better if we can print parameters num_rows and num_cols besides GetTransposeReordering_14_265_1576, it may be more intuitive to choose the threshold based on num_rows and num_cols.


The output format is fixed and also generic for other operators. I will change it if other operators also have
this kind of needs.

csukuangfj:
# Benchmark Results for IndexAdd1D

`K2_SEED=1`

MultiSplit is not faster than our current implementation, so I will remove it for IndexAdd and keep it only for
GetTransposeReodering.

## with MultiSplit
```
op_name,dtype,device,problem_size,number_of_iterations,elapsed_us_per_iteration
IndexAdd1D,int32,kCpu,10,500,0.800133
IndexAdd1D,int32,kCpu,100,500,1.821995
IndexAdd1D,int32,kCpu,500,500,5.708218
IndexAdd1D,int32,kCpu,1000,500,12.047768
IndexAdd1D,int32,kCpu,2000,500,25.907993
IndexAdd1D,int32,kCpu,5000,500,55.459976
IndexAdd1D,int32,kCpu,10000,500,132.425797
IndexAdd1D,int32,kCpu,50000,200,731.775757
IndexAdd1D,int32,kCuda,10,500,189.149307
IndexAdd1D,int32,kCuda,100,500,207.107452
IndexAdd1D,int32,kCuda,500,500,200.612549
IndexAdd1D,int32,kCuda,1000,500,202.656052
IndexAdd1D,int32,kCuda,2000,500,213.698288
IndexAdd1D,int32,kCuda,5000,500,289.130829
IndexAdd1D,int32,kCuda,10000,500,299.169983
IndexAdd1D,int32,kCuda,50000,200,249.831848
IndexAdd1D,float,kCpu,10,500,0.706196
IndexAdd1D,float,kCpu,100,500,1.798153
IndexAdd1D,float,kCpu,500,500,6.646156
IndexAdd1D,float,kCpu,1000,500,13.014317
IndexAdd1D,float,kCpu,2000,500,24.866104
IndexAdd1D,float,kCpu,5000,500,60.870171
IndexAdd1D,float,kCpu,10000,500,157.862183
IndexAdd1D,float,kCpu,50000,200,822.705017
IndexAdd1D,float,kCuda,10,500,206.052353
IndexAdd1D,float,kCuda,100,500,240.781830
IndexAdd1D,float,kCuda,500,500,212.909058
IndexAdd1D,float,kCuda,1000,500,212.137100
IndexAdd1D,float,kCuda,2000,500,229.809601
IndexAdd1D,float,kCuda,5000,500,317.291260
IndexAdd1D,float,kCuda,10000,500,281.757446
IndexAdd1D,float,kCuda,50000,200,320.104462
```

## With `atomicAdd`
```
op_name,dtype,device,problem_size,number_of_iterations,elapsed_us_per_iteration
IndexAdd1D,int32,kCpu,10,500,0.734329
IndexAdd1D,int32,kCpu,100,500,1.636028
IndexAdd1D,int32,kCpu,500,500,7.345676
IndexAdd1D,int32,kCpu,1000,500,14.644146
IndexAdd1D,int32,kCpu,2000,500,29.421806
IndexAdd1D,int32,kCpu,5000,500,56.109905
IndexAdd1D,int32,kCpu,10000,500,111.427780
IndexAdd1D,int32,kCpu,50000,200,732.965454
IndexAdd1D,int32,kCuda,10,500,5.484160
IndexAdd1D,int32,kCuda,100,500,5.415168
IndexAdd1D,int32,kCuda,500,500,5.396288
IndexAdd1D,int32,kCuda,1000,500,5.485184
IndexAdd1D,int32,kCuda,2000,500,5.449344
IndexAdd1D,int32,kCuda,5000,500,5.484352
IndexAdd1D,int32,kCuda,10000,500,7.238592
IndexAdd1D,int32,kCuda,50000,200,5.557281
IndexAdd1D,float,kCpu,10,500,0.669956
IndexAdd1D,float,kCpu,100,500,1.741886
IndexAdd1D,float,kCpu,500,500,6.268024
IndexAdd1D,float,kCpu,1000,500,13.950348
IndexAdd1D,float,kCpu,2000,500,23.530006
IndexAdd1D,float,kCpu,5000,500,58.936119
IndexAdd1D,float,kCpu,10000,500,148.161880
IndexAdd1D,float,kCpu,50000,200,786.554810
IndexAdd1D,float,kCuda,10,500,5.429376
IndexAdd1D,float,kCuda,100,500,5.473216
IndexAdd1D,float,kCuda,500,500,5.420415
IndexAdd1D,float,kCuda,1000,500,5.446976
IndexAdd1D,float,kCuda,2000,500,5.522176
IndexAdd1D,float,kCuda,5000,500,8.593280
IndexAdd1D,float,kCuda,10000,500,5.446400
IndexAdd1D,float,kCuda,50000,200,66.459999
```

csukuangfj:
Ready for review.

qindazhu:
> # Benchmark Results for IndexAdd1D
> `K2_SEED=1`
> 
> MultiSplit is not faster than our current implementation, so I will remove it for IndexAdd and keep it only for
> GetTransposeReodering.
> 
> ## with MultiSplit
> ```
> op_name,dtype,device,problem_size,number_of_iterations,elapsed_us_per_iteration
> IndexAdd1D,int32,kCpu,10,500,0.800133
> IndexAdd1D,int32,kCpu,100,500,1.821995
> IndexAdd1D,int32,kCpu,500,500,5.708218
> IndexAdd1D,int32,kCpu,1000,500,12.047768
> IndexAdd1D,int32,kCpu,2000,500,25.907993
> IndexAdd1D,int32,kCpu,5000,500,55.459976
> IndexAdd1D,int32,kCpu,10000,500,132.425797
> IndexAdd1D,int32,kCpu,50000,200,731.775757
> IndexAdd1D,int32,kCuda,10,500,189.149307
> IndexAdd1D,int32,kCuda,100,500,207.107452
> IndexAdd1D,int32,kCuda,500,500,200.612549
> IndexAdd1D,int32,kCuda,1000,500,202.656052
> IndexAdd1D,int32,kCuda,2000,500,213.698288
> IndexAdd1D,int32,kCuda,5000,500,289.130829
> IndexAdd1D,int32,kCuda,10000,500,299.169983
> IndexAdd1D,int32,kCuda,50000,200,249.831848
> IndexAdd1D,float,kCpu,10,500,0.706196
> IndexAdd1D,float,kCpu,100,500,1.798153
> IndexAdd1D,float,kCpu,500,500,6.646156
> IndexAdd1D,float,kCpu,1000,500,13.014317
> IndexAdd1D,float,kCpu,2000,500,24.866104
> IndexAdd1D,float,kCpu,5000,500,60.870171
> IndexAdd1D,float,kCpu,10000,500,157.862183
> IndexAdd1D,float,kCpu,50000,200,822.705017
> IndexAdd1D,float,kCuda,10,500,206.052353
> IndexAdd1D,float,kCuda,100,500,240.781830
> IndexAdd1D,float,kCuda,500,500,212.909058
> IndexAdd1D,float,kCuda,1000,500,212.137100
> IndexAdd1D,float,kCuda,2000,500,229.809601
> IndexAdd1D,float,kCuda,5000,500,317.291260
> IndexAdd1D,float,kCuda,10000,500,281.757446
> IndexAdd1D,float,kCuda,50000,200,320.104462
> ```
> 
> ## With `atomicAdd`
> ```
> op_name,dtype,device,problem_size,number_of_iterations,elapsed_us_per_iteration
> IndexAdd1D,int32,kCpu,10,500,0.734329
> IndexAdd1D,int32,kCpu,100,500,1.636028
> IndexAdd1D,int32,kCpu,500,500,7.345676
> IndexAdd1D,int32,kCpu,1000,500,14.644146
> IndexAdd1D,int32,kCpu,2000,500,29.421806
> IndexAdd1D,int32,kCpu,5000,500,56.109905
> IndexAdd1D,int32,kCpu,10000,500,111.427780
> IndexAdd1D,int32,kCpu,50000,200,732.965454
> IndexAdd1D,int32,kCuda,10,500,5.484160
> IndexAdd1D,int32,kCuda,100,500,5.415168
> IndexAdd1D,int32,kCuda,500,500,5.396288
> IndexAdd1D,int32,kCuda,1000,500,5.485184
> IndexAdd1D,int32,kCuda,2000,500,5.449344
> IndexAdd1D,int32,kCuda,5000,500,5.484352
> IndexAdd1D,int32,kCuda,10000,500,7.238592
> IndexAdd1D,int32,kCuda,50000,200,5.557281
> IndexAdd1D,float,kCpu,10,500,0.669956
> IndexAdd1D,float,kCpu,100,500,1.741886
> IndexAdd1D,float,kCpu,500,500,6.268024
> IndexAdd1D,float,kCpu,1000,500,13.950348
> IndexAdd1D,float,kCpu,2000,500,23.530006
> IndexAdd1D,float,kCpu,5000,500,58.936119
> IndexAdd1D,float,kCpu,10000,500,148.161880
> IndexAdd1D,float,kCpu,50000,200,786.554810
> IndexAdd1D,float,kCuda,10,500,5.429376
> IndexAdd1D,float,kCuda,100,500,5.473216
> IndexAdd1D,float,kCuda,500,500,5.420415
> IndexAdd1D,float,kCuda,1000,500,5.446976
> IndexAdd1D,float,kCuda,2000,500,5.522176
> IndexAdd1D,float,kCuda,5000,500,8.593280
> IndexAdd1D,float,kCuda,10000,500,5.446400
> IndexAdd1D,float,kCuda,50000,200,66.459999
> ```

What's the data distribution of your benchmark? The original approach (with `AtomicAdd`) is supposed to be faster if there's no so much duplicate values in the `index`.

csukuangfj:
https://github.com/k2-fsa/k2/blob/55f7d2e0a646825f2784450b21ed66252097cd22/k2/csrc/test_utils.cu#L82-L95
```cpp
 std::vector<int32_t> indexes(dim);
  int32_t start = allow_minus_one ? -1 : 0;
  for (int32_t &i : indexes) {
    int32_t tmp = RandInt(-max_value, max_value);
    i = std::max(tmp, start);
  }
```

It is generated randomly, so there should be duplicates. `MultiSplit` involves several kernel calls, which should be
considered as overhead.

k2/csrc/test_utils.h
```diff
@@ -98,6 +98,22 @@ void CheckRowSplits(RaggedShape &shape,
...
+   @param [in] allow_minus_one
+                        If true, the returned array will contain values
+                        in the range [-1, max_value]; [0, max_value] otherwise.
+   @param [in] dim        It specifies the length of the returned array.
+   @param [in] max_value  It specifies the maximum value the returned array can
+                          contain.
+
+   @return  Return a 1-D array with the given `dim`.
+ */
+Array1<int32_t> GenerateRandomIndexes(ContextPtr context, bool allow_minus_one,
```
qindazhu, line 114: I think we don't really need this function, RandomUniformArray does exactly what we do here, you can just pass min_value = 0 / -1 to RandomUniformArray to get the result.

qindazhu: Oh, I see, you want to generate many -1s here

k2/csrc/multisplit.h
```diff
@@ -0,0 +1,191 @@
...
+  K2_CHECK_GT(num_elements, 0);
+  K2_CHECK_GT(num_buckets, 0);
+  int32_t log_buckets = static_cast<int32_t>(ceilf(log2f(num_buckets)));
+  cudaStream_t stream = context->GetCudaStream();
+
+  constexpr int32_t kMultiSplitNumWarps = 8;
+  int32_t num_threads = kMultiSplitNumWarps * 32;
+  int32_t num_blocks = (num_elements + num_threads - 1) / num_threads;
+
+  Array1<uint32_t> d_mask(context, num_elements, 0);
```
qindazhu, line 118: why do we need init `d_mask` with 0?

k2/csrc/multisplit.h
```diff
@@ -0,0 +1,191 @@
...
+  const uint32_t *mask_data = in_keys;
+
+  if (bucket_mapping_func != nullptr) {
```
qindazhu, line 170: I would suggest we declare `d_mask` in the `if` branch so that we don't do memory allocation when `bucket_mapping_func`. you can do as below:
```
Array<uint32_t> d_mask;
if(bucket_mapping_func != nullptr){
d_mask = Array1<unint32_t>(context, num_elements);
....
}
```

k2/csrc/multisplit.h
```diff
@@ -0,0 +1,191 @@
...
+    K2_CUDA_SAFE_CALL(MarkBinsGeneral<<<num_blocks, num_threads, 0, stream>>>(
+        in_keys, num_elements, *bucket_mapping_func, d_mask.Data()));
+    mask_data = d_mask.Data();
+  }
+
+  Array1<uint32_t> d_out(context, num_elements, 0);
```
qindazhu, line 176: I think we don't need init it with 0

k2/csrc/multisplit.h
```diff
@@ -0,0 +1,191 @@
...
+  size_t temp_storage_bytes = 0;
+
+  K2_CUDA_SAFE_CALL(cub::DeviceRadixSort::SortPairs(
```
qindazhu, line 179: if `bucket_mapping_func == nullptr`, we are doing exactly `sorting keys` here, so we can just call `cub::DeviceRadixSort::SortKeys`, that would be much faster the current approach (i.e. no extra memory allocation)

k2/csrc/multisplit.h
```diff
@@ -0,0 +1,191 @@
...
+  Array1<uint64_t> d_key_value_pairs(context, num_elements);
+  K2_CUDA_SAFE_CALL(
+      PackingKeyValuePairs<<<num_blocks, num_threads, 0, stream>>>(
+          num_elements, in_keys, in_values, d_key_value_pairs.Data()));
+
+  Array1<uint32_t> d_out(context, num_elements, 0);
+  size_t temp_storage_bytes = 0;
+
+  K2_CUDA_SAFE_CALL(cub::DeviceRadixSort::SortPairs(
+      nullptr, temp_storage_bytes, mask_data, d_out.Data(),
```
qindazhu, line 136: Again, if `bucket_mapping_func == nullptr` , we can just calll `SortPairs`, there's no need to pack and unpack.

qindazhu: But, for the case `bucket_mapping_func != nullptr`, I doubt if the caller really need `keys_out`, they may only cares the `values_out`. Anyway, they can always pass `mapped_keys` as keys and `Range(0, num_elems)` as values here, then they can gather `keys_out` and `values_out` with the re-ordered `Range(0, num_elems)` within only one lambda, much simpler than current `pack` and `unpack` approach.

k2/csrc/multisplit.h
csukuangfj: > we can just call cub::DeviceRadixSort::SortKeys

How can we get a sorted value as well as a sorted key at the same time?

k2/csrc/multisplit.h
csukuangfj: You are right, it's not needed.

k2/csrc/multisplit.h
```diff
@@ -0,0 +1,191 @@
...
+ * @param [in] out_keys     Pointer to the output keys array. If not NULL,
+ *                          it will contain a sorted version of `in_keys`. The
+ *                          sort is performed according to the bucket
+ *                          information of the input elements and it is sorted
+ *                          in ascending order.
+ * @param [in] out_values   Pointer to the output values array. If not NULL,
+ *                          it will contain a sorted version of `in_values`.
+ */
+template <typename Lambda>
+void MultiSplitKeyValuePairs(ContextPtr context, int32_t num_elements,
```
qindazhu, line 105: With below comments, I would suggest we change the funcion name to SortPairs, no `bucket_mapping_func`.
In the case where in_keys are the true key, we just call `cub::SortPairs`.
In the case where in_keys are the mapped keys, user can just pass values as `Range(0, num_elements)` and get `keys_out` and `values_out` in another lambda, of course we can wrapp the two steps into a single function (something like `GetSortedIndex`)

qindazhu: Or  they don't care about `keys_out` they can just pass into `values` here.

k2/csrc/multisplit.h
csukuangfj: what if in_keys and out_keys share the same memory?

k2/csrc/multisplit.h
csukuangfj: what if one of `out_keys` and `out_values`  is a nullptr?

k2/csrc/ragged_ops.cu
```diff
@@ -1143,7 +1144,26 @@ Array1<int32_t> GetTransposeReordering(Ragged<int32_t> &src, int32_t num_cols) {
...
+  Array1<int32_t> order = Range(context, num_elements, 0);
+  Array1<int32_t> ans(context, num_elements);
+
+  auto bucket_mapping_lambda =
+      [values_data] __device__(uint32_t i) -> uint32_t {
+    return values_data[i];
+  };
+
+  MultiSplitKeysOnly(context, num_elements, num_buckets, &bucket_mapping_lambda,
+                     reinterpret_cast<const uint32_t *>(order.Data()),
```
qindazhu, line 1163: How about we just pass `values_data` as the keys and pass order as the values to call `MultiSplitKeyValuePaires`? No extra memory allocation compared to current approach but saving a mapping lambda?

k2/csrc/multisplit.h
qindazhu: 1. Let's test if `cub` supports case where `in_keys == out_keys`. I guess it may support (as it does in ExclusiveScan)
2. If it doesnot support, for the case where `bucket_mapping_func != nullptr`, usually the user has already generated the mapped keys, we don't need to generate it again (e.g. in GetTransposeReodering), so we can just call `SortPairs` with the mapped keys as the key and the keys as the values

k2/csrc/multisplit.h
qindazhu: It's meaningless if a user sort key-value pairs but don't need the output, in that case (one of them is nullptr), what they should call is SortKeys

k2/csrc/multisplit.h
csukuangfj: The function name is from CUDPP and I think MultiSplit is a well known
algorithm and it helps explain the meaning of the arguments.

k2/csrc/ragged_ops.cu
csukuangfj: I think we want to get a permutation of [0, 1, 2, 3, 4, ..., N].

If we pass `values_data`, the returned value would be a permutation of `values_data`.

k2/csrc/multisplit.h
csukuangfj: With a bucket mapping function, it is more flexible without introducing much overhead, I think.

qindazhu: I'm OK with the name even though we use Sort approach here instead of Historgram approach (as you said before it only supports bins < = 32), the point is, as I commented before, if `bucket_mapping_func==nullptr`, we can just call SortPairs ( I noticed you have updated this); If `bucket_mapping_func != nullptr`, we still don't need the pack-then-unpack approach, as I said here:

> In the case where in_keys are the true key, we just call cub::SortPairs.
In the case where in_keys are the mapped keys, user can just pass values as Range(0, num_elements) and get keys_out and values_out in another lambda, of course we can wrapp the two steps into a single function (something like GetSortedIndex)

We can save one kenerl call here and another advantage is it can support any value types. Another point is, the caller usually have generated the mapped_keys (as in `GetTransposeReorder`), we don't need map it again with `bucket_mapping_func`.

k2/csrc/multisplit_test.cu
```diff
@@ -0,0 +1,102 @@
...
+    return i;
+  };
+
+  uint32_t (*bucket_mapping)(uint32_t) = nullptr;
+  for (int32_t i = 0; i != 2; ++i) {
+    Array1<int32_t> out_keys(context, k.size());
+
+    if (i == 0) {
+      MultiSplitKeysOnly(context, num_elements, num_buckets, bucket_mapping,
+                         reinterpret_cast<const uint32_t *>(in_keys.Data()),
```
qindazhu, line 83: @csukuangfj wondering if you tested with `&in_keys == &out_keys`? just curious if it  works in cub.

k2/csrc/ragged_ops.cu
qindazhu: No, I mean, you pass `value_data` as the keys (as you did with `bucket_mapping_lambda`?) and pass `order = Range(context, num_elements, 0) ` as `values` (`ans` here) to call `SortKeyValuePairs`

csukuangfj: I use RadixSort directly in this case.

k2/csrc/multisplit_test.cu
csukuangfj: Yes, it works.

k2/csrc/multisplit.h
csukuangfj: I think MultiSplit should be a general interface. `GetTransposeReordering` is just a special case.
I just used `RadixSort` directly in `GetTransposeReordering` and now `MultiSplit` is not
used at all in k2. I am going to remove it.

k2/csrc/multisplit_test.cu
qindazhu: Thanks

csukuangfj:
It turns out `MultiSplit` is not needed in k2 and I have removed it.
RadixSort is used to implement `GetTransposeReordering` and the benchmark result shows that
it is still faster than the current segmented sort based approach.

csukuangfj:
Ready to review. There is now much less code to review.

k2/csrc/ragged_ops.cu
```diff
@@ -1143,7 +1144,32 @@ Array1<int32_t> GetTransposeReordering(Ragged<int32_t> &src, int32_t num_cols) {
...
+  Array1<int32_t> ans = Range(context, num_elements, 0);
+
+  cudaStream_t stream = context->GetCudaStream();
+
+  size_t temp_storage_bytes = 0;
+  K2_CUDA_SAFE_CALL(cub::DeviceRadixSort::SortPairs(
+      nullptr, temp_storage_bytes, src.values.Data(), tmp_values.Data(),
+      ans.Data(), ans.Data(), num_elements, 0, log_buckets, stream));
+
+  Array1<int8_t> d_temp_storage(context, temp_storage_bytes);
```
qindazhu, line 1164: Not sure if it works to pass `ans.Data()` as `tmp_values` above and allocate d_temp_storage and tmp_values together within in an Array1...though it may looks strange but save a memory allocation.

qindazhu: Forget it, current code is much cleaner.

qindazhu:
+1 (though it's a pity that cudpp does not bucket >=32 with historgram approach)

csukuangfj:
You can check how Multisplit is implemented in CUDPP by visiting the following link:
https://github.com/cudpp/cudpp/blob/9dc7357ee818498a87bc2f68f9d887c298e0e3f6/src/cudpp/app/multisplit_app.cu#L2098
```cpp
      if (numBuckets <= 32)
        multisplit_key_value(d_keys, d_values, plan->m_d_temp_keys,
            plan->m_d_temp_values, numElements, numBuckets, ms_context,
            CustomBucketMapper(bucketMappingFunc), true);
      else
        reducedBitSortKeyValue(d_keys, d_values, numElements, numBuckets,
            CustomBucketMapper(bucketMappingFunc), plan);
      break;
```

`reducedBitSortKeyValue` is essentially a wrapper around `RadixSort`.
https://github.com/cudpp/cudpp/blob/9dc7357ee818498a87bc2f68f9d887c298e0e3f6/src/cudpp/app/multisplit_app.cu#L1981

qindazhu:
mm, that's why I said it's a pity they don't implement it with **Historgram** approach (instead of Sort approach) for bucket > 32..I was not doubting...I guess maybe there's some difficulty to implement it with historgram approach for bucket>32.

k2/csrc/ragged_ops.cu
csukuangfj: sounds reasonable, fixed, thanks

qindazhu: Ah, thanks, maybe add documentation to say why we do this, i.e. save one memory allocation.

PR #576, opened 2021-01-09
Intersect generic

danpovey:
This PR adds GPU-based intersection.
This is currently suited for graphs of moderate out-degree, as it has an exhaustive (not logarithmic) search process.
I will later add a version or option to use some kind of binary or n-ary search.

It also does not treat epsilons specially.  In the declaration there is a comment explaining how I plan to eventually do that, by making the epsilon self-loops explicit and having a separate epsilon-filtering stage afterward if needed.


k2/csrc/fsa_algo.h
```diff
@@ -243,6 +267,60 @@ bool Intersect(FsaOrVec &a_fsas, int32_t properties_a, FsaOrVec &b_fsas,
...
+                     in `b_fsas`; elements will all be >= 0.
+    @return  Returns composed FsaVec; will satisfy `ans.Dim0() == b_fsas.Dim0()`.
+
+  CAUTION:
+    All states in the result will be accessible except possibly
+    the final-states; but they won't necessarily all be
+    coaccessible (i.e. be able to reach the end), so you may want to call
+    Connect() afterward if that matters to you.
+ */
+FsaVec IntersectDevice(FsaVec &a_fsas, int32_t properties_a,
```
csukuangfj, line 317: It works on both CPU and GPU. Is there a reason to add `Device` to the name?

danpovey: It's because it's the implementation optimized for device (while our other Intersect() would probably be faster for CPU).  We can take as implicit in k2 that code for device will also work on CPU.

k2/csrc/intersect_dense.cu
```diff
@@ -0,0 +1,848 @@
...
+
+    std::vector<int32_t> range(num_fsas_);
+    // fill with 1, 2, .. num_fsas_.
+    std::iota(range.begin(), range.end(), 1);
+    std::vector<RaggedShape> arc_scores_prefixes =
+        GetPrefixes(arc_scores_.shape, range);
+
+    ContextPtr c_cpu = GetCpuContext();
+
+    // This vector, of length T_ + 1, tells us, for each frame 0 <= t <= T, how
```
csukuangfj, line 546: > This vector, of length T_ + 1

It is not necessary of length `T_ + 1`.

`T_` is the longest sequence in the DenseFsaVec, **BUT** it may happen
that this sequence is never used. That is, `a_to_b_map` excludes this
sequence.

When this case happens,   line 562 will cause out-of-bound error,
like the one in https://github.com/k2-fsa/k2/issues/693

PR #585, opened 2021-01-13
Reordering check

danpovey:


danpovey:
This check is not needed, and I disabled it for speed, but there are a couple functions here that might prove useful, so merging.

PR #587, opened 2021-01-13
Start working on backprop for getting arc-scores, will be needed for MBR etc.

danpovey:


danpovey:
This is far from ready, just adding it so you guys know what I am working on.

danpovey:
Guys, it would be nice if someone could help me with finishing and debugging parts of this PR, 
particularly the Python changes.
You can see that I've made a change to the interface of GetBackwardScores(), so it no longer takes
the total scores, and to GetArcScores(), but I haven't propagated those to the python level or tested
them at that level.

And you'll also see where I am going with the changes to how the autograd works.  Some of them
are basically cosmetic changes, others are to enable our future work with things like MBR training, so
we can backprop the arc scores (as opposed to just the total scores).  The C++ interfaces and code
to enable this, is not 100% finished yet; let me first make sure the interfaces are right and then at least
sketch the implementation; I may not debug it.

qindazhu:
Would like to do this, I guess I need to merge this PR locally and work based on this? ( I suppose current code may not build successfully)

danpovey:
Yes.

qindazhu:
OK, taking, thanks

danpovey:
@qindazhu please see my latest commit, I hope it's not too disruptive.
I realized that because the arcs themselves have scores, the use of `arc_scores` for the arc posteriors is confusing,
so I renamed that to `arc_post`; there may be a few instances  not changed still.
Also I made a fix to the C++ backprop interface of GetArcPost(), and sketched out what the python-level backprop
code will look like.

danpovey:
Another TODO, once any compilation issues are fixed again, is to create a branch of snowfall that will work with the current changes, still just doing LF-MMI, but taking into account the small interface changes we made, and checking that it still works.

It would also be nice to propagate the `inner_labels`, called `phones`, in the training and decoding graphs, so we have access to the phone sequences, but just repeated once instead of multiple times.  (These can't really be worked out from the input symbols because of repeats of the same phone).  This will eventually be needed for phone-level MBR training.

qindazhu:
I have built and tested this PR locally, all tests get passed, so I think you can merge it now, I'll work on the C++ code now based on it.

danpovey:
Haowen, for the C++, I propose that you implement BackpropGetArcScores(), I implement BackpropGetBackwardScores() which is more complicated and may actually require the interface to be fixed a bit, and then you implement BackpropGetForwardScores().

k2/python/k2/fsa_algo.py
```diff
@@ -60,7 +61,10 @@ def top_sort(fsa: Fsa) -> Fsa:
 
 
 def intersect(a_fsa: Fsa, b_fsa: Fsa,
-              treat_epsilons_specially: bool = True) -> Fsa:
+              treat_epsilons_specially: bool = True,
+              seqframe_idx_name: Optional[str] = None,
```
csukuangfj, line 65: Could you add some doc for these arguments?

k2/python/k2/fsa_algo.py
```diff
@@ -571,3 +575,10 @@ def invert(fsa: Fsa) -> Fsa:
         ragged_arc, aux_labels, _ = _k2.invert(fsa.arcs, fsa.aux_labels,
                                                need_arc_map)
         return Fsa(ragged_arc, aux_labels)
+
+
+def create_sparse(row_indexes: torch.Tensor, col_indexes: torch.Tensor,
```
csukuangfj, line 580: Can this one be removed as there is a similar version in utils.py that is with more detailed doc?

qindazhu:
> and then you implement BackpropGetForwardScores().

 OK

k2/python/k2/fsa_algo.py
danpovey: Sure, remove it pls.

k2/python/k2/fsa_algo.py
danpovey: Sorry, remove this, it should only have been in intersect_dense.

k2/python/k2/fsa_algo.py
danpovey: BTW we can have those extra args in both the pruned and un-pruned version of dense intersection.

k2/python/k2/autograd.py
```diff
@@ -388,6 +515,19 @@ def intersect_dense_pruned(a_fsas: Fsa, b_fsas: DenseFsaVec,
         frame for any given intersection/composition task. This is advisory,
         in that it will try not to exceed that but may not always succeed.
         You can use a very large number if no constraint is needed.
+      seqframe_idx_name:
+        If set (e.g. to 'seqframe'), an attribute in the output will be created
+        that encodes the sequence-index and the frame-index within that sequence;
+        this is equivalent to a row-index into b_fsas.values, or, equivalently,
+        an element in b_fsas.shape.
+        # TODO: actually implement this.   It can be done by dividing the
```
csukuangfj, line 523: How do we handle the arcs that enter the final state? 

Should we set `seqframe_idx` and `frame_idx` of them to -1?

danpovey: No, the indexes will be set the same way as others, no special case.  In the data that we give to the dense FSA, those frames are like normal frames.

k2/csrc/ragged_ops_inl.h
```diff
@@ -434,6 +434,158 @@ Ragged<T> Index(Ragged<T> &src, Ragged<int32_t> &indexes, bool remove_axis) {
...
+    PairOutputIterator<T> output_iter(output_data);
+    PairMaxOp<T> op;
+    Pair<T> initial_pair { initial_value, -1 };
+
+    // This code is based on the example here:
+    // https://nvlabs.github.io/cub/structcub_1_1_device_segmented_reduce.html
+    std::size_t temp_storage_bytes = 0;
+
+    // the first time is to determine temporary device storage requirements
+    K2_CUDA_SAFE_CALL(cub::DeviceSegmentedReduce::Reduce(
```
qindazhu, line 576: Just a question, wondering if we need the max values as well, if so, maybe we can call `cub::DeviceSegmentedReduce::ArgMax` https://nvlabs.github.io/cub/structcub_1_1_device_segmented_reduce.html#a2c61f69772a76c383ed8d34c6e5a2194 directly

danpovey: We'd need to define an appropriate pair type, which would be a slight hassle.  In general you can easily look up the max value. (I considered that approach.)

BTW the current code, I realized, is redundant: the output iterator stores an offset, which is not needed (can just move the pointer).  Perhaps you could fix that?

qindazhu: OK, will do

k2/python/k2/autograd.py
csukuangfj: Please have a look at the following pull-request
https://github.com/danpovey/k2/pull/2

k2/csrc/ragged.h
```diff
@@ -212,6 +216,59 @@ class RaggedShape {
   std::vector<RaggedShapeLayer> layers_;
 };
 
+
+template <typename T, int MAX_DIM>
+struct ArrayAccessor {
+  T data[MAX_DIM];
+  ArrayAccessor() { }
```
csukuangfj, line 223: I would suggest removing the default constructor and the copy constructor.
The default behavior provided by the compiler is good enough for this case.

The struct becomes a non-POD type after adding them.

k2/csrc/ragged.h
```diff
@@ -212,6 +216,59 @@ class RaggedShape {
...
+};
+
+
+// call this variable `xxx_row_splits_acc`
+template <int MAX_LAYERS>
+struct RowSplitsAccessor {
+  int32_t *ptrs[MAX_LAYERS];  // these are indexed by layer, from 0.
+
+  // row_splits_acc(1) == shape.RowSplits(1), for instance.
+  int32_t *operator () (int32_t layer) { return ptrs[layer - 1]; }
```
csukuangfj, line 239: Do we need to add
```cpp
K2_DCHECK_GE(layer, 1);
K2_DCHECK_LE(layer, MAX_LAYERS);
```

Also, I would propose
```cpp
int32_t *ptr[MAX_LAYERS] = {nullptr};
```

k2/csrc/ragged_ops.cu
```diff
@@ -1871,4 +1871,144 @@ Array1<int32_t> CoveringShapeForwardMap(RaggedShape &src,
   return ans;
 }
 
+
+
+void RaggedShapeAxis0Splitter::Init(RaggedShape &src) {
+  int32_t num_layers = src.NumLayers(),
+    num_layers_out = num_layers - 1,
+              dim0 = src.Dim0();
+  K2_CHECK(num_layers_out <= 4);  // If this fails, add something to the 4s and
```
csukuangfj, line 1880: `K2_CHECK_LE` will give more detailed error info, i.e., it will print the value of `num_layers_out` on failure. `K2_CHECK` only prints `num_layers_out <= 4` but
does not tell us what the value `num_layer_out` is.

k2/csrc/ragged.h
danpovey: I needed to add it because it needs a __host__ __device__ copy constructor (to be transferable to GPU) but if that exists then it doesn't get a default constructor.

danpovey: I don't believe a __device__ copy constructor gets added by default?  (May be wrong though.)

danpovey: oh you're right, it seems to still work after removing them.

csukuangfj: BTW,

```cpp
ArrayAccessor() { }
```
and
```cpp
ArrayAccessor() = default;
```

are not the same. I would recommend the second approach (using `=default`) if you really want to add it.

csukuangfj:
```cpp
template <typename T, int MAX_DIM>
struct ArrayAccessor {
  T data[MAX_DIM];
};

void TestArrayAccessor() {
  ContextPtr c = GetCudaContext();

  ArrayAccessor<int32_t, 2> acc;
  acc.data[0] = 10;
  acc.data[1] = 20;
  Array1<int32_t> arr(c, 2);
  int32_t *arr_data = arr.Data();
  K2_EVAL(
      c, 2, lamda_set, (int32_t i) { arr_data[i] = acc.data[i]; });
  K2_LOG(INFO) << arr;
}
```

prints `[10, 20]`

k2/csrc/ragged.h
```diff
@@ -212,6 +216,59 @@ class RaggedShape {
...
+
+  RowSplitsAccessor(RaggedShape &src);
+
+  __host__ __device__ RowSplitsAccessor(
```
csukuangfj, line 236: Can this one and the one below for row ids also be removed?

k2/csrc/ragged.h
danpovey: This class is intended to be used mostly on GPU; I don't want checks like this as they're slow and bulk up the compiled code.
I don't think it's necessary to initialize unused elements; it'll likely be a segfault if one is used regardless of the actual value.

k2/csrc/ragged.h
```diff
@@ -212,6 +216,52 @@ class RaggedShape {
...
+        }*/
+
+};
+
+// call this variable `xxx_row_ids_acc`
+template <int MAX_LAYERS>
+struct RowIdsAccessor {
+  int32_t *ptrs[MAX_LAYERS];  // these are indexed by layer, from 0.
+
+  // row_ids_acc(1) == shape.RowSplits(1), for instance.
```
csukuangfj, line 249: typo: `shape.RowIds(1)`

k2/csrc/ragged_ops.h
```diff
@@ -82,6 +81,26 @@ void LogSumPerSublist(Ragged<T> &src, T initial_value, Array1<T> *dst_values) {
...
+  Output to an array `max_values` the arg-max within each sub-list along the
+  last axis of `src` i.e. the max taken over the last axis), i.e. the index
+  within `src.values` of the maximum element of that sub-list, or -1
+  if the sub-list was empty or had value <= `initial_value`.  [or maybe
+  < `initial_value`, must check, that may be undefined.]
+
+     @param [in] src            Input ragged array; must have src.NumAxes()
+                                >= 2. src.values is allowed to be empty.
+     @param [out] argmax        Array to which the argmax indexes will be written.
+                                max_values->Dim() == src.TotSize(src.NumAxes() - 1).
```
csukuangfj, line 95: Should this be
```cpp
argmax->Dim() == src.TotSize(src.NumAxes() - 2)
```
?

k2/csrc/ragged_ops_inl.h
```diff
@@ -434,6 +434,158 @@ Ragged<T> Index(Ragged<T> &src, Ragged<int32_t> &indexes, bool remove_axis) {
...
+struct PairMaxOp {
+  __device__ Pair<T> operator() (const Pair<T> &a,
+                                 const Pair<T> &b) const {
+    // NOTE: could specialize this via a union, if T == int32_t, might be
+    // marginally faster.
+    if (a.t > b.t || (a.t == b.t && a.idx > b.idx))
+      return a;
+    else return b;
+  }
+  PairMaxOp() { }
```
csukuangfj, line 522: I have tested and verified locally that you can remove the explicitly provided constructor and copy constructor. 
They are not needed.

k2/csrc/ragged_ops.h
```diff
@@ -82,6 +81,26 @@ void LogSumPerSublist(Ragged<T> &src, T initial_value, Array1<T> *dst_values) {
   ApplyOpPerSublist<T, LogAdd<T>>(src, initial_value, dst_values);
 }
 
+
+/*
+  Output to an array `max_values` the arg-max within each sub-list along the
+  last axis of `src` i.e. the max taken over the last axis), i.e. the index
+  within `src.values` of the maximum element of that sub-list, or -1
+  if the sub-list was empty or had value <= `initial_value`.  [or maybe
```
csukuangfj, line 89: I think it should be `or had value < initial_value`. The implementation prefers the latter one if there is a tie.

k2/csrc/ragged_ops_inl.h
qindazhu: Yeah, it does't need as we did in all ops in utils.h, I have removed them locally.  Also addressed other issues of `ArgMax`, will make a PR later together.

k2/csrc/ragged.h
```diff
@@ -212,6 +216,52 @@ class RaggedShape {
   std::vector<RaggedShapeLayer> layers_;
 };
 
+
+template <typename T, int MAX_DIM>
+struct ArrayAccessor {
```
csukuangfj, line 221: There is an `Array2Accessor` in `array.h`. Shall we also move this one to `array.h`?

k2/csrc/ragged_ops.cu
```diff
@@ -1871,4 +1871,144 @@ Array1<int32_t> CoveringShapeForwardMap(RaggedShape &src,
   return ans;
 }
 
+
+
+void RaggedShapeAxis0Splitter::Init(RaggedShape &src) {
+  int32_t num_layers = src.NumLayers(),
+    num_layers_out = num_layers - 1,
+              dim0 = src.Dim0();
+  K2_CHECK_LE(num_layers_out, 4);  // If this fails, add something to the 4s and
```
csukuangfj, line 1880: `4` and `5` are magic numbers here. Shall we given them a name, e.g.,  place the following code
```cpp
constexpr int32_t kMaxNumLayersOut = 4;
```
in the header?

k2/csrc/ragged_ops.cu
```diff
@@ -1871,4 +1871,144 @@ Array1<int32_t> CoveringShapeForwardMap(RaggedShape &src,
...
+  // set composite_row_splits_ and also those elements of
+  // the output row_splits which are bound to be zero.
+  K2_EVAL(c, dim0 + 1, lambda_set_composite_row_splits, (int32_t i) -> void {
+      int32_t cur_pos = i;
+      composite_row_splits_acc(0, i) = cur_pos;
+      for (int32_t l = 0; l < num_layers; l++) {
+        // The following statement sets the zero at the beginning of each
+        // row_splits, plus a final zero that we write to avoid an if-statement.
+        row_splits_out_acc.data[l][cur_pos + i] = 0;
+        cur_pos = src_row_splits_acc.ptrs[l][cur_pos];
```
csukuangfj, line 1908: `RowSplitsAccessor::operator()` is never used. Should we remove it? Seems it is good enough to access its `ptrs` directly.

k2/csrc/ragged_ops_inl.h
qindazhu: Removed in https://github.com/danpovey/k2/pull/4, also removed copy constructor of those Iterator and added `__forceinline__`

k2/csrc/ragged_ops.cu
```diff
@@ -1871,4 +1871,144 @@ Array1<int32_t> CoveringShapeForwardMap(RaggedShape &src,
...
 
+
+
+void RaggedShapeAxis0Splitter::Init(RaggedShape &src) {
+  int32_t num_layers = src.NumLayers(),
+    num_layers_out = num_layers - 1,
+              dim0 = src.Dim0();
+  K2_CHECK_LE(num_layers_out, 4);  // If this fails, add something to the 4s and
+                                   // 5s here and in the header.
+  K2_CHECK_GT(num_layers, 1);
```
csukuangfj, line 1882: Should it be
```cpp
K2_CHECK_GT(num_layers, 2);
```

k2/csrc/ragged_ops.h
qindazhu: Sure, seems it's a historical issue, also updated the documentation of other RaggedOps

k2/csrc/ragged_ops.cu
danpovey: would rather keep for now in case someone wants to access it with the same ordering as RowSplits(1) etc.

k2/csrc/ragged_ops.cu
danpovey: OK with me, someone could do that, I've had compilation issues with constexpr in the past so am cautious about it.

k2/csrc/ragged.h
danpovey: This is not an array in the same sense, it's a small array accessor (that can be transferred with the lambda).

k2/csrc/ragged.h
danpovey: actually, initialization is convenient for debugging, will add = {nullptr}

danpovey:
It would be nice if someone can help me by writing tests for BackpropGetBackwardScores().
The easiest way would be as follows (for log_semiring=true): set the `backward_scores_deriv_in` to all zeros except for a 1 at the start-state.   Then do BackpropGetBackwardScores(), and the returned derivative should be identical to the posterior as obtained by doing GetArcPost() and exponentiating.  LMK if it fails, or check in the failing test, I can help debug.

qindazhu:
> writing tests for BackpropGetBackwardScores().


Will do after this PR https://github.com/danpovey/k2/pull/4 get merged (hope it can get merged ASAP as the `arc-scores` branch gets larger and larger, we may need to reslove conflicts again and again...)

danpovey:
Merging so we don't diverge too far.  But will keep the commits separate.

qindazhu:
mm, seems you forget merge https://github.com/danpovey/k2/pull/4 before meging..I have closed it and created another PR https://github.com/k2-fsa/k2/pull/593 with same code to master..

csukuangfj:
I think we have to update snowfall as well. `get_tot_scores` is no longer a method of the module `k2`.

qindazhu:
I have created a PR to do this last week https://github.com/k2-fsa/snowfall/pull/73

PR #588, opened 2021-01-13
Add random generators of uniform distribution for CPU and CUDA

csukuangfj:


danpovey:
Great!
Will look to-morrow.

k2/csrc/rand.cu
```diff
@@ -0,0 +1,159 @@
...
+  return state;
+}
+
+template <typename FloatType>
+static void RandCpu(int32_t dim, FloatType *out) {
+  static_assert(std::is_same<FloatType, float>::value ||
+                    std::is_same<FloatType, double>::value,
+                "");
+  // std::uniform_real_distribution returns a number in
+  // the interval [low, high], but we want (low, high].
```
qindazhu, line 58: `uniform_real_distribution` returns `[low, high)`

k2/csrc/rand.h
```diff
@@ -0,0 +1,70 @@
...
+ *                      It can be either a CPU context or a CUDA context.
+ *
+ * @param [in] seed     The target seed.
+ */
+void SetSeed(ContextPtr context, uint64_t seed);
+
+/* Fill the given array with random numbers from a uniform distribution on
+ * the interval (0.0, 1.0].
+ *
+ * 0 is exclusive and 1 is inclusive.
```
qindazhu, line 41: Wondering why do we need this version, i.e. (0, 1], I was thinking we don't really care the boundary and the version `[0,1)` would be enough

qindazhu: this is also the default implementation of `uniform_real_distribution`

k2/csrc/rand.h
```diff
@@ -0,0 +1,70 @@
...
+ * 0 is inclusive and 1 is exclusive.
+ *
+ * `FloatType` can be either `float` or `double`.
+ *
+ * @param [in]  context  It specifies the device on which the random
+ *                       numbers are generated.
+ * @param [in]  dim      The dimension of the returned array.
+ */
+template <typename FloatType>
+Array1<FloatType> Rand(ContextPtr context, int32_t dim) {
```
qindazhu, line 62: Why not just change the implementation of https://github.com/k2-fsa/k2/blob/7f01b9c2427fd1be037fc100740264e4dc66c155/k2/csrc/array_ops.h#L306, I think this is more general interface if we want a random array.

danpovey:
changing the interface is OK with me, but if  it uses half-open intervals
like [begin,end), please check all usages of it and maybe modify the 'end'
args by adding 1
if they are integers, there may be cases where it might cause a crash.

k2/csrc/rand.h
csukuangfj: `curand` returns (0, 1]. To make the behavior consistent for CPU and CUDA, `(0, 1]` is used here.

k2/csrc/rand.h
csukuangfj: 1. It's easier to add new code.
2. Changing it would change its semantics since it says `min_value` is included


BTW, there are several problems with `RandUniformArray1`:

 - When T is a real type, `max_value` is actually not included, but the documentation says it is included
 - It does not cache the states of the generator, which means several calls to`RandUniformArray1`
   with identical `seed`, `min_value`, and `max_value` will return an identical result.
- It is inefficient to re-construct the generator for each call

k2/csrc/rand.h
csukuangfj: Will try to reverse (0, 1] to [0, 1) for the CUDA part.

csukuangfj:
Currently, only uniform distribution is implemented. Do we need to support other kinds of distributions? @danpovey

danpovey:
Let's add them only as required.

danpovey:
Thanks!  Merge when you think  it's ready

csukuangfj:
Merging.

PR #589, opened 2021-01-15
Fix array test index bug.

yaguanghu:
There seems to be a mistake in the test logic of array3. Is my understanding right?


qindazhu:
Yeah, you're right, thanks! 

(BTW,  code under `k2/csrc/host` are the old code we wrote that work only on CPU, while code directly under `k2/csrc` are the latest code we write and maintain which work for both GPU and CPU, we may also wrap some code under `k2/csrc/host` to `k2/csrc/host_shim.h` for temporary usage, but definitely we may remove those host code finally. )

yaguanghu:
OK, I see. Maybe I can help to do this.
I noticed that there is not Array3 defination and implementation in k2/csrc/array.h, isn't is used?

qindazhu:
We don't need Array3 for now, instead we use `RaggedShape` and `Ragged` to describe Fsa(Vec).

Issue #567, opened 2021-01-04
Simplify header format?


danpovey:
Is there opportunity to simplify the header format?  I believe we have some doxygen-style markup in there that we are not actually making use of (?).  I suggest to change, for example:
```
/**                                                                                                                                                                                               
 * @brief                                                                                                                                                                                         
 * host_shim                                                                                                                                                                                      
 *                                                                                                                                                                                                
 * @copyright                                                                                                                                                                                     
 * Copyright (c)  2020  Xiaomi Corporation (authors: Daniel Povey, Haowen Qiu)                                                                                                                    
 *                                                                                                                                                                                                
 * @copyright                                                                                                                                                                                     
 * See LICENSE for clarification regarding multiple authors                                                                                                                                       
 */
```
to:
```
/*                                                                                                                                                                                               
 * Copyright (c)  2020  Xiaomi Corporation (authors: Daniel Povey, Haowen Qiu)                                                                                                                    
 *
 * See LICENSE for clarification regarding multiple authors                                                                                                                                       
 */
```
I know we have doxygen-style comments, but I actually do that for the sake of human readability as text
and to have a consistent format, rather than an expectation that we'll ever compile documentation with doxygenm.

qindazhu:
IIRC, @songmeixu changed all headers from the below one to the above one in previous PR, not sure if he has any comment about this or wants to take the task.

yaguanghu:
I can take this task.

danpovey:
Great, thanks!

PR #590, opened 2021-01-16
Simplify header format.

yaguanghu:
#567 
I keep the previous code comments after copyright. Should it be deleted or moved to somewhere else?


danpovey:
Looks great!  Will merge.

PR #597, opened 2021-01-19
Fixed some issues of Backprop* functions and added test for BackpropGetBackwardScores

qindazhu:
Will add other tests in the next PR, just make a PR to commit current fixes, especially the NaN issue.

### DIFF
--- a/k2/csrc/fsa_utils.cu
+++ b/k2/csrc/fsa_utils.cu
@@ -1211,7 +1211,7 @@ void BackpropGetArcPost(FsaVec &fsas, Ragged<int32_t> &incoming_arcs,
   // compute backward_scores_deriv
   Array1<int32_t> incoming_arc_post_deriv =
       arc_post_deriv[incoming_arcs.values];
-  Ragged<FloatType> ragged_backward_scores_deriv(fsas.shape,
+  Ragged<FloatType> ragged_backward_scores_deriv(incoming_arcs.shape,
                                                  incoming_arc_post_deriv);
   SumPerSublist(ragged_backward_scores_deriv, 0, backward_scores_deriv);
   // set the forward_scores_deriv for the final state and backward_scores_deriv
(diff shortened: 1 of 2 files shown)


k2/csrc/fsa_utils.cu
```diff
@@ -1616,27 +1640,24 @@ Array1<FloatType> BackpropGetForwardScores(
...
       Ragged<FloatType> leaving_arc_deriv(leaving_arc_batch.shape);
       FloatType *leaving_arc_deriv_data = leaving_arc_deriv.values.Data();
 
       K2_EVAL(
-          c, num_arcs, lambda_set_arc_deriv_etc, (int32_t arc_idx)->void {
+          c, leaving_arc_batch.NumElements(), lambda_set_arc_deriv_etc,
+          (int32_t arc_idx)->void {
             int32_t fsas_arc_idx012 = leaving_arc_batch_data[arc_idx];
-            Arc arc = arcs[fsas_arc_idx012];
+            const Arc &arc = arcs[fsas_arc_idx012];
```
csukuangfj, line 1660: Why to use a reference here? Is `Arc arc` using local memory?

qindazhu: i think we don't need to copy to local memory here

danpovey:
Cool!  Merge if you think it's ready.

k2/csrc/fsa_utils.cu
```diff
@@ -1237,6 +1237,17 @@ void BackpropGetArcPost(FsaVec &fsas, Ragged<int32_t> &incoming_arcs,
       });
 }
 
+template <>
```
csukuangfj, line 1240: This is a specialization, not an instantiation.

`BackpropGetArcPost` is never compiled.

Fixed in #599.

qindazhu: Yeah, noted I fixed in PR https://github.com/k2-fsa/k2/pull/600, there may be conflicts with you, sorry.

PR #602, opened 2021-01-20
Change build_ctc_topo to put the blank symbol into `labels`.

csukuangfj:


danpovey:
LGTM, perhaps one of the other guys can merge if they think it's OK.

k2/python/tests/ctc_gradients_test.py
```diff
@@ -23,8 +23,8 @@ def build_ctc_topo(tokens: List[int]) -> k2.Fsa:
     symbols to a single output symbol.
 
     Caution:
-      The resulting topo is an FST. Epsilons are on the left
-      side (i.e., ilabels) and tokens are on the right side (i.e., olabels)
+      The resulting topo is an FST. Epsilons are on the right
+      side (i.e., olabels) and tokens are on the left side (i.e., olabels)
```
qindazhu, line 27: left side is ilabels?

danpovey:
Actually the whole comment is not well worded IMO.
Maybe  "A token which appears once on the right side (i.e. olabels) may
appear multiple times on the
    left side (ilabels), possibly with epsilons in between."
would be better.

danpovey:
.. or maybe "blanks in between"

qindazhu:
I would suggest we clarify `blank` and `epsilon` in the documentation as well so readers will not be confused.

k2/python/tests/ctc_gradients_test.py
csukuangfj: fixed, thanks.

csukuangfj:
> Actually the whole comment is not well worded IMO.

Thanks, changed as you suggested.

csukuangfj:
> I would suggest we clarify blank and epsilon in the documentation as well so readers will not be confused.

Do you have any idea of how to make it clearer?

yaguanghu:
Maybe keep `epsilon` in ilables is a solution? Then we can use `1` or `num_of_tokens + 1` as `blank` index.

qindazhu:
> Maybe keep `epsilon` in ilables is a solution? Then we can use `1` or `num_of_tokens + 1` as `blank` index.

I think we have used `blank == 0` in C++ intersect code, so defining it as other values may be not a good idea. `when 0 appears on left side, it represents blank` is fine with me.

k2/python/tests/ctc_gradients_test.py
```diff
@@ -36,20 +37,43 @@ def build_ctc_topo(tokens: List[int]) -> k2.Fsa:
...
             if i == j:
-                rules += f'{i} {i} 0 {tokens[i]} 0.0\n'
+                arcs += f'{i} {i} {tokens[i]} 0 0.0\n'
             else:
-                rules += f'{i} {j} {tokens[j]} {tokens[j]} 0.0\n'
-        rules += f'{i} {final_state} -1 -1 0.0\n'
-    rules += f'{final_state}'
-    ans = k2.Fsa.from_str(rules)
-    ans = k2.arc_sort(ans)
+                # we assume that eps has the same ID, i.e., 0, with
```
qindazhu, line 46: I suppose we can delete these comments with above `CAUTION`?

csukuangfj: fixed.

csukuangfj:
Merging.

PR #604, opened 2021-01-21
added assertion for test of tropical version BackpropGetBackwardScores

qindazhu:


qindazhu:
OK, mering as they are just testings and work well.

PR #606, opened 2021-01-21
Add linear fst wrapper.

yaguanghu:
As mentioned in [#77](https://github.com/k2-fsa/snowfall/issues/77)

### DIFF
--- a/k2/python/k2/fsa_algo.py
+++ b/k2/python/k2/fsa_algo.py
@@ -37,15 +37,18 @@ def linear_fsa(symbols: Union[List[int], List[List[int]]]) -> Fsa:
     return fsa
 
 
-def linear_fst(symbols: Union[List[int], List[List[int]]]) -> Fsa:
-    '''Construct an linear FST from symbols.
+def linear_fst(symbols: Union[List[int], List[List[int]]],
+               aux_symbols: Union[List[int], List[List[int]]]) -> Fsa:
+    '''Construct an linear FST from symbols and it's corresponding
+       auxiliary symbols.
 
     Note:
       The scores of arcs in the returned FST are all 0.
-      The input and output labels are the same in this function.
 
     Args:
       symbols:
+        A list of integers or a list of list of integers.
+      aux_symbols:
         A list of integers or a list of list of integers.
 
     Returns:
@@ -54,12 +57,16 @@ def linear_fst(symbols: Union[List[int], List[List[int]]]) -> Fsa:
     '''
     ragged_arc = _k2.linear_fsa(symbols)
     if isinstance(symbols[0], List):
-        for sym in symbols:
+        assert isinstance(aux_symbols[0],
+                          List), 'aux_symbols and symbols does not match.'
+        for sym in aux_symbols:
             sym.append(-1)  # -1 == kFinalSymbol
-        aux_labels = torch.IntTensor(symbols)
+        print(aux_symbols)
+        aux_labels = torch.IntTensor(aux_symbols)
+        print(aux_labels)
     else:
-        symbols.append(-1)  # -1 == kFinalSymbol
-        aux_labels = torch.IntTensor(symbols)
+        aux_symbols.append(-1)  # -1 == kFinalSymbol
+        aux_labels = torch.IntTensor(aux_symbols)
     fsa = Fsa(ragged_arc, aux_labels=aux_labels)
     return fsa
 
@@ -89,7 +96,8 @@ def top_sort(fsa: Fsa) -> Fsa:
     return sorted_fsa
 
 
-def intersect(a_fsa: Fsa, b_fsa: Fsa,
+def intersect(a_fsa: Fsa,
+              b_fsa: Fsa,
               treat_epsilons_specially: bool = True) -> Fsa:
     '''Compute the intersection of two FSAs on CPU.
 
--- a/k2/python/k2/fsa_algo.py
+++ b/k2/python/k2/fsa_algo.py
@@ -40,7 +40,7 @@ def linear_fsa(labels: Union[List[int], List[List[int]]]) -> Fsa:
 def linear_fst(labels: Union[List[int], List[List[int]]],
                aux_labels: Union[List[int], List[List[int]]]) -> Fsa:
     '''Construct an linear FST from labels and its corresponding
-       auxiliary labels.
+    auxiliary labels.
 
     Note:
       The scores of arcs in the returned FST are all 0.
@@ -58,7 +58,7 @@ def linear_fst(labels: Union[List[int], List[List[int]]],
     ragged_arc = _k2.linear_fsa(labels)
     if isinstance(labels[0], List):
         assert isinstance(aux_labels[0],
-                          List), 'aux_labels and labels does not match.'
+                          List), 'aux_labels and labels do not match.'
         flattened_labels = []
         for aux in aux_labels:
             flattened_labels.extend(aux + [-1])  # -1 == kFinalSymbol
--- a/k2/python/k2/fsa_algo.py
+++ b/k2/python/k2/fsa_algo.py
@@ -42,6 +42,7 @@ def linear_fst(symbols: Union[List[int], List[List[int]]]) -> Fsa:
 
     Note:
       The scores of arcs in the returned FST are all 0.
+      The input and output labels are the same in this function.
 
     Args:
       symbols:
@@ -49,7 +50,7 @@ def linear_fst(symbols: Union[List[int], List[List[int]]]) -> Fsa:
 
     Returns:
       An FST if the input is a list of integers.
-      A vector of FSAs if the input is a list of list of integers.
+      A vector of FSTs if the input is a list of list of integers.
     '''
     ragged_arc = _k2.linear_fsa(symbols)
     if isinstance(symbols[0], List):
--- a/k2/python/k2/fsa_algo.py
+++ b/k2/python/k2/fsa_algo.py
@@ -186,15 +186,16 @@ def compose(a_fsa: Fsa,
     '''Compute the composition of two FSAs (currently on CPU).
 
     Note:
-      If there is no `aux_labels` in the input FSAs, it is
-      equivalent to :func:`k2.intersect`.
-      The difference from :func:`k2.intersect` is when a_fsa has the
-      `aux_labels` attribute set.  These are interpreted as output labels
-      (olabels), and the composition involves matching the olabels of a with
-      the ilabels of b.  This is implemented by intersecting the inverse of
+      `a_fsa.aux_labels` is required to be defined.
+
+      For both FSAs, the `aux_labels` attribute is interpreted as output labels,
+      (olabels), and the composition involves matching the olabels of a_fsa with
+      the ilabels of b_fsa.  This is implemented by intersecting the inverse of
       a_fsa (a_fsa_inv) with b_fsa, then replacing the ilabels of the result
       with the original ilabels on a_fsa which are now the aux_labels of
-      a_fsa_inv.
+      a_fsa_inv.  If `b_fsa.aux_labels` is not defined, `b_fsa` is treated as an
+      acceptor (as in OpenFST), i.e. its olabels and ilabels are assumed to be
+      the same.
 
     Args:
       a_fsa:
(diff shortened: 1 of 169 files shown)


k2/python/k2/fsa_algo.py
```diff
@@ -36,6 +37,33 @@ def linear_fsa(symbols: Union[List[int], List[List[int]]]) -> Fsa:
     return fsa
 
 
+def linear_fst(symbols: Union[List[int], List[List[int]]]) -> Fsa:
```
csukuangfj, line 40: @danpovey 
Can you imagine in which situation is this function helpful?

```python
fsa.aux_labels = fsa.labels.clone()
```
can achieve the same goal in case it is really needed.

danpovey: I agree, there is not really any point to this if the ilabels and olabels are the same, because compose() will treat it the same as a transducer with separate olabels and ilabels.

yaguanghu: As #554 mentioned, if an FSA does not have aux_labels, compose is equivalent to intersect. So should we modify compose function to treat an acceptor the same as a transducer?

danpovey: I believe it already does.

yaguanghu: It seems not like that.  
See https://github.com/k2-fsa/k2/blob/bbeb3ea65b1bfe9aec6ba62a06d42a6c78ad19a0/k2/python/k2/fsa_algo.py#L203-L207
I don't thinks this logic will treat an acceptor the same as a transducer.

qindazhu: Agree, the above API is meaningless.

I was thinking we were talking about API like below https://github.com/k2-fsa/snowfall/issues/77#issuecomment-764368954:

```
linear_fst(symbols, aux_symbols) where symbols could be a list or a list of list , aux_symbols could be a list or a ragged tensor (2 axes) from single FSA, and be a list of list or a ragged tensor (3 axes).
```
though I prefer approach `fsa.aux_labels = some_tensor/ragged_tensor` for all cases, but I'm fine with above API.

danpovey:
I think you are mistaken about that.

k2/python/k2/fsa_algo.py
yaguanghu: OK, this api is more general. I will revise it.

danpovey:
Actually you might be right.   I think perhaps these lines:
```
 if not hasattr(b_fsa, 'aux_labels'):
        return intersect(a_fsa, b_fsa, treat_epsilons_specially)
```
should be removed.

danpovey:
Wait, no, I was wrong again, the code later on is wrong for that.
Where it says:
```
   out_fsa.labels = index(a_fsa_inv.aux_labels, a_arc_map)
    out_fsa.aux_labels = index(b_fsa.aux_labels, b_arc_map)
```
it would have to say:
```
    out_fsa.aux_labels = (index(b_fsa.aux_labels, b_arc_map) if defined b_fsa.aux_labels
                                        else out_fsa.labels)
    out_fsa.labels = index(a_fsa_inv.aux_labels, a_arc_map)

```

if inner_labels is not None:
        # out_fsa.`inner_labels` = out_fsa.labels
        setattr(out_fsa, inner_labels, out_fsa.labels)
    out_fsa.labels = index(a_fsa_inv.aux_labels, a_arc_map)
    out_fsa.aux_labels = index(b_fsa.aux_labels, b_arc_map)

danpovey:
Wait, no, there are more problems.  I was sure the logic was right at some point.  I'll make a PR.

k2/python/k2/fsa_algo.py
qindazhu: sorry we can just use list of list (of list)here instead of ragged tensor, Then the input would be more pythonic

csukuangfj:
The `aux_labels` of an FSA is usually embedded in the string form. In which case will it be separated from the string form and be passed as a list of integers to `linear_fst`?

danpovey:
I can imagine situations where someone might have two linear sequences with the same length, corresponding to input and output.

qindazhu:
> In which case will it be separated from the string form and be passed as a list of integers to linear_fst?

Can't python list contain string or char?

BTW, I just thought it may be helpful we can create a function that converts Python `List [of List [of List]]`  to tensor or a ragged tensor. It should be a template, i.e. works for str or int, even float.

csukuangfj:
> Can't python list contain string or char?



A Python list can contain any Python object, including `str`.

csukuangfj:
Could you write some tests for linear_fst ?

Thursday, 21 January 2021, 21:04 +0800 from notifications@github.com  <notifications@github.com>:

k2/python/k2/fsa_algo.py
```diff
@@ -36,6 +37,38 @@ def linear_fsa(symbols: Union[List[int], List[List[int]]]) -> Fsa:
...
+      The scores of arcs in the returned FST are all 0.
+
+    Args:
+      symbols:
+        A list of integers or a list of list of integers.
+      aux_symbols:
+        A list of integers or a list of list of integers.
+
+    Returns:
+      An FST if the input is a list of integers.
```
qindazhu, line 55: I would to say `symbols` (and `aux_symbols)` instead of `input`

k2/python/k2/fsa_algo.py
```diff
@@ -36,6 +37,38 @@ def linear_fsa(symbols: Union[List[int], List[List[int]]]) -> Fsa:
...
+      A vector of FSTs if the input is a list of list of integers.
+    '''
+    ragged_arc = _k2.linear_fsa(symbols)
+    if isinstance(symbols[0], List):
+        assert isinstance(aux_symbols[0],
+                          List), 'aux_symbols and symbols does not match.'
+        for sym in aux_symbols:
+            sym.append(-1)  # -1 == kFinalSymbol
+        aux_labels = torch.IntTensor(aux_symbols).reshape(-1,)
```
qindazhu, line 64: I wonder the behavior here: what's the size of `aux_labels`? it equals to `len(sym)` before appending or after appending? 

BTW, I guess it may be better if we also append -1 to aux_labels as well, so that when we do `invert` on the resulted linear_fst, everything would be fine naturally.

k2/python/k2/fsa_algo.py
```diff
@@ -36,6 +37,38 @@ def linear_fsa(symbols: Union[List[int], List[List[int]]]) -> Fsa:
     return fsa
 
 
+def linear_fst(symbols: Union[List[int], List[List[int]]],
+               aux_symbols: Union[List[int], List[List[int]]]) -> Fsa:
+    '''Construct an linear FST from symbols and it's corresponding
+       auxiliary symbols.
+
```
qindazhu, line 44: Maybe better to add a Caution here to say that we don't support aux_labels as a ragged tensor for now. We can fix it later if we really find we need `linear_fst` to handle this case (i.e. aux_symbols is ragged tensor)

danpovey: I prefer to use `labels` instead of `symbols` here.  I would like to standardize on only using the word 'symbols' in the context of 'symbol table', and use 'labels' elsewhere.

qindazhu: +2, can't agree any more.

csukuangfj: This is a historical issue. 

The C++ API has been using `symbols` and `linear_fsa` just follows the convention. Later on, `symbols` in Python
means symbol table but then argument name for `k2.linear_fsa` remains up to now.

danpovey:
ok.  we should fix.

k2/python/k2/fsa_algo.py
yaguanghu: The size of aux_labels is `len(sym)` after appending, because we should keep the size of aux_labels and arcs the same.
I appened -1 to aux_symbols and then converted it to torch tensor as `aux_labels`

k2/python/k2/fsa_algo.py
yaguanghu: I will change `symbols` to `labels` also in `linear_fsa`.

k2/python/k2/fsa_algo.py
csukuangfj: Does `torch.IntTensor` handle the case when each sublist in `aux_symbols` has a
different length?

k2/python/k2/fsa_algo.py
```diff
@@ -17,25 +18,58 @@
...
     '''
-    ragged_arc = _k2.linear_fsa(symbols)
+    ragged_arc = _k2.linear_fsa(labels)
     fsa = Fsa(ragged_arc)
     return fsa
 
 
+def linear_fst(labels: Union[List[int], List[List[int]]],
+               aux_labels: Union[List[int], List[List[int]]]) -> Fsa:
+    '''Construct an linear FST from labels and it's corresponding
```
csukuangfj, line 42: its corresponding

k2/python/k2/fsa_algo.py
```diff
@@ -17,25 +18,58 @@
...
+       auxiliary labels.
```
csukuangfj, line 43: Please remove the three extra spaces before `auxiliary labels`.

k2/python/k2/fsa_algo.py
```diff
@@ -17,25 +18,58 @@
...
+        A list of integers or a list of list of integers.
+
+    Returns:
+      An FST if the labels is a list of integers.
+      A vector of FSTs if the input is a list of list of integers.
+    '''
+    ragged_arc = _k2.linear_fsa(labels)
+    if isinstance(labels[0], List):
+        assert isinstance(aux_labels[0],
+                          List), 'aux_labels and labels does not match.'
```
csukuangfj, line 61: typo: do not match

k2/python/k2/fsa_algo.py
yaguanghu: fixed it.

danpovey: I think it should be k2.RaggedInt, no?  
also rename all symbols->labels.

k2/python/k2/fsa_algo.py
```diff
@@ -17,25 +18,58 @@
...
+    ragged_arc = _k2.linear_fsa(labels)
+    if isinstance(labels[0], List):
+        assert isinstance(aux_labels[0],
+                          List), 'aux_labels and labels do not match.'
+        flattened_labels = []
+        for aux in aux_labels:
+            flattened_labels.extend(aux + [-1])  # -1 == kFinalSymbol
+        aux_labels = torch.IntTensor(flattened_labels)
+    else:
+        aux_labels.append(-1)  # -1 == kFinalSymbol
```
csukuangfj, line 67: This changes the input `aux_labels` **in-place**. 

I would suggest making a copy of it.

k2/python/k2/fsa_algo.py
yaguanghu: Currently `compose` function seems to support `torch.tensor` only, see
https://github.com/k2-fsa/k2/blob/bbeb3ea65b1bfe9aec6ba62a06d42a6c78ad19a0/k2/python/k2/fsa_algo.py#L209

k2/python/tests/linear_fst_test.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
```
csukuangfj, line 1: Please also add this file to `python/tests/CMakeLists.txt`.

k2/python/k2/fsa_algo.py
danpovey: Even if torch.IntTensor handles that, which I don't think it does, our code internally does not, it expects our own ragged tensor.

danpovey: That is specific to 'compose'.  The FSA format in general supports aux_labels being a k2.RaggedInt.

yaguanghu: So if I use `_k2.RaggedInt`, it would not work with `compose`, is my understanding right?

csukuangfj: > I think it should be k2.RaggedInt, no?

Currently, `RaggedInt` has only two constructors available in Python:
```
    1. _k2.RaggedInt(arg0: str)
    2. _k2.RaggedInt(shape: k2::RaggedShape, values: at::Tensor)
```

Shall we add another constructor supporting `List[List[int]]` ?

yaguanghu: I can take this task 😊

csukuangfj:
+1

danpovey:
I wrote this about 24 hours ago by email, but it bounced due to rate-limit 😡.
"yes, another constructor would be good."

yaguanghu:
I will work on it. But can we merge this change first?

danpovey:
Is this change needed for something?

yaguanghu:
It's not yet. I will append the revised `RaggedInt` constructor to this change.

danpovey:
I guess it's OK to merge this.. if someone wants to.

yaguanghu:
I expect to finish the contructor of `RaggedInt` tomorrow, please wait for that.

csukuangfj:
> I expect to finish the contructor of RaggedInt tomorrow, please wait for that.

Thanks. Let's merge it first. You can create a new pull-request to address it.

Issue #478, opened 2020-12-09
index_select optimization possible


danpovey:
Don't do this just now as I am changing the code where it fails to print more debug info..
The way SimpleRaggedIndexSelect1D tests that each sub-list has only one nonzero element is via a reduction.
There is an easier way that also naturally gives the output.

It's:
  Initialize array `output` of size ragged_input.Dim0(); set it to zero.
  For each element of the ragged array:  if (element is nonzero) set `output[row_of_this_elem] = element` 

  For each element of the ragged array:  if (element is nonzero and  `output[row_of_this_elem] != element` ) there is an error.

Just letting you know it's an available programming pattern.

danpovey:
MM sorry, just realized that's not quite right in case there are duplicate elements in a list, but you could replace `element` in the above with `element_index` and it would work correctly, except there is one more indexing operation to get the output; and you'd have to initialize the array with -1's for that to work.

qindazhu:
Thanks, learned. BTW, just to confirm, These two loops are two differnet kenerls (one by one), right? (otherwise we may get a read-write mismatch on one kernel)
```
For each element of the ragged array: if (element is nonzero) set output[row_of_this_elem] = element

For each element of the ragged array: if (element is nonzero and output[row_of_this_elem] != element ) there is an error.
```
So there are three kenerls in total (another is init the array  with `-1`), the original approach has 3 kenerls as well, I guess the gain may come from drop `SumPerSublist`?

Anyway, learned a new pattern, thanks

danpovey:
Yes, they need to be in different kernels.

There are other issues with SimpleRaggedIndexSelect.
 Firstly, I don't really like this kind of code to be in the torch/ directory- the core implementation code should be in csrc/.
But also it's unnecessarily complicated.  We can just index the tensor with the `values` of the ragged tensor, use that to construct a `Ragged<T>` object using the indexed elements, and then use a simpler operation like `Array1<T> SelectNonzeroElements(Ragged<T> &src);`

danpovey:
I was changing that code but let me not try any more as they would just be overwritten.
I had been trying to make the error message more informative:
```
  const int32_t *counts_data = counts.Data();
  Array1<int32_t> status(context, 1, 0);
  int32_t *status_data = status.Data();
  K2_EVAL(
      context, counts.Dim(), lambda_check_status, (int32_t i)->void {
        if (counts_data[i] > 1) status_data[0] = 1 + i;
      });
  int32_t s = status[0];
[...]

PR #607, opened 2021-01-21
Refactor code of SimpleRaggedIndexSelect

qindazhu:
Fixed some issues metioned by https://github.com/k2-fsa/k2/issues/478. Using approach https://github.com/k2-fsa/k2/issues/478#issuecomment-741527608 to re-implement `SimpleRaggedIndexSelect` (I found `Array1<T> SelectNonzeroElements(Ragged<T> &src)` approach would not be applicable for the requirements, as for empty rows or rows which contains no non-zero elements, we need still output a value 0)

### DIFF
--- a/k2/csrc/tensor_ops.h
+++ b/k2/csrc/tensor_ops.h
@@ -9,6 +9,7 @@
 #define K2_CSRC_TENSOR_OPS_H_
 
 #include "k2/csrc/array.h"
+#include "k2/csrc/ragged.h"
 #include "k2/csrc/tensor.h"
 
 namespace k2 {
@@ -57,6 +58,8 @@ Tensor Cast(Tensor src, Dtype new_dtype);
                      shape (indexes.Dim(), src.Dim(1), src.Dim(2), ...),
                      i.e. with the same num-axes as `src` and
                      axis 0's dim replaced with indexes.Dim().
+                     Noted the ans would be contiguous even though `src`
+                     is not contiguous.
  */
 Tensor Index(Tensor &src, Array1<int32_t> &indexes,
              bool allow_minus_one = true);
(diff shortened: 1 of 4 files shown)


k2/csrc/tensor_ops.h
```diff
@@ -106,6 +109,26 @@ Tensor IndexAdd(Tensor &src, Array1<int32_t> &indexes, int32_t dim,
 void IndexAdd(Tensor &src, Array1<int32_t> &indexes, bool allow_minus_one,
               Tensor *dest);
 
+/*
+  Returns a 1-D Tensor that is a result of indexing 1-D `src` with Ragged array
+  `indexes` whose NumAxes() is 2. ans.Dims()[0] will equal to indexes.Dims0() as
```
danpovey, line 114: you mean Dim0()..

k2/csrc/tensor_ops.h
```diff
@@ -106,6 +109,26 @@ Tensor IndexAdd(Tensor &src, Array1<int32_t> &indexes, int32_t dim,
...
+  we suppose there is at most one non-zero element in `src` for any indexes
+  sub-list in `indexes`.
+
+     @param [in] src  Source 1-D tensor, to be indexed.
+     @param [in] indexes   Indexes to use whose NumAxes() == 2, for any
+                      sub-list `i` in `indexes`, we suppose there is at most
+                      one non-zero element in `src` and we'll set ans[i]
+                      with that non-zero element; if all elements for
```
danpovey, line 122: Please be more precise...  explain what we will set ans[i] to, e.g. is it src[that_element] or that_element?

k2/csrc/tensor_ops.h
qindazhu: Here ans is a `k2::Tensor` so I use `ans.Dims()`

k2/csrc/tensor_ops.h
danpovey: ... also (this is probably my fault for suggesting this interface, but...) it might be more elegant to create a function that would create an `Array1<int32_t>` containing (the nonzero element) if it exists, else -1; and then use our normal IndexSelect() to index with that array.  But this interface is OK too I guess.

danpovey:
I meant indexes.Dims0().

k2/csrc/tensor_ops.h
qindazhu: Oh, yes, sorry, will update

k2/csrc/tensor_ops.h
qindazhu: I suupose you mean return the indexes (or -1) of the nonzero values... will change to that, thanks!

danpovey: wait.. I think I realized what happened here.
This was supposed to be  an optimziation.

danpovey: So don't change the code.. just clarify the documentation.

qindazhu: OK, then the current PR would be ready to merge with my last commit, noted I changed `elements` in documentation to `values in src`, so there should be no misunderstanding

PR #608, opened 2021-01-21
Fix composition w.r.t. b having no aux_labels

danpovey:


danpovey:
@yaguanghu  if you have a use-case for this, would you mind testing it?

k2/python/k2/fsa_algo.py
```diff
@@ -200,11 +201,7 @@ def compose(a_fsa: Fsa,
     '''
     assert a_fsa.is_cpu()
     assert b_fsa.is_cpu()
-    if not hasattr(a_fsa, 'aux_labels'):
-        return intersect(a_fsa, b_fsa, treat_epsilons_specially)
-
-    if not hasattr(b_fsa, 'aux_labels'):
-        return intersect(a_fsa, b_fsa, treat_epsilons_specially)
+    assert hasattr(a_fsa, 'aux_labels')
```
csukuangfj, line 204: From https://github.com/k2-fsa/k2/pull/554#discussion_r549059387

https://github.com/k2-fsa/k2/blob/d5ce23c44daf814eff40f2a1c28ab210f4c2bee3/k2/python/k2/fsa_algo.py#L148-L150

I was assuming that if either FSA does not have aux_labels, `compose`
is equivalent to `intersect`.

The user has to `invert` the input `a_fsas` himself/herself before
calling `intersect`.

csukuangfj: I don't think there exist problems with the current code in the master branch.

danpovey: It only does something nontrivial if a_fsa.aux_labels is defined, so if not I'd rather the user just called intersect().

danpovey: If a_fsa.aux_labels are defined and b_fsa.aux_labels are not defined, the current code is wrong. `intersect` will treat a_fsa as being the wrong way round.

csukuangfj: > intersect will treat a_fsa as being the wrong way round.

That is what I wrote before:

> The user has to invert the input a_fsas himself/herself before calling intersect.

k2/python/k2/fsa_algo.py
```diff
@@ -221,10 +218,14 @@ def compose(a_fsa: Fsa,
     if inner_labels is not None:
         # out_fsa.`inner_labels` = out_fsa.labels
         setattr(out_fsa, inner_labels, out_fsa.labels)
+    out_fsa.aux_labels = (index(b_fsa.aux_labels, b_arc_map)
+                          if defined b_fsa.aux_labels
+                          else out_fsa.labels)
```
csukuangfj, line 223: It's fine if you change it in this way. But I think it should be
```python
else index(b_fsa.labels, b_arc_map)
```

k2/python/k2/fsa_algo.py
danpovey: Ah I see.  But the interface doesn't make sense in that case-- it's inconsistent and not what the user would expect.

danpovey:
Let me merge this now, if there are problems we'll deal with them later.

k2/python/k2/fsa_algo.py
```diff
@@ -221,10 +218,14 @@ def compose(a_fsa: Fsa,
     if inner_labels is not None:
         # out_fsa.`inner_labels` = out_fsa.labels
         setattr(out_fsa, inner_labels, out_fsa.labels)
+    out_fsa.aux_labels = (index(b_fsa.aux_labels, b_arc_map)
+                          if defined b_fsa.aux_labels
```
csukuangfj, line 222:
```python
if hasattr(b_fsa, 'aux_labels')
```

k2/python/k2/fsa_algo.py
```diff
@@ -150,15 +150,16 @@ def compose(a_fsa: Fsa,
...
     Note:
-      If there is no `aux_labels` in the input FSAs, it is
-      equivalent to :func:`k2.intersect`.
-      The difference from :func:`k2.intersect` is when a_fsa has the
-      `aux_labels` attribute set.  These are interpreted as output labels
-      (olabels), and the composition involves matching the olabels of a with
+      `a_fsa.aux_labels` is required to be defined.
+
+      For both FSAs, the `aux_labels` attribute is interpreted as output
+      labels, (olabels), and the composition involves matching the olabels of a with
```
qindazhu, line 156: `olabels of a`  should be  `olabels of a_fsa`, also `ilabels of b_fsa`

csukuangfj:
The changes are not covered by the unit tests, I believe.
As I have been assuming that `k2.compose` is used by FSTs exclusively, there are no tests for `hasattr(b_fsa, 'aux_labels') == False`

danpovey:
sorry, will fix those in separate pull request.  but I have a meeting in a second so no time to add unit-test... please add if you want via separae PR.

danpovey:
BTW the reason I want to handle b_fsa.aux_labels not being defined, is so that we have the same kind of support for "acceptors being equivalent to transducers with both labels the same" that OpenFST uses.

danpovey:
BTW please merge that other PR if you think it's OK.

csukuangfj:
I see. I agree the current fix is less error-prone for users.

yaguanghu:
Sure, I will test this PR and add unit-test if needed.

csukuangfj:
@yaguanghu 
Are you going to write tests for this change?

Issue #326, opened 2020-11-10
from_str problem


danpovey:
```
'''Create an Fsa from a string in the k2 format.                                                                                                                                  
     (See also from_openfst). 
```

danpovey:
Sorry something wrong happened, will try again.
from_str has an issue that it cannot read either:
   - the Fsa with no states
or:
   - the Fsa with a start state and final-state but no arcs.
because it tries to figure out the format from the 1st arc.  That doesn't work if there are no arcs.

Also it would be nice to have the  following clarification to the usage message, if its OK with the style:
```
'''Create an Fsa from a string in the k2 format.                                                                                                                                  
        (See also from_openfst). 
```

danpovey:
... we definitely need to test this for the FSA with no states, and the FSA with no arcs but the final-state could be various values greater than zero, e.g.  1 or 2.  Please always keep in mind corner cases like this !

danpovey:
Incidentally, I'm torn on whether the `acceptor` arg for `from_str` should default to false (as for openfst) or true (which would be more  natural for k2 itself).  Maybe false for consistency would be easier?  Or make it true but also make the openfst one be true or undefined?  IDK.

csukuangfj:
> the Fsa with no states
>or:
>the Fsa with a start state and final-state but no arcs.
>

The problem is that `from_str` invokes 
https://github.com/k2-fsa/k2/blob/72a8768beaf8f55ffecc9de30fa18c05370e9f04/k2/csrc/fsa.h#L204


The vector contains `Arc` information. If there are no arcs, it is not possible to keep the state information.

danpovey:
OK, that function was written earlier, when the text format was different (no final-state) and when my plan was to use arrays of arcs to serialize FSAs.  Now things have changed.  We can give the num-states (== final-state + 1) to this function as an arg, and change it so it no longer attempts to work that out.

danpovey:
@csukuangfj  you'll work on this at some point, right?

csukuangfj:
Yes, I'll start this once `k2.index` is finished. Sorry, I am busy this week.

songmeixu:
Hi, @csukuangfj . I want to ask if this problem has been solved. If you have not had time to deal with it, would you like to let me try to solve it?

csukuangfj:
@songmeixu 

sure, thank you.

PR #611, opened 2021-01-21
FsaFromArray1: deal with num_arcs=0, but num_states != 0.

songmeixu:
To solve issue #326

- Add 'num_states' as an arg to FsaFromArray1()
- Deal with 'num_arcs=0' in FsaFromArray1()
- Get 'num_states' in impl of FsaFromString(), and pass to FsaFromArray1()

### DIFF
--- a/k2/csrc/fsa.cu
+++ b/k2/csrc/fsa.cu
@@ -232,67 +232,84 @@ int32_t GetFsaBasicProperties(const Fsa &fsa) {
   return ans;
 }
 
-Fsa FsaFromArray1(Array1<Arc> &array, bool *error) {
+Fsa FsaFromArray1(Array1<Arc> &arcs, bool *error, int32_t final_state) {
   NVTX_RANGE(K2_FUNC);
-  const Arc *arcs_data = array.Data();
-  ContextPtr &c = array.Context();
-  int32_t num_arcs = array.Dim();
-  // We choose to return an Fsa with no states and no arcs.  We could also have
-  // chosen to return an Fsa with 2 states and no arcs.
-  if (num_arcs == 0)
-    return Fsa(EmptyRaggedShape(c, 2), Array1<Arc>(c, 0));
+  const Arc *arcs_data = arcs.Data();
+  ContextPtr &c = arcs.Context();
+  int32_t num_arcs = arcs.Dim();
   *error = false;
+
+  if (num_arcs == 0 && final_state == -1) {
+    K2_LOG(WARNING) << "Could not convert Array1 to FSA, "
+                       "there are no states";
+    *error = true;
+    return Fsa();
+  }
 
   // If the FSA has arcs entering the final state, that will
   // tell us what the final-state id is.
   // If there are no arcs entering the final-state, we let the final state be
-  // (highest numbered state that has arcs leaving it) + 1, so num_states
+  // (highest numbered state that has arcs leaving it) + 1, so final_state
   // (highest numbered state that has arcs leaving it) + 2.
 
-  // element 0 is num-states, element 1 is error flag that's set to
+  // element 0 is final-state, element 1 is error flag that's set to
   // 0 on error.
 
-  Array1<int32_t> num_states_array(c, 2, -1);
-  int32_t *num_states_data = num_states_array.Data();
+  Array1<int32_t> final_state_array(c, 2, -1);
+  int32_t *final_state_data = final_state_array.Data();
 
   Array1<int32_t> row_ids1(c, num_arcs);  // maps arc->state.
   int32_t *row_ids1_data = row_ids1.Data();
 
+  // from arcs:
+  // get 1) arc->state maps
+  //     2) final_state,
+  // and 3) check final state
   K2_EVAL(
       c, num_arcs, lambda_misc, (int32_t i)->void {
         row_ids1_data[i] = arcs_data[i].src_state;
         if (arcs_data[i].label == -1) {
-          int32_t final_state = arcs_data[i].dest_state;
-          int32_t old_value = num_states_data[0];
-          if (old_value >= 0 && old_value != final_state + 1)
-            num_states_data[1] = 0;  // set error flag.
-          num_states_data[0] = final_state + 1;
+          int32_t final_state_temp = arcs_data[i].dest_state;
+          int32_t old_value = final_state_data[0];
+          if (old_value >= 0 && old_value != final_state_temp) {
+            final_state_data[1] = 0;  // set error flag.
+          }
+          final_state_data[0] = final_state_temp;
         }
       });
-  num_states_array = num_states_array.To(GetCpuContext());
-  int32_t num_states = num_states_array[0], error_flag = num_states_array[1];
+
+  final_state_array = final_state_array.To(GetCpuContext());
+  int32_t final_state_in_arcs = final_state_array[0],
+          error_flag = final_state_array[1];
+
   if (error_flag == 0) {
     K2_LOG(WARNING) << "Could not convert tensor to FSA, there was a problem "
-                       "working out the num-states in the FSA, num_states="
-                    << num_states;
+                       "working out the final-state in the FSA, final_state="
+                    << final_state_in_arcs;
     *error = true;
     return Fsa();
   }
-  if (num_states == -1) {
-    // there was no final arc, so let the final state be the highest-numbered
-    // state that is referenced, plus one.
 
-    Array1<int32_t> max_state(c, num_arcs);
-    int32_t *max_state_data = max_state.Data();
-    K2_EVAL(
-        c, num_arcs, lambda_get_dest_state, (int32_t i)->void {
-          int32_t dest = arcs_data[i].dest_state, src = arcs_data[i].src_state;
-          max_state_data[i] = (dest > src ? dest : src);
-        });
-    Array1<int32_t> max_state0 = max_state.Range(0, 1);
-    Max(max_state, 0, &max_state0);
-    num_states = max_state[0] + 2;
+  if (final_state_in_arcs <= 0) {  // Cannot get valid final_state from arcs.
+    if (final_state <= 0) {  // Cannot get valid final_state from input also.
+      K2_LOG(WARNING) << "Could not convert tensor to FSA, there was a problem "
+                         "working out the final-state in the FSA. No arcs "
+                         "with label -1, and the input final_state is -1.";
+      *error = true;
+      return Fsa();
+    }
+
+    if (num_arcs == 0) {
+      // Get final_state from input, but no arcs.
+      return Fsa(EmptyRaggedShape(c, 2), Array1<Arc>(c, 0));
+    }
+  } else if (final_state == -1) {  // Get final_state from arcs.
+    final_state = final_state_in_arcs;
   }
+
+  K2_CHECK_LE(final_state_in_arcs, final_state) <<
+    "The final_state get from arcs should <= the input final_state."
+    "(The input one may count dead states)";
 
   if (!ValidateRowIds(row_ids1)) {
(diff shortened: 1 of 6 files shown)


k2/csrc/fsa.cu
```diff
@@ -233,17 +233,24 @@ int32_t GetFsaBasicProperties(const Fsa &fsa) {
...
 
+  if (num_states <= 0) {
+    K2_LOG(WARNING) << "Could not convert Array1 to FSA, "
+                       "there are no states";
+    *error = true;
+    return Fsa();
+  }
+
+  if (num_arcs == 0) {
+    return Fsa(EmptyRaggedShape(c, num_states), Array1<Arc>(c, 0));
```
songmeixu, line 251: Here, I'm not sure about how to deal with the super final state. Should it be added in this case (num_arcs == 0)?

k2/csrc/fsa.h
```diff
@@ -205,8 +205,37 @@ std::ostream &operator<<(std::ostream &os, const DenseFsaVec &dfsavec);
...
+       no arcs).
+     - Else if there were some states but no arcs, we'll return the FSA (with
+       'num_states' and no arcs).
+     - Otherwise, we'll let `final_state` be the highest-numbered state
+       that has any arcs leaving or entering it, plus one.  (This FSA
+       has no successful paths but still has states.)
+
+    @param [in] arc         Source tensor.  Caution: the returned FSA will share
+                            memory with this tensor, so don't modify it afterward!
+    @param [in] num_states  The number of states for the returned FSA.
```
danpovey, line 232: I think we need to be clear about what this means... remember that there can be states which have no arcs leaving them, in the middle of other states.  Perhaps what we really mean is that the final_state is the num_states - 1?

k2/csrc/fsa_utils.cu
```diff
@@ -156,11 +162,12 @@ static Fsa K2AcceptorFromStream(std::istringstream &is) {
     }
   }
 
+  num_states = states_set.size();
```
danpovey, line 165: I am concerned about this line; remember, not all states may be seen, that would still be a valid FST.  (e.g. states 0,1,2,10,11,12 are seen).

danpovey (commented):
couple comments...
thanks!

songmeixu (commented):
Here, I'm not sure about how to deal with the super final state. Should it be added in this case (num_arcs == 0)?

k2/csrc/fsa_utils.cu
csukuangfj: Should it be the following:
```cpp
num_states = largest state number in states_set + 1;
```
?

k2/csrc/fsa_utils.cu
```diff
@@ -143,11 +146,14 @@ static Fsa K2AcceptorFromStream(std::istringstream &is) {
...
       int32_t symbol = StringToInt(splits[2]);
       float score = StringToFloat(splits[3]);
+      states_set.insert(src_state);
+      states_set.insert(dest_state);
       arcs.emplace_back(src_state, dest_state, symbol, score);
     } else if (num_fields == 1u) {
       //   0
       // final_state
-      (void)StringToInt(splits[0]);  // this is a final state
+      int32_t final_state = StringToInt(splits[0]);  // this is a final state
```
csukuangfj, line 155: As we assume that the final state always has the largest state number,
is it possible to use 
```cpp
num_states = final_state + 1;
```
?

csukuangfj:
It would be nice if the current changes are covered by some tests.

k2/csrc/fsa.cu
danpovey: No, if there is no start state there should be no super-final state (0 or 2 states are allowed, but not 1 state).

k2/csrc/fsa_utils.cu
qindazhu: If understood correctly, the only purpose we use `states_set` is to get the largest number?? why not just use a single int32_t variable (init with -1 and check >-1 finally) to do this.

k2/csrc/fsa_utils.cu
danpovey: Yes, I think so, but we should probably check that it is the highest numbered state, because it's an error if not.

csukuangfj:
@songmeixu 
Wondering if you have some time to continue with this pull-request?

songmeixu:
> @songmeixu
> Wondering if you have some time to continue with this pull-request?

Y. I am on it today.

csukuangfj:
Here are some suggestions regarding the comments from https://github.com/k2-fsa/k2/issues/326

(1)

> Incidentally, I'm torn on whether the acceptor arg for from_str should default to false (as for openfst) or true (which would be more natural for k2 itself). Maybe false for consistency would be easier? Or make it true but also make the openfst one be true or undefined? IDK.

The above comments are about
https://github.com/k2-fsa/k2/blob/580645b88c5fc4d64cf2a4dd41681fef731fcad6/k2/python/k2/fsa.py#L1056
and
https://github.com/k2-fsa/k2/blob/580645b88c5fc4d64cf2a4dd41681fef731fcad6/k2/python/k2/fsa.py#L1103

We have to change ` def from_str(cls, s: str) -> 'Fsa': ` to deal with 

> the Fsa with no states

Since the current implementation of `from_str` tries to figure out whether the string is an acceptor or a transducer
from the first line of the string, it will abort for an empty string.

One possible solution is to change
https://github.com/k2-fsa/k2/blob/580645b88c5fc4d64cf2a4dd41681fef731fcad6/k2/python/k2/fsa.py#L1091-L1098

to let the user provide an extra argument `acceptor` to `from_str`, just like `from_openfst`.

The C++ parts that need to be changed are
https://github.com/k2-fsa/k2/blob/580645b88c5fc4d64cf2a4dd41681fef731fcad6/k2/csrc/fsa_utils.cu#L161
and
https://github.com/k2-fsa/k2/blob/580645b88c5fc4d64cf2a4dd41681fef731fcad6/k2/csrc/fsa_utils.cu#L228

For an empty string, `finished` should be `false` after the while loop and we can return an empty FSA at this point.

I think the above approach can solve the first issue mentioned in #326 

> - the Fsa with no states

---

(2) The second issue is
> - the Fsa with a start state and final-state but no arcs.

According to the below comment from #326 
> Now things have changed. We can give the num-states (== final-state + 1) to this function as an arg, and change it so it no longer attempts to work that out.

We need to change the following function
https://github.com/k2-fsa/k2/blob/580645b88c5fc4d64cf2a4dd41681fef731fcad6/k2/csrc/fsa.cu#L235
to accept an extra argument `int32_t final_state`. Since `FsaFromArray1` has the final state available as an argument,
its implementation can be simplified a bit by removing the code about figuring out the final state.

----

Hope that the above comments are helpful for resolving #326

songmeixu:
hi, @csukuangfj

About the 1) issue, my solution is to check the `max_state`. And it works also for openfst cases.

About the 2) issue, with 1) solved as above, I add a check about `num_arcs` before `fsafromarray1` call. If it's 0, just return a FSA with two states.

Do they make senses?

csukuangfj:
Can you add tests to cover your changes?






Monday, 8 February 2021, 21:24 +0800 from notifications@github.com  <notifications@github.com>:

songmeixu:
> Can you add tests to cover your changes? 

On it.

ps: The action about CI seems very slow and broken accidentally.

csukuangfj:
> ps: The action about CI seems very slow and broken accidentally.

Please ignore the style check error for CI. I just restarted the failed jobs and everything seems fine.

Yes, the CI is very slow. Building k2 takes time as it contains lots of templates and it has to compile for 7 archs separately.
It takes several minutes on our server even if it has lots of cpu cores. `GitHub actions` uses only one job to compile
k2 because of its limited RAM. That's why it is even slower for GitHub actions.

k2/csrc/fsa_utils.cu
```diff
@@ -228,14 +236,20 @@ static Fsa K2TransducerFromStream(std::istringstream &is,
   K2_CHECK_EQ(finished, true) << "The last line should be the final state";
```
danpovey, line 236: (this relates to a few lines above)...
I don't think we should be ignoring the final_state, as in:
` (void)StringToInt(splits[0]);`.  Instead, I think it's better if the interface of  FsaFromArray1 is extended to
take the final-state.  Otherwise we may do something very unexpected if the input is invalid.

danpovey: .. this code is a bit of a mess because once I had the idea that the best way to represent FSAs was simply arrays of arcs, keeping the final-state out of it, but this turned out to be not really practical.

k2/csrc/fsa.h
```diff
@@ -205,8 +205,40 @@ std::ostream &operator<<(std::ostream &os, const DenseFsaVec &dfsavec);
...
+       just one implicit super final state (set id = 1) and no arcs), and set error = false.
+     - Otherwise, we'll let `final_state` be the highest-numbered state
+       that has any arcs leaving or entering it, plus one.  (This FSA
+       has no successful paths but still has states.)
 
-Fsa FsaFromArray1(Array1<Arc> &arc, bool *error);
+    @param [in] array       Source arcs. Caution: the returned FSA will share
+                            memory with this tensor, so don't modify it
+                            afterward!
+    @param [in] num_states  Here, the num_states means the number of all states
```
danpovey, line 233: Let's change this interface so the id of the final-state is provided, instead of the num_states.  THen there will be no confusion about whether the states are contiguous.

k2/csrc/fsa_utils.cu
```diff
@@ -477,6 +506,8 @@ static Fsa OpenFstTransducerFromStream(std::istringstream &is,
   }
 
   K2_CHECK(is.eof());
+  K2_CHECK_GT(max_state, 0) << "Invalid input stream: "
+                            << "\nOpenFST transducer expects valid states.";
```
danpovey, line 511: .. BTW in order to support multiple aux_labels, I would be OK to just make the `aux_labels` arg to this function an array2, which would probably make the code changes quite minimal.
We can assume that the caller knows how many aux_labels there are.

csukuangfj:
@songmeixu 
FYI: From https://github.com/k2-fsa/k2/pull/444#issuecomment-737866684

> Guys, can you get out of the habit of doing force-push? It makes it hard to view commits.
Better to merge and not rebase while you are working on a branch; I can do 'squash and merge' when merging from github.

songmeixu:
> @songmeixu
> FYI: From [#444 (comment)](https://github.com/k2-fsa/k2/pull/444#issuecomment-737866684)
> 
> > Guys, can you get out of the habit of doing force-push? It makes it hard to view commits.
> > Better to merge and not rebase while you are working on a branch; I can do 'squash and merge' when merging from github.

Hi, due to I recommitted too many times to meet the need. Instead of merging with old ones, I force pushed the new one. I will follow "merge" style in next PR. THX!

csukuangfj:
> > @songmeixu
> 
> > FYI: From [#444 (comment)](https://github.com/k2-fsa/k2/pull/444#issuecomment-737866684)
> 
> > 
> 
> > > Guys, can you get out of the habit of doing force-push? It makes it hard to view commits.
> 
> > > Better to merge and not rebase while you are working on a branch; I can do 'squash and merge' when merging from github.
> 
> 
> 
> Hi, due to I recommitted too many times to meet the need. Instead of merging with old ones, I force pushed the new one. I will follow "merge" style in next PR. THX! 

It can be squashed into one single commit by GitHub automatically at the end, no matter how many commits there are in the pull request.

danpovey:
Thanks!  Let me just test with snowfall before merging..

songmeixu:
> It can be squashed into one single commit by GitHub automatically at the end, no matter how many commits there are in the pull request.

danpovey:
Thanks!
I merged this indirectly via #690

PR #612, opened 2021-01-22
Add Python tutorials for fsa_algo

csukuangfj:
A preview is available at https://k3.readthedocs.io/en/latest/python_tutorials/fsa_algo.html

Screen Shot 2021-01-22 at 10 30 09 AM


docs/source/python_tutorials/fsa_algo.rst
```diff
@@ -0,0 +1,217 @@
...
+
+
+arc_sort
+~~~~~~~~
+
+:func:`k2.intersect` and :func:`k2.compose` require two
+arc-sorted inputs. You can use :func:`k2.arc_sort` to convert
+an unsorted FSA/FsaVec into a sorted FSA/FsaVec.
+
+An FSA is sorted if for each state, its leaving arcs are sorted
```
qindazhu, line 75: In k2 now, what we really do in arc-sort is: sort on label first then on dest_state

docs/source/python_tutorials/fsa_algo.rst
```diff
@@ -0,0 +1,217 @@
...
+    :align: center
+    :figwidth: 600px
+
+.. figure:: images/c_fsa_intersect.svg
+    :alt: c_fsa
+    :align: center
+    :figwidth: 600px
+
+:func:`k2.intersect` has an optional argument ``treat_epsilons_specially``.
+Its default value is `True`. If it is set to `False`, then the label `0`
```
qindazhu, line 144: It may be better to clarify the beavior when `treat_epsilons_specially` is true or false, or add the link https://github.com/k2-fsa/k2/blob/9725fa2bd723acd13277249fa4719a851d9a36ea/k2/csrc/fsa_algo.h#L209 here?

csukuangfj: >  If it is set to False, then the label 0 is treated as a normal label.

In addition, an example is followed by the above statement. Do you have a better way to make it more understandable?

docs/source/python_tutorials/images/fsa_invert.svg
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
```
qindazhu, line 2: Just a question: why we don't make picture above `fsa_aux.png` and `fsa_aux_symbols.png` as svg format?

csukuangfj: It is feasible.

docs/source/python_tutorials/fsa_algo.rst
qindazhu: I just feel our definition of `treat_epsilons_specially` has more meanings than users thought, user may be cannot figure out the meaning correctly from the examples. Anyway, adding the link to the documentation about `treat_epsilon_especially` Dan wrote in fsa_algo.h is fine with me.

csukuangfj: - when `treat_epsilons_specially` is True, 0 is treated as epsilons
- when `treat_epsilons_specially` is False, 0 is treated as normal labels

---


> I just feel our definition of treat_epsilons_specially has **more** meanings than users thought

What are the extra meanings besides the above ones?

qindazhu: > when treat_epsilons_specially is True, 0 is treated as epsilons
    when treat_epsilons_specially is False, 0 is treated as normal labels

mm, we should clarify how the treat_epsilons_specially controls the beavior when matching labels in intersect. If you check the documentation @danpovey wrote in fsa_algo.h, you'll see that we also handle case where input changed by AddEpsilonSelfLoops. Anyway, I don't think users can get the meaining well above documentation.

csukuangfj: > You'll see that we also handle case where input changed by AddEpsilonSelfLoops.


I don't think there is code doing this kind of handling in `Intersect`. All it does is
- when treat_epsilons_specially is True, it also handles epsilon transitions
- when treat_epsilons_specially is False, epsilons are treated as normal labels. They are not special any more.

 
What @danpovey  documents is to explain in which case the behavior is correct if treat_epsilons_specially is False, I think.

---

> we should clarify how the treat_epsilons_specially controls the beavior when matching labels in intersect.

Can you explain why the following two statements confuse the users? I'd like to improve it.

- when treat_epsilons_specially is True, 0 is treated as epsilons
- when treat_epsilons_specially is False, 0 is treated as normal labels

danpovey:
Looks great to me!

PR #614, opened 2021-01-22
Wrap IntersectDevice.

### LINKED ISSUE #609: Wrap GPU intersection code
Need to modify the composition and intersection code in python, so that they also work for inputs  on GPU, if treat_epsilons_specially == false.

csukuangfj:
Closes #609


danpovey:
Cool-- looks OK to me.
In my mind we'd be calling this from the normal intersect() if it detected that it was on GPU.  But I suppose we could just have it print a warning and have the user do this.

csukuangfj:
>In my mind we'd be calling this from the normal intersect() if it detected that it was on GPU

 The interface of `IntersectDevice` is different from `Intersect`. `IntersectDevice` is more powerful.

danpovey:
Oh, I see, at the C++ level, the Intersect() wrapper does call IntersectDevice().

danpovey:
This is OK with me.

csukuangfj:
Will merge.

PR #615, opened 2021-01-22
support `size` in k2.create_sparse.

### LINKED ISSUE #613: create_sparse size
We should add the 'size' argument to `create_sparse`, to set the size of the posterior, e.g. it would be a tuple like (num_frames, highest_phone_plus_one).  To be passed into torch.sparse_coo_tensor.

csukuangfj:
Fixes #613


danpovey:
LGTM!

PR #616, opened 2021-01-24
Fix bugs in k2.compose.

csukuangfj:
`Fsa.labels` returns a reference, so we have to clone it.

### DIFF
--- a/k2/python/k2/fsa_algo.py
+++ b/k2/python/k2/fsa_algo.py
@@ -94,8 +94,7 @@ def top_sort(fsa: Fsa) -> Fsa:
     return sorted_fsa
 
 
-def intersect(a_fsa: Fsa,
-              b_fsa: Fsa,
+def intersect(a_fsa: Fsa, b_fsa: Fsa,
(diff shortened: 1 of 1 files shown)


k2/python/k2/fsa_algo.py
```diff
@@ -252,9 +251,13 @@ def compose(a_fsa: Fsa,
...
-                          if hasattr(b_fsa, 'aux_labels')
-                          else out_fsa.labels)
+
+    if hasattr(b_fsa, 'aux_labels'):
+        out_fsa.aux_labels = index(b_fsa.aux_labels, b_arc_map)
+    else:
+        # need a clone here since `Fsa.labels` is a reference
+        out_fsa.aux_labels = out_fsa.labels.clone()
+
     out_fsa.labels = index(a_fsa_inv.aux_labels, a_arc_map)
```
csukuangfj, line 261: Without `clone()`, `out_fsa.aux_labels` shares the same memory with `out_fsa.labels`. Any change done to either one will be reflected in the other one. `out_fsa` is always an acceptor.

danpovey:
LGTM.

PR #618, opened 2021-01-25
Implement SegmentedExclusiveSum.

### LINKED ISSUE #617: Segmented ExclusiveSum
This is an op I need for something: a segmented version of ExclusiveSum(), like a scan-not-reduce version of SumPerSublist().

csukuangfj:
Closes #617 

Ready for review.

---

A large part of the code is modified from CUDPP, which has a very.different.code.style from k2. All code from CUDPP is placed inside the folder `k2/csrc/cudpp` leaving the code style unchanged.

### DIFF
--- a/k2/csrc/ragged_ops_inl.h
+++ b/k2/csrc/ragged_ops_inl.h
@@ -573,17 +573,19 @@ void SegmentedExclusiveSum(Ragged<T> &src, Array1<T> *dst) {
   uint32_t *flags_data = flags.Data();
 
   Array1<int32_t> row_splits = src.RowSplits(src.NumAxes() - 1);
+  Array1<int32_t> row_ids = src.RowIds(src.NumAxes() - 1);
   const int32_t *row_splits_data = row_splits.Data();
+  const int32_t *row_ids_data = row_ids.Data();
   K2_EVAL(
-      c, row_splits.Dim() - 1, set_flags, (int32_t i)->void {
-        int32_t begin = row_splits_data[i];
-        int32_t end = row_splits_data[i + 1];
-        if (begin < end) {
-          flags_data[begin] = begin == 0 ? 0 : 1;
-
-          for (int32_t k = begin + 1; k != end; ++k) {
-            flags_data[k] = 0;
-          }
+      c, dim, set_flags, (int32_t idx01)->void {
+        int32_t idx0 = row_ids_data[idx01];
+        int32_t idx0x = row_splits_data[idx0];
+        int32_t idx0x_next = row_splits_data[idx0 + 1];
+        if (idx0x < idx0x_next) {
+          if (idx01 == idx0x)
+            flags_data[idx01] = idx01 == 0 ? 0 : 1;
+          else
+            flags_data[idx01] = 0;
         }
       });
 
--- a/k2/csrc/ragged_ops_inl.h
+++ b/k2/csrc/ragged_ops_inl.h
@@ -20,6 +20,7 @@
 #include <vector>
 
 #include "k2/csrc/array_ops.h"
+#include "k2/csrc/cudpp/cudpp.h"
 #include "k2/csrc/macros.h"
 #include "k2/csrc/moderngpu_allocator.h"
 #include "moderngpu/kernel_segsort.hxx"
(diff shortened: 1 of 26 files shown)


danpovey:
Cool!  Looks good to me.

danpovey:
It would be nice if you can get some clarity on whether the input and output arrays are allowed to be the same.
Although I know it's not likely to be well-documented enough.

csukuangfj:
https://github.com/k2-fsa/k2/blob/63e9321e8a42d7c0606f57c3c9d401ad771f1be1/k2/csrc/ragged_ops.h#L1310-L1311

https://github.com/k2-fsa/k2/blob/63e9321e8a42d7c0606f57c3c9d401ad771f1be1/k2/csrc/cudpp/cudpp.h#L31-L32

Does the above doc looks clear?

csukuangfj:
It supports accumulation **in-place**. 

The doc describes this explicitly and the tests cover it.

danpovey:
Great!

k2/csrc/ragged_ops_inl.h
```diff
@@ -559,6 +560,59 @@ void ArgMaxPerSublist(Ragged<T> &src, T initial_value, Array1<int32_t> *dst) {
...
+    SegmentedExclusiveSum(c, src.values.Data(), dim, flags_data, dst->Data());
+    return;
+  }
+
+  K2_CHECK_EQ(c->GetDeviceType(), kCpu);
+
+  const T *src_values_data = src.values.Data();
+  T *dst_data = dst->Data();
+
+  T sum = 0;
```
qindazhu, line 603: It's better if we can put below lines into a if cpu branch, so that we can get the result for cpu in one iteration (with `flags` stuff).

danpovey: .. it effectively is a CPU branch, no?  
Might be clearer if it was in an else { }.  It did slightly confuse me.  But no big deal.

qindazhu:
OK，after looking over the code, I think we can implement `SegmentedExclusiveSum` more clearly without cudpp. As for cudpp version, we still need an extra array `flag` and an extra kenerl to set `flag`, so here we have two arrays (one for `flag` and one for `ans`) and two kenerls, we can implement `SegmentedExclusiveSum` by:

```
Array1<T> exclusive_sum(c, dim);
Exclusive(src.values, &exclusive_sum);
T* ans_data = ans.Data();
start a lambda kenerl, ans_data[idx01] = exclusive_sum[idx01] - exclusive_sum[idx0x]
```
It still uses two kenerls and one extra array, but much simpler. I even think it may faster than the current version as exclusive_sum may be faster than SegmentedScan in cudpp and there's no if-else in the second kenerl. @danpovey what do you think?

danpovey:
That seems to make sense, I suppose it comes down to the speed difference.
In principle, it should be possible to do a much smaller reduction, since
you only need reduction over small ranges typically, but possibly cudpp is
implementing it
as a single big reduction anyway.  Also it's a lot of code.

csukuangfj:
>  I even think it may be faster than the current version

Can you benchmark the method you propose against the current approach?

qindazhu:
I'm adding it (though I believe the code simplity is much more important here)

qindazhu:
FYI, the approach I proposed would be much faster than the current one for long arrays……will make a PR.
```
ops_name, ragged_dim0, ragged_num_elements,....

SegmentedExclusiveSumCuDpp111_521,int32,kCuda,50,100,12.959999
SegmentedExclusiveSumCuDpp112_1005,int32,kCuda,100,100,12.960320
SegmentedExclusiveSumCuDpp208_2004,int32,kCuda,200,50,30.202242
SegmentedExclusiveSumCuDpp557_5001,int32,kCuda,500,20,30.927999
SegmentedExclusiveSumCuDpp1063_10024,int32,kCuda,1000,10,30.895996
SegmentedExclusiveSumCuDpp11024_100050,int32,kCuda,10000,1,36.576000

SegmentedExclusiveSum111_521,int32,kCuda,50,100,20.448959
SegmentedExclusiveSum112_1005,int32,kCuda,100,100,20.384642
SegmentedExclusiveSum208_2004,int32,kCuda,200,50,20.510078
SegmentedExclusiveSum557_5001,int32,kCuda,500,20,20.350399
SegmentedExclusiveSum1063_10024,int32,kCuda,1000,10,20.678398
SegmentedExclusiveSum11024_100050,int32,kCuda,10000,1,27.136000
```

danpovey:
mm, can you remind me what the last element is (time taken vs. speed)?

qindazhu:
Have talked offline, just noted here, it's average microseconds per iteration, updated in above table

PR #619, opened 2021-01-26
 Random paths

danpovey:


danpovey:
have not tried to compile this, very WIP.

k2/csrc/array_ops.h
```diff
@@ -198,6 +198,15 @@ T MaxValue(const Array1<T> &src) {
   return MaxValue(src.Context(), src.Dim(), src.Data());
 }
 
+template <typename T>
+T Sum(const Array1<T> &src) {
```
qindazhu, line 202: The function name `Sum` is a little confused, as `ExclusiveSum` does not count the last element in `src`.

k2/csrc/fsa_utils.h
```diff
@@ -521,6 +521,110 @@ FsaVec FsaVecFromArcIndexes(FsaVec &fsas, Ragged<int32_t> &best_arc_indexes);
 Ragged<int32_t> ComposeArcMaps(Ragged<int32_t> &step1_arc_map,
                                Ragged<int32_t> &step2_arc_map);
 
+
+/*
+  Return a ragged array that represents the cumulative distribution function
+  (cdf) of the probability of entering arcs leaving each state of `fsas`.
```
qindazhu, line 527: `entering arcs leaving` -> `arcs leaving`?

k2/csrc/fsa_utils.cu
```diff
@@ -2096,4 +2097,395 @@ void FixNumStates(FsaVec *fsas) {
...
+  K2_CHECK_LE(fsas.NumAxes(), 3);
+  int32_t state_axis = (fsas.NumAxes() == 3 ? 1 : 0);
+  ContextPtr c = fsas.Context();
+  int32_t num_states = fsas.TotSize(state_axis),
+      num_arcs = fsas.NumElements();
+  Array1<FloatType> state_post(c, num_states);
+  // Use lowest() instead of -infinity() to initialize the sum, to avoid
+  // -infinity - (-infinity) = NaN; but actually this shouldn't really be an
+  // issue as we shouldn't be taking paths with scores equal to -infinity.
+  LogSumPerSublist(fsas,
```
qindazhu, line 2113: I guess `fsas` here is Ragged<FloatType>(fsas.shape, arc_post)?

k2/csrc/fsa_utils.cu
```diff
@@ -2096,4 +2097,395 @@ void FixNumStates(FsaVec *fsas) {
...
+                   std::numeric_limits<FloatType>::lowest(),
```
qindazhu, line 2114: just noted we can also here init with `-infinity` and check `state_post_data != -inifinity` below, I have used this approach for ComputeScores/ArcPost related functions, it works well. But you version would also work well I believe.

k2/csrc/fsa_utils.cu
```diff
@@ -2096,4 +2097,395 @@ void FixNumStates(FsaVec *fsas) {
...
+  Array1<FloatType> arc_inv_tots(c, num_states);
+  FloatType *arc_inv_tots_data = arc_inv_tots.Data();
+  K2_EVAL(c, num_states, lambda_set_inv_tots, (int32_t i) {
+      int32_t begin_arc = fsas_row_splits2[i],
+          end_arc = fsas_row_splits2[i + 1];
+      FloatType inv_tot = 1.0;
+      if (end_arc > begin_arc) {
+        FloatType this_tot = arc_cdf_data[end_arc - 1] +
+            arc_pdf_data[end_arc - 1],
+            inv_tot;
```
qindazhu, line 2164: `inv_tot` has been defined above.

k2/csrc/fsa_utils.cu
```diff
@@ -2096,4 +2097,395 @@ void FixNumStates(FsaVec *fsas) {
...
+template <typename FloatType>
+Ragged<int32_t> RandomPaths(Ragged<FsaVec> &fsas,
+                            Array1<FloatType> &arc_cdf,
+                            const Array1<int32_t> &num_paths,
+                            Ragged<int32_t> &state_batches) {
+  using namespace random_path_internal;
+  K2_CHECK_EQ(fsas.NumAxes(), 3);
+  K2_CHECK_EQ(fsas.NumElements(), arc_cdf.Dim());
+  K2_CHECK_EQ(fsas.Dim0(), num_paths.Dim());
+  K2_CHECK(IsCompatible(fsas, arc_cdf));
```
qindazhu, line 2212: Just noted as we have defined `GetContext` long long ago, so it would be much simpler to use it for context checking stuff, i.e. `ContextPtr c = GetContext(fsas, arc_cdf, num_paths, state_batches)`

k2/csrc/fsa_utils.cu
```diff
@@ -2096,4 +2097,395 @@ void FixNumStates(FsaVec *fsas) {
...
+  K2_CHECK(IsCompatible(fsas, state_batches));
+
+
+  ContextPtr c = fsas.Context();
+  int32_t num_fsas = fsas.Dim0();
+
+  // For each FSA, work out the number of batches of states that it has,
+  // which will tell us how much memory we need to allocate.
+  Array1<int32_t> num_state_batches(c, num_fsas);
+  num_state_batches = -1;  // so we can more easily detect errors..
```
qindazhu, line 2223: maybe it's more clear to pass -1 as the 3rd parameter in above construction code of `num_state_batches`? though they are equivalent..

k2/csrc/fsa_utils.cu
```diff
@@ -2096,4 +2097,395 @@ void FixNumStates(FsaVec *fsas) {
...
+         num_states = state_batches_row_splits2[this_batch_start + i + 1] -
+          state_batches_row_splits2[this_batch_start + i];
+      if (num_states == 0) {
+        num_state_batches_data[i] = 0;
+      } else {
+        int32_t prev_batch_start = state_batches_row_splits1[b - 1],
+            prev_num_states = state_batches_row_splits2[prev_batch_start + i + 1] -
+              state_batches_row_splits2[prev_batch_start + i];
+        if (prev_num_states != 0)
+          num_state_batches_data[i] = 0;
```
qindazhu, line 2239: I guess this function has not finished, but just a question, wondering why not we pass `GetStatesBatch(fsas, false)` as `state_batches` here? Is is becasue we called `GetStatesBatch(fsas, true)` for all GetForward/Backward computation? Can we cache another variable like `state_batch_transpose_` in Python code?

k2/csrc/array_ops.h
danpovey: Will fix.

csukuangfj: Fixed in #623

k2/csrc/fsa_utils.cu
danpovey: Mm.  it's true that it would be more efficient.  But let's wait to see whether there's any problem here. 
When we're ready to test things at the python level I'll look at the nsys output.

k2/csrc/array_ops.h
qindazhu: mm, I don't think we need the function to sum all values in Array, I just feel what we want is the last element of `exclusive_sum` as @danpovey implemented, I commented as I think the name `Sum` may be confusing here.

csukuangfj: You don't need to sum all values. You can specify a range to sum.

danpovey: No, I meant all values.  The user could specify a range by calling it with a sub-range of the array, if needed.

k2/csrc/fsa_utils.cu
```diff
@@ -2149,4 +2150,407 @@ void FixNumStates(FsaVec *fsas) {
...
+  */
+  FloatType *arc_cdf_data = arc_cdf.Data();
+  // `arc_inv_tots` will contain the inverses of the totals of the arc_post values
+  // leaving each state, if those totals were nonzero and finite; and otherwise, 1.0.
+  Array1<FloatType> arc_inv_tots(c, num_states);
+  FloatType *arc_inv_tots_data = arc_inv_tots.Data();
+  K2_EVAL(c, num_states, lambda_set_inv_tots, (int32_t i) {
+      int32_t begin_arc = fsas_row_splits2_data[i],
+          end_arc = fsas_row_splits2_data[i + 1];
+      FloatType inv_tot = 1.0;
```
csukuangfj, line 2213: I would suggest 
```cpp
FloatType inv_tot = FloatType(1.0);
```

`1.0` is by default of type `double`.

k2/csrc/fsa_utils.cu
```diff
@@ -2149,4 +2150,407 @@ void FixNumStates(FsaVec *fsas) {
...
+      arc_post -= state_post_data[state_idx];
+      arc_pdf_data[i] = exp(arc_post);
+    });
+
+
+  Ragged<FloatType> arc_pdf_ragged(fsas.shape, arc_pdf);
+
+  Array1<FloatType> arc_cdf(c, num_arcs);
+  // TODO: check, the following works with in-place.
+  ExclusiveSumPerSublist(arc_pdf_ragged, 0, &arc_cdf);
```
csukuangfj, line 2190: Can `arc_pdf` be reused even if this function supports in-place operation?

The last element of `arc_pdf` will be needed. But if it is modified in place,
then the last element of `arc_pdf` is overwritten and its original value is lost.

k2/csrc/fsa_utils.h
```diff
@@ -521,6 +521,110 @@ FsaVec FsaVecFromArcIndexes(FsaVec &fsas, Ragged<int32_t> &best_arc_indexes);
...
+  Return pseudo-randomly chosen paths through acyclic FSAs.  (Actually the paths
+  are deterministic, taken at fixed intervals through a certain cdf).
+
+    @param [in] fsas  An FsaVec (3 axes) that we are sampling from.
+    @param [in] arc_cdf  The result of calling GetArcCdf() with `fsas`.
+    @param [in] num_paths  An array giving the number of paths that is
+                        requested for each Fsa, with num_paths.Dim() ==
+                        fsas.Dim0().  Must be zero for any Fsa that
+                        is equivalent to the empty Fsa (e.g. its
+                        GetTotScores() entry is zero).
```
csukuangfj, line 565: For an empty FSA, is its total score equal to `negative infinity`, not zero?

k2/csrc/fsa_utils.h
```diff
@@ -521,6 +521,110 @@ FsaVec FsaVecFromArcIndexes(FsaVec &fsas, Ragged<int32_t> &best_arc_indexes);
...
+                      -infinity.
+    @param [in] state_batches  The result of calling GetStateBatches()
+                      on `fsas`.  Is needed so we can know the maximum
+                      possible length of each path, to know how much memory to
+                      allocate.
+
+   @return  Returns a ragged tensor with 3 axes: [fsa][path][arc],
+            containing arc-indexes (idx012) into `fsas`,
+             with `ans.Dim0() == fsas.Dim0()`,
+            `ans.TotSize(1) == Sum(num_paths)`.  Each bottom-level
```
csukuangfj, line 610: `num_paths` is a scalar here.

k2/csrc/fsa_utils.cu
csukuangfj: I believe it is yes.

k2/csrc/utils_inl.h
```diff
@@ -42,6 +42,35 @@ void ExclusiveSum(ContextPtr c, int32_t n, const SrcPtr src, DestPtr dest) {
...
+
+template <typename SrcPtr, typename DestPtr>
+void InclusiveSum(ContextPtr c, int32_t n, const SrcPtr src, DestPtr dest) {
+  K2_CHECK_GE(n, 0);
+  DeviceType d = c->GetDeviceType();
+  using SumType = typename std::decay<decltype(dest[0])>::type;
+  if (d == kCpu) {
+    SumType sum = 0;
+    for (int32_t i = 0; i != n; ++i) {
+      auto prev = src[i];  // save a copy since src and dest
```
csukuangfj, line 54:
```cpp
sum += src[i];
dest[i] = sum;
```

For inclusive sum, I believe we don't need to save a copy into `prev`.

k2/csrc/fsa_utils.cu
danpovey: The last element is 1.0, so we don't need it.

k2/csrc/fsa_utils.cu
```diff
@@ -2149,4 +2150,407 @@ void FixNumStates(FsaVec *fsas) {
...
+       g.sync();
+
+       FloatType p = shared_data->p;
+       int32_t begin_arc_idx01x = shared_data->begin_arc_idx01x,
+           num_arcs = shared_data->num_arcs;
+
+       if (num_arcs == 0) {
+         // If we reach a state with no arcs leaving it, it should be the final
+         // state; otherwise something went wrong.
+         K2_CHECK_EQ(fsas_row_ids2_data[begin_arc_idx01x], final_state);
```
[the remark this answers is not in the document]
danpovey: I think I made it check anyway because it's not in the loop...

k2/csrc/fsa_utils.cu
csukuangfj: Currently, you are using it to compute `inv_tot`.

csukuangfj: https://github.com/k2-fsa/k2/blob/94049f185e84076ee04a3c844fb8e5926a9fbf30/k2/csrc/fsa_utils.cu#L2215-L2216

Had it been reused, `arc_pdf_data[end_arc - 1]` 's previous value is lost.

danpovey: ah yes you are right, won't make it in place.

k2/csrc/fsa_utils.cu
```diff
@@ -2552,5 +2554,37 @@ Ragged<int32_t> RandomPaths(FsaVec &fsas,
...
+Ragged<int32_t> RandomPaths(FsaVec &fsas,
+                            Array1<FloatType> &arc_cdf,
+                            int32_t num_paths,
+                            Array1<FloatType> &tot_scores,
+                            Ragged<int32_t> &state_batches) {
+  ContextPtr c = GetContext(fsas, arc_cdf, tot_scores, state_batches);
+  int32_t num_fsas  = fsas.Dim0();
+  Array1<int32_t> num_paths_array(c, num_fsas);
+  int32_t *num_paths_data = num_paths_array.Data();
+  FloatType *tot_scores_data = tot_scores.Data();
```
csukuangfj, line 2567: `const FloatType *tot_scores_data`, as it is not going to be changed.

danpovey:
This should be ready to merge.  The tests pass, although for now I commented out the single-precision test as it fails due to an issue in SegmentedExclusiveSum().

qindazhu:
@danpovey wondering if there's any other consideration to compute cdf except giving arcs with large weights more chance to generate, if no, can we use weighted random sampling algorithm to choose arc? http://utopia.duth.gr/~pefraimi/research/data/2007EncOfAlg.pdf#:~:text=Algorithms%20A-Res%20and%20A-ExpJ%20can%20be%20used%20for,pass%2C%20k%20instances%20of%20A-Res%20or%20A-ExpJ%20respectively.

danpovey:
This algorithm is basically optimal already (and well suited to GPUs).  It's not sampling according to the weight of the arc, but the posterior probability on that arc according to the distribution over paths, weighted by their likelihood.

k2/csrc/fsa_utils.cu
```diff
@@ -2149,4 +2150,453 @@ void FixNumStates(FsaVec *fsas) {
...
+           // interval_end - interval_start > 0.0, p - interval_start >= 0.0,
+           // and (interval_end - interval_start) >= (p - interval_start).
+           p = (p - interval_start) / (interval_end - interval_start);
+
+           int32_t next_state_idx01 = arcs[arc_idx012].dest_state + state_idx0x,
+               next_arc_idx01x = fsas_row_splits2_data[next_state_idx01],
+               next_arc_idx01x_next = fsas_row_splits2_data[next_state_idx01 + 1];
+           shared_data->begin_arc_idx01x = next_arc_idx01x;
+           shared_data->num_arcs = next_arc_idx01x_next - next_arc_idx01x;
+           shared_data->p = p;
```
csukuangfj, line 2451: Do you need a `break` here?

danpovey: Well spotted!  If I had checked my email more often I would have saved time debugging..

PR #621, opened 2021-01-26
Implemented SegmentedExclusiveSum without cudpp, be more faster

qindazhu:
Will remove those cudpp file in a seperate PR.

### DIFF
--- a/k2/csrc/benchmark/ragged_ops_benchmark.cu
+++ b/k2/csrc/benchmark/ragged_ops_benchmark.cu
@@ -66,11 +66,61 @@ static void RegisterBenchmarkGetTransposeReordering(DeviceType device_type) {
   }
 }
 
+template <typename T>
+static BenchmarkStat BenchmarkSegmentedExclusiveSum(int32_t dim,
+                                                    DeviceType device_type) {
+  ContextPtr context;
+  if (device_type == kCpu) {
+    context = GetCpuContext();
+  } else {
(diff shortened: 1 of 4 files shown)


csukuangfj:
+2

danpovey:
LGTM.. I assume you will run some comparison...

qindazhu:
> LGTM.. I assume you will run some comparison...

Yeah, you can check it here https://github.com/k2-fsa/k2/pull/618#issuecomment-767343598, merging.

danpovey:
I'm afraid this was problematic for large-sized inputs.
In testing of some FSA algorithms,  I discovered a situation where the SegmentedExclusiveSum of an array with nonnegative elements had some elements which were small and negative.

danpovey:
I think we may need to revert to the cudpp approach.

danpovey:
(This is for floats.  The current approach is probably OK for integers, due to properties of modular arithmetic.)

qindazhu:
mm, OK, will do

danpovey:
Mm, actually, if we used doubles (and we do in practice), this might not be such an issue.
IDK...

qindazhu:
Let me revert to cudpp for now (and only for float and double, as the current approach would be faster for integer), I will try to implement an version `SegmentScan` based on modern gpu later, I found yesterday that it may be posibble to implement with modern gpu by using only one kenerl like `SegmentReduce` which is expected to be faster.

PR #622, opened 2021-01-26
Added random tests for SegmentedExclusiveSum and deleted those cudpp files

qindazhu:


k2/csrc/ragged_test.cu
```diff
@@ -2559,24 +2559,42 @@ TEST(RaggedShapeOpsTest, RaggedShapeAxis0Splitter) {
...
+                                 //
+                                 0, 5, 11, 18};
+      CheckArrayData(dst, expected);
+
+      // &src.values == dst
+      SegmentedExclusiveSum(src, &src.values);
+      CheckArrayData(src.values, expected);
+    }
+    {
+      // random case, we assume the implementation for cpu is correct and only
```
csukuangfj, line 2582: For testing, I would suggest using ExclusiveSum to compute the ground truth for each sublist separately.

k2/csrc/test_utils.h
```diff
@@ -53,33 +53,37 @@ bool ApproxEqual(FloatType a, FloatType b, double delta = 0.001) {
 }
 
 template <typename T>
-void ExpectEqual(const std::vector<T> &expected, const std::vector<T> &actual) {
+void ExpectEqual(const std::vector<T> &expected, const std::vector<T> &actual,
+                 double abs_error = 0.001) {
```
csukuangfj, line 57: `abs_error` is never used.

k2/csrc/ragged_test.cu
qindazhu: mm, the underlying cpu are the same, so there's no difference. Anyway, we have tested cpu code with simple case, so I believe it's fine to just compare the result of GPU version and CPU version.

k2/csrc/test_utils.h
qindazhu: It's template function, I used it in below float and double version

csukuangfj: template functions support overload.

```cpp
template<typename T>
void ExpectEqual(const std::vector<T> &expected, const std::vector<T> &actual);

template<typename T>
void ExpectEqual(const std::vector<T> &expected, const std::vector<T> &actual, T abs_error);
```
should be fine.

qindazhu: mm..thinking about how you will call the two versions, you must use a if-else based on T. for simplity, let's keep the current version...but you remind me to add documentation here to say the parameter just for float and double.

PR #627, opened 2021-01-27
Remove old code in utils.cu

danpovey:
This enables `nsys profile` to work, i.e. so that we can see the kernels running.
It seems those tools don't support dynamic parallelism (calling __global__ functions from __device__ functions)
and that is true even if that code is never reached.

### DIFF
--- a/k2/csrc/utils.cu
+++ b/k2/csrc/utils.cu
@@ -18,55 +18,6 @@
 
 namespace k2 {
 
-// See FillValues() where this is invoked.  It fills a region with
-// a constant value.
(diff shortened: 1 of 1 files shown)


qindazhu:
Just wondering how do you find they are related? Searching on Web?

BTW, it seems Nsight system supports CDP now? 
 https://forums.developer.nvidia.com/t/nsight-systems-doesnt-profile-kernels/80430/3
https://stackoverflow.com/questions/18719824/how-to-read-profiling-results-of-cuda-dynamic-parallelism-with-nsight
Just noted, though we may not need it now.

danpovey:
Justin Luitjens suggested it to me in an email.

PR #632, opened 2021-01-28
add double support for SegmentedExclusiveSum.

csukuangfj:


k2/csrc/ragged_ops_inl.h
```diff
@@ -573,7 +573,7 @@ void SegmentedExclusiveSum(Ragged<T> &src, Array1<T> *dst) {
   if (c->GetDeviceType() == kCuda) {
     // there's roundoff problem for float type with the below implementation in
     // else branch.
-    if (std::is_same<float, T>::value) {
+    if (std::is_same<float, T>::value || std::is_same<double, T>::value) {
```
qindazhu, line 576: use `std::is_floating_point`?

csukuangfj: I believe we don't support `long double`.

qindazhu: Apprently we did not support a lot of types even before or now (and usually we would not check this). Anyway, it's up to you as it's just a suggestion. 

LGTM, you can merge if you think it's OK.

PR #634, opened 2021-01-28
Wrap Append to Python

csukuangfj:
`Append` is used to concatenate ragged tensor attributes of Fsa for `k2.create_fsa_vec`.


danpovey:
LGTM!

PR #635, opened 2021-01-28
Implement UniqueSequences().

danpovey:
.. will be needed in n-best list processing.
Also extend `Index(RaggedShape, Array1..)` to take an `axis` argument.

### DIFF
--- a/k2/csrc/ragged_ops.h
+++ b/k2/csrc/ragged_ops.h
@@ -133,17 +133,18 @@ void OrPerSublist(Ragged<T> &src, T initial_value, Array1<T> *or_values) {
 }
 
 /*
-  Sort each sub-list in `src`, with operator `<`, and output the order to
+  Sort each sub-list in `src`, (with operator `<` if Op is LessThan<T>, and
+  '>' if Op is GreaterThan<T>; and output the order to
   `order`. CAUTION: don't rely on this being a stable sort for now. Will
   eventually make the operator customizable, in which case this would become a
   wrapper.
 
       @param [in] src   Ragged array with 2 axes.
       @param [out] order   List of indexes that we'll use to give `src`
-                      a sorted order; will be resized if its size is
-                      not src.values.Dim().  If you do
+                      a sorted order; must already be allocated with size equal
+                      to src.values.Dim().  If you do
                         src.values = src.values[*order]
-                      then src.values will be sorted.
+                      then src.values will be sorted (per sublist).
  */
 template <typename T, typename Op>
 void SortSublists(Ragged<T> &src, Array1<int32_t> *order);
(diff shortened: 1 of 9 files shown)


k2/csrc/ragged_ops.h
```diff
@@ -427,6 +412,14 @@ Ragged<int32_t> AddPrefixToRagged(Ragged<int32_t> &src,
  */
```
csukuangfj, line 412: > Read algorithms like in compose.cc to understand why.

What do you mean by saying `compose.cc`? There is no such file in k2.

k2/csrc/ragged_ops.h
```diff
@@ -1321,6 +1331,25 @@ template <typename T>
...
+  the earliest axes will be ignored.
+
+     @param [in] src  Source ragged tensor
+
+     @return   Returns a tensor with the same number of axes as `src` and
+            possibly fewer elements due to removing repeated sequences on the
+            last axis (and with the last-but-one indexes possibly in a different
+            order).
+
+  CAUTION: the current implementation does not completely remove the possibility
```
csukuangfj, line 1347: If several sequences have the same hash, the current implementation
keeps only one of them.

Repeated sequences would have an identical hash value. Why does it say
> the current implementation does not completely remove the possibility of repeated sequnces,

k2/csrc/ragged_ops.h
danpovey: can remove this sentence, must have moved things around.

k2/csrc/ragged_ops.h
danpovey: mm that was a mistake, it does not completely guarantee that all unique sequences will 
be present in the output.  Perhaps you can fix..

csukuangfj: Will fix it.

k2/python/csrc/torch/ragged.cu
```diff
@@ -151,7 +151,7 @@ static void PybindRaggedTpl(py::module &m, const char *name) {
         Array1<int32_t> indexes_array = FromTensor<int32_t>(indexes);
         Array1<int32_t> value_indexes;
 
-        Ragged<T> ans = Index(src, indexes_array,
+        Ragged<T> ans = Index(src, 0, indexes_array,
```
csukuangfj, line 154: It does not handle the case when `indexes_array` contains -1, which will result in OOM.

danpovey: Oh that's interesting.  I was just looking into the same issue for a different reason.

danpovey: Let me fix it.

csukuangfj: `index_array` is from the arc map, which may contain `-1`, violating the pre-condition of `Index`.

csukuangfj: thanks

Issue #555, opened 2020-12-27
Accelerating a common pattern


danpovey:
I have a proposal that could be used to accelerate a common programming pattern.

We frequently do something like this:

Allocate an array `array` of size N+1, get its data as array_data
Do
  `K2_EVAL(c, N, lambda_something, (int32_t i) -> void { ... });`
Do `ExclusiveSum(array, &array);`.

In cases where the amount of data is small enough that the kernel called in ExclusiveSum() only has to load the data from memory once, this is not ideal.  There is room for optimization by doing something like:
```
   K2_EVAL_AND_EXCSUM(c, N, array_data, lambda_something, (int32_t i) -> int32_t { 
    ... return something; 
      });
```
and we'd have to figure out the implementation detaills.  It might end up being an if-statement, as in, it would expand  to:
 ```
   if (N > some_fixed_threshold) {
       Eval the lambda separately and then call ExclusiveSum separately
   } else {
       call ExclusiveSum with some kind of object that evaluates the lambda.  It would probably 
       involve some class, templated on the lambda type, that wraps the lambda with an 
      operator []; would need both __host__ and __device__
       implementations, possibly the same.
   }
```
... this may not be 100% ideal, as in, it would be better if we knew for sure which sizes would end up causing multiple evaluations of the operator[], but I suspect that would be too much work and/or complexity.

Obviously this would need to be tested via benchmarks to make sure it's actually faster.

This shouldn't be prioritized above getting LF-MMI to work though.
BTW I have to stay home a couple days due to virus testing..

danpovey:
A related common pattern  is when we set the 'keep' array of a Renumbering object, which could be accelerated in a similar way.  I think a templated constructor would be the easiest way to implement it.

PR #640, opened 2021-01-29
using pattern trasform_scan to merge two kenerls into one

qindazhu:
Just found the way to optimize https://github.com/k2-fsa/k2/issues/555 with `transform_scan` in modern gpu, it turns out that it would be faster for most cases after I run multiple experiments (noted it seems  the speed on GPU is not so stable somtimes, so we'd better run enough cases and times to compare the performace)

We need to replace those pattern code in our codebase, but let me do this one by one while optimizing each ops.

-----------------------------------------------
The header below is `TransExclusiveSum_Old/New_dim` where `Old` is the old pattern that uses two kenerls while `New` uses one kenerl, specially, the below benchmark is run with:

**TransformExclusiveSum_Old**:
```
K2_EVAL(
      c, dim, lambda_multiple2,
      (int32_t i)->void { ans_data[i] = src_data[i] * 2; });
  ExclusiveSum(*ans, ans);
```
**TransformExclusiveSum_New**:
```
 K2_TRANS_EXCSUM(
      c, dim, ans_data, lambda_multiple2,
      (int32_t i)->int32_t { return src_data[i] * 2; });
```

```
TransExclusiveSum_Old_100,,kCuda,100,20,15.945600
TransExclusiveSum_Old_200,,kCuda,200,20,15.963200
TransExclusiveSum_Old_500,,kCuda,500,20,16.103998
TransExclusiveSum_Old_1000,,kCuda,1000,20,16.303999
TransExclusiveSum_Old_2000,,kCuda,2000,20,16.126400
TransExclusiveSum_Old_5000,,kCuda,5000,20,15.940801
TransExclusiveSum_Old_10000,,kCuda,10000,20,15.808001
TransExclusiveSum_Old_20000,,kCuda,20000,20,15.927999
TransExclusiveSum_Old_50000,,kCuda,50000,20,15.952001
TransExclusiveSum_Old_100000,,kCuda,100000,20,16.171200
TransExclusiveSum_Old_500000,,kCuda,500000,20,15.187199
TransExclusiveSum_Old_1000000,,kCuda,1000000,20,24.180799
TransExclusiveSum_Old_5000000,,kCuda,5000000,20,110.766396

TransExclusiveSum_New_100,,kCuda,100,20,4.302400
TransExclusiveSum_New_200,,kCuda,200,20,4.358400
TransExclusiveSum_New_500,,kCuda,500,20,4.280000
TransExclusiveSum_New_1000,,kCuda,1000,20,4.468800
TransExclusiveSum_New_2000,,kCuda,2000,20,4.993600
TransExclusiveSum_New_5000,,kCuda,5000,20,8.179199
TransExclusiveSum_New_10000,,kCuda,10000,20,13.382401
TransExclusiveSum_New_20000,,kCuda,20000,20,13.470400
TransExclusiveSum_New_50000,,kCuda,50000,20,13.307199
TransExclusiveSum_New_100000,,kCuda,100000,20,13.335999
TransExclusiveSum_New_500000,,kCuda,500000,20,11.816001
TransExclusiveSum_New_1000000,,kCuda,1000000,20,20.940800
TransExclusiveSum_New_5000000,,kCuda,5000000,20,92.555199
```

### DIFF
--- a/k2/csrc/benchmark/array_ops_benchmark.cu
+++ b/k2/csrc/benchmark/array_ops_benchmark.cu
@@ -9,6 +9,7 @@
 #include "k2/csrc/array_ops.h"
 #include "k2/csrc/benchmark/benchmark.h"
 #include "k2/csrc/test_utils.h"
+#include "moderngpu/kernel_scan.hxx"
 
 namespace k2 {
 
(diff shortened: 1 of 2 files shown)


qindazhu:
Will delete benchmark code in next commit as those code should not be merged (I just commit them to show the example), you can still check the code in the first commit.

danpovey:
What will happen if we need an "extra element" at the end of the exclusive sum?
Do we need an explicit if-statement?

danpovey:
In the majority of cases we do need that extra statement, and in the majority of cases the kernel would crash if evaluated on the extra element.  We may need 2 versions of the macro, one where the extra element is needed and one where not.

Also we will need to modify class Renumbering to support this.    E.g. give it a function WritableNew2old(), which would just allocate the array but not write anything to it.

qindazhu:
RE extra element, as I declared in the documentation and the example, current version always suppose we have an extra element in the provided `ans`, BTW, it seems we don't need a version that there's no extra element, but I can add it if you think we need.
https://github.com/k2-fsa/k2/blob/c25247c308c04087f049d11293f28d0d7c68b92b/k2/csrc/macros.h#L117-L130

RE  Renumbering, let me do this later, I was planning to implement it with a `transform_compact` pattern which is also a two-steps approach (one called `upsweep` which label `keep` data, the other called `downsweep` which copy data, but it seems it will use some cache between the two steps ), I prefer to refactor `Renumbering` after I did some benchmark on these two different approaches.

danpovey:
Great!
You can merge when you think is best...

qindazhu:
Thanks，let me merge this PR and go forward for other possible optimizations.

PR #643, opened 2021-01-29
Wrap RandomPaths.

### LINKED ISSUE #639: Wrap RandomPaths()
We need to wrap RandomPaths() to Python, in class Fsa.  This is not differentiable, the outputs are integers.
Currently I think we don't need to cache the cdf, because it's only needed by RandomPaths() which will typically only be called once.
```
    def get_random_paths(self, use_double_scores: bool, num_paths: int) -> k2.RaggedInt:
        '''Compute pseudo-random paths through the FSAs in this vector of FSAs 
        (this object must have 3 axes, `self.arcs.num_axes() == 3`) 

         Args:  
             use_double_scores:
                  If true, do computation with double-precision, else float (single-precision)
            num_paths:
                 Number of paths requested through each FSA.  FSAs that have no successful
                paths will have zero paths returned.
         Returns:
                Returns a k2.RaggedInt with 3 axes: [fsa][path][arc_pos]; the final sub-lists
             (indexed with arc_pos) are sequences of arcs starting from the start state and
             terminating in the final state.  The values are arc_idx012, i.e. arc indexes.
         '''
         pass                                   
```

csukuangfj:
Closes #639 

~Don't merge. The tests don't pass as CPU and CUDA produce different results for the same input.~

### DIFF
--- a/k2/python/tests/random_paths_test.py
+++ b/k2/python/tests/random_paths_test.py
@@ -37,9 +37,9 @@ class TestRandomPaths(unittest.TestCase):
                 self.assertEqual(str(path), '[ [ [ 0 1 ] [ 0 1 ] ] ]')
 
     def test_single_fsa_case2(self):
-        devices = [torch.device('cuda')]
-        #  if torch.cuda.is_available():
-        #      devices.append(torch.device('cuda', 0))
+        devices = [torch.device('cpu')]
+        if torch.cuda.is_available():
+            devices.append(torch.device('cuda', 0))
 
         for device in devices:
             for use_double_scores in (True,):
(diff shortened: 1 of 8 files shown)


k2/python/tests/random_paths_test.py
```diff
@@ -0,0 +1,125 @@
...
+                '''
+                fsa = k2.Fsa.from_str(s).to(device)
+                fsa_vec = k2.create_fsa_vec([fsa])
+
+                path = k2.random_paths(fsa_vec,
+                                       use_double_scores=use_double_scores,
+                                       num_paths=1)
+                assert path.num_axes() == 3
+                # p is 0.5, it should select the last leaving arc of each state
+                self.assertEqual(str(path), '[ [ [ 1 3 5 ] ] ]')
```
csukuangfj, line 63: @danpovey 
CPU gives [ 1, 3, 5], while CUDA returns [1, 2, 4].

danpovey: Your test logic is not quite right-- it "zooms in" to the distribution each time it takes an arc-- but I did find a bug.  Thanks!  Will push a fix.

danpovey: .. the CPU code was failing to zoom in.

csukuangfj: > Your test logic is not quite right

My test code is written according to the CPU version. Will update it once you fix it.

k2/python/tests/random_paths_test.py
```diff
@@ -0,0 +1,159 @@
...
+                #  iter 2, p is (0.5 - 0.5) / (1 - 0.5) = 0, select the first
+                #          leaving arc of state 2
+                # path 1
+                #  iter 0, p is 0.75, select the second leaving arc of state 0
+                #  iter 1, p is (0.75 - 0.5) / (1 - 0.5) = 0.5, select the
+                #          second leaving arc of state 1
+                #  iter 2, p is (0.5 - 0.5) / (1 - 0.5) = 0, select the
+                #          first leaving arc of state 1
+                assert path.num_axes() == 3
+                self.assertEqual(str(path), '[ [ [ 0 3 4 ] [ 1 3 4 ] ] ]')
```
csukuangfj, line 84: @danpovey 
CPU and CUDA still generate different results for this case.

For path 0,
- CPU returns [ 0 3 4]
- CUDA returns [1 2 4]

danpovey: I'll look into it.

csukuangfj: Fixed.

csukuangfj:
Ready to review.

danpovey:
LGTM, thanks!

PR #650, opened 2021-02-02
Change Eval() and EvalDevice() to use 1 kernel not 2, to reduce compi…

danpovey:
…lation time.


danpovey:
I haven't really tested the speed difference but I assume it would be small (and I am testing that everything works, with a separate PR).

This aims to reduce the compilation time, to have 1 kernel compiled per K2_EVAL() instead of 2.

PR #656, opened 2021-02-04
Implement k2.fsa_from_unary_function_ragged

### LINKED ISSUE #654: Unary FSA algo- consolidate code
I think we can create an interface like the following, which will avoid a lot of code duplication and
make it easier to support backprop for properties in a more general way.

```
def fsa_from_unary_function_tensor(src_fsa: Fsa, dest_arcs: _k2.RaggedArc, arc_map: Tensor):
    ''' Create an Fsa object, including autograd logic and propagating properties from the
       source FSA.  This is intended to be called from unary functions on FSAs where the arc_map is
       a Tensor of int32 (i.e. not ragged).

     Args:
       src_fsa:  The source Fsa, i.e. the arg to the unary function
       dest_arcs:  The raw output of the unary function, as output by whatever C++
                 algorithm we used
      arc_map:  A map from arcs in `dest_arcs` to the corresponding arc-index in `src_fsa`,
                 or -1 if the arc had no source arc (e.g. added epsilon self-loops).
   Returns:
          Returns the resulting Fsa, with properties propagated appropriately, and autograd
          handled.
     '''
      pass
      # note: would use set_scores_phantom with k2.index(src_fsa.scores, arc_map). 
      # I propose to make sure k2.index() handles autograd when the input is floating point.

```
The above case is fairly straightforward in terms of what we need to do for autograd.

We can also do something like the above for when the arc_map is a _k2.RaggedInt, e.g.
call it fsa_from_unary_function_ragged(...).  This would be called from epsilon removal, and
eventually from determinization.

Here's my proposal for what we can do with the properties propagation where the arc_map
is ragged.
   - Property was tensor of int32_t -> output will be _k2.RaggedInt, same as what k2.index() does, no autograd needed.
   - Property was ragged tensor of int32_t: for now we will treat all such properties as if they were sequences to be appended.  k2.index() seems to do with we need right now, goes to index_ragged() -> _k2.index().  I am not 100% happy with this just being called "index", as it also has `remove_axis = true` in the C++ code.  But OK for now I guess.
   - Property was tensor of float -> we sum the source elements for each sub-list, so output will be a tensor of float.  We can call this something like  `index_and_sum(src:Tensor, indexes:_k2.RaggedInt) -> Tensor`, which would handle the backprop, and could perhaps be used for the scores as well as floating-point tensor properties.  Currently k2.index() does this summation, I believe, but it does not support backprop.  I think it would make more sense to call it index_and_sum() instead of index(), since it is not pure indexing.

 For the implementation of `index_and_sum()`: the forward could be done with SumPerSublist() where the source is a ragged tensor formed from (indexes.shape, input_tensor[indexes.values]).  The backward could be done by doing input_tensor_grad.index_add(0, indexes.values, output_tensor_grad[indexes.row_ids]). Actually we'll use k2's own index_add and indexing operations, which support int32 indexes.  What I just described would only work for 1-d tensors of float, not 2-d; I suppose we can handle the 2-d case later on, if needed.

csukuangfj:
Closes #654

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ cmake_minimum_required(VERSION 3.8 FATAL_ERROR)
 
 project(k2 CUDA CXX)
 
-set(K2_VERSION "0.3.1")
+set(K2_VERSION "0.3.3")
 
 # ----------------- Supported build types for K2 project -----------------
 set(ALLOWABLE_BUILD_TYPES Debug Release RelWithDebInfo MinSizeRel)
--- a/k2/python/csrc/torch/ragged.cu
+++ b/k2/python/csrc/torch/ragged.cu
@@ -185,6 +185,18 @@ static void PybindRaggedImpl(py::module &m) {
         return Index(src_array, indexes);
       },
       py::arg("src"), py::arg("indexes"));
+
+  m.def(
+      "index_and_sum",
+      [](torch::Tensor src, Ragged<int32_t> &indexes) -> torch::Tensor {
+        K2_CHECK_EQ(src.dim(), 1) << "Expected dim: 1. Given: " << src.dim();
+        Array1<float> src_array = FromTensor<float>(src);
+        Ragged<float> ragged = Index(src_array, indexes);
+        Array1<float> ans_array(ragged.Context(), ragged.Dim0());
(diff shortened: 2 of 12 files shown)


csukuangfj:
Ready for review.

danpovey:
Great work!!  Looks good to me!

PR #662, opened 2021-02-08
Upload generated wheels to k2-fsa.org

csukuangfj:


csukuangfj:
Generated wheels are now uploaded to http://k2-fsa.org/nightly/ . Will setup GitHub actions to build k2 every night.

A screenshot is given below

Screen Shot 2021-02-09 at 8 01 18 PM


They are built with `torch==1.7.1`. Users can select a version to install based on the version of CUDA and Python they are using.

For example,

```
pip install k2==0.3.3+cu110.dev20210209 -f http://k2-fsa.org/nightly/ --trusted-host k2-fsa.org
```

installs `k2` for CUDA 11.0. `pip` will select the proper Python version automagically from 3.6, 3.7, and 3.8.

---

Since k2-fsa.org is using HTTP and there is no certificate for it, we have to add the option `--trusted-host k2-fsa.org` to `pip install`.

@danpovey  Can we apply for a certificate so that we can use HTTPS for k2-fsa.org ?

PR #669, opened 2021-02-20
Support torch.float64 and torch.int64 in k2.index_select.

csukuangfj:


danpovey:
Thanks!!  Looks good!

PR #670, opened 2021-02-23
avoid overwriting `inner_labels` in k2.compose.

csukuangfj:


csukuangfj:
Needed for  https://github.com/k2-fsa/snowfall/pull/106

https://github.com/k2-fsa/snowfall/blob/f6d3dcb4b0f1ab83024ba1fd9e8f92fe372db60d/snowfall/training/mmi_mbr_graph.py#L137

By default, `den_lats` does not have the attribute `phones`. This pull-request tries to avoid overwriting
the `phones` attribute for the `num_graph`.

PR #678, opened 2021-03-04
[NOT_FOR_MERGE_] Eval with row ids

danpovey:


danpovey:
@qindazhu if you can get EvalWithRowIds working I can finish this so it will be memory efficient.

PR #679, opened 2021-03-04
Add linear space support to NormalizePerSublist

csukuangfj:
https://github.com/k2-fsa/snowfall/pull/106#discussion_r582544398

> Basically we will be doing another pass of MMI training, except instead of summing the tot_scores we will compute a weighted sum where the weight is (num_copies_this_path / num_paths_this_seq). 

This pullrequest can be used to compute `(num_copies_this_path / num_paths_this_seq)`. Since it does not need
to support autograd in this case, autograd is not implemented at present.


danpovey:
Great!

PR #680, opened 2021-03-05
Memory efficiency in IntersectDense()

danpovey:


danpovey:
Even though this isn't optimally fast (because we need RowSplitsToRowIdsRange() to work properly),
I'll merge this because it resolves the OOM issue in the Aishell recipe.

PR #681, opened 2021-03-05
[WIP] Compute row_ids chunk by chunk

qindazhu:


qindazhu:
@danpovey I think it's hard to do this efficiently with current interface of `RowSplitsToRowIdsRang` as we can not search `row_start` and `row_end` for each chunk/range efficiently. Instead, I search those `row_start` and `row_end` for all chunks in one kernel in below code. If you think this approach is fine, I will copy those code to `EvalWithRowIds` tomorrow.

danpovey:
Mm.  It seems it will make the calling code quite complicated.
I have another idea.  Instead of segmenting by element we can segment by row.
It will make the pieces of slightly uneven size but I think it's OK.
Basically it would require to make a version of RowSplitsToRowIds() that would work
with an offset.  Let me think about the right interface..

qindazhu:
Closed as we don't need this, thanks.

PR #682, opened 2021-03-05
Change how we split up rows/elements in IntersectDense()

danpovey:


danpovey:
@qindazhu I've made some changes so that we don't need any nontrivial programming for the row-ids stuff.  Thanks for looking into it!  I thought it would be easier than it actually was...

qindazhu:
OK. BTW, if we have `row_offset`, we can also pass it to `RowSplitsToRowIdsOffset` and get the real `row_ids` (i.e. the value in it starting from row_offset instead of 0), as I did here

https://github.com/qindazhu/k2/blob/ea97ef576b76963ec8ac591279b17189b9ad1fb3/k2/csrc/array_ops_test.cu#L125-L131.

Just mention it, it depends on your needs.

danpovey:
Mm.  When you have a chance, perhaps you could make a PR to improve the
efficiency of the current code?

PR #686, opened 2021-03-16
Support ragged aux_labels in k2.compose

csukuangfj:
It is needed in
```python
k2.compose(decoding_lats, word_fsas)
```
where `aux_labels` in `decoding_lats` is a ragged tensor.

This pull-request extends `k2.compose` to support that.

---

[EDITED]: Mark it as WIP. Will merge after testing it in snowfall.

Used in https://github.com/k2-fsa/snowfall/pull/106

### DIFF
--- a/k2/csrc/ragged_ops.h
+++ b/k2/csrc/ragged_ops.h
@@ -1349,6 +1349,8 @@ Array1<T> ComputeHash(Ragged<int32_t> &src);
      @param [out] new2old_indexes
                        If not NULL, on return new2old_indexes[i] contains
                        the original input sublist for the i-th output sublist.
+                       If `src` has 2 axes, this array contains `src_idx0`;
+                       if `src` has 3 axes, this array contains `src_idx01`.
 
      @return   Returns a tensor with the same number of axes as `src` and
             possibly fewer elements due to removing repeated sequences on the
--- a/k2/csrc/ragged_ops.h
+++ b/k2/csrc/ragged_ops.h
@@ -1346,6 +1346,9 @@ Array1<T> ComputeHash(Ragged<int32_t> &src);
                        repeats (i.e., multiplicity) of each output sequence.
                        The caller does not need to pre-allocate it. It is
                        allocated inside the function.
+     @param [out] new2old_indexes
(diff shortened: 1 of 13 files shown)


csukuangfj:
Ready for review.

k2/csrc/ragged_ops.h
```diff
@@ -1346,6 +1346,9 @@ Array1<T> ComputeHash(Ragged<int32_t> &src);
                        repeats (i.e., multiplicity) of each output sequence.
                        The caller does not need to pre-allocate it. It is
                        allocated inside the function.
+     @param [out] new2old_indexes
```
danpovey, line 1349: Each output sublist could have appeared multiple times in the input, so is there a rule for which one is listed?

Should probably specify that this would be an idx0 if `src` had 2 axes, and an idx01 if `src` had 3 axes.

csukuangfj: > Each output sublist could have appeared multiple times in the input, so is there a rule for which one is listed?

Sublists of a seq are reordered by their hash values by calling `SortSublists`.
Although the implementation of`SortSublists` does not mention that it is stable, the Python tests
show that the first one is kept, other repeats are discarded.

I think we don't care which one is kept since their contents are identical (assuming there are no hash conflicts).

csukuangfj: > Should probably specify that this would be an idx0 if src had 2 axes, and an idx01 if src had 3 axes.

Thanks. Done.

danpovey: If it just puts an arbitrary one of the inputs, you should specify that.  (Also specify if this is nondeterministic, i.e.
may give different outputs for the same input on different run.s).

csukuangfj: > If it just puts an arbitrary one of the inputs, you should specify that. (Also specify if this is nondeterministic, i.e.
may give different outputs for the same input on different run.s).

Fixed. It is described in the comment that it uses a deterministic algorithm.

danpovey (approved):
LGTM!

PR #688, opened 2021-03-20
export k2 version information

csukuangfj:
```python
>>> import k2
>>> k2.__version__
'0.3.3'
>>> k2.__dev_version__
'0.3.3.dev20210320'
>>> k2.__build_type__
'Release'
>>> k2.__git_sha1__
'5ee74c7e22340f26788894a224b0cadcc2f9c898'
>>> k2.__git_date__
'Thu Mar 18 22:22:17 2021'
```

### DIFF
--- a/k2/python/k2/version.py
+++ b/k2/python/k2/version.py
@@ -7,6 +7,11 @@ import os
 import torch  # noqa
 import _k2
 
+__build_type__ = _k2.version.build_type
+__git_date__ = _k2.version.git_date
+__git_sha1__ = _k2.version.git_sha1
+__version__ = _k2.version.__version__
+
 
 def main():
     '''Collect the information about the environment in which k2 was built.
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,8 @@ def get_long_description():
 def get_cuda_version():
     import torch
     from torch.utils import collect_env
-    running_cuda_version = collect_env.get_running_cuda_version(collect_env.run)
+    running_cuda_version = collect_env.get_running_cuda_version(
+        collect_env.run)
     cuda_version = torch.version.cuda
     if running_cuda_version is not None:
         assert cuda_version in running_cuda_version, \
@@ -81,12 +82,14 @@ def get_short_description():
     return 'FSA/FST algorithms, intended to (eventually) be interoperable with PyTorch and similar'
 
 
+with open('k2/python/k2/__init__.py', 'a') as f:
+    f.write(f"__dev_version__ = '{get_package_version()}'\n")
+
 dev_requirements = [
     'clang-format==9.0.0',
     'flake8==3.8.3',
     'yapf==0.27.0',
 ]
-
 
 setuptools.setup(
     python_requires='>=3.6',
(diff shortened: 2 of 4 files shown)


setup.py
```diff
@@ -81,13 +82,15 @@ def get_short_description():
     return 'FSA/FST algorithms, intended to (eventually) be interoperable with PyTorch and similar'
 
 
+with open('k2/python/k2/__init__.py', 'a') as f:
+    f.write(f"__dev_version__ = '{get_package_version()}'\n")
```
csukuangfj, line 86: I am adding a new attribute to `k2` with this approach.
Perhaps @pzelasko you can also add an attribute to lhotse to contain its git sha1.

setup.py
```diff
@@ -120,3 +121,12 @@ def get_short_description():
...
     ],
 )
+
+# remove the line __dev_version__ from k2/python/k2/__init__.py
+with open('k2/python/k2/__init__.py', 'r') as f:
+    lines = f.readlines()
+
+with open('k2/python/k2/__init__.py', 'w') as f:
+    for line in lines:
+        if '__dev_version__' not in line:
```
csukuangfj, line 131: The added entry is removed so that it does not exist in the source tree,
which means it is always synchronized with the latest commit automatically.

k2/python/k2/version.py
```diff
@@ -7,6 +7,11 @@
 import torch  # noqa
 import _k2
 
+__build_type__ = _k2.version.build_type
```
csukuangfj, line 10: Do we need other attributes?

setup.py
pzelasko: Thanks I’ll check later if I can follow your approach

k2/python/k2/version.py
danpovey: I think this is enough-- thanks!
I'll wait for someone else to add related features to snowfall though, e.g. to print such info at the start of logs.

setup.py
pzelasko: I don't think this can work for Lhotse. It is easier in K2 because there is a C++ build that embeds the version + date every time you run it, so the Python code can query the native library for that. 

E.g. if I add a utility that queries git for sha1, it is going to crash when you `pip install lhotse`, because it is not a repository then. The best I can offer currently is `lhotse.__version__`, but it is going to change rarely. I'm not sure how other python packages deal with that, I'll look into it.

csukuangfj: I think it feasible.

I just created a demo repo to verify it. The repo is at the address https://github.com/csukuangfj/kk_test

If you install it using
```bash
pip install -vv git+https://github.com/csukuangfj/kk_test.git
```

After the installation, you will find a file `/some/path/to/your/venv/lib/python3.8/site-packages/kk_test/__init__.py`
containing
```
__version__ = '1.0.0'
__git_sha1__ = '96d9701521d62bf1bd127452e5aac6bc6029df79
```

---

Even though I don't show how to install it from PyPI, I think it is.always.possible to add
some file to `lhotse-x.x.x-py3-none-any.whl ` while you're building it or
add some file to `lhotse-x.x.x.tar.gz` when you create it.

The `some file` can contain whatever you want to put.  In the setup.py, it can figure out whether it
is able to get the `git sha1` from the repo; if not, just read it from `some file`.

PR #690, opened 2021-03-23
Aux labels etc

danpovey:
This builds on #611 to support multiple auxiliary labels when reading in FSAs.
Because of interface changes in k2.Fsa.from_str (which I feel were necessary), there is unfortunately
no backward or forward compatibility and we have to 
apply the following update to snowfall:
https://github.com/k2-fsa/snowfall/pull/136

### DIFF
--- a/k2/csrc/fsa.h
+++ b/k2/csrc/fsa.h
@@ -205,8 +205,41 @@ std::ostream &operator<<(std::ostream &os, const DenseFsaVec &dfsavec);
 */
 Fsa FsaFromTensor(Tensor &t, bool *error);
 
+/*
+  Create an FSA from a Array1. The Array1 is expected to has N arcs (the format
+  is src_state, dest_state, symbol, cost). The cost is not really an int32_t, it
+  is a float. This code will print an error message and output 'true' to
+  'error', and return an empty FSA (with no states or arcs) if 't' was not
+  interpretable as a valid FSA. These requirements for a valid FSA are:
+    - src_state values on the arcs must be non-decreasing
+    - all arcs with -1 as the label must be to a single state (call this
+      final_state) which has no arcs leaving it
+    - final_state, if not given, it need be calculated in this func. And it
+      must be numerically greater than any state which has arcs leaving it, and
+      >= any state that has arcs entering it.
+  If there are no arcs with -1 on the label, here is how we determine the final
+  state:
+     - If there were no states at all in the FSA we'll return the empty FSA
+       (with no arcs), and set error = true. (As the k2-fsa definition is there
+       at least must be a super final state.)
+     - Else if there were some states but no arcs, we'll return the FSA (with
+       just one implicit super final state (set id = 1) and no arcs), and set
+       error = false.
+     - Otherwise, we'll let `final_state` be the highest-numbered state
+       that has any arcs leaving or entering it, plus one.  (This FSA
+       has no successful paths but still has states.)
 
-Fsa FsaFromArray1(Array1<Arc> &arc, bool *error);
+    @param [in] arcs       Source arcs. Caution: the returned FSA will share
+                            memory with this tensor, so don't modify it
+                            afterward!
+    @param [out] error      Error flag.  On success this function will write
+                            'false' here; on error, it will print an error
+                            message to the standard error and write 'true' here.
+    @param [in] final_state This indicate the final-state ID (with default: -1,
+                            means find it by `arcs` in this func).
+    @return                 The resulting FSA will be returned.
+*/
+Fsa FsaFromArray1(Array1<Arc> &arcs, bool *error, int32_t final_state = -1);
 
 /*
   Returns a single Tensor that represents the FSA; this is just the vector of
--- a/k2/csrc/fsa_utils.cu
+++ b/k2/csrc/fsa_utils.cu
@@ -110,8 +110,7 @@ static void SplitStringToVector(const std::string &in, const char *delim,
 
 /* Create an acceptor from a stream, assuming the acceptor is in the k2 format:
 
-   src_state1 dest_state1 label1 score1
-   src_state2 dest_state2 label2 score2
+   src_state dest_state label [aux_label1 aux_label2 ... ] score
    ... ...
    final_state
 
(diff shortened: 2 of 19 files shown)


k2/csrc/fsa.h
```diff
@@ -205,8 +205,41 @@ std::ostream &operator<<(std::ostream &os, const DenseFsaVec &dfsavec);
 */
 Fsa FsaFromTensor(Tensor &t, bool *error);
 
+/*
+  Create an FSA from a Array1. The Array1 is expected to has N arcs (the format
+  is src_state, dest_state, symbol, cost). The cost is not really an int32_t, it
```
csukuangfj, line 210: I think `k2` is using `score` (or `weight`), not `cost`.

k2/csrc/fsa.h
```diff
@@ -205,8 +205,41 @@ std::ostream &operator<<(std::ostream &os, const DenseFsaVec &dfsavec);
...
+  is a float. This code will print an error message and output 'true' to
+  'error', and return an empty FSA (with no states or arcs) if 't' was not
```
csukuangfj, line 212: The `t` in `if t was not` is not defined anywhere.

k2/csrc/fsa.h
```diff
@@ -205,8 +205,41 @@ std::ostream &operator<<(std::ostream &os, const DenseFsaVec &dfsavec);
...
+       has no successful paths but still has states.)
 
-Fsa FsaFromArray1(Array1<Arc> &arc, bool *error);
+    @param [in] arcs       Source arcs. Caution: the returned FSA will share
+                            memory with this tensor, so don't modify it
+                            afterward!
+    @param [out] error      Error flag.  On success this function will write
+                            'false' here; on error, it will print an error
+                            message to the standard error and write 'true' here.
+    @param [in] final_state This indicate the final-state ID (with default: -1,
```
csukuangfj, line 238: typo: `This indicates`.

k2/csrc/fsa.h
```diff
@@ -205,8 +205,41 @@ std::ostream &operator<<(std::ostream &os, const DenseFsaVec &dfsavec);
...
+                            means find it by `arcs` in this func).
```
csukuangfj, line 239: typo: `means to find it`.

k2/csrc/fsa_utils.cu
```diff
@@ -110,8 +110,7 @@ static void SplitStringToVector(const std::string &in, const char *delim,
 
 /* Create an acceptor from a stream, assuming the acceptor is in the k2 format:
 
-   src_state1 dest_state1 label1 score1
-   src_state2 dest_state2 label2 score2
+   src_state dest_state label [aux_label1 aux_label2 ... ] score
```
csukuangfj, line 113: If `aux_label_x` is present, does it assume that every arc has the same number of `aux_labels`?

I guess so as the output argument is `Array2`, not `Ragged<int32_t>`.
It would be great if it is documented.

k2/csrc/fsa_utils.cu
```diff
@@ -137,277 +145,108 @@ static Fsa K2AcceptorFromStream(std::istringstream &is) {
...
       // final_state
-      (void)StringToInt(splits[0]);  // this is a final state
-      finished = true;               // set finish
+      final_state = StringToInt(splits[0]);
+      max_state = std::max(max_state, final_state);
+      if (final_state > 0) {
+        finished = true;  // set finish
+      }
     } else {
       K2_LOG(FATAL) << "Invalid line: " << line
```
csukuangfj, line 177: The error message also needs to be updated.

k2/csrc/fsa_utils.cu
```diff
@@ -120,14 +119,23 @@ static void SplitStringToVector(const std::string &in, const char *delim,
    the final state and set its label to -1.
 
    @param [in]  is    The input stream that contains the acceptor.
-
+   @param  [in]  num_aux_labels  The number of auxiliary labels to expect
+                     per arc; 0 == acceptor, 1 == transducer, but may be more.
+   @param [out] aux_labels_out  If num_aux_labels > 0, this will be
+                     assigned to,  with a new array on CPU of shape
+                     (num_arcs, num_aux_labels).
```
csukuangfj, line 126: The comment says its shape is `(num_arcs, num_aux_labels)`, but the implementation
at line 193 is
```cpp
    *aux_labels_out = Array2<int32_t>(c, num_aux_labels,
                                      (int32_t)arcs.size());
```
The code stores the `aux_labels` per arc non-contiguously. Is it intentional?
Just feel it is unnatural.

k2/python/csrc/torch/fsa.cu
```diff
@@ -98,12 +98,13 @@ static void PybindFsaUtil(py::module &m) {
...
-        Fsa fsa = FsaFromString(s, openfst, acceptor ? nullptr : &aux_labels);
+        Array2<int32_t> aux_labels;
+        Fsa fsa = FsaFromString(s, openfst, num_aux_labels,
+                                &aux_labels);
         torch::optional<torch::Tensor> tensor;
-        if (aux_labels.Dim() > 0) tensor = ToTensor(aux_labels);
+        if (num_aux_labels != 0) tensor = ToTensor(aux_labels);
         return std::make_pair(fsa, tensor);
       },
       py::arg("s"), py::arg("acceptor") = true, py::arg("openfst") = false,
```
csukuangfj, line 110: This line has to be updated to match the actual arguments and their default values
```cpp
py::arg("s"), py::arg("num_aux_labels") = 0, py::arg("openfst") = false,
```

k2/python/k2/fsa.py
```diff
@@ -1061,19 +1062,17 @@ def shape(self) -> Tuple[int, ...]:
             raise ValueError(f'Unsupported num_axes: {self.arcs.num_axes()}')
 
     @classmethod
-    def from_str(cls, s: str) -> 'Fsa':
+    def from_str(cls, s: str, num_aux_labels: int = 0,
+                 aux_label_names: List[str] = ['aux_labels',
```
csukuangfj, line 1066: Are there any reasons to set these kinds of default names? 
`Optional[List[str]]` seems to be more natural.

`num_aux_labels` and `len(aux_label_names)` contain the
same information, though it is not checked in the code.

k2/python/tests/fsa_test.py
```diff
@@ -227,7 +285,7 @@ def test_transducer_from_str(self):
             5 0  1  50  -8.2
             6
         '''
-        fsa = k2.Fsa.from_str(_remove_leading_spaces(s))
+        fsa = k2.Fsa.from_str(_remove_leading_spaces(s), num_aux_labels=1)
```
csukuangfj, line 288: The test code covers only the case for `num_aux_labels=1`.

I am worrying that most of the code is actually not tested.

k2/csrc/fsa_utils.cu
danpovey: .. this is kind of implicit because `num_aux_labels` is an arg to the function.

k2/csrc/fsa_utils.cu
danpovey: Comment was wrong.  It is intentional, because the array will need to be separated into multiple arrays for different aux_label names.

k2/python/k2/fsa.py
danpovey: mm, I see what you mean...  I'll make it so at most one should be specified.

k2/python/k2/fsa.py
```diff
@@ -1096,34 +1095,35 @@ def from_str(cls, s: str) -> 'Fsa':
...
-            acceptor = False
-        arcs, aux_labels = _k2.fsa_from_str(s, acceptor, False)
-        ans = Fsa(arcs, aux_labels=aux_labels)
-        return ans
+        try:
+            arcs, aux_labels = _k2.fsa_from_str(s, num_aux_labels, False)
+            ans = Fsa(arcs)
+            if aux_labels is not None:
+                for i in range(aux_labels.shape[0]):
+                    setattr(ans, aux_label_names[i], aux_labels[i, :])
```
csukuangfj, line 1103: I see. `row_i` is for `aux_label_i`.

Issue #696, opened 2021-03-24
Crash "invalid device function" when running snowfall recipe


RuABraun:
Hi,
I want to run snowfall's librispeech recipe. I'm getting a crash when training starts that seems to have to do with k2:

```
# python -m torch.distributed.launch --nproc_per_node=1 ./mmi_bigram_train.py --world_size 1
World size: 1 Rank: 0                                                                                                                                                                                     
[F] /idiap/temp/rbraun/code/k2/k2/csrc/eval.h:134:void k2::EvalDevice(cudaStream_t, int32_t, LambdaT&) [with LambdaT = __nv_dl_wrapper_t<__nv_dl_tag<k2::RaggedShape (*)(const k2::RaggedShape&, int), k2::
Unsqueeze, 1>, int*, int>; cudaStream_t = CUstream_st*; int32_t = int] Check failed: e == cudaSuccess (98 vs. 0)  Error: invalid device function.                                                         
                                                        
                                     
[ Stack-Trace: ]                  
/idiap/temp/rbraun/code/k2/build/lib/libk2_log.so(k2::internal::GetStackTrace()+0x34) [0x7fdc7e038054]                                                                                                    
/idiap/temp/rbraun/code/k2/build/lib/libk2context.so(k2::internal::Logger::~Logger()+0x2a) [0x7fdc7e1ab3ca]
/idiap/temp/rbraun/code/k2/build/lib/libk2context.so(void k2::EvalDevice<__nv_dl_wrapper_t<__nv_dl_tag<k2::RaggedShape (*)(k2::RaggedShape const&, int), &k2::Unsqueeze, 1u>, int*, int> >(CUstream_st*, in
t, __nv_dl_wrapper_t<__nv_dl_tag<k2::RaggedShape (*)(k2::RaggedShape const&, int), &k2::Unsqueeze, 1u>, int*, int>&)+0x28a) [0x7fdc7e325a3a]
/idiap/temp/rbraun/code/k2/build/lib/libk2context.so(k2::Unsqueeze(k2::RaggedShape const&, int)+0x74e) [0x7fdc7e303c0e]                                                                                   
/idiap/temp/rbraun/code/k2/build/lib/libk2context.so(k2::FsaToFsaVec(k2::Ragged<k2::Arc> const&)+0x1fc) [0x7fdc7e1eceec]
[...]

PR #699, opened 2021-03-26
Free CUDA memory in a correct way when PYTORCH_NO_CUDA_MEMORY_CACHING is set

### LINKED ISSUE #698: Crash in CUDA 11
The following minimal demo will crash for CUDA 11 + Python 3.9 + PyTorch 1.7.1 + latest k2 (for both Debug & Release build)

~When it runs with `cuda-memcheck`, the process just hangs and seems never to terminate.~

When running with `cuda-memcheck`, it prints no errors after the crash.

The reason for the crash is similar to the one mentioned in https://github.com/k2-fsa/k2/issues/696#issuecomment-806060787

Note that `k2.closure` uses `torch.nonzero` internally.

https://github.com/k2-fsa/k2/blob/08a2bfeab04feaa95ad9a58bf67fb9e2b169cf29/k2/python/k2/fsa_algo.py#L660

The crash in snowfall also uses `torch.nonzero` https://github.com/k2-fsa/snowfall/blob/4a909a3a609d5a3444b14fc40d779f217e1263c1/egs/librispeech/asr/simple_v1/mmi_bigram_train.py#L62
```python
finite_indexes = torch.nonzero(mask).squeeze(1)
```

It will not crash for the same code built with CUDA 10.1

## Crash log
```
  File "/ceph-fj/open-source/k2/build_debug_11/./ab.py", line 23, in <module>
    main()
  File "/ceph-fj/open-source/k2/build_debug_11/./ab.py", line 19, in main
    ans = k2.closure(fsa)
  File "/root/fangjun/open-source/k2/k2/python/k2/fsa_algo.py", line 677, in closure
    new_value = fix_aux_labels(value, fsa.arcs.row_splits(1), arc_map)
  File "/root/fangjun/open-source/k2/k2/python/k2/fsa_algo.py", line 663, in fix_aux_labels
    minus_one_index[minus_one_index > src_start_state_last_arc_index] += 1
RuntimeError: invalid shape dimension -16711680
```

## Demo

```python
#!/usr/bin/env python3

import torch
import k2


def main():
    device = torch.device('cuda', 0)

    s = '''
        0 1 1 0.1
        1 2 2 0.2
        2 3 -1 0.3
        3
    '''
    fsa = k2.Fsa.from_str(s).to(device).requires_grad_(True)

    fsa.aux_labels = torch.tensor([10, 20, -1], dtype=torch.int32).to(device)
    ans = k2.closure(fsa)


if __name__ == '__main__':
    main()
```

csukuangfj:
Closes #698 

Also 

1. Apply the macro `-DTORCH_API_INCLUDE_EXTENSION_H` globally
2. Add `--compiler-options -rdynamic --compiler-options -lineinfo` for debug build to support `cuda-memcheck`
3. Fix crashes caused by CUB in CUDA 11
---

~Marked as WIP as it needs more test.~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,8 @@ endif()
 
 string(TOUPPER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_UPPERCASE)
 if("${CMAKE_BUILD_TYPE_UPPERCASE}" STREQUAL "DEBUG")
+  # refer to https://docs.nvidia.com/cuda/cuda-memcheck/index.html#compilation-options
+  # The two options is for cuda-memcheck' stack backtrace feature more useful.
   set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --compiler-options -rdynamic --compiler-options -lineinfo")
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,6 +28,11 @@ elseif(NOT CMAKE_BUILD_TYPE IN_LIST ALLOWABLE_BUILD_TYPES)
     choose one from ${ALLOWABLE_BUILD_TYPES}")
 endif()
 
+string(TOUPPER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_UPPERCASE)
+if("${CMAKE_BUILD_TYPE_UPPERCASE}" STREQUAL "DEBUG")
+  set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --compiler-options -rdynamic --compiler-options -lineinfo")
+endif()
+
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 option(BUILD_SHARED_LIBS "Whether to build shared or static lib" ON)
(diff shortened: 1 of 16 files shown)


CMakeLists.txt
```diff
@@ -28,6 +28,11 @@ elseif(NOT CMAKE_BUILD_TYPE IN_LIST ALLOWABLE_BUILD_TYPES)
     choose one from ${ALLOWABLE_BUILD_TYPES}")
 endif()
 
+string(TOUPPER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE_UPPERCASE)
+if("${CMAKE_BUILD_TYPE_UPPERCASE}" STREQUAL "DEBUG")
```
danpovey, line 32: Perhaps you could comment that we added these so that cuda-memcheck would give more useful
stack traces?  We don't want to think it was required, later on.

csukuangfj: Thanks, comments are added.

csukuangfj:
This also fixes the CUDA 11 crash reported in #696 
- https://github.com/k2-fsa/k2/issues/696#issuecomment-806060787

k2/csrc/pytorch_context.cu
```diff
@@ -14,6 +14,43 @@
 #include "k2/csrc/pytorch_context.h"
 
 namespace k2 {
+// CAUTION: This is a workaround to free the CUDA memory
```
csukuangfj, line 17: It is fixed in
https://github.com/pytorch/pytorch/commit/394b720e38455c2c07e6a31ef816edd0c5afbeb4

But before k2 uses a PyTorch release that contains that fix, we will keep this workaround.

Issue #701, opened 2021-03-29
Proposal for slight refactoring


danpovey:
Guys,
I have a proposal to slightly modify some parts of our codebase to make future maintenance easier.
The proposal is that Array1 and Array2 would store their dtype, probably as our `Dtype` (which is
an enum class).  Then we'd introduce a template parameter "Any" or "Generic" (which would just
be an empty class), and we'd allow casting to and from regular and Any type, e.g.

```
template<typename T>
 class Array1 <T> {
   ...
   Array1<k2::Any> &Any() { return *reinterpret_cast<Array1<Any>*>(this); }
   template <typename T> Array1<T> AsType() {
        // assert this type is Any and this->dtype_ matches T.
        return *reinterpret_cast<Array1<T*>(this);
    }
}
```
... and in certain circumstances we might provide an implementation for type Any, which
would dispatch over the concrete types.

Since the template parameter of type Ragged just maps to the template parameter of Array1, we'd also
be able to instantiate Ragged with type Any, although of course we'd need to provide implementations
that would do dispatching, wherever we use this.

The idea, then, is that when we deal with ragged arrays from python, we'd always, or at least
usually, actually have it as type `_k2.Ragged == k2::Ragged<Any>`, and we'd do the dispatching from C++, maybe
from inside our k2 library instead of at the pytorch wrapping level.

We might have a type k2.Ragged which would add various python utilities to those implemented at the C++ level, such as for pickling and unpickling via __getstate__ and __setstate__, which would enable I/O via torch.load and torch.save.

csukuangfj:
I would like to implement the following items first

- [ ] `Array1<k2::Any> &Any() { return *reinterpret_cast<Array1<Any>*>(this); }`
- [ ] `template <typename T> Array1<T> AsType()`
- [ ] `instantiate Ragged with type Any`
- [ ] when we deal with ragged arrays from python, we'd always, or at least usually, actually have it as type _k2.Ragged == k2::Ragged<Any>
- [ ] do the dispatching from C++, maybe from inside our k2 library instead of at the pytorch wrapping level.

---

As for serialization with `torch.save/torch.load`, it may need some investigation to support storage sharing
across serialization. Different ragged tensors may share the underlying memory, so only one copy needs to be saved.

danpovey:
Cool, but I have already been drafting some code... let me work on it a little first...

PR #702, opened 2021-03-29
Generic arrays

danpovey:
This is an early draft of the generic-array proposal..

### DIFF
--- a/k2/csrc/CMakeLists.txt
+++ b/k2/csrc/CMakeLists.txt
@@ -86,10 +86,11 @@ add_library(test_utils SHARED test_utils.cu)
 target_link_libraries(test_utils PUBLIC context gtest)
 
 # please sort the source files alphabetically
-set(cuda_tests
-  algorithms_test
+set(  algorithms_test
   array_ops_test
   array_test
+  cuda_tests
+  dtype_test
   fsa_algo_test
   fsa_test
   fsa_utils_test
--- a/k2/csrc/CMakeLists.txt
+++ b/k2/csrc/CMakeLists.txt
@@ -34,6 +34,7 @@ set(context_srcs
   fsa.cu
   fsa_algo.cu
   fsa_utils.cu
+  hash.cu
   host_shim.cu
   intersect.cu
   intersect_dense.cu
@@ -78,7 +79,9 @@ target_link_libraries(context PUBLIC k2_nvtx)
 target_link_libraries(context PUBLIC moderngpu)
 if(K2_USE_PYTORCH)
   target_link_libraries(context PUBLIC ${TORCH_LIBRARIES})
+  target_link_libraries(context PUBLIC ${TORCH_DIR}/lib/libtorch_python.so)
 endif()
+target_include_directories(context PUBLIC ${PYTHON_INCLUDE_DIRS})
 
 #---------------------------- Test K2 CUDA sources ----------------------------
 
--- a/k2/csrc/CMakeLists.txt
+++ b/k2/csrc/CMakeLists.txt
@@ -86,7 +86,8 @@ add_library(test_utils SHARED test_utils.cu)
 target_link_libraries(test_utils PUBLIC context gtest)
 
 # please sort the source files alphabetically
-set(  algorithms_test
+set(cuda_tests
+  algorithms_test
   array_ops_test
   array_test
   cuda_tests
--- a/k2/csrc/CMakeLists.txt
+++ b/k2/csrc/CMakeLists.txt
@@ -90,7 +90,6 @@ set(cuda_tests
   algorithms_test
   array_ops_test
   array_test
-  cuda_tests
   dtype_test
   fsa_algo_test
   fsa_test
--- a/k2/csrc/array.h
+++ b/k2/csrc/array.h
@@ -304,7 +304,8 @@ class Array1 {
      this is a CPU array, it would have much less overhead to index the Data()
      pointer. */
   T operator[](int32_t i) const {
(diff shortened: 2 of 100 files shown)


danpovey:
Relates to #701

danpovey:
This works to the extents that the test run, i.e. things that worked before still seem to work.
This adds very little real functionality because almost nothing has dispatching code, and
some implementation code doesn't even have asserts that type != Any, so that code would crash if
reached.

It does demonstrate one way to do dispatching via a macro, although that code is not tested.
My intention is that when we expose code to python we make the types Ragged<Any>, and 
we can either do the dispatching in k2/python/csrc/torch/, or create, in k2/csrc/, dispatching versions of the
actual functions by specializing the templates (in cases where this seems like it might be generally
useful, such as indexing).
I don't want to get into the habit of instantiating things for more types than we really need; this 
will make compilation slower than it already is.  So instead of using the macros, I'd be OK with 
switches or if-statements; or, if something is valid for only one type, just an assertion is OK,
and then use `something.Specialize<int32_t>()` or whatever the type is.

We may want to rename some interface functions to take out the "float" and "double".

This is probably not super urgent.  IIRC the reason this came to my mind is that I wanted to be able to do I/O
for ragged arrays in a generic way.

k2/csrc/array_ops.cu
```diff
@@ -491,5 +491,12 @@ void RowSplitsToRowIdsOffset(const Array1<int32_t> &row_splits_part,
   RowSplitsToRowIds(row_splits, row_ids_part);
 }
 
+template <>
+Array2<Any> ToContiguous(const Array2<Any> &src) {
+  Array2<Any> ans;
+  FOR_REAL_AND_INT32_TYPES(src.GetDtype(), T,
+                           ans = ToContiguous(src.Specialize<T>()).Generic());
+  return Array2<Any>();  // Silence warning
```
csukuangfj, line 499: Should we return `ans` here? 

Otherwise, the `ans` set inside the macro is never used.

k2/csrc/array.h
```diff
@@ -49,6 +55,40 @@ class Array1 {
                                        byte_offset_);
   }
 
+  // Return a reference to this viewed as `Any` type (for when we
+  // want a generic array without type informatiopn)
+  // Reverse this with Specialize().
+  Array1<Any> &Generic() {
+    return *reinterpret_cast<Array1<Any>*>(this);
+  }
+  Array1<Any> &Generic() const {
```
csukuangfj, line 64: I am afraid this won't compile if you instantiate it.

It is not possible to bind a `const lvalue reference` to a `non-const lvalue reference`.

It also shows that a large part of the change is not covered by unittests.

k2/csrc/CMakeLists.txt
```diff
@@ -90,6 +90,8 @@ set(cuda_tests
   algorithms_test
   array_ops_test
   array_test
+  cuda_tests
```
csukuangfj, line 93: For each xxx_test, it's expected that there is a file xxx_test.cu

`cuda_tests` can be removed as it is just a variable here.

```
set(variable_name
  value1
  value2
  ...
)
```

danpovey:
I am fixing errors with this, testing more of it, and making sure it compiles.  We'll need this at some point, and the longer I leave it the more work it will be to resolve conflicts later.

danpovey:
OK, I fixed the style_check issues, this is ready to merge as far as I am concerned.

k2/csrc/array.h
```diff
@@ -49,6 +55,44 @@ class Array1 {
...
+  // a template definition already.  It is an odd syntax.
+  template <typename U>
+  Array1<U> &Specialize() {
+    static_assert(std::is_same<T, Any>::value,
+                  "generic arrays not supported here");
+    K2_CHECK_EQ(dtype_, DtypeOf<U>::dtype);
+    return *reinterpret_cast<Array1<U>*>(this);
+  }
+  template <typename U>
+  Array1<U> &Specialize() const {
```
csukuangfj, line 87: Will this compile if you instantiate it?

The return value is a const lvalue, but the return type is an lvalue reference.

k2/csrc/array.h
```diff
@@ -454,17 +529,57 @@ class Array2 {
...
+    return *reinterpret_cast<const Array2<U>*>(this);
+  }
+
+  // Return a reference to this viewed as `Any` type (for when we
+  // want a generic array without type informatiopn)
+  // Reverse this with Specialize().
+  Array2<Any> &Generic() {
+    return *reinterpret_cast<Array2<Any>*>(this);
+  }
+  Array2<Any> &Generic() const {
```
csukuangfj, line 564: Since its return type is not a const lvalue reference, I am afraid
it will result in a compile error if you use this function.

The compiler does not complain at present because you are not using it.

k2/csrc/array.h
```diff
@@ -454,17 +529,57 @@ class Array2 {
...
+                       dtype_);
     } else {
       auto region = NewRegion(region_->context, static_cast<size_t>(dim0_) *
-                                                    static_cast<size_t>(dim1_) *
-                                                    ElementSize());
-      Array1<T> array(dim0_ * dim1_, region, 0);
+                              static_cast<size_t>(dim1_) *
+                              ElementSize());
+      Array1<T> array(dim0_ * dim1_, region, 0, dtype_);
+      static_assert(!K2_TYPE_IS_ANY(T),
```
csukuangfj, line 581: Should we move this static_assert to the beginning of this function?
I think that would be much clearer to say this function is not for `T==Any`.

csukuangfj:
+2

PR #708, opened 2021-04-06
Refactor hash code

danpovey:
This PR is a refactoring of the hash code which will make it easier to later extend it so we can store
more 'value' bits to enable larger composition.
@csukuangfj I discovered that we were wasting 2 bits, so although this is mostly just refactoring, it will
enable you to process slightly larger inputs.  I'll do the main changes to-morrow, but I can merge this now.


csukuangfj:
> it will enable you to process slightly larger inputs.


Thanks!

With an output beam size 3, the WER for test-clean is around 5.8%.

PR #711, opened 2021-04-08
Fix sorted intersection to be more rigorous about num bits needed.

danpovey:


danpovey:
Fangun, it would be nice if you could test with this version of the code.
I need to make similar changes to the non-sorted intersection, we can merge both together.

csukuangfj:
Ok, testing it.

csukuangfj:
> Fangun, it would be nice if you could test with this version of the code.


This pull-request has the same behavior as the master. That is, it can produce the same the WER as the master.
However, it does not support a larger output beam

danpovey:
OK.  This was a code reorganization to be more rigorous in certain rare conditions.  I'll merge.

danpovey:
FYI, for the longer term, the way we should probably address the intersection blow-up is by doing "pruned composition" (Kaldi has algorithms like this).  It's actually a relatively simple extension of the current composition code, where you'd need to pass in 'backward probs' (cost to the end from each state) for the lattice.

PR #713, opened 2021-04-10
Implement PruneOnArcPost.

csukuangfj:
It is used to prune arcs to avoid OOM when doing LM rescoring with the whole lattice.


danpovey:
thanks!!

PR #715, opened 2021-04-14
Implement k2.append to concatenate a list of FsaVec.

csukuangfj:
It is useful in lm rescoring with a large `num_paths`.
Basically, we will split a large FsaVec into a batch of small FsaVec and do `k2.intersect_device` on these batches.
In the end, we concatenate the resulting FsaVec. In this way, we can avoid CUDA OOM.


danpovey:
Lgtm!

PR #717, opened 2021-04-17
Discounted cumulative sum

danpovey:
This is something I wanted for a layer type I was working on.
Technically it does not have a lot to do with k2 directly, but in order to make it work I had to make some changes to the Tensor class, so what I suggest (mostly because it will save me time) is that we merge this if there are no objections, and after a few months if this function doesn't get any use we can delete that code.

### DIFF
--- a/k2/csrc/tensor.cu
+++ b/k2/csrc/tensor.cu
@@ -33,9 +33,8 @@ Shape::Shape(const std::vector<int32_t> &dims)
     strides_[i] = strides_[i + 1] * dims_[i + 1];
   }
 
-  num_element_ = ComputeNumElement();
+  num_elements_ = ComputeNumElements();
   is_contiguous_ = true;  // always be true here as we compute strides from dims
-  storage_size_ = ComputeStorageSize();
 }
 
 Shape::Shape(const std::vector<int32_t> &dims,
@@ -46,14 +45,13 @@ Shape::Shape(const std::vector<int32_t> &dims,
   K2_CHECK_EQ(static_cast<int32_t>(strides.size()), num_axes_);
   std::copy(dims.begin(), dims.end(), dims_);
   std::copy(strides.begin(), strides.end(), strides_);
-  num_element_ = ComputeNumElement();
+  num_elements_ = ComputeNumElements();
   is_contiguous_ = ComputeIsContiguous();
-  storage_size_ = ComputeStorageSize();
 }
 
-int64_t Shape::ComputeNumElement() const {
+int64_t Shape::ComputeNumElements() const {
   NVTX_RANGE(K2_FUNC);
-  if (num_axes_ == 0) return 0;
+  if (num_axes_ == 0) return 1;  // scalar
 
   int64_t elements = 1;
   for (int32_t i = 0; i < num_axes_; ++i) {
@@ -62,43 +60,50 @@ int64_t Shape::ComputeNumElement() const {
   return elements;
 }
 
-int64_t Shape::ComputeStorageSize() const {
+void Shape::GetReachableElems(int64_t *begin_out, int64_t *end_out) const {
   NVTX_RANGE(K2_FUNC);
-  if (num_axes_ == 0) return 0;
-
-  int64_t size = 1;
+  int64_t begin = 0,
+      end = 1;
   for (int32_t i = 0; i < num_axes_; ++i) {
-    if (dims_[i] == 0)
-      return 0;
-    else
-      size += (dims_[i] - 1) * (int64_t)strides_[i];
+    if (dims_[i] == 0) {
+      goto empty_output;
+    } else if (strides_[i] > 0) {
+      end += (dims_[i] - 1) * (int64_t)strides_[i];
+    } else {
+      begin += (dims_[i] - 1) * (int64_t)strides_[i];
+    }
   }
-  K2_CHECK_GE(size, 0);
-  return size;
+  *begin_out = begin;
+  *end_out = end;
+  return;
+empty_output:
+  *begin_out = 0;
+  *end_out = 0;
 }
+
+int64_t Shape::StorageSize() const {
+  int64_t begin, end;
+  GetReachableElems(&begin, &end);
+  return end - begin;
+}
+
 
 bool Shape::ComputeIsContiguous() const {
   NVTX_RANGE(K2_FUNC);
-
-  // It may happen that all strides are zero,
-  // i.e., the tensor contains only one element.
-  // In this case, the tensor is contiguous.
-  int32_t s = 0;
-  for (int32_t i = num_axes_ - 1; i >= 0; --i) {
-    K2_CHECK_GE(strides_[i], 0);
-    s += strides_[i];
-  }
-  if (s == 0) return true;
-
   int64_t z = 1;
   for (int32_t i = num_axes_ - 1; i >= 0; --i) {
-    K2_CHECK_GE(strides_[i], z);
     if (dims_[i] != 1) {
       if (strides_[i] != z) return false;
       z *= dims_[i];
     }
   }
   return true;
+}
+
+void Shape::SetStride(int32_t axis, int32_t stride) {
+  K2_CHECK_LT(static_cast<uint32_t>(axis), static_cast<uint32_t>(num_axes_));
+  strides_[axis] = stride;
+  ComputeIsContiguous();
 }
 
 std::ostream &operator<<(std::ostream &os, const Shape &shape) {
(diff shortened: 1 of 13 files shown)


danpovey:
... some of these changes in Tensor are just cosmetic.  Others to support negative stride, and to allow "scalars" (Tensors with no axes).

k2/csrc/tensor.cu
```diff
@@ -62,34 +60,38 @@ int64_t Shape::ComputeNumElement() const {
...
   NVTX_RANGE(K2_FUNC);
-  if (num_axes_ == 0) return 0;
-
-  int64_t size = 1;
+  int64_t begin = 0,
+      end = 1;
   for (int32_t i = 0; i < num_axes_; ++i) {
-    size += (dims_[i] - 1) * (int64_t)strides_[i];
+    if (dims_[i] == 0) {
+      goto empty_output;
```
csukuangfj, line 69: Should we avoid `goto`?

k2/csrc/tensor_ops.cu
```diff
@@ -507,4 +507,249 @@ Tensor SimpleRaggedIndexSelect1D(Tensor &src, Ragged<int32_t> &indexes) {
...
+    } else {
+      DiscountedCumSumCpuImpl<double>(N, T,
+                                      src.Data<double>(), src_stride0,
+                                      gamma.Data<double>(), gamma_stride0,
+                                      dest->Data<double>(), dest_stride0,
+                                      stride1);
+
+    }
+  } else {
+        K2_LOG(ERROR) << "This algorithm only instantiated for float and double; type is "
```
csukuangfj, line 729: I would suggest `K2_LOG(FATAL)` to terminate the program (i.e., to throw an exception).

k2/csrc/tensor_ops.cu
```diff
@@ -507,4 +507,249 @@ Tensor SimpleRaggedIndexSelect1D(Tensor &src, Ragged<int32_t> &indexes) {
...
+      block_dim(ThreadsPerBlock, 1, 1);
+  K2_CUDA_SAFE_CALL(DiscountedCumSumKernel<Real, ThreadsPerBlock>
+                    <<<grid_dim, block_dim, 0, stream>>>(N, T, x, x_stride0,
+                                                         gamma, gamma_stride0,
+                                                         y, y_stride0, stride1));
+}
+
+
+template <typename Real>
+void DiscountedCumSumCpuImpl(int N, int T,
```
csukuangfj, line 658: Should we make it `static` as it is supposed to be local in this translation unit?
(Also for the above CUDA kernel.)

k2/csrc/tensor_ops.cu
```diff
@@ -507,4 +507,249 @@ Tensor SimpleRaggedIndexSelect1D(Tensor &src, Ragged<int32_t> &indexes) {
...
+      y_0 = x_0
+      y_i = x_i + y_{i-1} gamma_i
+   for 0 < i < T.  (This is done as a generic inclusive-scan/inclusive-sum with a special
+   reduction op).
+
+  See DiscountedCumSumElement and CombineCumSumOp for how we use a special operator to
+  do this as an inclusive-sum.
+
+  The tensors involved must be 2-dimensional with dimensions (N, T) where N is
+  the batch size and T the time duration, and the stride on the T dim must be 1.
```
csukuangfj, line 559: > the stride on the T dim must be 1.

I think this constraint no longer exists.

Also, the documentation about the `stride1` for the arguments `gamma` and `y` should
also be updated.

k2/csrc/tensor.h
```diff
@@ -197,9 +230,13 @@ class Tensor {
 
   ContextPtr &Context() const { return impl_->data->context; }
 
+  TensorImplPtr Impl() { return impl_; }
```
csukuangfj, line 233: Should we append `const` to `Implt()`?

k2/csrc/tensor_test.cu
```diff
@@ -188,14 +229,25 @@ TEST(TensorTest, Tensor) {
...
+    int32_t n = t.NumElements();
     const int32_t *t_data = t.Data<int32_t>();
-    for (int32_t i = 0; i != t.Nelement(); ++i) {
+    for (int32_t i = 0; i != t.NumElements(); ++i) {
       EXPECT_EQ(t_data[i], expected_data[i]);
     }
+
+    for (int32_t axis = -3; axis < 3; axis++) {
+      Tensor t_flip = Flip(t, axis);
+      int32_t *t_flip_data = t_flip.Data<int32_t>();
```
csukuangfj, line 244: I would suggest `const int32_t *t_flip_data`.

k2/csrc/tensor_ops_test.cu
```diff
@@ -325,10 +326,69 @@ static void TestSimpleRaggedIndexSelect1D() {
...
+    Array2<float> y(cuda_context, M, T);
+    y = -10.0;
+
+    bool flip = (i % 2 == 1);
+
+    Array2<float> x_cpu = x.To(cpu_context),
+        gamma_cpu = gamma.To(cpu_context),
+        y_cpu(cpu_context, M, T);
+
+    Tensor x_ten = x.ToTensor(),
```
csukuangfj, line 360: I would suggest using

https://github.com/k2-fsa/k2/blob/0bcc330db4b566a22a431e692b10880a82c4b1ed/k2/csrc/tensor_ops_test.cu#L59-L60

to generate tensors with stride0 != 1.

The current test covers only a single case: `stride0 == 1`.

k2/csrc/tensor_ops_test.cu
```diff
@@ -325,10 +326,69 @@ static void TestSimpleRaggedIndexSelect1D() {
...
 }
 
+
+TEST(Tensor, DiscountedCumSum) {
+  for (int32_t i = 0; i < 4; i++) {
+    int32_t M = RandInt(0, 1000),
+        T = RandInt(0, 2000);  // TODO: increase.
+    while (M * T > 10000) {  // don't want test to take too long.
+      M /= 2;
+      T /= 2;
```
csukuangfj, line 343: It may happen that outside the while loop,  one of `M` or `T` is 0, which will
make the test fail.

k2/csrc/tensor_ops_test.cu
```diff
@@ -325,10 +326,69 @@ static void TestSimpleRaggedIndexSelect1D() {
...
 
+
 TEST(Index, SimpleRaggedIndexSelect1D) {
   TestSimpleRaggedIndexSelect1D<float>();
   TestSimpleRaggedIndexSelect1D<double>();
   TestSimpleRaggedIndexSelect1D<int32_t>();
 }
 
+
+TEST(Tensor, DiscountedCumSum) {
```
csukuangfj, line 337: Should we make the function a template?
It currently tests only the type `float`. `double` is not tested.

csukuangfj:
OK with me to merge it.

k2/csrc/tensor_ops_test.cu
danpovey: Because it's flipping, it's testing stride1 == -1 as well...

k2/csrc/tensor_ops_test.cu
danpovey: In general things are supposed to work for zero-size arrays...

k2/csrc/tensor_ops_test.cu
danpovey: i did this. also addressed other comments.  will merge.

k2/csrc/tensor_ops.cu
```diff
@@ -507,4 +507,248 @@ Tensor SimpleRaggedIndexSelect1D(Tensor &src, Ragged<int32_t> &indexes) {
...
+
+    @param [in] N    The batch size, i.e. number of separate sequences.  We expect
+               that N <= gridDim.x * gridDim.y.
+    @param [in] T    The sequence length.  There is no constraint on the sequence
+                     length; the kernel deals with ThreadsPerBlock items at a time,
+                     and takes care of T > ThreadsPerBlock by looping.
+    @param [in] x    Pointer to the x input data, which is an array of shape (N,T)
+    @param [in] x_stride0  Stride along axis 0 of of the `x` data
+    @param [in] gamma   Pointer to the gamma input data, which is an array of shape (N,T);
+                     the stride along the N axis is `gamma_stride` and the stride along
```
csukuangfj, line 577: > and the stride along the T axis is 1.

It is still not updated. Also for line 582.

k2/csrc/tensor_ops.cu
```diff
@@ -507,4 +507,248 @@ Tensor SimpleRaggedIndexSelect1D(Tensor &src, Ragged<int32_t> &indexes) {
...
+      cur_sum = x[t * stride1] + cur_sum * gamma[t * stride1];
+      y[t * stride1] = cur_sum;
+    }
+  }
+}
+
+
+void DiscountedCumSum(const Tensor &src, const Tensor &gamma, Tensor *dest) {
+  // check contexts compatible:
+  if (!(IsCompatible(src, gamma) && IsCompatible(src, *dest))) {
```
csukuangfj, line 675: Also, the following `LOG(ERROR)` -> `LOG(FATAL)`.

PR #719, opened 2021-04-19
Add NVTX to cover all functions.

csukuangfj:


danpovey:
After discussion, we'll remove this NVTX stuff from functions that are less than about 100 instructions with no system calls or significant library calls. 
My concern is (a) bulking up the qdrep files, and (b) slowing the program down.

danpovey (approved):
Looks OK to me--thanks!

PR #720, opened 2021-04-19
Add Fsa.convert_attr_to_ragged_().

csukuangfj:
Related to https://github.com/k2-fsa/snowfall/pull/166#discussion_r615519499

> I think it would be more elegant to make this a function invoked like den_graphs.convert_attr_to_ragged('aux_labels', remove_eps=True).
I imagine this will still have [-1] on the final-arcs, which is probably what we want
as the num graphs likely have this.


danpovey (approved):
LGTM, thanks!

PR #726, opened 2021-04-21
Implement ComposeArcMaps for 1-D arrays.

csukuangfj:
Related to https://github.com/k2-fsa/snowfall/issues/165

I am testing it in snowfall.


k2/csrc/fsa_utils.h
```diff
@@ -527,6 +527,21 @@ FsaVec FsaVecFromArcIndexes(FsaVec &fsas, Ragged<int32_t> &best_arc_indexes);
 Ragged<int32_t> ComposeArcMaps(Ragged<int32_t> &step1_arc_map,
                                Ragged<int32_t> &step2_arc_map);
 
+/*
+  Compose arc maps from two successive FSA operations which give arc maps as
+  a 1-D array.
```
danpovey, line 532: I notice the Index() function in array_ops.h is the same as this except it sets the value to 0 instead of -1.
How about adding an extra arg to that function that is the value to be used if the index is -1, defaulting to 0?
I prefer to use general-purpose operations.

csukuangfj: Yes, I agree.

k2/csrc/array_ops.cu
```diff
@@ -304,10 +304,39 @@ Array1<int32_t> GetCounts(ContextPtr c, const int32_t *src_data,
...
+
+    constexpr std::size_t kTreshold = (static_cast<std::size_t>(1) << 33);
+
+    if (temp_storage_bytes < kTreshold) {
+      RegionPtr temp_storage = NewRegion(c, temp_storage_bytes);
+      K2_CHECK_CUDA_ERROR(cub::DeviceHistogram::HistogramEven(
+          temp_storage->data, temp_storage_bytes, src_data, ans_data, n + 1, 0,
+          n, src_dim, c->GetCudaStream()));
+    } else {
+      // split the array and do a recursive call
```
csukuangfj, line 316: Relates to https://github.com/NVIDIA/cub/issues/288

k2/csrc/fsa_utils.h
csukuangfj: Fixed.

k2/csrc/array_ops.cu
danpovey: kThreshold.

danpovey:
You can merge this when it's convenient for you, looks OK.

k2/csrc/array_ops.cu
csukuangfj: Fixed. Thanks

PR #727, opened 2021-04-22
Refactor ragged ops

danpovey:
This fixes an issue Fangjun found where we exceeded the int16_t size constraint of GetTaskRedirect().
I have refactored some things to not use GetTaskRedirect() (which is not a good algorithm, especially since
we had to remove dynamic parallelism to work around an Nsight Systems bug).

I have also reduced the number of kernels used by quite a bit-- should be considerably faster, although I have not 
benchmarked it.


danpovey:
I want to merge this since it resolves a problem Fangjun had, BUT, can someone please create a benchmark for
Stack and/or Cat in ragged_ops_benchmark.cu and get some numbers?
Of this PR vs. current master == a16d375d0457fd015828ab56d76deedf354d5.
I notice that ragged_ops_benchmark.cu is super slow, perhaps that can be fixed too by changing the num iters.
I'm afraid I am not super familiar with the benchmarking code.
I am hoping someone will volunteer for this.

danpovey:
It is stacking on axis 0, and also indexing arrays on axis 0, which would be affected by this.

PR #732, opened 2021-04-30
Add `allow_truncate` to DenseFsaVec.__init__

csukuangfj:
See https://github.com/k2-fsa/snowfall/pull/181#discussion_r623684445

> I think the easiest way to resolve this would be to add an option with default:
> `allow_truncate=0`
> to `DenseFsaVec.__init__()` in k2, so it would truncate the last few frames from a segment if necessary to
> make it fit within the nnet output; we could set this to 2 or 3, for instance. It seems to me that this
> would be very easy to implement. The only thing that would not be quite right is in mmi.py where we do:
> 
> ```
>         tot_score, tot_frames, all_frames = get_tot_objf_and_num_frames(
>             tot_scores,
>             supervision_segments[:, 2]
>         )
> ```
> 
> supervision_segments[:, 2] may not have quite the right lengths. But we could easily add a member function to the
> DenseFsaVec object in Python, that would provide the correct length (e.g. we could save something in the constructor, to avoid an extra GPU->CPU transfer).

---

Ready for review.

### DIFF
--- a/k2/python/k2/dense_fsa_vec.py
+++ b/k2/python/k2/dense_fsa_vec.py
@@ -116,18 +116,19 @@ class DenseFsaVec(object):
                                  device='cpu',
                                  dtype=torch.int32)
         row_splits[1:] = torch.tensor(last_frame_indexes) + 1
-        # minus one to exclude the fake row [0, -inf, -inf, ...]
-        duration = row_splits[1:] - row_splits[:-1] - 1
 
         row_splits = row_splits.to(device)
         self.dense_fsa_vec = _k2.DenseFsaVec(scores, row_splits)
         self.scores = scores  # for back propagation
-        self._duration = duration.to(device)
 
     @property
     def duration(self) -> torch.Tensor:
         '''Return the duration of each seq.
         '''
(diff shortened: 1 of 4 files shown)


k2/python/k2/dense_fsa_vec.py
```diff
@@ -116,18 +116,19 @@ def __init__(self,
...
         row_splits = row_splits.to(device)
         self.dense_fsa_vec = _k2.DenseFsaVec(scores, row_splits)
         self.scores = scores  # for back propagation
-        self._duration = duration.to(device)
 
     @property
     def duration(self) -> torch.Tensor:
         '''Return the duration of each seq.
         '''
+        if not hasattr(self, '_duration'):
```
danpovey, line 128: I would prefer to cache the CPU version of this, in the object.
CUDA->CPU transfers really slow down the pipeline because they force it to wait to sync the device.

csukuangfj: Fixed.

danpovey:
Thanks!!  Merging.

PR #734, opened 2021-05-01
Build conda packages for k2.

csukuangfj:
Closes  #733 

## Usage

To install a pre-build conda package, use

~conda install -c k2-fsa k2~

To build a conda package from source, you can run

```bash
export K2_TORCH_VERSION=1.7.1
export K2_CUDA_VERSION=10.1

# Go to the source tree of k2
./scripts/build_conda.sh
```

You will see the following output at the end
```
...
# Automatic uploading is disabled
# If you want to upload package(s) to anaconda.org later, type:

anaconda upload /root/fangjun/software/anaconda3/2020.11/conda-bld/linux-64/k2-0.3.3.dev20210501-cu101+torch1.7.1+py38.tar.bz2

# To have conda build upload to anaconda.org automatically, use
# $ conda config --set anaconda_upload yes
anaconda_upload is not set.  Not uploading wheels: []
...
```

I've uploaded them to <https://anaconda.org/k2-fsa/k2>

---

## TODOs

- [x] Add documentation, describing how to install k2 with `conda install -c k2-fsa k2`
- [x] Add GitHub actions to build conda packages automatically and upload them to <https://anaconda.org/>
- [x] Support various version combinations of Python, PyTorch, and CUDA
  - [x] Python versions: 3.6, 3.7, 3.8, 3.9
  - [x] PyTorch versions: 1.5.0, 1.5.1, 1.6.0, 1.7.0, 1.7.1, 1.8.0
  - [x] CUDA versions: 10.1, 10.2, 11.0, 11.1

PR #740, opened 2021-05-14
Fix building pip packages.

csukuangfj:
Now k2 can be installed by

```bash
  python3 setup.py install
```

---

I have removed `scripts/build_pip.sh`.


setup.py
```diff
@@ -17,22 +50,72 @@
...
+        # build/lib.linux-x86_64-3.8
+        os.makedirs(self.build_lib, exist_ok=True)
+
+        k2_dir = os.path.dirname(os.path.abspath(__file__))
+
+        cmake_args = os.environ.get('K2_CMAKE_ARGS', '')
+        make_args = os.environ.get('K2_MAKE_ARGS', '')
+
+        if cmake_args == '':
+            cmake_args = '-DCMAKE_BUILD_TYPE=Release'
```
pzelasko, line 93: I think it was already suggested somewhere, but wouldn't `RelWithDebInfo` help solving issues reported by the users by giving more informative stack traces / gdb / cuda-gdb information?

csukuangfj: It was mentioned in https://github.com/k2-fsa/k2/pull/734#discussion_r629095496

> I am not sure whether it's worthwhile to change -O3 to -O2. If users want to report issues, I would
recommend them to use Debug.

I often see projects are released with `Release` build, not `RelWithDebInfo` build.

pzelasko: Alright, missed that one. Cool, makes sense.

danpovey:
LGTM!  (Not thorough check...)

csukuangfj:
Will merge in the morning.

PR #741, opened 2021-05-17
Fix k2.to_str() when aux_labels is a ragged tensor.

csukuangfj:


danpovey:
LGTM.
We may want some way of printing out such aux_labels at some point (even if it can't be used to create another FSA).

csukuangfj:
> We may want some way of printing out such aux_labels at some point (even if it can't be used to create another FSA).

`print(fsa)` or `str(fsa)` output the aux_labels and other attributes of an Fsa, if there are any.

PR #743, opened 2021-05-20
Expand ragged labels and read/write ragged attributes in text form

danpovey:


k2/csrc/fsa_algo.cu
```diff
@@ -1115,6 +1115,61 @@ FsaOrVec ExpandArcs(FsaOrVec &fsas, RaggedShape &labels_shape,
...
+  } else {
+    K2_CHECK_EQ(fsas.NumAxes(), 2);
+    int32_t num_arcs = fsas.TotSize(1), num_states = fsas.Dim0();
+    K2_EVAL(c, num_arcs, lambda_fix_final_labels_2axis, (int32_t arc_idx01) -> void {
+        // we name this as if it is an aux_label, but it could have any name.
+        int32_t cur_aux_label = labels_data[arc_idx01 * labels_stride];
+        const Arc &arc = arcs_data[arc_idx01];
+        if (arc.dest_state + 1 == num_states) {
+          // dest_state is final-state.
+          K2_DCHECK_EQ(arc.label, -1);
```
csukuangfj, line 1157: Should we also add this check to the kernel for `num_axes==3`?

k2/csrc/fsa_algo_test.cu
```diff
@@ -149,6 +150,44 @@ TEST(ArcSort, NonEmptyFsaVec) {
...
+    FsaVec fsas = Stack(0, 2, fsa_ptrs.data(), nullptr);
+
+
+    Array1<int32_t> labels(context, "[ 0 -1 1 2 3 4 5 0 ]"),
+        correct_labels(context, "[ 0 0 1 2 3 4 5 -1 ]");
+
+    FixFinalLabels(fsa, labels.Data(), 1);
+    EXPECT_EQ(true, Equal(labels, correct_labels));
+
+
```
csukuangfj, line 180: Maybe you forgot to call `FixFinalLabels` for the FsaVec `fsas`.

Also, you need to save a copy of `labels` for later reference when testsing `fsas` as it is modified in
line 177
```cpp
FixFinalLabels(fsa, labels.Data(), 1);
```

k2/python/k2/fsa_algo.py
```diff
@@ -920,3 +920,60 @@ def prune_on_arc_post(fsas: Fsa, threshold_prob: float,
     out_fsa = k2.utils.fsa_from_unary_function_tensor(fsas, ragged_arc,
                                                       arc_map)
     return out_fsa
+
+
+def expand_ragged_labels(
```
csukuangfj, line 925: Shall we change it to `expand_ragged_attributes`?

The `labels` of an Fsa is always a normal tensor. Only the attributes
may be ragged.

k2/python/tests/expand_ragged_labels_test.py
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env python3
+#
+# Copyright (c)  2021  Xiaomi Corp.       (authors: Fangjun Kuang, Daniel Povey)
+#
+# See ../../../LICENSE for clarification regarding multiple authors
+
+# To run this single test, use
+#
+#  ctest --verbose -R  expand_arcs_test_py
```
csukuangfj, line 9: expand_arcs_test_py --> expand_ragged_labels_test_py

k2/python/tests/expand_ragged_labels_test.py
```diff
@@ -0,0 +1,178 @@
...
+                        dest.scores,
+                        torch.tensor([10, 20, 0, 0, 0, 30, 0],
+                                     dtype=torch.float32,
+                                     device=device)))
+
+                assert torch.all(
+                    torch.eq(
+                        dest.int_attr,
+                        torch.tensor([1, 2, 0, 0, 0, 3, 0], dtype=torch.int32, device=device)))
+                _k2.fix_final_labels(dest.arcs, dest.int_attr)
```
csukuangfj, line 145: Do we need to add a note in the function `k2.expand_ragged_labels`
saying that the user may need to call `_k2.fix_final_labels`?

BTW, `_k2.fix_final_labels` is not exported to the `k2` module in Python.

csukuangfj: Or as another alternative, can we add an extra argument to `_k2.expand_ragged_labels`
to accept also a list of normal tensors. We can change them in a single kernel, which is more efficient.
And it is less error-prone as there is no need for the user to call `_k2.fix_final_labels`.

danpovey: Regarding needing to call _k2.fix_final_labels:
  The reason I'm not mentioning this is that we don't have a general requirement for attributes, that for final arcs they must have the value -1.  I am enforcing this in this case because it seems likely to be what would be wanted, but I may change my mind about this in future and let them be zero (as I think I envisaged at one point early on).
Here, I just wanted to test the function; I don't have an immediate use for it.

csukuangfj:
+2

danpovey:
I know this isn't ideal, but I have extended this PR to include additional features: namely, to support reading and writing ragged attributes as part of the text representation of FSTs, included on the same line as the arcs.
This should make it easier to interface with external scripts and software.

PR #749, opened 2021-05-31
Some refactoring about tensor properties, make them handled more gene…

danpovey:
…rally, and make filler symbol configurable

### DIFF
--- a/k2/python/k2/fsa.py
+++ b/k2/python/k2/fsa.py
@@ -342,6 +342,10 @@ class Fsa(object):
             if name == 'labels':
                 assert value.dtype == torch.int32
                 self.arcs.values()[:, 2] = value
+                # fix_final_labels() will change 0's to -1's and vice versa to
+                # ensure that constraints on where final-labels should appear,
+                # are satisfied.
+                _k2.fix_final_labels(self.arcs, None)
                 self.__dict__['_properties'] = None
                 # access self.properties which will do a validity check on the
                 # modified FSA after getting the properties
--- a/k2/python/k2/fsa.py
+++ b/k2/python/k2/fsa.py
@@ -216,18 +216,17 @@ class Fsa(object):
         _ = self.properties
 
     def to_str(self, openfst: bool = False) -> str:
-        aux_labels = []  # actually a list of all integer tensor attributes, not
-                         # just the one named 'aux_labels'
+        extra_labels = []
         ragged_labels = []
         for name, value in sorted(self.named_tensor_attr(include_scores=False)):
             if isinstance(value, torch.Tensor) and value.dtype == torch.int32:
-                aux_labels.append(value)
+                extra_labels.append(value)
             elif isinstance(value, _k2.RaggedInt):
                 ragged_labels.append(value)
 
         if self.arcs.num_axes() == 2:
             ans = 'k2.Fsa: ' + _k2.fsa_to_str(self.arcs, openfst=openfst,
-                                              aux_labels=aux_labels,
+                                              extra_labels=extra_labels,
                                               ragged_labels=ragged_labels)
         else:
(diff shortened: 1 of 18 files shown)


k2/python/k2/fsa.py
```diff
@@ -259,6 +259,28 @@ def __str__(self) -> str:
...
+            fsa.foo_filler = -1
+        which will mean the "filler" for attribute fsa.foo is -1; and
+        this will get propagated when you do FSA operations, like any
+        other non-tensor attribute.  The filler is the value that means
+        "nothing is here" (like epsilon).
+        Caution: you should use a value that is castable to float and back
+        to integer without loss of precision, because currently the
+        `default_value` parameter of `index_select` in ./ops.py is a float.
+        '''
+        try:
```
csukuangfj, line 276: How about
```python
assert attribute_name != 'aux_labels', 'you cannot set the filler for aux_labels'

ans = getattr(self, attribute_name + '_filler', 0)

return ans
```

danpovey: This doesn't work because we still *get* the filler for 'aux_labels', we just expect it to not be set.  Will change it, though, to also allow it to be set to 0.

csukuangfj: In that case, 
```python
ans = getattr(self, attribute_name + '_filler', 0) 
assert attribute_name != 'aux_labels' or ans == 0, 'you cannot set the filler for aux_labels' 
return ans
```
My intention is that

- (1) You don't need to use `try ... except` here. `getattr` accepts a default value

- (2) I usually see people using `assert some_expr, "some message"` in Python.
People in C/C++ use `assert(some_expr && "some message");`

k2/python/k2/fsa.py
```diff
@@ -791,6 +821,55 @@ def requires_grad_(self, requires_grad: bool) -> 'Fsa':
...
+           It is OK if src_name and/or dest_name equals 'labels' or 'scores',
+           but these special attributes won't be deleted.
+        '''
+        assert src_name != dest_name
+        try:
+            value = getattr(self, src_name)
+            if src_name == 'labels':
+                value = value.clone()
+            setattr(self, dest_name, value)
+            try:
```
csukuangfj, line 849: If `src_name` is `scores`, you're deleting it, which does not match the `NOTE` given
in the comment.

k2/python/k2/fsa.py
```diff
@@ -791,6 +821,55 @@ def requires_grad_(self, requires_grad: bool) -> 'Fsa':
...
+                deleted.
+        Returns: `self`
+
+        Note:
+           It is OK if src_name and/or dest_name equals 'labels' or 'scores',
+           but these special attributes won't be deleted.
+        '''
+        assert src_name != dest_name
+        try:
+            value = getattr(self, src_name)
```
csukuangfj, line 845: The implementation does not enforce that `src_name` is from the tensor attribute,
though the comment requires that it MUST be a tensor attribute.

Users may use it incorrectly.

k2/python/k2/utils.py
```diff
@@ -440,7 +447,8 @@ def fsa_from_unary_function_tensor(src: Fsa, dest_arcs: _k2.RaggedArc,
 
 
 def fsa_from_unary_function_ragged(src: Fsa, dest_arcs: _k2.RaggedArc,
-                                   arc_map: _k2.RaggedInt) -> Fsa:
+                                   arc_map: _k2.RaggedInt,
+                                   remove_filler: bool = True) -> Fsa:
```
csukuangfj, line 451: This newly added argument is not documented at all.

k2/python/k2/fsa.py
danpovey: OK, I will check. (We also allow 'labels' as a special case, which I'll note).

PR #752, opened 2021-06-04
Fix top-sort test failure (there can be just 1 arc)

danpovey:


danpovey:
it was a failure on windows CPU-only build.

Issue #755, opened 2021-06-07
FsaReplace operation


danpovey:
```
/*
   Replace, in `index`, labels symbol_range_begin <= label < symbol_range_begin+src.Dim0()
   with the Fsa indexed `label - symbol_range_begin` in `src`, identifying the source and destination
   states of the arc in `src` with the initial and final states in
`src[label - symbol_range_begin]`.
   Arcs with labels outside this range are just copied.  Caution: the
result may not be a valid Fsa  because labels on final-arcs in `src` (which will be -1) may end
up on non-final arcs in the result; you can use FixFinalLabels() to fix this.

  @param [in] src   FsaVec containing individual Fsas that we'll be
         inserting into the result.  No FSA in `src` may have arcs
        entering its initial state; this function will crash if this requirement
         is violated.
  @param [in] index  Fsa or FsaVec (2 or 3 axes) that dictates the overall
                    structure of the result (the result will have the same number 
                   of axes as`index`.
  @param [in] symbol_range_begin  Beginning of the range (interval) of
             symbols that are to  be replaced with Fsas.  Symbols numbered
              symbol_range_begin <= i < src.Dim0()
                will be replaced with the Fsa in `src[i - symbol_range_begin]`
  @param [out,optional] arc_map_src  If not nullptr, will be set to a new
array that
             maps from arc-indexes in the result to the corresponding arc
             in `src`, or -1 if there was no such arc (for out-of-range symbols in `index`)
   @param [out,optional] arc_map_index  If not nullptr, will be set to a new array
            that maps from arc-indexes in the result to the arc in `index`
            that it originates from, only if it includes the weight from that arc in `index`; and -1
            otherwise).  For arcs that result from inserting an Fsa in `src`, (say, src[i]) they include the
            weight from the arc in `index` if the arc was from the initial state in src[i].

*/
FsaOrVec ReplaceFsa(FsaVec src, FsaOrVec index, int32_t symbol_range_begin,
                                      Array1<int32_t> *arc_map_src =
nullptr,
                                      Array1<int32_t> *arc_map_index =
nullptr);
```

This was discussed here https://github.com/k2-fsa/snowfall/issues/191
but creating an issue in k2 which is where it belongs.  I made a few changes to the 
draft interface.

csukuangfj:
>   @param [in] symbol_range_begin  Beginning of the range (interval) of
             symbols that are to  be replaced with Fsas.  Symbols numbered
[...]

PR #759, opened 2021-06-08
Implement FsaReplace operation

pkufool:
Implements FsaReplace operation  #755

### DIFF
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -1277,7 +1277,9 @@ FsaOrVec ReplaceFsa(FsaVec &src, FsaOrVec &index, int32_t symbol_range_begin,
   // state itself"), then `num_ostates_for[idx01] = 1`, meaning "keep the
   // original state".  Otherwise, idx1 - 1 represents an arc_idx2 [into `index`]
   // and we set `num_ostates_for[idx01] = max(0, state_num-1)`, where state_num
-  // is the states count of the fsa in `src` that would repalce into this arc.
+  // is the states number of the fsa in `src` that would repalce into this arc,
+  // the final state of this fsa will identify with the dest-state of this arc,
+  // so we minus 1.
   Array1<int32_t> num_ostates_for(c, foo_size + 1);
   int32_t *num_ostates_for_data = num_ostates_for.Data();
   const Arc *index_arcs_data = index.values.Data();
@@ -1362,19 +1364,22 @@ FsaOrVec ReplaceFsa(FsaVec &src, FsaOrVec &index, int32_t symbol_range_begin,
           num_arcs = orig_num_arcs;
         } else {
           // All inserted states have the same num of arcs as in the src.
+          // note: the prefix `index_` means it is an idxXXX w.r.t. `index`.
+          // the prefix `src_` means the variable is an idxXXX w.r.t. `src`.
           int32_t index_arc_idx2 = idx2 - 1,
                   index_arc_idx01x = index_row_splits2_data[idx01],
                   index_arc_idx012 = index_arc_idx01x + index_arc_idx2,
                   index_label = index_arcs_data[index_arc_idx012].label,
-                  src_idx0 = index_label - symbol_range_begin;
-          K2_CHECK_GE(src_idx0, 0);
-          K2_CHECK_LT(src_idx0, num_src_fsas);
-          int32_t src_idx1 = idx3,
-                  src_idx0x = src_row_splits1_data[src_idx0],
-                  src_idx01 = src_idx0x + src_idx1,
-                  src_idx01x = src_row_splits2_data[src_idx01],
-                  src_idx01x_next = src_row_splits2_data[src_idx01 + 1],
-                  src_num_arcs = src_idx01x_next - src_idx01x;
+                  src_fsa_idx0 = index_label - symbol_range_begin;
+          K2_CHECK_GE(src_fsa_idx0, 0);
+          K2_CHECK_LT(src_fsa_idx0, num_src_fsas);
+          int32_t src_state_idx1 = idx3,
+                  src_state_idx0x = src_row_splits1_data[src_fsa_idx0],
+                  src_state_idx01 = src_state_idx0x + src_state_idx1,
+                  src_arc_idx01x = src_row_splits2_data[src_state_idx01],
+                  src_arc_idx01x_next =
+                    src_row_splits2_data[src_state_idx01 + 1],
+                  src_num_arcs = src_arc_idx01x_next - src_arc_idx01x;
           num_arcs = src_num_arcs;
         }
         num_oarcs_data[idx0123] = num_arcs;
@@ -1388,7 +1393,7 @@ FsaOrVec ReplaceFsa(FsaVec &src, FsaOrVec &index, int32_t symbol_range_begin,
   RaggedShape full_shape =
       ComposeRaggedShapes(to_ostates_shape, ostate_to_oarcs);
 
-  // for the lower-order row-splits and row-ids, use tot_row_{splits,idx}n_data
+  // for the lower-order row-splits and row-ids, use tot_row_{splits,ids}n_data
   const int32_t *full_row_splits4_data = full_shape.RowSplits(4).Data(),
                 *full_row_ids4_data = full_shape.RowIds(4).Data();
   int32_t tot_oarcs = full_shape.NumElements();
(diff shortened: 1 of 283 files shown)


pkufool:
the index fsa
image
the src fsas, index 0, 1, 2
image
image
image
the replaced fsa with `symbol_range_begin=0`
image

k2/csrc/fsa_algo.cu
```diff
@@ -1246,6 +1246,271 @@ void InvertHost(FsaOrVec &src, Ragged<int32_t> &src_aux_labels, FsaOrVec *dest,
...
+        if (arc_map_index_data)
+          arc_map_index_data[idx01234] = arc_index_map_idx;
+        oarcs_data[idx01234] = oarc;
+      });
+  // remove current axes 1 and 2... [after removing axis 1, old axis 2 becomes
+  // axis 1, so remove axis 1 twice].
+  RaggedShape temp = RemoveAxis(full_shape, 1);
+  return FsaVec(RemoveAxis(temp, 1), oarcs);
+}
+
```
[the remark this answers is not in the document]
csukuangfj: You can take the existing function "ExpandArcs" as a template for implementation.

pkufool:
@danpovey  I only handle the attributes (aux_label) of index fsa and ignore the attributes in src.
 The other thing, I am not quite sure whether there is a problem with the autograd, since `fsa_from_unary_function_tensor` only deal with the scores of index. Any suggestions?

danpovey:
You should be using 
 _propagate_aux_labels_binary_function
I think, rather than fsa_from_unary_function_tensor.  Actually this is poorly named, we should probably call it fsa_from_bimary_function, and instead of giving the Fsa as an arg, give it the ragged tensor and have it construct the Fsa from that.  You'll see it's only called  twice, currently, so easy to do.

See the testing code for things like intersection, to make sure the derivatives are propagated back to the input FSAs in the correct way.

k2/csrc/fsa_algo.h
```diff
@@ -659,6 +660,41 @@ void InvertHost(FsaOrVec &src, Ragged<int32_t> &src_aux_labels, FsaOrVec *dest,
...
                                 Array1<int32_t> *arc_map = nullptr);
 
+
+/*
+  Replace, in `index`, `labels symbol_range_begin <= label < symbol_range_begin+src.Dim0()`
+  with the Fsa indexed `label - symbol_range_begin` in `src`, identifying the
+  source and destination states of the arc in `src` with the initial and
+  final states in `src[label - symbol_range_begin]`. Arcs with labels outside
+  this range are just copied. Labels on final-arcs in `src` (which will be -1)
+  would be set to 0(epsion) in the result.
```
danpovey, line 670: 0 (epsilon)

Also please edit it so the whitespace lines up (is all on the same cursor position)
I originally wrote this in one of these github windows, and you can't line it up because
the GitHub text window uses non-fixed-width characters.

k2/python/k2/utils.py
```diff
@@ -514,3 +514,90 @@ def fsa_from_unary_function_ragged(src: Fsa, dest_arcs: _k2.RaggedArc,
     k2.autograd_utils.phantom_index_and_sum_scores(dest, src.scores, arc_map)
 
     return dest
+
+
+def fsa_from_bimary_function_tensor(
```
danpovey, line 519: bimary -> binary

k2/csrc/fsa_algo_test.cu
```diff
@@ -1053,4 +1054,67 @@ TEST(FsaAlgo, TestRemoveEplsionSelfRandomFsas) {
...
+                        "  [ [ 0 1 21 21.0 0 2 22 22.0 ] "
+                        "    [ 1 3 -1 0.0 ] [2 3 -1 0.0] [ ] ] "
+                        "  [ [ 0 1 31 31.0 0 2 33 33.0 ] "
+                        "    [ 1 2 32 32.0 ] [ 2 3 -1 0.0 ] [ ] ] ]").To(c);
+    FsaVec index = FsaVec(" [ [ [ 0 1 1 1.0 0 2 2 2.0 ] [ 1 3 3 3.0 ] "
+                          "     [ 2 4 4 4.0 2 5 -1 0.0 ] [ 3 5 -1 0.0 ] "
+                          "     [ 4 5 -1 0 ] [ ] ] ]").To(c);
+    Array1<int32_t> arc_src_map, arc_index_map;
+    FsaVec replaced_fsa =
+        ReplaceFsa(src, index, 0, &arc_src_map, &arc_index_map);
```
danpovey, line 1101: Are you ever testing this where the 3rd arg is nonzero?

k2/csrc/fsa_algo.h
```diff
@@ -659,6 +660,41 @@ void InvertHost(FsaOrVec &src, Ragged<int32_t> &src_aux_labels, FsaOrVec *dest,
...
+                       structure of the result (the result will have the same
+                       number of axes as`index`.
+
+    @param [in] symbol_range_begin  Beginning of the range (interval) of symbols
+                                    that are to  be replaced with Fsas.
+                                    Symbols numbered `symbol_range_begin <= i < src.Dim0() + symbol_range_begin`
+                                    will be replaced with the Fsa in `src[i - symbol_range_begin]`
+
+    @param [out,optional] arc_map_src  If not nullptr, will be set to a new
+                                       array that maps from arc-indexes in the
```
danpovey, line 684: Actually (my fault) this comment is not quite accurate, or at least not specific.
Because this arc_map will be used in autograd, we need to clarify that it is only set for the 
initial arc corresponding each label in `src` that is replaced.  I assume this is an initial epsilon
in the output that leads into the start-state of the element of `index`?  (Otherwise I'm wondering
how we deal with the situation where there are arcs into the start-state of the FSA in `index`).

k2/csrc/fsa_algo_test.cu
danpovey: Also: in order to properly test the semantics of this, especially with randomized FSAs, it would be more ideal to check its equivalence with composition.  This operation is intended as a cheaper replacement for composition; you should be able to figure out the details of this.  However, the equivalence may only hold after removing epsilons in the output.  (This may be easier to test at the python level).

danpovey (commented):
A few more comments..
Adding testing code that's randomized may be a hassle, but it will be a good opportunity to test your knowledge of the framework and to learn more.

k2/csrc/fsa_algo_test.cu
pkufool: > Are you ever testing this where the 3rd arg is nonzero?

I test it in python level, will add the test case here.


> Also: in order to properly test the semantics of this, especially with randomized FSAs, it would be more ideal to check its equivalence with composition. This operation is intended as a cheaper replacement for composition; you should be able to figure out the details of this. However, the equivalence may only hold after removing epsilons in the output. (This may be easier to test at the python level).

Will figure out the relationship of replace and composition and try to add test case.

danpovey:
Also please check carefully that the comments match the implementation.
For instance I think the comments say we identify the start-state of the FSA in `src` with the
source-state of the arc, which I think is not true now because of the epsilon.    Perhaps could say something like:

The destination state of the arc in `index` is identified with the final-state of the corresponding FSA in `src`, and the arc in `index` will become an epsilon arc leading to a new state in the output that is a copy of the start-state of the corresponding FSA in `src`.

k2/csrc/fsa_algo.h
pkufool: In my current implementation, the following `index` and `src`
|index|src|
|--|--|
|image|image|

replace with `symbol_range_begin=2` is:

image

and `arc_map_src = [-1, -1, 0, 1, 2, 3, -1]`, mapping arcs to src FsaVec, all of the arcs inherit from `index` would be -1.
`arc_map_index = [0, 1, -1, -1, -1, -1, 2]`, mapping arcs to index Fsa, arcs be replaced(epsilon arcs that leading to the start-state of corresponding fsa in `src`) and arcs with label out-of-range will have values, the others will be -1.

As for the situation where there are arcs into the state-state of the FSA in `index`,  current implementation can not handle it appropriate, will fix it.

k2/csrc/fsa_algo_test.cu
pkufool: @danpovey  Sorry, I still haven't figured out how replace act as a replacement of composition, would you please show me more details, or, some simple test cases, thanks!

danpovey: Consider what happens if you compose the "index" FSA with an FSA, say f, that has the following structure:
the start-state of f has an arc directly to final-state with -1 label, and:
 For labels `symbol_range_begin <= s < src.Dim0()`:
    There is an arc with ilabel == s and olabel == 0, from state 0 of f to a state in f that corresponds 
  to the start-state of the
  FSA f_s = src[s-symbol_range_begin].  The FSA f_s has its labels as the olabels of f, with epsilons as the
  ilabels; and its final-state is identified with the start-state of f.  Other states in f_s are just copied to new 
  states in f.  The -1 labels in f_s would be converted to 0 olabels in f.
To get epsilons treated correctly in composition assuming treat_epsilons_specially=false, we'd have to
add epsilon self-loops to `src` before doing the composition.  (Note: things wouldn't work quite right
if symbol_range_begin <= 0). 
Anyway, the result of composing src with f should be equivalent to the result of the `replace` operation.

pkufool: A simple example to show whether I understood right.

For the replace operation, the following `index` and `src`

|index|src{0,1}|
|--|--|
|image|image|
||image|

with `symbol_range_begin=1` the output is
image

For the compose operation, I construct fsa `f` as the explanation above:

|inverted index|f|
|--|--|
|image|image|

then I did `k2.invert(k2.compose(k2.invert(index), f))`, the output is

image

the only difference is the olabel of the arc points to the final state.

So, if I want to check the equivalence between replace and composition, I should construct a fsa `f` from src, then compare the result of `k2.replace(index, src)` and `k2.invert(k2.compose(k2.invert(index), f))`

danpovey: This looks plausible although I haven't checked carefully.
What I find strange is that your index and src have olabels.  olabels should not be needed for doing the indexing or the comparison with the composed result.
Ultimately you should be able to remove the olabels and compare it as a simple FSA.

danpovey:
I guess this  is ready to merge?

pkufool:
> I guess this is ready to merge?

No, still have some problems when testing the equivalence of compose and replace in random fsa. It seems that when there are multiple arcs leaving a state with the same label (not deterministic) and multiple arcs entering a state with the same lable, the composition result will merge these arcs, so structures of the replace result and composition result may be a little different.  But I think these little difference of structures is not matter, their `tot_score` are the same.

Also, I found that when I intersect an FsaVec with Fsa, the result of intersecting an FsaVec and Fsa is different from the result of intersecting the Fsa one by one, say `result1 = k2.intersect(fsa_vec, fsa)`, `result2 = k2.intersect(fsa_vec[i], fsa)` then `result1[i]` is different from `result2`, structures are the same, scores are different.  It would be nice if you could explain this, or I could look into the source code and find by my self. Thanks!

danpovey:
_propagate_aux_labels_binary_function() [which is badly named I think, we should call it _propagate_attributes_binary_function()] overwrites the scores, this is part of what enables the
autograd mechanism to work.  fsa_algo.py:157,
```
            # The following will actually overwrite `scores` with the same                                                                                                                              
            # value it had before; but this enables the autograd to work since                                                                                                                          
            # we do it using torch mechanisms.                                                                                                                                                          
            value = index_select(a_value, a_arc_map, default_value=filler) \
                    + index_select(b_value, b_arc_map, default_value=filler)
            setattr(out_fsa, name, value)
```
... so the error  could be in the arc maps.  I think the following may be the problem: in csrc/fsa:algo.cu:206,
```
    ok = ok && ans;
    int32_t this_num_arcs = creator.GetArcOffsetFor(i + 1) - arc_offset;
    if (arc_map_a) {
      int32_t arc_offset_a = a_fsas_row_splits12_data[i * stride_a];
      for (int32_t i = 0; i < this_num_arcs; i++)
        this_arc_map_a[i] += arc_offset_a;
    }
    if (arc_map_b) {
      int32_t arc_offset_b = b_fsas_row_splits12_data[i * stride_b];
      for (int32_t i = 0; i < this_num_arcs; i++)
        this_arc_map_b[i] += arc_offset_b;
    }
```
.. the code is not checking that this_arc_map_a[i] or this_arc_map_b[i]  is not -1 prior to increasing the value.

danpovey:
.. also, it looks like the default is treat_epsilons_specially=True, which only works on CPU.  It's better if possible to not use that setting, so the code can work on GPU also.  (To handle epsilons, you can sometimes add epsilon self-loops to the FSA on one side or the other).

danpovey:
> the composition result will merge these arcs,

I don't believe this is the way `compose` is supposed to work... I think they are supposed to remain separate arcs.

pkufool:
> I don't believe this is the way `compose` is supposed to work... I think they are supposed to remain separate arcs.

I think this only occurs when all of the src-state, dest-state and label are same, `k2.random_fsa()` may generate that kind of fsa, see the example below, there are two arcs connect state 0, 1 with label 10.
|index|replace result|intersect result|
|--|--|--|
|image|image|image|

danpovey:
OK, and does this happen with CPU or GPU intersection?

pkufool:
> OK, and does this happen with CPU or GPU intersection?

Both

danpovey:
> > I don't believe this is the way `compose` is supposed to work... I think they are supposed to remain separate arcs.
> 
> I think this only occurs when all of the src-state, dest-state and label are same, `k2.random_fsa()` may generate that kind of fsa, see the example below, there are two arcs connect state 0, 1 with label 10.
> 
> index	replace result	intersect result
> image	image	image

What you show here is expected.  It looks to me like the two results would be equivalent in both tropical and log semirings.

pkufool:
I think this is ready to merge.

danpovey:
Great, thanks!

PR #760, opened 2021-06-10
Hash speed fix

danpovey:


danpovey:
This is an improvement in the hashing algorithm, so that the increment in buckets that it takes when there is a hash collision is more random.
Previously, we had a problem when doing LM rescoring, that the hash key was of the form:
`state_in_lm_fst * n  +  state_in_lattices`
and that meant that one big contiguous range of keys, of the form:
`id_of_unigram_backoff_state * n  +  state_in_lattices`, for `0 <= state_in_lattices < num_states_in_lattices`,
was always occupied, because all lattice states match with the unigram backoff state.  The problem is that the 
increment in bucket-index that it took when there was a collision, was always quite small (8 in this particular case
I tested).  So it would have to increment thousands of times to escape from that fully-occupied range, if it happened
that something else hashed to that same range causing a collision.
I fixed this by making the amount by which in increments much more random, so that things can more quickly 'escape from' that range.

I am also changing the code so that in release mode, the user can still do
`export K2_SYNC_KERNELS=1`
and have it synchronize kernels.  There is very little extra overhead from enabling this (one if-statement per kernel launch) and it's convenient.

galv:
Regarding K2_SYNC_KERNELS, I expect that you can use CUDA_LAUNCH_BLOCKING=1 for effectively the same thing. Main benefit would be that it works for existing libraries like cusparse and cublas (which you could call with K2_EVAL, to clarify, but you would have to include a redundant synchronize call to get the correct error from cusparse or cublas), which I suspect K2 may want to integrate at some point for block-diagonal matrices (among other things).

https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#concurrent-execution-host-device

danpovey:
Ah OK, that's useful.

PR #762, opened 2021-06-11
Add `invalidate_cache` to `Fsa`.

csukuangfj:


danpovey:
BTW, when we assign to the scores, we don't need to invalidate the leaving_arc_batches and entering_arc_batches.

csukuangfj:
> BTW, when we assign to the scores, we don't need to invalidate the leaving_arc_batches and entering_arc_batches.

Thanks! Fixed in #763 .

PR #763, opened 2021-06-12
Invalidate only cached entries related to scores.

csukuangfj:


k2/python/tests/fsa_test.py
```diff
@@ -1115,6 +1115,24 @@ def test_convert_attr_to_ragged(self):
             expected = k2.RaggedInt('[ [1] [] [-1] ]')
             assert str(fsa.tensor_attr2) == str(expected)
 
+    def test_invalidate_cache(self):
+        s = '''
+            0 1 1 0.1
+            1 2 -1 0.2
+            2
+        '''
+        fsa = k2.Fsa.from_str(s)
```
danpovey, line 1124: I think we need to invalidate `incoming_arcs` as well.

csukuangfj: `incoming_arcs` depends only on the topology of the FSA, I think. Do we really need to also invalidate it?

https://github.com/k2-fsa/k2/blob/327b084a2b24739a532e6a876f0d350ad1843bc2/k2/csrc/fsa_utils.cu#L878-L905

danpovey: Ah, you  seem to be right.

PR #770, opened 2021-06-20
Support PyTorch 1.9.0

csukuangfj:


k2/csrc/array_ops_test.cu
```diff
@@ -266,10 +266,18 @@ void TestExclusiveSumArray1(int32_t num_elem) {
   }
 }
 
+#ifndef _MSC_VER
+// It results in the following error SOMETIMES on windows
```
csukuangfj, line 270: @danpovey 

This test is disabled for windows since it fails sporadically and is not reproducible locally.

danpovey (commented):
Cool !  LGTM!

PR #773, opened 2021-06-30
Fix a bug in Fsa._invalidate_cache.

csukuangfj:
We also need to remove "entering_arcs" since
it may be set in get_forward_scores(). See
https://github.com/k2-fsa/k2/blob/926f160f3d4dc986b5f556852f65cbaebb78a717/k2/python/k2/fsa.py#L603-L618

---

I am encountering this bug because the WER of different lm_scale values are the same when rescoring with the whole lattice.
This pull request fixes it.

### DIFF
--- a/k2/python/k2/fsa.py
+++ b/k2/python/k2/fsa.py
@@ -249,17 +249,17 @@ class Fsa(object):
             pattern = re.compile('score')
             to_remove = []
 
-            for key in self._cache:
+            for key in self.__dict__['_cache']:
                 if pattern.search(key):
                     to_remove.append(key)
 
-            if 'entering_arcs' in self._cache:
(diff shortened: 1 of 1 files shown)


k2/python/k2/fsa.py
```diff
@@ -249,12 +249,17 @@ def _invalidate_cache_(self, scores_only: bool = True) -> None:
             pattern = re.compile('score')
             to_remove = []
 
-            for key in self.__dict__['_cache']:
```
danpovey, line 252: I think self.__dict__['cache'] was faster, because we avoid going through our __getitem__() code.

csukuangfj: ok, reverted.

PR #775, opened 2021-07-06
Fix score calculation error when intersecting FsaVec and Fsa on cpu

pkufool:
See https://github.com/k2-fsa/k2/pull/759#issuecomment-874357866

### DIFF
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -203,17 +203,13 @@ bool Intersect(FsaOrVec &a_fsas, int32_t properties_a, FsaOrVec &b_fsas,
     int32_t this_num_arcs = creator.GetArcOffsetFor(i + 1) - arc_offset;
     if (arc_map_a) {
       int32_t arc_offset_a = a_fsas_row_splits12_data[i * stride_a];
-      for (int32_t i = 0; i < this_num_arcs; i++) {
-        this_arc_map_a[i] = this_arc_map_a[i] == -1 ? this_arc_map_a[i] :
-                            this_arc_map_a[i] + arc_offset_a;
-      }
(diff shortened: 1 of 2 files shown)


k2/csrc/fsa_algo.cu
```diff
@@ -203,13 +203,17 @@ bool Intersect(FsaOrVec &a_fsas, int32_t properties_a, FsaOrVec &b_fsas,
     int32_t this_num_arcs = creator.GetArcOffsetFor(i + 1) - arc_offset;
     if (arc_map_a) {
       int32_t arc_offset_a = a_fsas_row_splits12_data[i * stride_a];
-      for (int32_t i = 0; i < this_num_arcs; i++)
-        this_arc_map_a[i] += arc_offset_a;
+      for (int32_t i = 0; i < this_num_arcs; i++) {
```
csukuangfj, line 206: Could you also add a test to cover it?

k2/csrc/fsa_algo.cu
```diff
@@ -203,13 +203,17 @@ bool Intersect(FsaOrVec &a_fsas, int32_t properties_a, FsaOrVec &b_fsas,
...
+        this_arc_map_a[i] = this_arc_map_a[i] == -1 ? this_arc_map_a[i] :
```
csukuangfj, line 207: I would suggest
Screen Shot 2021-07-06 at 3 50 38 PM
which is clearer, I think. You can keep it if you like it.

k2/csrc/fsa_algo.cu
pkufool: Ok

pkufool:
@csukuangfj  can this PR merge now?

csukuangfj:
Thanks!

PR #776, opened 2021-07-07
Build ctc graph from symbols in batch mode

pkufool:
See https://github.com/k2-fsa/snowfall/issues/220

The python api and results are as follows:

image

### DIFF
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -468,8 +468,8 @@ FsaVec CtcGraphs(const Ragged<int32_t> &symbols,
   K2_EVAL(
       c, num_states, lambda_set_num_arcs, (int32_t state_idx01)->void {
         int32_t fsa_idx0 = fts_row_ids1_data[state_idx01],
-                // we minus fsa_idx0 here, because we adding one more state, the
-                // final state for each fsa
+                // we minus fsa_idx0 here, because we are adding one more state,
+                // the final state for each fsa
                 sym_state_idx01 = state_idx01 / 2 - fsa_idx0,
                 remainder = state_idx01 % 2,
                 current_num_arcs = 2;  // normally there are two arcs, self-loop
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -483,7 +483,12 @@ FsaVec CtcGraphs(const Ragged<int32_t> &symbols,
             current_num_arcs = 0;
           } else {
             int32_t current_symbol = symbol_data[sym_state_idx01],
-                    next_symbol = symbol_data[sym_state_idx01 + 1];
+                    // we set the next symbol of the last symbol to -1, so
+                    // the following if clause will always be true
+                    next_symbol = (sym_state_idx01 + 1) == sym_final_state ?
+                                  -1 : symbol_data[sym_state_idx01 + 1];
+            // symbols must be not equal to -1, which is specially used in k2
+            K2_CHECK_NE(current_symbol, -1);
             // if current_symbol equals next_symbol, we need a blank state
             // between them, so there are two arcs for this state
             // otherwise, this state will point to blank state and next symbol
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -441,12 +441,16 @@ FsaVec CtcGraphs(const Ragged<int32_t> &symbols,
   Array1<int32_t> num_states_for(c, num_fsas + 1);
   int32_t *num_states_for_data = num_states_for.Data();
   const int32_t *symbol_row_split1_data = symbols.RowSplits(1).Data();
+  // symbols indexed with [fsa][symbol]
+  // for each fsa we need `symbol_num * 2 + 1 + 1` states, `symbol_num * 2 + 1`
+  // means that we need a blank state on each side of a symbol state, `+ 1` is
+  // for final state in k2
   K2_EVAL(
       c, num_fsas, lambda_set_num_states, (int32_t fsa_idx0)->void {
-        int32_t state_idx0x = symbol_row_split1_data[fsa_idx0],
-                state_idx0x_next = symbol_row_split1_data[fsa_idx0 + 1],
-                state_num = state_idx0x_next - state_idx0x;
-        num_states_for_data[fsa_idx0] = state_num * 2 + 2;
+        int32_t symbol_idx0x = symbol_row_split1_data[fsa_idx0],
+                symbol_idx0x_next = symbol_row_split1_data[fsa_idx0 + 1],
+                symbol_num = symbol_idx0x_next - symbol_idx0x;
+        num_states_for_data[fsa_idx0] = symbol_num * 2 + 2;
       });
 
   ExclusiveSum(num_states_for, &num_states_for);
@@ -460,20 +464,30 @@ FsaVec CtcGraphs(const Ragged<int32_t> &symbols,
   const int32_t *fts_row_splits1_data = fsa_to_states.RowSplits(1).Data(),
                 *fts_row_ids1_data = fsa_to_states.RowIds(1).Data(),
                 *symbol_data = symbols.values.Data();
+  // set the arcs number for each state
   K2_EVAL(
       c, num_states, lambda_set_num_arcs, (int32_t state_idx01)->void {
         int32_t fsa_idx0 = fts_row_ids1_data[state_idx01],
+                // we minus fsa_idx0 here, because we adding one more state, the
+                // final state for each fsa
                 sym_state_idx01 = state_idx01 / 2 - fsa_idx0,
                 remainder = state_idx01 % 2,
-                current_num_arcs = 2;
-        if (remainder) {
+                current_num_arcs = 2;  // normally there are two arcs, self-loop
+                                       // and arc points to the next state
+                                       // blank state always has two arcs
+        if (remainder) {  // symbol state
           int32_t sym_final_state =
                     symbol_row_split1_data[fsa_idx0 + 1];
+          // There is no arcs for final states
           if (sym_state_idx01 == sym_final_state) {
             current_num_arcs = 0;
           } else {
             int32_t current_symbol = symbol_data[sym_state_idx01],
                     next_symbol = symbol_data[sym_state_idx01 + 1];
+            // if current_symbol equals next_symbol, we need a blank state
+            // between them, so there are two arcs for this state
+            // otherwise, this state will point to blank state and next symbol
+            // state, so we need three arcs here.
             if (current_symbol != next_symbol)
               current_num_arcs = 3;
           }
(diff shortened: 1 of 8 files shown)


pzelasko:
Maybe a naive question -- what is the difference between this implementation, and doing it via Python-level k2 operations? I remember there was some code from @csukuangfj for MMI that batched all the FSA ops for a given list of utterances. Just curious what is the gain.

k2/python/tests/ctc_graph_test.py
```diff
@@ -0,0 +1,62 @@
...
+        cls.devices = [torch.device('cpu')]
+        if torch.cuda.is_available() and k2.with_cuda:
+            cls.devices.append(torch.device('cuda', 0))
+            if torch.cuda.device_count() > 1:
+                torch.cuda.set_device(1)
+                cls.devices.append(torch.device('cuda', 1))
+
+    def test(self):
+        for device in self.devices:
+            fsa_vec = k2.ctc_graph([[1,2,2],[1,2,3]])
```
csukuangfj, line 42: Could you also test the case when the input is a ragged tensor?

csukuangfj:
> Maybe a naive question -- what is the difference between this implementation, and doing it via Python-level k2 operations? I remember there was some code from @csukuangfj for MMI that batched all the FSA ops for a given list of utterances. Just curious what is the gain.

There are several differences:

(1) There are no optional silences before and after each word in this pull-request.
(2) This pull-request constructs the resulting graph directly, instead of creating it from `k2.intersect`. It is presumably faster
as it involves fewer kernel calls, though it needs benchmarks.
(3) For CTC training, we don't need to construct the graph `ctc_topo` anymore with this pull-request.
(4) The labels in this pull-request are input tokens, and its aux_labels are also tokens/epsilons. While in snowfall, the labels are phones and aux_labels are words/epsilons.
(5) In snowfall, it supports words with multiple pronunciations since there is a lexicon, while this pull-request
supports only one.
(6) In snowfall, it supports both MMI and CTC. This pull-request supports only CTC training since there is no bigram P.

k2/csrc/fsa_algo.h
```diff
@@ -478,6 +475,21 @@ Fsa LinearFsa(const Array1<int32_t> &symbols);
  */
 FsaVec LinearFsas(const Ragged<int32_t> &symbols);
 
+/*
+  Create an FsaVec containing ctc graph FSAs, given a list of sequences of
+  symbols
+
+    @param [in] symbols  Input symbol sequences (must not contain
```
csukuangfj, line 482: Do we need to add a check inside the kernel that none of the input symbols is -1?

k2/python/tests/ctc_graph_test.py
```diff
@@ -0,0 +1,62 @@
...
+    def setUpClass(cls):
+        cls.devices = [torch.device('cpu')]
+        if torch.cuda.is_available() and k2.with_cuda:
+            cls.devices.append(torch.device('cuda', 0))
+            if torch.cuda.device_count() > 1:
+                torch.cuda.set_device(1)
+                cls.devices.append(torch.device('cuda', 1))
+
+    def test(self):
+        for device in self.devices:
```
csukuangfj, line 41: Seems that `device` is not used.

k2/csrc/fsa_algo.h
pkufool: I add a checking in the kernel `set_num_arcs`, which will enumerate all the symbols. I think it's enough.

k2/python/csrc/torch/fsa_algo.cu
```diff
@@ -661,6 +661,48 @@ static void PybindFixFinalLabels(py::module &m) {
...
+         int32_t /*unused_gpu_id*/)
+        -> std::pair<FsaVec, torch::optional<torch::Tensor>> {
+        DeviceGuard guard(symbols.Context());
+        Array1<int32_t> arc_map;
+        FsaVec graph = CtcGraphs(symbols, need_arc_map ? &arc_map : nullptr);
+        torch::optional<torch::Tensor> tensor;
+        if (need_arc_map) tensor = ToTorch(arc_map);
+        return std::make_pair(graph, tensor);
+      },
+      py::arg("labels"), py::arg("need_arc_map") = true, py::arg("gpu_id"));
```
csukuangfj, line 703:
```suggestion
      py::arg("symbols"), py::arg("need_arc_map") = true, py::arg("gpu_id"));
```

k2/python/csrc/torch/fsa_algo.cu
```diff
@@ -661,6 +661,48 @@ static void PybindFixFinalLabels(py::module &m) {
...
+      py::arg("gpu_id") = -1,
+      R"(
+  If gpu_id is -1, the returned FsaVec is on CPU.
+  If gpu_id >= 0, the returned FsaVec is on the specified GPU.
+      )");
+
+  m.def(
+      "ctc_graph",
+      [](const Ragged<int32_t> &symbols, bool need_arc_map = true,
+         int32_t /*unused_gpu_id*/)
```
csukuangfj, line 694: The last argument has no default value but it is after an argument with default value.
It's not valid C++, I think.

danpovey:
I think there should be a boolean option to specify the type of CTC topology: "standard" or "simplified", where the "standard" one makes the blank mandatory between a pair of identical symbols.

k2/csrc/fsa_algo.cu
```diff
@@ -430,6 +430,154 @@ FsaVec LinearFsas(const Ragged<int32_t> &symbols) {
...
+        if (remainder) {  // symbol state
+          int32_t sym_final_state =
+                    symbol_row_split1_data[fsa_idx0 + 1];
+          // There is no arcs for final states
+          if (sym_state_idx01 == sym_final_state) {
+            current_num_arcs = 0;
+          } else {
+            int32_t current_symbol = symbol_data[sym_state_idx01],
+                    // we set the next symbol of the last symbol to -1, so
+                    // the following if clause will always be true
```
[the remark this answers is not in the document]
pkufool: It means that the last symbol state would always have 3 arcs. we handle `next_symbol` here, first, to avoid segment fault error, second, to confirm that `current_symbol != next_symbol` so we will assign the last symbol state 3 arcs.

will explain more in the docs.

k2/csrc/fsa_algo.cu
```diff
@@ -430,6 +430,154 @@ FsaVec LinearFsas(const Ragged<int32_t> &symbols) {
...
+  K2_EVAL(
+      c, num_fsas, lambda_set_num_states, (int32_t fsa_idx0)->void {
+        int32_t symbol_idx0x = symbol_row_split1_data[fsa_idx0],
+                symbol_idx0x_next = symbol_row_split1_data[fsa_idx0 + 1],
+                symbol_num = symbol_idx0x_next - symbol_idx0x;
+        num_states_for_data[fsa_idx0] = symbol_num * 2 + 2;
+      });
+
+  ExclusiveSum(num_states_for, &num_states_for);
+  Array1<int32_t> &fsa_to_states_row_splits = num_states_for;
```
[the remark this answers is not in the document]
pkufool: After doing ExclusiveSum, `num_states_for` is actually the `row_splits`, using different names here just for easy understanding,  we'll use `fsa_to_states_row_splits` to construct ragged_shape below.

k2/python/csrc/torch/fsa_algo.cu
pkufool: I saw the error in actions, it could compile successfully on linux, will fix it.

pkufool:
> I think there should be a boolean option to specify the type of CTC topology: "standard" or "simplified", where the "standard" one makes the blank mandatory between a pair of identical symbols.

Ok, will add the option.

pkufool:
Add the option `standard`, default True,  the standard one makes the blank mandatory between a pair of identical symbols.

An example to demonstrate their difference is as follow:

image

danpovey:
Great, thanks!
Feel free to merge when you guys think it's OK.

k2/csrc/fsa_algo.cu
```diff
@@ -430,6 +430,159 @@ FsaVec LinearFsas(const Ragged<int32_t> &symbols) {
...
+                current_num_arcs = 2;  // normally there are two arcs, self-loop
+                                       // and arc pointing to the next state
+                                       // blank state always has two arcs
+        if (remainder) {  // symbol state
+          int32_t sym_final_state =
+                    symbol_row_split1_data[fsa_idx0 + 1];
+          // There is no arcs for final states
+          if (sym_state_idx01 == sym_final_state) {
+            current_num_arcs = 0;
+          } else {
```
csukuangfj, line 484:
```suggestion
          } else if(!standard) {
            current_num_arcs = 3;
          } else {
            // same as before the latest change
          }
```
For non-standard topo, `current_num_arcs` is always 3. Put it into
a separate if statement can save some work.

k2/python/k2/fsa_algo.py
```diff
@@ -962,3 +962,60 @@ def expand_ragged_attributes(
...
+        identical symbols. Default True.
+      device:
+        Optional. It can be either a string (e.g., 'cpu',
+        'cuda:0') or a torch.device.
+        If it is None, then the returned FSA is on CPU. It has to be None
+        if `symbols` is an instance of :class:`k2.RaggedInt`, the returned
+        FSA will on the same device as `k2.RaggedInt`.
+
+    Returns:
+        An FsaVec contains the returned ctc graphs, with `Dim0()` the same as
```
csukuangfj, line 993:
```suggestion
        An FsaVec containing the returned ctc graphs, with `Dim0()` the same as
```

k2/csrc/fsa_algo.cu
```diff
@@ -430,6 +430,159 @@ FsaVec LinearFsas(const Ragged<int32_t> &symbols) {
...
+              -1 : symbol_data[sym_state_idx01 + 1];
+          // for standard topology, the symbol state can not point to next
+          // symbol state if the next symbol is identical to current symbol.
+          if (current_symbol == next_symbol && standard) {
+            K2_CHECK_LT(arc_idx2, 2);
+            arc.label = arc_idx2 == 0 ? 0 : current_symbol;
+            arc.dest_state = arc_idx2 == 0 ? state_idx1 + 1 : state_idx1;
+          } else {
+            switch (arc_idx2) {
+              case 0:   // the arc points to blank state
```
csukuangfj, line 556:
```suggestion
              case 0:   // the arc pointing to blank state
```

k2/csrc/fsa_algo.cu
```diff
@@ -430,6 +430,159 @@ FsaVec LinearFsas(const Ragged<int32_t> &symbols) {
...
+                arc.label = 0;
+                arc.dest_state = state_idx1 + 1;
+                break;
+              case 1:   // the self loop arc
+                arc.label = current_symbol;
+                arc.dest_state = state_idx1;
+                break;
+              case 2:  // the arc points to next symbol state
```
csukuangfj, line 564:
```suggestion
              case 2:  // the arc pointing to the next symbol state
```

PR #779, opened 2021-07-13
[WIP] Suffix arrays, needed for nbest processing

danpovey:


mahsa7823:
The file k2/csrc/nbest_test.cu seems to be missing in this branch.

pkufool:
@danpovey  
I finish my first draft, of course, there will be more test case.

I have a question here about the sentence boundary when finding the best matching positions.

For example, we have  inputs as follows:
tokens: `[ [ L, O, V, E, #], [L, I, V, E, #], [L, I, F, E, #], [M, O, V, E, #]]`
counts: `[ [1, 1, 1, 1, 1], [1, 1, 1, 1, 1], [0, 0, 0, 0, 0], [0, 0, 0, 0, 0]]`
scores: `[ [1, 2, 3, 4, 5], [6, 7, 8, 9, 10], [0, 0, 0, 0, 0], [0, 0, 0, 0, 0]]`

In current implementation, the outputs are (with max_order = 2):
mean: `[[3.5, 2, 3, 4, 5], [6, 7, 5.5, 6.5, 7.5], [6, 7, 5.5, 6.5, 7.5,],[5.5, 2, 3, 4, 5 ]]`
var: `[[6.25, 0, 0, 0, 0],[0, 0, 6.25, 6.25, 6.25,],[0, 0, 8.25, 6.25, 6.25],[8.25, 0, 0, 0, 0]]`
counts_out: `[ [2, 1, 1, 1, 1], [1, 1, 2, 2, 2], [1, 1, 0, 2, 2], [0, 1, 1, 1, 1]]`
ngram_order: `[[ 2, 1, 2, 2, 2], [2, 2, 1, 2, 2], [2, 2, 0, 1, 2], [0, 1, 2, 2, 2]]`

Let's see the `L` in `LIFE#`, its best matching position is 5 (`L` in `LIVE#`) with lcp `VE#L`.  so the outputs of this `L` are 6(mean), 0(var), 1(counts_out), 2(ngram_order, min(4, max_order)). 

But I think the best matching positions  should be `L` in `LOVE#` and `L` in `LIVE#` with lcp `L`,  and the outputs are
3.5(mean, (1 + 6) / 2),  6.25(var), 2(counts_out), 1(ngram_order).

Which is the expected one?

danpovey:
Sorry, I responded to this earlier but somehow my response was lost, perhaps when I closed a tab.
Brief summary for posterity: yes, it would be more ideal to never match further than the BOS boundary,
but I didn't ask you to implement it with that refinement because for our present purposes it doesn't matter:
once we get to BOS, the scores should generally match exactly, so it wouldn't matter much for our purposes.
Incidentally, @csukuangfj : we can include the count, as well as the mean, var and longest-matched n-gram order, in
the stats that we provide to the network that preducts the rescored scores; so there are in total 5 inputs; the (count, mean, var, ngram_order; and the original score from before rescoring).

PR #781, opened 2021-07-14
Support replicate in k2.ragged.pad

csukuangfj:
Also,
(1) Fix a bug in k2.index_and_sum. Originally, it did not support -1 in indexes. This pull-request fixes that.
(2) Rename k2.ragged.pad_ragged to k2.ragged.pad

---

The mode `replicate` is added to support implementing top-k extraction from `Nbest`.
When an utterance has less than k paths, we repeat its last path so that each utterance has
exactly `k` paths.


pkufool (approved):
+1

PR #782, opened 2021-07-14
Ignore strides CHECK in FromTorch when input tensor is empty

pkufool:
See https://github.com/k2-fsa/snowfall/pull/227#issuecomment-877945713

### DIFF
--- a/k2/csrc/ragged_ops.cu
+++ b/k2/csrc/ragged_ops.cu
@@ -1000,7 +1000,7 @@ RaggedShape Transpose(RaggedShape &src, Array1<int32_t> *value_indexes) {
   ContextPtr c = src.Context();
   int32_t src_dim0 = src.Dim0(), src_tot_size1 = src.TotSize(1);
   if (src_dim0 <= 0) {
-    *value_indexes = Array1<int32_t>(c, 0);
+    if (value_indexes) *value_indexes = Array1<int32_t>(c, 0);
     return src;
   }
   int32_t src_dim1 = src_tot_size1 / src_dim0;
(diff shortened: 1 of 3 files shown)


csukuangfj:
@glynpu 
Could you try this pull request and see if it solves the issue in
https://github.com/k2-fsa/snowfall/pull/227#issuecomment-873777500

pkufool:
I think this PR do solve the issue (https://github.com/k2-fsa/snowfall/pull/227#issuecomment-873777500), apart from the stride checking problem, there is also a bug in `Transpose` that return "null" `index_values` when the input is empty FsaVec (with no fsa at all). 
Actually, there are two kinds of Array1 with Dim() equals to zero, one is constructed from default constructor (`region_ == nullptr`), the other is constructed from `Array1(ContextPtr c, int32_t size)` with size equals to zero. And, many of the functions can not handle the first one as input, because we would call `GetContext()` at beginning which causes crash. Do we need a CHECK at these functions or return directly when the input Array1 with a nullptr region.

pkufool:
I want to add a test case to cover the empty FsaVec, but I didn't find a way to construct a empty FsaVec.

I thought there are some kinds of FsaVec, the first, FsaVec without any fsa (this kind cause this crash), the second, FsaVec contains empty fsas without states, the third, FsaVec contains empty fsas with states(surely no arcs).

danpovey:
> I think this PR do solve the issue ([k2-fsa/snowfall#227 (comment)](https://github.com/k2-fsa/snowfall/pull/227#issuecomment-873777500)), apart from the stride checking problem, there is also a bug in `Transpose` that return "null" `index_values` when the input is empty FsaVec (with no fsa at all).
> Actually, there are two kinds of Array1 with Dim() equals to zero, one is constructed from default constructor (`region_ == nullptr`), the other is constructed from `Array1(ContextPtr c, int32_t size)` with size equals to zero. And, many of the functions can not handle the first one as input, because we would call `GetContext()` at beginning which causes crash. Do we need a CHECK at these functions or return directly when the input Array1 with a nullptr region.

An Array1 that's default-constructed, with no region, is considered invalid.  That is, it should not appear as an input argument to functions, and should not be produced as output from a function that had valid inputs.  It can be considered an error, or just lead to a crash, if passed into a function.

danpovey:
@pkufool feel free to merge if you think this is ready.

csukuangfj:
@danpovey Could you give @pkufool merge permissions for this repo?  I guess he does not have merge permissions.

PR #786, opened 2021-07-19
Suffix arrays, needed for nbest processing

pkufool:
See  https://github.com/k2-fsa/snowfall/issues/232 and https://github.com/k2-fsa/k2/pull/779


pkufool:
@danpovey What should the mean and variance be when there is no best-matching-position (tightest interval is root interval),  zeros?

danpovey:
It's OK to use the root interval itself.
If the root interval has zero stats (which should not happen), you can use zeros.

k2/csrc/CMakeLists.txt
```diff
@@ -86,7 +88,7 @@ else()
 endif()
 
 # the target
-add_library(context ${context_srcs})
+add_library(context ${context_srcs} ${context_cc_srcs})
```
csukuangfj, line 91:
```suggestion
add_library(context ${context_srcs})
```

k2/python/k2/utils.py
```diff
@@ -656,3 +656,93 @@ def random_fsa_vec(min_num_fsas: int = 1,
...
+              [ [ hi, my, name, is, eos ], [ bye, my, name, is, eos ] ], ... ]
+
+        where the words would actually be represented as integers,
+        The eos symbol is required if this code is to work as intended
+        (otherwise this code will not be able to recognize when we have reached
+        the beginnings of sentences when comparing histories).
+        bos symbols are allowed but not required.
+
+      scores:
+        An one dim torch.tensor with scores.size() == tokens.NumElements(),
```
csukuangfj, line 696:
```suggestion
        A one dim torch.tensor with scores.size() == tokens.NumElements(),
```

PR #787, opened 2021-07-22
Add k2.ctc_topo

csukuangfj:
## Standard ctc_topo

It's deterministic. However, its number of arcs is `O(n^2)`, where `n` is the number of tokens.

## Modified ctc_topo

It's non-deterministic. However, its number of arcs is `O(n)`, where `n` is the number of tokens.

## modified graph = compose(modified_ctc_topo, transcript)

Differences between `standard_graph` and `modified_graph`:
  - In the standard graph, there is a **mandatory** blank between the two repeated symbols `o` in `zoo` (at state 4)
  - There is no such a constraint in the `modified_graph`

For the following input: `[z z <blk> <blk> o o <blk> o <blk>]`, since there is a blank in the input, when it is intersected with the `standard/modified_graph`, the outputs should be the same (shown below), that is, they are **equivalent** in this case
inputs

## standard lattice = intersect(standard_graph, input)
standard_lattice

## modified lattice = intersect(modified_graph, input)
modified_lattice

For the following input: `[z z <blk> <blk> o o  o <blk>]`, since there are no blanks between `o` in the inputs, when it is
intersected with the standard graph, the output is empty; the output for intersecting with the modified graph is not empty (given below)

inputs2

## standard lattice2 = intersect(standard_graph, input2)
It's empty

## modified lattice2 = intersect(modified_graph, input2)
modified_lattice2

---

The code for producing the above figures is available at
https://github.com/k2-fsa/k2/blob/5ef66965509109009f4927fa5cadcf61fe3173ae/k2/python/tests/ctc_topo_test.py#L32

### DIFF
--- a/k2/python/k2/ctc_loss.py
+++ b/k2/python/k2/ctc_loss.py
@@ -30,25 +30,12 @@ class CtcLoss(nn.Module):
     `torch.CtcLoss`.
     '''
 
-    def __init__(self):
-        super().__init__()
-
-    def forward(self,
-                decoding_graph: Fsa,
-                dense_fsa_vec: DenseFsaVec,
-                output_beam: float = 10,
-                reduction: Literal['none', 'mean', 'sum'] = 'sum',
-                use_double_scores: bool = True,
-                target_lengths: Optional[torch.Tensor] = None) -> torch.Tensor:
-        '''Compute the CTC loss given a decoding graph and a dense fsa vector.
-
+    def __init__(self,
+                 output_beam: float,
+                 reduction: Literal['none', 'mean', 'sum'] = 'sum',
+                 use_double_scores: bool = True):
+        '''
         Args:
-          decoding_graph:
-            An FsaVec. It can be the composition result of a CTC topology
-            and a transcript.
-          dense_fsa_vec:
-            It represents the neural network output. Refer to the help
-            information in :class:`k2.DenseFsaVec`.
           output_beam:
              Beam to prune output, similar to lattice-beam in Kaldi.  Relative
              to best path of output.
@@ -61,6 +48,26 @@ class CtcLoss(nn.Module):
           use_double_scores:
             True to use double precision floating point in computing
             the total scores. False to use single precision.
+        '''
+        super().__init__()
+        assert reduction in ('none', 'mean', 'sum')
+        self.output_beam = output_beam
+        self.reduction = reduction
+        self.use_double_scores = use_double_scores
+
+    def forward(self,
+                decoding_graph: Fsa,
+                dense_fsa_vec: DenseFsaVec,
+                target_lengths: Optional[torch.Tensor] = None) -> torch.Tensor:
+        '''Compute the CTC loss given a decoding graph and a dense fsa vector.
+
+        Args:
+          decoding_graph:
+            An FsaVec. It can be the composition result of a CTC topology
+            and a transcript.
+          dense_fsa_vec:
+            It represents the neural network output. Refer to the help
+            information in :class:`k2.DenseFsaVec`.
           target_lengths:
             Used only when `reduction` is `mean`. It is a 1-D tensor of batch
             size representing lengths of the targets, e.g., number of phones or
@@ -69,20 +76,20 @@ class CtcLoss(nn.Module):
           If `reduction` is `none`, return a 1-D tensor with size equal to batch
           size. If `reduction` is `mean` or `sum`, return a scalar.
         '''
-        assert reduction in ('none', 'mean', 'sum')
-        lattice = intersect_dense(decoding_graph, dense_fsa_vec, output_beam)
+        lattice = intersect_dense(decoding_graph, dense_fsa_vec,
+                                  self.output_beam)
 
         tot_scores = lattice.get_tot_scores(
-            log_semiring=True, use_double_scores=use_double_scores)
+            log_semiring=True, use_double_scores=self.use_double_scores)
         loss = -1 * tot_scores
         loss = loss.to(torch.float32)
 
-        if reduction == 'none':
+        if self.reduction == 'none':
             return loss
-        elif reduction == 'sum':
+        elif self.reduction == 'sum':
             return loss.sum()
         else:
-            assert reduction == 'mean'
+            assert self.reduction == 'mean'
             loss /= target_lengths
             return loss.mean()
 
@@ -121,5 +128,8 @@ def ctc_loss(decoding_graph: Fsa,
       If `reduction` is `none`, return a 1-D tensor with size equal to batch
       size. If `reduction` is `mean` or `sum`, return a scalar.
     '''
-    return CtcLoss()(decoding_graph, dense_fsa_vec, output_beam, reduction,
-                     use_double_scores, target_lengths)
+    m = CtcLoss(output_beam=output_beam,
+                reduction=reduction,
+                use_double_scores=use_double_scores)
+
+    return m(decoding_graph, dense_fsa_vec, target_lengths)
--- a/k2/python/k2/ctc_loss.py
+++ b/k2/python/k2/ctc_loss.py
@@ -62,7 +62,7 @@ class CtcLoss(nn.Module):
             the total scores. False to use single precision.
           target_lengths:
             Used only when `reduction` is `mean`. It is a 1-D tensor of batch
-            size. It represent lengths of the targets, e.g., number of phones or
+            size representing lengths of the targets, e.g., number of phones or
             number of word pieces in a sentence.
         Returns:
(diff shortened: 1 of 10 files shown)


k2/python/k2/fsa_algo.py
```diff
@@ -1011,3 +1011,78 @@ def ctc_graph(symbols: Union[List[List[int]], k2.RaggedInt],
     aux_labels = k2.index(symbol_values, arc_map)
     fsa = Fsa(ragged_arc, aux_labels=aux_labels)
     return fsa
+
+def ctc_topo(max_token: int, standard: bool) -> k2.Fsa:
```
danpovey, line 1015: I think it would be clearer to have 'modified: bool = False' as the arg.

danpovey:
.. also, I think I changed my mind about having a k2 version of CtcLoss.  I guess it would be a convenient way to compare k2 with PyTorch ctc; and people can always look at how it's implemented if they want.

csukuangfj:
ok, I will add k2.CtcLoss

csukuangfj:
I just added `k2.CtcLoss` and its functional form `k2.ctc_loss`. It supports the same semantics `reduction` used in `torch.CtcLoss`. Unit-tests are given to ensure that `k2.CtcLoss` produces the same output as `torch.CtcLoss` when given the same input.

k2/python/k2/fsa_algo.py
csukuangfj: Thanks. Fixed.

k2/python/k2/ctc_loss.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def __init__(self):
+        super().__init__()
+
+    def forward(self,
+                decoding_graph: Fsa,
+                dense_fsa_vec: DenseFsaVec,
+                output_beam: float = 10,
+                reduction: Literal['none', 'mean', 'sum'] = 'sum',
+                use_double_scores: bool = True,
```
pzelasko, line 41: I think that arguments `output_beam`, `reduction`, and `use_double_scores` could be moved into the constructor instead -- this would seem more consistent with `torch.CTCLoss` https://pytorch.org/docs/stable/generated/torch.nn.CTCLoss.html

csukuangfj: Thanks. Fixed.

danpovey:
OK to merge?

pzelasko:
Is this CTC topology 100% correct? I thought that CTC allows repeated token IDs without blank in between to be parsed as a single token instance, e.g. `[1, 1, 0]` could be a single token `[1]`, whereas `[1, 0, 1]` must be two tokens `[1, 1]`. But the former is not the case with inverted `k2.ctc_topo` used for intersection:

```python
H_inv = k2.arc_sort(k2.ctc_topo(1).invert())
token_ids_with_blanks = k2.arc_sort(k2.top_sort(
    k2.linear_fsa([1, 1, 0])
))
x = k2.intersect(token_ids_with_blanks, H_inv)
k2.to_dot(x)
```

image

The graph has some extra epsilon arcs, if we remove them, we can clearly see a path that interprets it as two instances of token `1`:

(BTW I'm not sure that remove_epsilon should remove arcs where label=0 but aux_label != 0?)

image

On the other hand, fig 3 from (Graves et al., 2006) paper suggests that would be interpreted as a single token instance (note the horizontal arrows in C, A, and T)

image

If I'm missing anything here, please LMK.

csukuangfj:
> If I'm missing anything here, please LMK.

0 is not a token ID. That is the key.

So 
```python3
token_ids_with_blanks = k2.arc_sort(k2.top_sort(
    k2.linear_fsa([1, 1, 0])
))
```
is not correct. Please use

```
token_ids_without_blanks = k2.arc_sort(k2.top_sort(
    k2.linear_fsa([1, 1])
))
```


Also, you can use `k2.compose()` so that you don't need to invert `H`. If you use `k2.intersect()`, you have
to invert `x` at the end.
(Please see https://github.com/k2-fsa/k2/blob/31e1307c3131d03384ef83c33472309d22d2e79d/k2/python/tests/ctc_loss_test.py#L209-L211)

---

Please see the following colab notebook I just created.
https://colab.research.google.com/drive/1Xc4eK4Ve5sA6HaH3j_WQWXJ4Q0WyW2yi?usp=sharing


Screen Shot 2021-11-25 at 07 47 18

You can see that there is a mandatory blank.

csukuangfj:
Note there is another function `k2.ctc_graph()`

Screen Shot 2021-11-25 at 07 52 01

pzelasko:
Hmm I meant when `[1, 1, 0]` correspond e.g. to greedily decoded model outputs (argmax over posteriors). It means when I'm going back from the predicted model sequence to recognized token IDs. My variable name `token_ids_with_blank` was misleading I think.

pzelasko:
In other words, if the most likely path from the model is `[1, 1, 0]`, then it should mean the model recognized one-token sequence `[1]`, but with the current CTC topo implementation, it seems to recognize two-token sequence `[1, 1]`.

csukuangfj:
> In other words, if the most likely path from the model is [1, 1, 0], then it should mean the model recognized one-token sequence [1], but with the current CTC topo implementation, it seems to recognize two-token sequence [1, 1].

To get `[1]` from `[1, 1, 0]`, I think the correct API to use is [torch.unique_consecutive](https://pytorch.org/docs/stable/generated/torch.unique_consecutive.html)

I don't understand why it is possible to get `[1]` by constructing an FSA from `[1, 1, 0]` and intersecting it
with a `ctc_topo`.

pzelasko:
Maybe I'm interpreting wrongly some of the operations or how CTC topo is intended to be used. But my understanding is that the original formulation of CTC allowed the model to predict `[1, 1, 0]` to indicate that token `[1]` is being recognized. See e.g. the illustration in https://distill.pub/2017/ctc/

I think that the CTC topology in k2 doesn't allow that, as it doesn't allow to collapse consecutive symbols `[1, 1]` into a single symbol `[1]`. Instead, if model predicted `[1, 1, 0]` then it'd be "collapsed" to token IDs `[1, 1]`.

csukuangfj:
>  But my understanding is that the original formulation of CTC allowed the model to predict [1, 1, 0] to indicate that token [1] is being recognized

Ah, I see what you meant.

---

> I think that the CTC topology in k2 doesn't allow that, as it doesn't allow to collapse consecutive symbols [1, 1] into a single symbol [1]

No, it does support that.

Please see the test code for.more.examples.
https://github.com/k2-fsa/k2/blob/12f591526a7afa3a3d55cc37c67576328cf926e5/k2/python/tests/ctc_topo_test.py#L72

You can also find the differences between `modified=True` and `modified=False`.

---

> Maybe I'm interpreting wrongly some of the operations or how CTC topo is intended to be used

One thing I think you need to remember is that `ctc_topo` converts repeated **input** symbols to a single symbol.
So, in your case, you don't need to `invert` `ctc_topo`.  Please pay attention to the `input` and `output` labels
of `ctc_topo`, i.e., `labels` and `aux_labels` in k2.

csukuangfj:
Also note that `0` in this case represents the `blank` symbol, not epsilon.

pzelasko:
Thanks, that makes sense! I confused the input and output labels in the output of `k2.ctc_topo`. 

But now that I don’t invert it, there are still two unique paths after the intersection, and it seems to me like only one of them (green circle) is „correct” for the original CTC formulation. See attached: 

9D557B2E-9521-48BD-BF43-A10F7521AB6C

csukuangfj:
> Thanks, that makes sense! I confused the input and output labels in the output of `k2.ctc_topo`. 
> 
> 
> 
> But now that I don’t invert it, there are still two unique paths after the intersection, and it seems to me like only one of them (green circle) is „correct” for the original CTC formulation. See attached: 
> 
> 
> 
> 9D557B2E-9521-48BD-BF43-A10F7521AB6C
> 
> 

Please see my above comment.

0 is a blank, not epsilon here.

You have to set the argument treat epsilon specially to False when invoking k2.intersect().

You can see the usage in the above test code.

pzelasko:
Thanks for the explanation, it's now clear to me. You have to disable implicit epsilon self-loops with `treat_epsilons_specially=False` to make it do the right thing.

pzelasko:
BTW could this arg be named `add_eps_self_loops=True` instead? I think it gives a better description of what it's doing.

csukuangfj:
> You have to disable implicit epsilon self-loops with treat_epsilons_specially=False to make it do the right thing.

It is not only about epsilon self-loops but also about epsilon non-self-loops.

It is about all epsilon transitions.

When `treat_epsilons_specially=False`, `0` is treated like a normal symbol, no matter where it appears.

PR #788, opened 2021-07-22
Fix CPU version AtomicCAS

pkufool:


danpovey:
This undoes an optimization (avoid memory read) that was no longer valid because of changes in calling code.  The hope is that since the function is inlined, the compiler will automatically do the optimization where it's actually applicable.

PR #790, opened 2021-07-23
Draft of pushing code for determinization

danpovey:


danpovey:
@pkufool I was hoping you could finish this.   I have not attempted to compile it.
It needs:
  - testing code
  - getting it to compile
  - wrapping code so the option is exposed to Python, and the option is available at that level.
The testing of the pushing is a little tricky, we can perhaps leave the C++ tests as they are currently, with the
no-pushing option, and just test from Python that the un-pushed and pushed versions are equivalent, plus eyeball
them to make sure they look right.

PR #791, opened 2021-07-25
Weight pushing for determinization

pkufool:
Finishing the remaining work in https://github.com/k2-fsa/k2/pull/790

### DIFF
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -311,26 +311,40 @@ void RemoveEpsilonAndAddSelfLoops(FsaOrVec &src, int32_t properties,
 }
 
 
-
-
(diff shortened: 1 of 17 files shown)


danpovey:
Great work!

pkufool:
I think weight pushing for determinization is working now, visualize results are as follows:

BTW, current implementation of weight pushing will produce _**negative**_ scores, does it matter? @danpovey 

|Weight pushing|result|
|--|--|
|Original|image|
|kNoWeightPushing|image|
|kLogWeightPushing|image|
|kTripicalWeightPushing|image|

danpovey:
The log and tropical weight pushing give the same score in this case, is it right?  I don't think it should be the same for this example.
Negative scores can be OK.

pkufool:
Sorry, I was mistaken, fixed.
The picture in comments above is updated.

PR #794, opened 2021-08-02
Fix some bugs of empty inputs

pkufool:
We met serveral crashes when decoding with single wav, as post in https://github.com/k2-fsa/snowfall/issues/239. I thought these crashes are all about empty inputs.

@luomingshuang 's issue locating in the `k2.index_select` function, occurs when the input `indexes` is an empty tensor, see the comments in source code for details.

~~@alucassch 's issue occuring in the `nbest_decoding`, is a bug in the `k2.ragged.arg_max_persublist`,. The function return `[-1]` with the empty input, which causes a strange behavior in the following code.~~

**Update:**  I made a mistake, the return value of `arg_max_persublist` is expected, this was also a bug in `k2.index(src, index)`.  If `src` is an empty tensor, `k2.index` should return an empty tensor,  in current implementation, `k2.index([], [-1])` would return `[0]`.
 

    argmax_indexes = k2.ragged.argmax_per_sublist(ragged_tot_scores)

    # Since we invoked `k2.ragged.unique_sequences`, which reorders
    # the index from `paths`, we use `new2old`
    # here to convert argmax_indexes to the indexes into `paths`.
    #
    # Use k2.index here since argmax_indexes' dtype is torch.int32
    best_path_indexes = k2.index(new2old, argmax_indexes)

    paths_2axes = k2.ragged.remove_axis(paths, 0)

    # best_paths is a k2.RaggedInt with 2 axes [path][arc_pos]
    best_paths = k2.index(paths_2axes, best_path_indexes)

### DIFF
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -399,6 +399,9 @@ FsaVec LinearFsas(const Ragged<int32_t> &symbols) {
   // row_splits2 maps from state_idx01 to arc_idx012; row_ids2 does the reverse.
   // We'll set them in the lambda below.
   Array1<int32_t> row_splits2(c, num_states + 1), row_ids2(c, num_arcs);
+  // When num_states equals to zero, the code below won't set the last value of
+  // row_splits2, we should initialize here, or it will be a random value.
+  if (num_states == 0) row_splits2 = 0;
 
   int32_t *row_ids2_data = row_ids2.Data(),
           *row_splits2_data = row_splits2.Data();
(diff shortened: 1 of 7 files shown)


pkufool:
I fixed an empty input issue in the `TransposeRagged` before https://github.com/k2-fsa/k2/pull/782, it was alse about the empty input. After reading lots of code, I am still confusing about the empty ragged.

danpovey:
Yes, in a 2-d ragged tensor, [ ] and [ [] ] are not the same but they both have zero elements.  Note: the text form does not tell you the number of axes, for example [ ] is a representation of a valid ragged tensor with any number of axes.

danpovey:
It looks fine to me.  If you find any case where you are really not sure what should happen, because in your opinion the comments on the interface do not properly define the behavior for some input, then let me know and we can discuss it.

PR #800, opened 2021-08-17
pickle ragged

pkufool:


k2/python/tests/ragged_test.py
```diff
@@ -56,23 +56,24 @@ def test_ragged_int_from_str(self):
...
-            torch.save(ragged_int, "ragged.pt")
-            ragged_reload = torch.load("ragged.pt")
-            self.assertEqual(ragged_int, ragged_reload)
-            os.remove("ragged.pt")
+            # test num_axes == 3
+            raggeds = ("[ [ [ ] ] ]", "[ [ [1 2] [3] ] [ [4 5] [6] ] ]")
+            for s in raggeds:
+                ragged_int = k2.RaggedInt(s).to(device)
+                torch.save(ragged_int, "ragged.pt")
+                ragged_reload = torch.load("ragged.pt")
```
danpovey, line 74: So I assume map_location would naturally work for this?

pkufool: Yes.
image

PR #802, opened 2021-08-17
Fix a bug in invalidating the caches of an FSA.

csukuangfj:
We should also remove arc_cdf and arc_post.

Found this bug during calling `k2.random_paths` with a scaled `lattice.scores`.

This patch should fix it.


danpovey:
LGTM.

PR #807, opened 2021-08-22
Support sum_per_sublist for int32

### LINKED ISSUE #805: `sum` interface
The sum_per_sublist function for ragged lists in k2 only supports float; I'd like it to also support int32.
Incidentally, I wonder whether it might be more torch-like to just call it `sum`, and have the axis specifiable, defaulting
to -1.  And to also have a class-member version of it.  So it could be called as k2.ragged.sum(ragged_array).
Of course this would have to fail, right now, for axis == 0, for instance.  But in future we might implement that case also.

pkufool:
Fix #805

### DIFF
--- a/k2/python/k2/ragged/ops.py
+++ b/k2/python/k2/ragged/ops.py
@@ -175,7 +175,7 @@ def pad(src: Union[_k2.RaggedInt, _k2.RaggedFloat],
     return _k2.pad_ragged(src, mode, value)
 
 
-def sum_per_sublist(src: _k2.RaggedFloat,
+def sum_per_sublist(src: Union[_k2.RaggedFloat, _k2.RaggedInt],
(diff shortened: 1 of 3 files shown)


k2/python/k2/ragged/ops.py
```diff
@@ -191,9 +191,11 @@ def sum_per_sublist(src: _k2.RaggedFloat,
 
     Args:
       src:
-        A ragged float tensor. Note that the sum is performed on the last axis.
+        A ragged float or int32 tensor.
+        Note that the sum is performed on the last axis.
```
csukuangfj, line 195: Could you also add documentation for `initial_value`?

csukuangfj: And document the behavior when one sublist is empty.

PR #808, opened 2021-08-22
Support dtype conversions of ragged tensors.

### LINKED ISSUE #806: '.to()' for ragged tensors
Can we please have the '.to()' of ragged tensors support changing dtype, not just device?  For now it would just support changing float and int32, later this may change.
This is low priority though.

csukuangfj:
Closes #806 

## TODOs
- [x] Add more tests

### DIFF
--- a/k2/csrc/array.h
+++ b/k2/csrc/array.h
@@ -237,6 +237,24 @@ class Array1 {
     return ans;
   }
 
+  // The ToType() macro will be expanded to
+  //
+  //  Array1<int32_t> ToInt() const;
+  //  Array1<float> ToFloat() const;
+  //  Array1<double> ToDouble() const;
+  //  Array1<int64_t> ToLong() const;
+  //
+  // which is roughly equivalent to the following template
+  //
+  //  template<typename U>
+  //  Array1<U> To() const;
+  //
+  // The purpose is to convert Array1<T> to Array1<U>, e.g.,
+  // convert Array1<int32_t> to Array1<float>.
+  //
+  // If T == U, then the array itself is returned; otherwise,
+  // a new array is returned.
+  //
 #define ToType(type, name)                                    \
   Array1<type> To##name() const {                             \
     if (std::is_same<type, T>::value)                         \
--- a/k2/csrc/array.h
+++ b/k2/csrc/array.h
@@ -237,8 +237,32 @@ class Array1 {
     return ans;
   }
 
-  // Copy from another array of the same dimension and type.
-  void CopyFrom(const Array1<T> &src);
+#define ToType(type, name)                                    \
+  Array1<type> To##name() const {                             \
+    if (std::is_same<type, T>::value)                         \
+      return *reinterpret_cast<const Array1<type> *>(this);   \
+                                                              \
+    Array1<type> ans(Context(), Dim(), DtypeOf<type>::dtype); \
+    if (Dim() == 0) return ans;                               \
+                                                              \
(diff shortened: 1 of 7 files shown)


k2/csrc/array.h
```diff
@@ -237,8 +237,32 @@ class Array1 {
     return ans;
   }
 
-  // Copy from another array of the same dimension and type.
-  void CopyFrom(const Array1<T> &src);
+#define ToType(type, name)                                    \
```
csukuangfj, line 240: I am using macros instead of templates here because it throws the following error
```
error: ‘__T0’ was not declared in this scope
```
if templates were used here. I don't know how to fix such kinds of errors, so macros are used here.

k2/python/csrc/torch/torch_util.h
```diff
@@ -203,45 +203,142 @@ struct TensorTag {};
...
+
+   @param pyclass      The object to move.
+   @param scalar_type  The type to convert to.
+
+   @return Return the passed object if it was of the given type;
+           otherwise, return a clone of the given object with the
+           given type.
+ */
+template <typename PyClass>
+py::object To(PyClass &pyclass, torch::ScalarType scalar_type) {
```
csukuangfj, line 255: `py::object` is used here so that we can return
- `Ragged<int32_t>`
- `Ragged<float>`
- etc
in the same C++ function.

k2/csrc/ragged.h
```diff
@@ -465,6 +465,20 @@ struct Ragged {
     return Ragged<T>(new_shape, new_values);
   }
 
+#define ToType(type, name)                                  \
```
danpovey, line 468: Hm, I was thinking you'd just overload the to() function to accept a torch type?
That's how I normally do type conversions in torch, e.g. tensor.to(torch.float32) or something.
This code is kind of hard to read and hard to discover.

csukuangfj: > That's how I normally do type conversions in torch, e.g. tensor.to(torch.float32) or something.

Yes, the current implementation supports that. Valid usages are:
```python
# create a ragged_tensor
ragged_tensor = ragged_tensor.to(torch.device('cuda', 0))
ragged_tensor = ragged_tensor.to(torch.int32)
```

> Hm, I was thinking you'd just overload the to() function to accept a torch type?

`torch.dtype` and `torch.device` cannot be accessed in C++, I think. Both of them can be accessed in
C++ using `py::object` and that's why I overload `k2::To`: (1) One function is to handle `torch.device`
(2) The other function is to handle `torch.dtype`.

The reason I am using `ToType` is to implement something like converting `Ragged<float>` to `Ragged<int32_t>`.
I tried to replace the macro with a template, but it throws some errors mentioned in the comment that I've no
idea how to solve them.

k2/csrc/array.h
danpovey: M, OK, please make a comment so it's quite clear what is going on.

k2/python/csrc/torch/torch_util.h
danpovey: OK.  I didn't realize that the other code was internal implementation, not interface.

k2/csrc/array.h
csukuangfj: ok, I just added some documentation to the macro `ToType` explaining what it is doing.

csukuangfj:
Not for merging as we are currently working on the `Any` type.

csukuangfj:
Closing via #812

PR #809, opened 2021-08-22
[WIP] Wrap ragged tensor

pkufool:
This pull request contains some fixes about ragged tensor:

- [X] add lists of lists constructor
- [x] make operations on ragged tensors more easily discoverable from the Python side.
- [ ] add indexing to RaggedInt.
- [x] make operations on ragged tensors more torch-like

**Not  ready for merge**, need more tests to make sure the wrapped Ragged tensor behaves the same as `_k2.Ragged`

### DIFF
--- a/k2/python/csrc/torch/ragged.cu
+++ b/k2/python/csrc/torch/ragged.cu
@@ -51,7 +51,7 @@ static void PybindRaggedTpl(py::module &m, const char *name) {
               }),
               py::arg("shape"), py::arg("values"));
 
-  pyclass.def(py::init([](const std::vector<std::vector<T>> lists,
+  pyclass.def(py::init([](const std::vector<std::vector<T>> &lists,
                           py::object device = py::str("cpu"))
     -> std::unique_ptr<PyClass> {
                 ContextPtr context = GetContext(device);
@@ -141,6 +141,41 @@ static void PybindRaggedTpl(py::module &m, const char *name) {
       },
       py::arg("axis"), py::arg("i"));
 
+  pyclass.def(
+      "index",
+      [](PyClass &self, int32_t axis, torch::Tensor indexes,
+         bool need_value_indexes =
+             true) -> std::pair<PyClass, torch::optional<torch::Tensor>> {
+        DeviceGuard guard(self.Context());
+        Array1<int32_t> indexes_array = FromTorch<int32_t>(indexes);
+        Array1<int32_t> value_indexes;
+
+        Ragged<T> ans = Index(self, axis, indexes_array,
+                              need_value_indexes ? &value_indexes : nullptr);
+        torch::optional<torch::Tensor> value_indexes_tensor;
+
+        if (need_value_indexes) value_indexes_tensor = ToTorch(value_indexes);
+
+        return std::make_pair(ans, value_indexes_tensor);
+      },
+      py::arg("axis"), py::arg("indexes"),
+      py::arg("need_value_indexes") = true);
+
+  pyclass.def(
+      "index",
+      [](PyClass &self, Ragged<int32_t> &indexes, bool remove_axis = false)
+      -> Ragged<T> {
+        DeviceGuard guard(self.Context());
+        return Index(self, indexes, remove_axis);
+      }, py::arg("indexes"), py::arg("remove_axis") = false);
+
+  pyclass.def(
+     "remove_axis",
+     [](PyClass &self, int32_t axis) -> PyClass {
+       DeviceGuard guard(self.Context());
+       return self.RemoveAxis(axis);
+     }, py::arg("axis"));
+
   pyclass.def("__eq__", [](const PyClass &self, const PyClass &other) -> bool {
     DeviceGuard guard(self.Context());
     return Equal(self, other);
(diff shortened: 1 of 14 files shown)


k2/python/csrc/torch/ragged.cu
```diff
@@ -49,6 +50,15 @@ static void PybindRaggedTpl(py::module &m, const char *name) {
                 return std::make_unique<PyClass>(shape, FromTorch<T>(values));
               }),
               py::arg("shape"), py::arg("values"));
+
+  pyclass.def(py::init([](const std::vector<std::vector<T>> lists,
```
csukuangfj, line 54:
```suggestion
  pyclass.def(py::init([](const std::vector<std::vector<T>> &lists,
```

k2/python/csrc/torch/ragged.cu
```diff
@@ -49,6 +50,15 @@ static void PybindRaggedTpl(py::module &m, const char *name) {
...
+                          py::object device = py::str("cpu"))
+    -> std::unique_ptr<PyClass> {
```
csukuangfj, line 56: Do we need a device guard here?

k2/python/csrc/torch/torch_util.cu
```diff
@@ -155,6 +155,18 @@ torch::Tensor ToTorch(Tensor &tensor) {
...
 }
 
+ContextPtr GetContext(py::object device_obj) {
+  auto device = torch::Device(static_cast<py::str>(device_obj));
+  if (device.type() == torch::kCPU) {
+    return GetCpuContext();
+  } else if (device.type() == torch::kCUDA) {
+    return GetCudaContext(device.index());
+  } else {
+    K2_LOG(FATAL) << "Unsupported devices.";
```
csukuangfj, line 165:
```suggestion
    K2_LOG(FATAL) << "Unsupported device: " << device.str();
```

k2/python/k2/fsa.py
```diff
@@ -1394,7 +1394,7 @@ def set_scores_stochastic_(self, scores) -> None:
 
         # Note we use `to` here since `scores` and `self.scores` may not
         # be on the same device.
-        self.scores = ragged_scores.values.to(self.scores.device)
+        self.scores = ragged_scores.values().to(self.scores.device)
```
csukuangfj, line 1397: `ragged_scores` here is an instance of `k2.ragged.RaggedFloat`, which
has a property `values`.

`k2.ragged.RaggedFloat` is a wrapper around `k2.RaggedFloat` and supports autograd.

pkufool:
**Not ready for reviewing** ,  just to confirm that I am doing the right way. @danpovey  suggested last week, we should make the operation on ragged tensor more easily discoverable and add torch-like interface.  I am adding python wrapper for `_k2.RaggedInt`, `_k2.RaggedFloat`  and _k2 functions.  so that, we will replace all the `_k2.RaggedInt` `_k2.RaggedFloat` with the wrapped one, and we will have two versions of ragged operations like pytorch.  
For example,
```
tensor = k2.RaggedInt([[1,2,3],[4,5],[6]])
# function in ragged
sum = k2.ragged.sum(tensor, axis=1)
# member function
sum = tensor.sum(axis=1)
```
This would be a big refactoring,  hope for you opinions.

danpovey:
I think it's a good idea.
There is one thing we should figure out, though.  
In k2 there is the "Any" type, which can be used as the template arg when you don't know the type of a tensor.
The idea was that we can have dynamic dispatching, although I never fully decided where this dispatching
would take place, i.e. would it be at the wrapper level in python/csrc/torch, or whether it would be at the level
of the individual operations in k2/csrc/.

If we were to start using this Any type, and just make k2.Ragged a wrapper for Ragged<Any>, it would
make the interface more Torch-like because we could specify dtypes like in torch code (we'd have to create
a special dtype for Arc, of course).  One possibility would be to, temporarily, keep RaggedInt and RaggedFloat
around to avoid breaking code too suddenly, while also leaving the Ragged wrapper, which would have different,
simpler and more Torch-like interfaces.

danpovey:
... another option, in case we decide it is easiest just to have most of this done in C++, is to create a "fake" parallel Python type just for documentation, so things can be more easily discoverable that way.  Or simply figure out a way to make nice documentation about these operations.   But it may be difficult to make the backprop for ops on float work correctly from C++.  In torch I think they do this with macros and a meta-build system of some kind, I don't think the code is that simple and easy.

csukuangfj:
> ... another option, in case we decide it is easiest just to have most of this done in C++, is to create a "fake" parallel Python type just for documentation, so things can be more easily discoverable that way. Or simply figure out a way to make nice documentation about these operations. But it may be difficult to make the backprop for ops on float work correctly from C++. In torch I think they do this with macros and a meta-build system of some kind, I don't think the code is that simple and easy.

I like this approach.

As for
> But it may be difficult to make the backprop for ops on float work correctly from C++.

I think it is doable and its code for backprop in C++ is as easy as the one we are using in Python.
For the meta-build system, it is also easy to support that in the current build system.

pkufool:
see #812

PR #812, opened 2021-08-25
Make k2 ragged tensor more PyTorch-y like.

csukuangfj:
This pull-request aims to kill `k2.RaggedInt` and `k2.RaggedFloat`.

---

## Usage example

```python
#!/usr/bin/env python3

import torch
import k2
import _k2

# TODO: will move _k2.ragged to k2.ragged

a = _k2.ragged.tensor([[1, 2], [3, 4.0]])
assert a.dtype == torch.float32

a = _k2.ragged.tensor([[1, 2], [3, 4]])
assert a.dtype == torch.int32

a = _k2.ragged.tensor([[1, 2], [3, 4]], dtype=torch.float64)
assert a.dtype == torch.float64
assert a.device == torch.device("cpu")

a = a.to(torch.device("cuda", 0))
assert a.device == torch.device("cuda", 0)

b = a.to(torch.int32)
b.dtype == torch.int32

a = b.to(torch.device("cpu")).to(torch.int64)
assert a.device == torch.device("cpu")
assert a.dtype == torch.int64

assert isinstance(a, k2.ragged.Tensor)
```

---

Will create a new class in C++ to wrap `Ragged<Any>`.


csukuangfj:
A preview of the documentation can be found at
https://k3.readthedocs.io/en/latest/python_api/tensor.html

Will add usage examples to it once the code is finished.

k2/python/csrc/torch/doc/any.h
```diff
@@ -0,0 +1,199 @@
+
+
+#ifndef K2_PYTHON_CSRC_TORCH_DOC_ANY_H_
+#define K2_PYTHON_CSRC_TORCH_DOC_ANY_H_
+
+namespace k2 {
+
+static constexpr const char *kRaggedAnyInitDataDoc = R"doc(
+Create a ragged tensor with two axes.
```
csukuangfj, line 9: I am trying to put the documentation in C++ headers.
Users can view the help information in a usual way, i.e.,
```python
>>> import k2.ragged
>>> help(k2.ragged.Tensor.__init__)
```

The output is given below:
Screen Shot 2021-08-28 at 18 14 17

---

Also, the doc is going to be hosted at
https://k2-fsa.github.io/k2/index.html

The reason is that now the doc depends on the C++ source code and we have to compile
 `_k2` to generate the doc.

---
A preview is available at 
https://csukuangfj.github.io/k2/python_api/tensor.html#k2.ragged.Tensor

csukuangfj: The doc is easier to discover as it is bound to the actual class. There is no need to create a fake class just for documentation purposes.

danpovey:
Great!

k2/python/csrc/torch/doc/any.h
```diff
@@ -0,0 +1,416 @@
...
+
+static constexpr const char *kRaggedAnyNeDoc = R"doc(
+Compare two ragged tensors.
+
+Caution:
+  The two tensors MUST have the same dtype. Otherwise,
+  it throws.
+
+>>> import torch
+impor>>> import k2.ragged as k2r
```
pkufool, line 231: delete prefix 'impor'.

k2/python/csrc/torch/doc/any.h
```diff
@@ -0,0 +1,416 @@
...
+static constexpr const char *kRaggedAnyInitStrDoc = R"doc(
+Create a ragged tensor from its string representation.
+
+An example string for a 2-axis ragged tensor is given below::
+
+    [ [1]  [2] ]
+
+An example string for a 3-axis ragged tensor is given below::
+
+    [ [[1] [2 3]]  [[2] [] [3, 4,]] ]
```
pkufool, line 47: no comma, [ [[1] [2 3]] [[2] [] [3, 4,]] ] > [ [[1] [2 3]] [[2] [] [3 4]] ]

k2/python/csrc/torch/doc/any.h
```diff
@@ -0,0 +1,416 @@
...
+>>> a = k2r.Tensor([[1], [2, 3, 5]])
+>>> a.dtype
+torch.int32
+>>> b = a.to(torch.float64)
+>>> b.dtype
+torch.float64
+
+Caution:
+  Currently, only support for dtypes ``torch.int32``, ``torch.float32``, and
+  ``torch.float64`` are implemented. We can support other types if needed.
```
pkufool, line 124: delete "are implemented"

k2/python/csrc/torch/doc/any.h
```diff
@@ -0,0 +1,416 @@
...
+static constexpr const char *kRaggedAnyNumelDoc = R"doc(
+Returns:
+  Return number of elements in this tensor. It equals to
+  `self.data.numel()`.
+>>> import torch
+>>> import k2.ragged as k2r
+>>> a = k2r.Tensor([[1], [], [3, 4, 5, 6]])
+>>> a.numel()
+5
+>>> b = k2r.Tensor('[ [[1] [] []]  [[2 3]]]')
```
pkufool, line 354: It is better to use the same constructor in one example.

k2/python/csrc/torch/v2/any.cu
```diff
@@ -206,14 +205,38 @@ void PybindRaggedAny(py::module &m) {
...
   SetMethodDoc(&any, "__setstate__", kRaggedAnySetStateDoc);
 
+  any.def("remove_axis", &RaggedAny::RemoveAxis, py::arg("axis"));
+  any.def("arange", &RaggedAny::Arange, py::arg("axis"), py::arg("begin"),
+          py::arg("end"));
+  any.def("remove_values_leq", &RaggedAny::RemoveValuesLeq, py::arg("cutoff"));
+  any.def("remove_values_eq", &RaggedAny::RemoveValuesEq, py::arg("target"));
+  any.def("argmax_per_sublist", &RaggedAny::ArgMaxPerSublist,
+          py::arg("initial_value"));
+  any.def("max_per_sublist", &RaggedAny::MaxPerSublist,
```
pkufool, line 215: To make the interface more torch-like, I think we should change this to `max` and add one more parameter like `axis`, as dan suggested before. So as other sublist operators.

Of course, we could keep two interface if back compatible is needed.

danpovey: Agreed, I think that would be a good idea.
It's OK to have an arg for axis, but only support -1 (or num_axes-1) for now.
Incidentally, in torch, `axis` is called `dim`.  We should probably stay consistent within k2 for now though.

csukuangfj: I agree to remove the `per-sublist` part.

All operations on the `values` are on the last axis, I think. Is there a need to add an extra argument `axis`?
How is it supposed to be used by users?

pkufool: I remembered that Dan said we would have a plan to support the operations on other axis, we could add this argument and set the default value to -1.

k2/python/csrc/torch/fsa_algo.cu
```diff
@@ -686,72 +699,93 @@ static void PybindReplaceFsa(py::module &m) {
 }
 
 static void PybindCtcGraph(py::module &m) {
+  m.def(
+      "ctc_graph",
+      [](RaggedAny &symbols, torch::optional<torch::Device> = {},
```
pkufool, line 704: I think we don't need device argument here.

k2/python/csrc/torch/fsa_algo.cu
```diff
@@ -686,72 +699,93 @@ static void PybindReplaceFsa(py::module &m) {
...
       "ctc_graph",
       [](const std::vector<std::vector<int32_t>> &symbols,
-          int32_t gpu_id = -1, bool modified = false, bool need_arc_map = true)
-        -> std::pair<FsaVec, torch::optional<torch::Tensor>> {
-        ContextPtr context;
-        if (gpu_id < 0)
-          context = GetCpuContext();
-        else
-          context = GetCudaContext(gpu_id);
+         torch::optional<torch::Device> device = {}, bool modified = false,
```
pkufool, line 722: What's the benefit of passing argument device by `torch::Device` and `std::string`,  why not use `py::object` to avoid implementing it twice.

we can get context like the code below, whether the passing in argument is `cuda:0` or `torch.device`.
```
ContextPtr GetContext(py::object device_obj) {
  auto device = torch::Device(static_cast<py::str>(device_obj));
  if (device.type() == torch::kCPU) {
    return GetCpuContext();
  } else if (device.type() == torch::kCUDA) {
    return GetCudaContext(device.index());
  } else {
    K2_LOG(FATAL) << "Unsupported device: " << device.str();
    return GetCpuContext();   // unreachable code
  }
}
```

k2/python/csrc/torch/fsa_algo.cu
csukuangfj: The purpose is to make the call `k2.linear_fsa` in Python as uniform as possible: You don't need to know the type of `lables`.
https://github.com/k2-fsa/k2/blob/8988489702372b7e331dcd6815d3e9bb162f4c84/k2/python/k2/fsa_algo.py#L66

k2/python/csrc/torch/fsa_algo.cu
csukuangfj: Yes, your approach would work, but it produces less beautiful documentation.

The current implementation produces the following doc

```python
import _k2
help(_k2.linear_fsa)
```

```
linear_fsa(...) method of builtins.PyCapsule instance
    linear_fsa(*args, **kwargs)
    Overloaded function.

    1. linear_fsa(labels: k2::RaggedAny, device: Optional[torch::Device] = None) -> k2::Ragged<k2::Arc>

    2. linear_fsa(labels: List[int], device: Optional[torch::Device] = None) -> k2::Ragged<k2::Arc>

    3. linear_fsa(labels: List[int], device: Optional[str] = None) -> k2::Ragged<k2::Arc>

    4. linear_fsa(labels: List[List[int]], device: Optional[torch::Device] = None) -> k2::Ragged<k2::Arc>

    5. linear_fsa(labels: List[List[int]], device: Optional[str] = None) -> k2::Ragged<k2::Arc>
 ```

pkufool: Cool！Thanks for explanation！

csukuangfj:
Ready for review and merge.

I know it contains lots of changes, though most of them are documentation (several thousands of lines of documentation).

All existing test cases are passed.  No test case is removed.

---

We need to add more tests to `k2/python/tests/ragged_tensor_test.py`. 

Help is wanted (A single test function added to that file is also appreciated)

csukuangfj:
Will add more tutorials to `docs/source/python_tutorials/ragged` when I have time.

k2/python/csrc/torch/doc/any.h
csukuangfj: Thanks, fixed.

danpovey:
Great work!
I looked it over briefly and it looks great.
I think it's OK to merge.

csukuangfj:
Merging.

Will release a new version tomorrow.

The documentation of this pull request can be found at
https://k2-fsa.github.io/k2/python_api/api.html#k2-ragged

It contains usage examples for all most every API.
Screen Shot 2021-09-07 at 8 30 53 PM

PR #819, opened 2021-09-06
Implements ArgMaxPerSubList for device with cub::ArgMax

pkufool:
There is a bug in current implementation of ArmMaxPerSubList which implements with `cub::Reduce`, the error occurs randomly, especially for the very unbalanced distributed data(size of sublists vary a lot). You can easily reproduce the bug with the code below.

```python
import k2
import torch
import random

res = []
# sublists with single element
for i in range(10000):
    res.append([random.random() * -100])
# sublist with huge elements
res.append([random.random() * -100 for x in range(5000)])
# sublists with single element
for i in range(100):
    res.append([random.random() * -100])

ragged_cpu = k2.ragged.create_ragged2(res)
ragged = ragged_cpu.to(torch.device("cuda:0"))

a = k2.ragged.argmax_per_sublist(ragged, -10000).to("cpu")
b = k2.ragged.argmax_per_sublist(ragged_cpu, -10000)

for i in range(len(a)):
    if a[i] != b[i]:
        print (a[i], b[i])
```

In this pull request, we implements the `ArgMaxPerSubList` with `cub::ArgMax`, see [here](https://nvlabs.github.io/cub/structcub_1_1_device_segmented_reduce.html#a2c61f69772a76c383ed8d34c6e5a2194), and we add a checking to confirm whether the results are right. We will see if `cub::ArgMax` works.

### DIFF
--- a/k2/csrc/ragged_ops_inl.h
+++ b/k2/csrc/ragged_ops_inl.h
@@ -602,7 +602,7 @@ void ArgMaxPerSublist(Ragged<T> &src, T initial_value, Array1<int32_t> *dst) {
     // of argmax results, we would recompute the error ones.
 
     Array1<int8_t> flag(c, 1, 0);
-    int8_t* flag_data = flag.Data();
+    int8_t *flag_data = flag.Data();
     K2_EVAL(
         c, src.NumElements(), lambda_check_argmax, (int32_t idx01) -> void {
           int32_t idx0 = row_ids[idx01];
@@ -624,9 +624,9 @@ void ArgMaxPerSublist(Ragged<T> &src, T initial_value, Array1<int32_t> *dst) {
             output_data[idx0] = max_idx;
           }
         });
-    flag_data = flag.To(GetCpuContext()).Data();
-    //flag_data = flag.Data();
-    if (flag_data[0] == 1)
+    flag = flag.To(GetCpuContext());
+    const int8_t *flag_data_c = flag.Data();
+    if (flag_data_c[0] == 1)
       K2_LOG(WARNING) << "There are errors in cub::ArgMax results,"
                       << "we have recomputed it.";
   }
--- a/k2/csrc/ragged_ops_inl.h
+++ b/k2/csrc/ragged_ops_inl.h
@@ -467,7 +467,7 @@ struct PairInputIterator {
   __device__ __forceinline__ PairInputIterator(const T *t, int32_t offset)
       : t_(t), offset_(offset) {}
   __device__ __forceinline__ Pair<T> operator[](int32_t idx) const {
-    return Pair<T>{t_[idx], idx + offset_};
+    return Pair<T>{t_[idx + offset_], idx + offset_};
   }
   __device__ __forceinline__ PairInputIterator operator+(int32_t offset) {
     return PairInputIterator{t_, offset + offset_};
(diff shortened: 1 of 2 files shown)


k2/csrc/ragged_ops_inl.h
```diff
@@ -566,24 +568,67 @@ void ArgMaxPerSublist(Ragged<T> &src, T initial_value, Array1<int32_t> *dst) {
...
+
+    Array1<int8_t> flag(c, 1, 0);
+    int8_t* flag_data = flag.Data();
+    K2_EVAL(
+        c, src.NumElements(), lambda_check_argmax, (int32_t idx01) -> void {
+          int32_t idx0 = row_ids[idx01];
+          if (output_data[idx0] == -1) return;
+          T max_value = values_data[output_data[idx0]];
+          if (values_data[idx01] > max_value) {
+            flag_data[0] = 1;
```
csukuangfj, line 612: Shall we remove `flag` and print the warning inside the `if`?
We can use a guard, i.e.,`idx1 == 0`, to let each sublist print at most one warning.

---

`flag` involves one cpu->gpu, and one gpu->cpu call, no matter whether there is an error in `ArgMax` or not.

k2/csrc/ragged_ops_inl.h
```diff
@@ -566,24 +568,67 @@ void ArgMaxPerSublist(Ragged<T> &src, T initial_value, Array1<int32_t> *dst) {
...
+          if (row_splits[idx0 + 1] == row_splits[idx0])
+            output_data[idx0] = -1;
+          else
+            output_data[idx0] = d_out_data[idx0].key + row_splits[idx0];
+        });
+
+
+    // TODO: Finally, we will remove the checking below.
+    // To check whether the argmax result of cub is right. If there are errors
+    // of argmax results, we would recompute the error ones.
```
csukuangfj, line 602:
```suggestion
    // of argmax results, we would recompute the erroneous ones.
```

k2/csrc/ragged_ops_inl.h
```diff
@@ -566,24 +568,67 @@ void ArgMaxPerSublist(Ragged<T> &src, T initial_value, Array1<int32_t> *dst) {
...
+              T elem = values_data[j];
+              if (elem >= val) {
+                val = elem;
+                max_idx = j;
+              }
+            }
+            output_data[idx0] = max_idx;
+          }
+        });
+    flag_data = flag.To(GetCpuContext()).Data();
```
csukuangfj, line 627:
```cpp
flag.To(GetCpuContext())
```
returns a temporary object and you are referring to a pointer to its member, which
is deallocated after this line.

Please assign a name to `flag.To(GetCpuContext())`

csukuangfj: Also, please define a `const` pointer to `Data()`.

k2/csrc/ragged_ops_inl.h
pkufool: It is hard to print the warning message inside the `if`, if we only allow to print once. We never know which element will go into the `if`, so `idx1==0` won't work.

I don't know the transferring cost between cpu and gpu precisely, I thought there was only one Byte. We will finally remove these code in one month or two, if we won't see any printout warnings.

k2/csrc/ragged_ops_inl.h
pkufool: "the error ones" means the ones with error.

k2/csrc/ragged_ops_inl.h
pkufool: Fixed, thanks.

k2/csrc/ragged_ops_inl.h
csukuangfj: "error" is a noun, not an adjective.

you can use "incorrect ones" or "erroneous ones", but not "error ones".

k2/csrc/ragged_ops_inl.h
csukuangfj: Then can we let the thread with "idx01" corresponding to the one having max value in a sub list print the warning?

Launching a kernel takes more time than that of transferring just one byte, I believe.

pkufool:
We use `cub::Reduce` a lot, it's hard to replace all these functions.  I didn't find the errors for other persublist functions with the unbalanced data. So we only fix `ArgMaxPerSubList` this time.

BTW, There will be little difference between cpu and gpu version of `sum_per_sublist`, I think it is the precision problem.

image

k2/csrc/ragged_ops_inl.h
```diff
@@ -566,24 +568,63 @@ void ArgMaxPerSublist(Ragged<T> &src, T initial_value, Array1<int32_t> *dst) {
     }
   } else {
     K2_CHECK_EQ(c->GetDeviceType(), kCuda);
-    argmax_internal::PairInputIterator<T> input_iter(values_data);
```
csukuangfj, line 569: Shall we remove everything inside the namespace `argmax_internal`?

danpovey: Since it's not used now I guess it's OK to remove it, we can always restore it from the git history if we need.
It still bothers me a little that we haven't found the bug though.

k2/csrc/ragged_ops_inl.h
danpovey: You guys do it how you want.
My two cents, though:
 - If it's an error condition, printing multiple warnings is fine.
 - Transferring a byte from GPU->CPU is actually extremely slow, much slower than launching a kernel, because it forces the GPU to be synced.  A kernel launch just gets queued and control is returned to the CPU without anything necessarily being run at all (at least, not immediately).

k2/csrc/ragged_ops_inl.h
csukuangfj: > It still bothers me a little that we haven't found the bug though.


I just found it!

The following implementation is buggy:
https://github.com/k2-fsa/k2/blob/818b138b33eabe440601df8910a2b97ac088594b/k2/csrc/ragged_ops_inl.h#L469-L471

`t_[idx]` should be changed to `t_[idx + offset_]`

danpovey: Oops !

k2/csrc/ragged_ops_inl.h
```diff
@@ -540,11 +540,13 @@ void ArgMaxPerSublist(Ragged<T> &src, T initial_value, Array1<int32_t> *dst) {
 
   int32_t last_axis = src.NumAxes() - 1;
   const Array1<int32_t> &row_splits_array = src.RowSplits(last_axis);
+  const Array1<int32_t> &row_ids_array = src.RowIds(last_axis);
```
csukuangfj, line 543: Stuff related to row_ids can be moved to the block belonging to the `#if 0` macro.

csukuangfj:
+2

PR #797, opened 2021-08-05
Update intersect_dense.cu

luomingshuang:
When I run the new mmi_bigram_{train, decode}.py, there is a bug such as the following. 
With kangwei's help, I remove some lines code in k2/csrc/intersect_dense.cu and the script can run successfully. 
So I make a PR to update the intersect_dense.cu. I'm not sure if there are other ways to modify it.
```
batch 2170, epoch 1/10 global average objf: 0.242149 over 28408356.0 frames (100.0% kept), current batch average objf: 0.255091 over 12872 frames (100.0% kept) avg time waiting for batch 0.005s
batch 2180, epoch 1/10 global average objf: 0.242089 over 28538822.0 frames (100.0% kept), current batch average objf: 0.192952 over 13173 frames (100.0% kept) avg time waiting for batch 0.005s
[F] /ceph-ly/open-source/k2/k2/csrc/intersect_dense.cu:851:lambda [](signed int)->void::operator()(signed int)->void block:[0,0,0], thread: [41,0,0] Check failed: tot_score_end == tot_score_start || fabs(tot_score_end - tot_score_start) < 1.0 -658531.687500 vs -658530.625000
/ceph-ly/open-source/k2/k2/csrc/intersect_dense.cu:851: lambda [](signed int)->void::operator()(signed int)->void: block: [0,0,0], thread: [41,0,0] Assertion `Some bad things happened` failed.
[F] /ceph-ly/open-source/k2/k2/csrc/array.h:329:T k2::Array1<T>::operator[](int32_t) const [with T = int; int32_t = int] Check failed: ret == cudaSuccess (710 vs. 0)  Error: device-side assert triggered. 


[ Stack-Trace: ]
/ceph-ly/open-source/k2/build/lib/libk2_log.so(k2::internal::GetStackTrace()+0x47) [0x7f4da74943c7]
/ceph-ly/open-source/k2/build/lib/libk2context.so(k2::Array1<int>::operator[](int) const+0xeb9) [0x7f4da77795b9]
/ceph-ly/open-source/k2/build/lib/libk2context.so(k2::Renumbering::ComputeOld2New()+0x14e) [0x7f4da777451e]
/ceph-ly/open-source/k2/build/lib/libk2context.so(k2::Renumbering::ComputeNew2Old()+0x7f8) [0x7f4da7775c98]
[...]

PR #820, opened 2021-09-08
Prune with max_arcs in IntersectDense

### LINKED ISSUE #815: k2.intersect_dense triggers ValueError: cannot create std::vector larger than max_size()
I'm occasionally encountering `ValueError: cannot create std::vector larger than max_size()` when calling `k2.intersect_dense`:
```
File "/opt/conda/lib/python3.8/site-packages/k2-1.6.dev20210825+cuda11.3.torch1.9.0a0-py3.8-linux-x86_64.egg/k2/autograd.py", line 550, in forward
    ragged_arc, arc_map_a, arc_map_b = _k2.intersect_dense(
ValueError: cannot create std::vector larger than max_size()
```
I don't think the error is exactly related to an OOM issue: e.g. I'm getting the ValueError with batch size 32 (occupied gpu memory around 50%) and not getting it with batch size 16 (occupied gpu memory around 30%).
Thus there is a possibility that this behavior is some kind of bug. I wonder if anyone else had this?

pkufool:
Close #815 and #797 .


k2/csrc/array.h
```diff
@@ -454,6 +456,8 @@ class Array1 {
 
   void Init(ContextPtr context, int32_t size, Dtype dtype) {
     K2_CHECK(K2_TYPE_IS_ANY(T) || dtype == DtypeOf<T>::dtype);
+    K2_CHECK_GE(size, 0) << "Array size MUST be great equal to 0, "
```
csukuangfj: “great than or equal to"

k2/csrc/intersect_dense.cu
```diff
@@ -230,6 +229,8 @@ class MultiGraphDenseIntersect {
       score_cutoffs = GetScoreCutoffs();
       score_cutoffs_data = score_cutoffs.Data();
       float **state_scores_data = state_scores_.Data();
+      Array1<int64_t> state_arcs(c_, product);
```
pkufool, line 232: @csukuangfj

PR #828, opened 2021-09-15
Add levenshtein alignment

pkufool:
* k2.levenshtein_graph

image

image

### DIFF
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -454,15 +454,15 @@ FsaVec LinearFsas(const Ragged<int32_t> &symbols) {
 }
 
 FsaVec LevenshteinGraphs(const Ragged<int32_t> &symbols,
-                         float self_loop_weight /* = -1 */,
+                         float penalty /* = -1 */,
                          Array1<int32_t> *aux_labels /*= nullptr*/,
-                         Array1<float> *weight_bias /*= nullptr*/) {
+                         Array1<float> *penalty_bias /*= nullptr*/) {
   NVTX_RANGE(K2_FUNC);
   K2_CHECK_EQ(symbols.NumAxes(), 2);
   ContextPtr &c = symbols.Context();
 
-  // For each fsa, the number of states will be number of `symbols + 2`, we plus
-  // 2 because we need an extra super final arc for each fsa.
+  // For each fsa, the number of states will be number of symbols plus 2, we
+  // plus 2 because we need an extra super final arc for each fsa.
   RaggedShape fsa_to_states = ChangeSublistSize(symbols.shape, 2);
 
   int32_t num_states = fsa_to_states.NumElements();
@@ -481,7 +481,7 @@ FsaVec LevenshteinGraphs(const Ragged<int32_t> &symbols,
                                        // the next state.
         if (state_idx01 == final_state - 1)
           current_num_arcs = 2;
-        if (state_idx01 == final_state)
+        else if (state_idx01 == final_state)
           current_num_arcs = 0;
         num_arcs_for_data[state_idx01] = current_num_arcs;
       });
@@ -506,10 +506,10 @@ FsaVec LevenshteinGraphs(const Ragged<int32_t> &symbols,
     *aux_labels = Array1<int32_t>(c, num_arcs);
     aux_labels_data = aux_labels->Data();
   }
-  float *weight_bias_data = nullptr;
-  if (weight_bias != nullptr) {
-    *weight_bias = Array1<float>(c, num_arcs);
-    weight_bias_data = weight_bias->Data();
+  float *penalty_bias_data = nullptr;
+  if (penalty_bias != nullptr) {
+    *penalty_bias = Array1<float>(c, num_arcs);
+    penalty_bias_data = penalty_bias->Data();
   }
 
   K2_EVAL(
@@ -525,7 +525,10 @@ FsaVec LevenshteinGraphs(const Ragged<int32_t> &symbols,
                 current_symbol = symbols_data[sym_state_idx01],
                 aux_labels_value = 0;
 
-        float weight_bias_value = 0;
+        K2_CHECK((current_symbol != 0) && (current_symbol != -1))
+          << "0 and -1 are not expected to be a symbol.";
+
+        float penalty_bias_value = 0;
         Arc arc;
         arc.src_state = state_idx1;
 
@@ -533,10 +536,10 @@ FsaVec LevenshteinGraphs(const Ragged<int32_t> &symbols,
           case 0:  // the self loop arc
             arc.label = 0;
             arc.dest_state = state_idx1;
-            arc.score = self_loop_weight;
+            arc.score = penalty;
             aux_labels_value = 0;
-            // Actually, it is self_loop_weight - (-1)
-            weight_bias_value = self_loop_weight + 1;
+            // Actually, it is penalty - (-1)
+            penalty_bias_value = penalty + 1;
             break;
           case 1:   // the arc pointing to next state with blank
             if (state_idx01 == final_state_idx01 - 1) {  // the arc pointing to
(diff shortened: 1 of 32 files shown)


pkufool:
When `need_weight_bias == True`:

image

k2/python/k2/fsa_algo.py
```diff
@@ -1069,3 +1072,58 @@ def ctc_topo(max_token: int,
     ragged_arc, aux_labels = _k2.ctc_topo(max_token, device, modified)
     fsa = Fsa(ragged_arc, aux_labels=aux_labels)
     return fsa
+
+def levenshtein_graph(symbols: Union[k2.RaggedTensor, List[List[int]]],
+        self_loop_weight: float = -1,
+        need_weight_bias: bool = False,
```
csukuangfj, line 1078: Maybe we need a better name for the argument `need_weight_bias`?

Also, I would recommend converting the returned  `weight_bias` into an attribute of the resulting FSA.

k2/python/k2/fsa_algo.py
```diff
@@ -1069,3 +1072,58 @@ def ctc_topo(max_token: int,
     ragged_arc, aux_labels = _k2.ctc_topo(max_token, device, modified)
     fsa = Fsa(ragged_arc, aux_labels=aux_labels)
     return fsa
+
+def levenshtein_graph(symbols: Union[k2.RaggedTensor, List[List[int]]],
+        self_loop_weight: float = -1,
```
pkufool, line 1077: @danpovey What would you suggest to name this argument, `self_loop_weight` may not be a good one.

csukuangfj: It represents insertion and deletion cost. How about calling it "ins_del_cost"?

k2/python/k2/fsa_algo.py
```diff
@@ -1022,6 +1022,9 @@ def ctc_graph(symbols: Union[List[List[int]], k2.RaggedTensor],
             [it for symbol in symbols for it in symbol],
             dtype=torch.int32,
             device=device)
+        symbols = k2.RaggedTensor(symbols)
```
csukuangfj, line 1025: This `if..else` can be simplified to
```python3
if isinstance(symbols, k2.RaggedTensor):
  assert device is None
else:
  symbols = k2.RaggedTensor(symbols, device=device)

symbol_values = symbols.values

```

k2/python/k2/fsa_algo.py
```diff
@@ -1069,3 +1072,58 @@ def ctc_topo(max_token: int,
...
+        if `symbols` is an instance of :class:`k2.RaggedTensor`, the returned
+        FSA will on the same device as `k2.RaggedTensor`.
+
+    Returns:
+        An FsaVec containing the returned levenshtein graphs, with "Dim0()"
+        the same as "len(symbols)"(List[List[int]]) or "dim0"(k2.RaggedTensor).
+        If `need_weight_bias` set to True, a tensor with weight bias on each arc
+        will be returned as well.
+    '''
+    if isinstance(symbols, k2.RaggedTensor):
```
csukuangfj, line 1114:
```suggestion
    if isinstance(symbols, k2.RaggedTensor):
      assert device is None
    else:
      symbols = k2.RaggedTensor(symbols, device=device)
```

k2/python/k2/fsa_algo.py
```diff
@@ -1069,3 +1072,58 @@ def ctc_topo(max_token: int,
...
+            - A list of list-of-integers, e..g, `[ [1, 2], [1, 2, 3] ]`
+            - An instance of :class:`k2.RaggedTensor`.
+              Must have `num_axes == 2`.
+
+      self_loop_weight:
+        The weight on the self loops arcs in the graphs, the main idea of this
+        weight is to set insertion and deletion penalty, which will affect the
+        shortest path searching produre.
+      need_weight_bias:
+        If set to True, a tensor with the bias to -1 on each arc will be
```
csukuangfj, line 1099: The documentation for `need_weight_bias` is not easy to understand.

k2/python/k2/fsa_algo.py
pkufool: How about naming it `penalty`, and the fsa's attribute named `penalty_bias`.

k2/csrc/fsa_algo.cu
```diff
@@ -453,9 +453,130 @@ FsaVec LinearFsas(const Ragged<int32_t> &symbols) {
       arcs);
 }
 
+FsaVec LevenshteinGraphs(const Ragged<int32_t> &symbols,
+                         float penalty /* = -1 */,
```
danpovey, line 457: I think ins_del_score is a good name, since it explains what it is.
I'm surprised the default is -1 though, instead of -0.5.  I thought we were going to multiply all scores by 0.5, so that the total cost equals the Levenshtein alignment score.

k2/python/csrc/torch/v2/ragged_any.cu
```diff
@@ -200,6 +203,7 @@ RaggedAny::RaggedAny(const std::string &s, py::object dtype /*=py::none()*/,
   }
 
   ContextPtr context = GetContext(device);
+  DeviceGuard guard(context);
```
pkufool, line 206: @csukuangfj  I think we need a device guard here.

pkufool:
`Intersect_device` like interface  to calculate levenshtein distance and corresponding alighments.


image

k2/python/k2/fsa_algo.py
```diff
@@ -1069,3 +1058,110 @@ def ctc_topo(max_token: int,
...
+    See https://github.com/k2-fsa/k2/pull/828 for more details about levenshtein
+    graph.
+
+    Args:
+      symbols:
+        It can be one of the following types:
+
+            - A list of list-of-integers, e..g, `[ [1, 2], [1, 2, 3] ]`
+            - An instance of :class:`k2.RaggedTensor`.
+              Must have `num_axes == 2`.
```
csukuangfj, line 1080: Also with dtype `torch.int32`?

k2/python/k2/fsa_algo.py
```diff
@@ -1069,3 +1058,110 @@ def ctc_topo(max_token: int,
...
+        the same as "len(symbols)"(List[List[int]]) or "dim0"(k2.RaggedTensor).
+    '''
+    if not isinstance(symbols, k2.RaggedTensor):
+        symbols = k2.RaggedTensor(symbols, device=device)
+
+    ragged_arc, aux_labels, score_offsets = _k2.levenshtein_graph(
+        symbols, ins_del_score, True)
+    fsa = Fsa(ragged_arc, aux_labels=aux_labels)
+    if ins_del_score_offset_attr is not None:
+        fsa.__setattr__(name=ins_del_score_offset_attr, value=score_offsets)
```
csukuangfj, line 1108:
```suggestion
        setattr(fsa, ins_del_score_offset_attr, score_offsets)
```

k2/python/k2/fsa_algo.py
```diff
@@ -1069,3 +1058,110 @@ def ctc_topo(max_token: int,
...
+    lattice = k2.remove_epsilon_self_loops(lattice)
+
+    alignment = k2.shortest_path(lattice, use_double_scores=True).invert_()
+    alignment.rename_tensor_attribute_("aux_labels_levenshtein", "aux_labels")
+
+    alignment.scores -= alignment.ins_del_score_offset
+    distance = -alignment.get_tot_scores(
+        use_double_scores=False, log_semiring=False)
+
+    return alignment, distance
```
csukuangfj, line 1167: `distance` is already contained in `alignment`.
Can we rename this function to `levenshtein_alignment` and just
return the `alignment`?
Users can get the actual alignment and distance from the return value.

Also, `levenshtein_distance` returns both the alignment and distance. I
was expecting that it returns only the distance.

csukuangfj:
+2

PR #829, opened 2021-09-16
Trigger GitHub actions manually.

csukuangfj:


csukuangfj:
This commit changes how GitHub actions is triggered.
Instead of running GitHub actions on every commit, we run it only when we add a label `ready`.

Issue #822, opened 2021-09-09
Disable support for torch 1.3.1


csukuangfj:
As we are doing autograd in C++ and because of the change introduced in https://github.com/pytorch/pytorch/pull/28287
the current code won't compile for torch 1.3.1 and below.

You can see that the CI starts to fail for torch 1.3.1:
https://github.com/k2-fsa/k2/actions/runs/1213679119

Shall we disable support for torch 1.3.1?

PR #831, opened 2021-09-17
Support printing ragged tensors in a more compact way.

csukuangfj:
Screen Shot 2021-09-17 at 11 43 51 AM

If a ragged tensor has many sublists, it is useful to print them in a more compact way;
otherwise, it does not fit into one screen.


csukuangfj:
Also, it disables support for `torch 1.3.1`. 

Closes #822

PR #832, opened 2021-09-24
fix: make python setup.py works for development mode

songmeixu:
namely, install by bellow cmds works:
`python setup.py develop` or `pip install -e .`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -207,6 +207,7 @@ setuptools.setup(
     long_description_content_type='text/markdown',
     url='https://github.com/k2-fsa/k2',
     package_dir={
+        '': 'k2/python',
         'k2': 'k2/python/k2',
         'k2.ragged': 'k2/python/k2/ragged',
         'k2.sparse': 'k2/python/k2/sparse',


csukuangfj:
Thanks for the contribution.

`setup.py` is mainly used to build wheels that can be installed by normal users.

As developers, I would recommend you to read https://k2-fsa.github.io/k2/installation/for_developers.html, which does not use `setup.py` at all.

songmeixu:
Yeah, I read that doc. But python develop mode is a common routine for developers, e.g. it requires nothing to do after you change python sources. And if c++ sources change, just pip install -e ., instead of firstly make and others. And u don't need an env shell to source.

danpovey:
I think it would be nice if setup.py were to support development mode, but I don't understand how this change accomplishes that.  Is the empty string supposed to be a key for that `package_dir` arg?  I'd like to understand what is going on.

songmeixu:
> I think it would be nice if setup.py were to support development mode, but I don't understand how this change accomplishes that. Is the empty string supposed to be a key for that `package_dir` arg? I'd like to understand what is going on.

I think the old code should work,
according to https://docs.python.org/3/distutils/setupscript.html#listing-whole-packages

But not, I think there's a bug of setuptool pkg, and make a PR: https://github.com/pypa/setuptools/pull/2803.

And we could wait for their answers, it's not critical for us anyway.

PR #839, opened 2021-10-03
WIP: Support torchscript.

csukuangfj:


k2/torch/bin/decode.cc
```diff
@@ -0,0 +1,89 @@
+#include <cassert>
+#include <cstdio>
```
danpovey, line 2: It would be nice if it were  more obvious from the binary itself what it does and how to use it.. or at least a pointer to where to find this info.

danpovey: mm, I see it has a usage message; is it accurate? (./bin/decode.py)... and what does this binary do?
I'm just wondering what kind of thing a jit file can contain, is it a nnet evaulation?  With fixed size?

csukuangfj: This is just a prototype. I will add more documentation later.

The intention of this file is to replicate what we are doing in `decode.py` in icefall.

k2/torch/bin/decode.cc
```diff
@@ -0,0 +1,89 @@
...
+
+  return torch::jit::pickle_load(f);
+}
+
+int main(int argc, char *argv[]) {
+  at::set_num_threads(1);
+  at::set_num_interop_threads(1);
+
+  std::string usage = R"(
+    ./bin/decode.py \
```
csukuangfj, line 39: Oh, a typo here. It should be `./bin/decode`

If you clone `k2` and do
```
mkdir build
cd build
cmake ..
make decode
```
you will find a binary in `./bin/decode`

k2/torch/bin/decode.cc
csukuangfj: > I'm just wondering what kind of thing a jit file can contain, is it a nnet evaulation? With fixed size?

As far as I know,  the torch scripted model behaves similarly to `torch.nn.Module`. You can vary the input shape, i.e.,
the `N` and `T` in `(N, T, C)`.

Checkpoints just save the parameters of a model, without topology information.
If you want to recover from some checkpoint, you have to first build a `nn.Module` through some Python code, 
which contains the definition of the model.

With a torch scripted model, all you need is a binary file ( I don't know what exactly is saved in it), load
it in C++, feed some input into it, and get the output for later processing.

pzelasko: the jit file has model weights and the model code translated into TorchScript intermediate representation (IR). AFAIK it doesn’t speed up the inference in any way unless the IR code is further optimized; e.g. it can be translated/exported to ONNX. 

Also I think it is possible to extend both TorchScript and ONNX with custom ops (by building shared library extensions: https://pytorch.org/tutorials/advanced/torch_script_custom_ops) — maybe it makes sense to build k2 custom ops and involve LM + decoding as part of the JIT model for deployment packaging, I guess a lot of people would find it very convenient (same could be done for kaldifeat for a single-file model that outputs transcripts directly). Some food for thought.

danpovey:
Great!!

k2/torch/bin/decode.cc
pzelasko: Oh, that’s too bad. I guess it’s possible to convert an FSA to a bunch of tensors and reconstruct it inside but it’s probably too much hassle…

danpovey:
OK, so if I understand correctly, what is happening here is that you are using TorchScript to encode the model itself, then just coding the FSA-related parts in C++ directly?

And I believe the current status of this is that, since we haven't finished creating a C++ version of the Fsa object that we use in Python, things like automatic handling of attributes will not be happening right now, and these have to be coded directly?

BTW, it still does worry me a little, that the attribute handling will become much more opaque once we convert it to C++.  But I also don't see another way to easily enable C++ deployment.

csukuangfj:
The purpose of this pull-request is to show that we can import a pre-trained neural network model from PyTorch and
use it with k2 in C++.

As for the attributes, this pull-request only handles `aux_labels`.

---

We will use `k2::RaggedArc` later, which handles attribute propagations automatically.
See https://github.com/k2-fsa/k2/blob/v2.0-pre/k2/python/csrc/torch/v2/ragged_arc.h

>  the attribute handling will become much more opaque once we convert it to C++

I don't think it is opaque. We can use it just like what we are doing in Python, i.e., `c_fsa = a_fsa.IntersectDense(b_fsa)`
will handle the attributes of `c_fsa` automatically.

`k2::RaggedArc` currently has Python dependencies so I am not using it in this pull-request.

Will remove Python dependencies from `k2::RaggedArc` later.

danpovey:
Ah I see, cool.
I am wondering whether a different name might be better, than RaggedArc.  Perhaps FsaClass?
Another possibility would be to rename Fsa,FsaVec,FsaOrVec to RawFsa, RawFsaVec, RawFsaOrVec, and rename RaggedArc to Fsa.  That would be long-term a better long-term solution, IMO, because the names would match those in Python, but would likely require more messing about with documentation and so on.

pkufool:
> Will remove Python dependencies from `k2::RaggedArc` later.

@csukuangfj  Have you started this stuff ?  If not, I want to have a try. The plan is to replace the `pyobject`  with `torch::IValue`.

csukuangfj:
> > Will remove Python dependencies from `k2::RaggedArc` later.
> 
> @csukuangfj Have you started this stuff ? If not, I want to have a try. The plan is to replace the `pyobject` with `torch::IValue`.

Sure, you can go ahead.

csukuangfj:
Now it supports both CTC decoding and HLG decoding using pre-computed features.

Will implement decoding directly by reading sound files using https://github.com/csukuangfj/kaldifeat

csukuangfj:
Because of the issue mentioned in https://github.com/pytorch/pytorch/issues/47493,
`k2/torch/csrc/deserialization.cu` causes compilation errors with NVCC when using torch 1.7.1

That issue has been fixed in https://github.com/pytorch/pytorch/pull/47492/files, which is
only available from torch 1.8.0.

I am afraid we have to use at least torch 1.8.0 for deployment.

csukuangfj:
Ready for review.

PR #846, opened 2021-10-14
import torch first in the smoke tests to prevent SEGFAULT

jtrmal:
This took me some time to figure out -- I'm not sure if there is a way how to prevent the segfault completely
```
(py3) [jtrmal@test4 k2]$ python3 -c "import torch; import _k2; print(_k2.__file__)"
/home/hltcoe/jtrmal/k2/build_release/lib/_k2.cpython-39-x86_64-linux-gnu.so
(py3) [jtrmal@test4 k2]$ python3 -c "import _k2; print(_k2.__file__)"
Segmentation fault
(py3) [jtrmal@test4 k2]$```

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -78,7 +78,7 @@ To build a release version, use:
   python3 -c "import k2; print(k2.__file__)"
   # It should print /some/path/k2/k2/python/k2/__init.py
 
-  python3 -c "import _k2; print(_k2.__file__)"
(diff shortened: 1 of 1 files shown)


lumaku:
I can reproduce this issue.
My best guess about this error is:
If you directly `import _k2`, there is no torch device specified. The device pointer is zero, any call to `THPDevice_New` will fail with segfault. That would explain why it works with `import torch`, as this sets a device.

PR #856, opened 2021-10-30
Use `torch::class_` to register custom classes.

csukuangfj:
I figured out how to modify the qualified name of a registered custom class, so `torch::class_` is used here directly.


danpovey:
clever!

PR #858, opened 2021-11-02
Implements Cpp version FsaClass

pkufool:
In this pull request, I implements a basic C++ version `FsaClass` supporting propagating attributes.  It only depends k2 core code and torch library.

To avoid making a too big pull request, will apply this `FsaClass` for decoding in another PR.


k2/torch/csrc/fsa_class.cu
```diff
@@ -0,0 +1,476 @@
...
+#include "k2/csrc/fsa_algo.h"
+#include "k2/csrc/fsa_utils.h"
+#include "k2/csrc/ragged_ops.h"
+#include "k2/torch/csrc/deserialization.h"
+#include "k2/torch/csrc/fsa_class.h"
+#include "k2/torch/csrc/utils.h"
+
+namespace k2 {
+
+FsaClass::FsaClass(const std::string &s,
```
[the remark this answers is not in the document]
pkufool: I copied this constructor for testing purpose, will clear the code and add some docs here.

k2/torch/csrc/fsa_class.cu
```diff
@@ -0,0 +1,476 @@
...
+      Ragged<int32_t> new_value =
+          Index<int32_t>(iter.second, arc_map, /*remove_axis*/ true);
+      SetRaggedTensorAttr(iter.first, new_value);
+    }
+  }
+}
+
+void FsaClass::CopyAttrs(std::vector<FsaClass> &srcs) {
+  // copy tensor attributes
+  std::unordered_map<std::string, int> tensor_attrs;
```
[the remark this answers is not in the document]
pkufool: Thanks, I didn't notice this.

k2/torch/csrc/fsa_class.cu
```diff
@@ -0,0 +1,476 @@
...
+FsaClass FsaClass::FromUnaryFunctionRagged(FsaClass &src,
+                                           const Ragged<Arc> &arcs,
+                                           Ragged<int32_t> &arc_map) {
+  FsaClass dest(arcs);
+  // Check the validation of the fsa, will trigger a fatal error if the fsa
+  // is not valid.
+  dest.Properties();
+  for (const auto &iter : src.tensor_attrs) {
+    if (iter.second.scalar_type() == torch::kInt32) {
+      torch::Tensor value = iter.second.clone();
```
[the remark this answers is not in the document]
pkufool: There is no need, fixed.

k2/torch/csrc/deserialization.h
```diff
@@ -22,9 +22,27 @@
...
+// TODO(fangjun): Make it a template
+struct RaggedIntHelper : public Ragged<int32_t>,
+                         public torch::CustomClassHolder {
+  using k2::Ragged<int32_t>::Ragged;
+  RaggedIntHelper(Ragged<int32_t> ragged) : Ragged<int32_t>(ragged) {}
+};
+
+struct RaggedRegister {
+  RaggedRegister() {
+    static std::once_flag register_ragged_int_flag;
```
csukuangfj, line 41: The `once_flag` and `call_once` are not needed if this class is moved to `.cu` file.

If it is invoked multiple times, it will throw in the second invocation.

k2/torch/csrc/deserialization.h
```diff
@@ -22,9 +22,27 @@
 #include <string>
 
 #include "k2/csrc/fsa.h"
+#include "k2/torch/csrc/utils.h"
 
 namespace k2 {
 
+void RegisterRaggedInt();
```
csukuangfj, line 29: This function and the following class `RaggedIntHelper` can be moved to `.cu` file.

Users don't need to know about them.

k2/torch/csrc/fsa_class.cu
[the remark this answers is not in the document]
csukuangfj: Please write a helper function in the test code to do that.

k2/torch/csrc/fsa_class.cu
```diff
@@ -0,0 +1,476 @@
...
+      auto masking =
```
csukuangfj, line 107: For decoding, we don't need to check for `-1` in the arc map, as that will never happen.

k2/torch/csrc/fsa_class.cu
```diff
@@ -0,0 +1,476 @@
...
+        SumPerSublist<T>(s, 0, &ans_array);
+        dest.SetTensorAttr(iter.first, Array1ToTorch(ans_array));
+      });
+    }
+  }
+  dest.CopyRaggedTensorAttrs(src, arc_map);
+  return dest;
+}
+
+FsaClass FsaClass::FromBinaryFunctionTensor(FsaClass &a_src, FsaClass &b_src,
```
csukuangfj, line 134: I think this function will not be used in the decoding.
We can add it later if it is needed.

Just want to make the decoding code as simple as possible.
Please add code that is only needed by decoding.

k2/torch/csrc/fsa_class.cu
```diff
@@ -0,0 +1,476 @@
...
+    }
+  }
+  return properties;
+}
+
+std::string FsaClass::PropertiesStr() /*const*/ {
+  return FsaPropertiesAsString(Properties());
+}
+
+torch::Tensor FsaClass::Arcs() {
```
csukuangfj, line 287: Do we need this function in decoding, and other functions like `PropertiesStr()`, `Labels()`?

k2/torch/csrc/fsa_class.cu
```diff
@@ -0,0 +1,476 @@
...
+
+torch::Tensor FsaClass::Labels() /*const*/ { return Arcs().index({"...", 2}); }
+
+void FsaClass::SetLabels(torch::Tensor labels) {
+  K2_CHECK_EQ(labels.numel(), fsa.NumElements());
+  K2_CHECK_EQ(labels.scalar_type(), torch::kInt32);
+  Labels().copy_(labels);
+}
+
+FsaClass FsaClass::ToOtherContext(const ContextPtr &context) const {
```
csukuangfj, line 310: Also, I am not sure whether this function and the following `To` will be needed.
We can always construct an FSA on a given device and use it to construct a
`FsaClass`. After that, we don't need to move it to another device during decoding, I think.

csukuangfj (changes requested):

k2/torch/csrc/fsa_class.cu
```diff
@@ -0,0 +1,476 @@
...
+    return;
+  }
+
+  if (IsRaggedInt(value)) {
+    SetRaggedTensorAttr(name, ToRaggedInt(value));
+    return;
+  }
+
+  K2_LOG(FATAL) << "Attribute type is not supported, name : " << name;
+  return;
```
csukuangfj, line 402: Please delete the trailing `return`. It is not needed.

k2/torch/csrc/fsa_class.cu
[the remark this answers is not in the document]
csukuangfj: After reading the test code, I find that you actually don't need this function in the test code.

You can always use a string to construct a `k2::Fsa` and use the following constructor
```
explicit FsaClass(const Ragged<Arc> &fsa)
```

By the way, please use `Fsa` to replace `Ragged<Arc>`.

csukuangfj (changes requested):

k2/torch/csrc/deserialization.cu
```diff
@@ -37,6 +37,8 @@
 
 namespace k2 {
 
+RaggedRegister ragged_register;
```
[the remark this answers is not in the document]
csukuangfj: Also, please add a comment here explaining why we define a global variable here.
Others may don't know the underlying reason.

k2/torch/csrc/deserialization.h
pkufool: Already, I will commit later.

k2/torch/csrc/fsa_class.cu
pkufool: We don't need this function in `Intersect`?

pkufool:
All right then, will remove all of code not needed currently, let's add then later if necessary.

k2/torch/csrc/fsa_class.cu
csukuangfj: No, I don't think we need `k2::Intersect` in decoding.

csukuangfj:
> All right then, will remove all of code not needed currently, let's add then later if necessary.

Thanks! Let us keep only the code that is actually needed at present.

pkufool:
I keep the `FromBinaryFunctionTensor` function, will need it to do rescoring later. I think it won't do any harm to keep it.

pkufool:
Shall we merge this PR? @csukuangfj

csukuangfj:
> Shall we merge this PR? @csukuangfj

Please resolve all the comments first.

k2/torch/csrc/deserialization.h
```diff
@@ -22,9 +22,18 @@
...
+#include "k2/torch/csrc/utils.h"
 
 namespace k2 {
 
+// A helper class to construct a Ragged<int32_t> from an archive
+// TODO(fangjun): Make it a template
+struct RaggedIntHelper : public Ragged<int32_t>,
+                         public torch::CustomClassHolder {
+  using k2::Ragged<int32_t>::Ragged;
+  explicit RaggedIntHelper(Ragged<int32_t> ragged) : Ragged<int32_t>(ragged) {}
```
csukuangfj, line 34:
```suggestion
  explicit RaggedIntHelper(const Ragged<int32_t> &ragged) : Ragged<int32_t>(ragged) {}
```

k2/torch/csrc/deserialization.h
```diff
@@ -22,9 +22,18 @@
 #include <string>
 
 #include "k2/csrc/fsa.h"
+#include "k2/torch/csrc/utils.h"
```
csukuangfj, line 25: Why do we need to include this header in this file? Where is it used?

pkufool: It may need by the code I previous commited,  can remove now.

PR #863, opened 2021-11-06
Refactor deserialization code

csukuangfj:
Support to load files saved in Python by `torch.save()` in C++,  especially for `torch.save(fsa.as_dict(), 'xxx.pt')`.

Also, add tests for deserialization code.


csukuangfj:
Note: Because of the issue (https://github.com/pytorch/pytorch/issues/67902), when a dict of tensors is saved in
`xxx.pt` and we want to load it to C++, the dict has at least two elements. Otherwise, it will throw.

k2/torch/csrc/deserialization.cu
```diff
@@ -368,41 +370,51 @@ k2::FsaClass LoadFsa(const std::string &filename) {
...
   torch::IValue ivalue = torch::jit::readArchiveAndTensors(
       "data", "", "", type_resolver, obj_loader,
-      /*device=*/c10::nullopt, *reader);
+      /*device=*/map_location, *reader);
 
 #else
   torch::IValue ivalue =
       torch::jit::readArchiveAndTensors("data", type_resolver, obj_loader,
-                                        /*device=*/c10::nullopt, *reader);
+                                        /*device=*/map_location, *reader);
```
pkufool, line 378: Since we can load file to a particular device, we indeed no need to implement `To(device)` for FsaClass. You could delete them.

k2/torch/csrc/fsa_class.h
```diff
@@ -124,6 +124,14 @@ struct FsaClass {
    */
   torch::IValue GetAttr(const std::string &name) /*const*/;
 
+  torch::Tensor GetTensorAttr(const std::string &name) const {
```
pkufool, line 127: I think we'd better not expose the implementation details(i.e. tensor, raggedtensor).

csukuangfj: Different from training, there are only one or two attributes per FSA during decoding. I am not sure whether it's a good way to use a dict to save the attributes and to use IValue to add one more indirection.

PR #866, opened 2021-11-08
fix evaluating kFsaPropertiesValid

pingfengluo:
maybe "properties & kFsaPropertiesValid != kFsaPropertiesValid" is wrong? referring C++ Operator Precedence


csukuangfj:
Thanks!

PR #867, opened 2021-11-08
Support rescoring with an n-gram LM during decoding

csukuangfj:
Usage
```
./bin/decode \
  --g ./G.pt \
  --use_gpu true \
  --method ngram-rescoring \
  --jit_pt ./cpu_jit.pt \
  --hlg ./HLG.pt \
   --word_table ./words.txt \
  ./1089-134686-0000.wav \
  ./1089-134686-0001.wav \
  ./1089-134686-0002.wav
```

Caution: You have to make the following modifications to `G.pt` (the one directly from icefall):
```python3
import torch
g = torch.load('G.pt')
# Add another entry to g
g['dummy'] = 1
torch.save(g, 'G.pt')
```

By default, `G.pt` is a dict containing only one tensor. But it throws because of the issue described in
https://github.com/pytorch/pytorch/issues/67902

We can add a dummy entry to it to make its size larger than 1 and then everything works fine.


csukuangfj:
Refer to the documentation https://icefall.readthedocs.io/en/latest/recipes/librispeech/conformer_ctc.html#deployment-with-c
for how to generate the required files, e.g., `cpu_jit.pt`.

csukuangfj:
I am merging it. If there are any issues, I will resolve them in later commits.

PR #870, opened 2021-11-08
Add attention rescoring.

csukuangfj:
~Still WIP.~


csukuangfj:
Now all the decoding methods that are used in Python can also be used in C++ (without python dependencies), namely
- decoding with a CTC topo
- decoding with an HLG graph
- decoding with an HLG graph + n-gram LM rescoring (i.e, whole-lattice-rescoring in Python)
- decoding with an HLG graph + n-gram LM rescoring + attention decoder rescoring (i.e., attention-decoder in Python)

You can find a pre-trained model at https://huggingface.co/csukuangfj/icefall-asr-librispeech-conformer-ctc-jit-bpe-500-2021-11-09

It also contains instructions for how to use it.

Will add a colab notebook for illustration.

---

**Caution**: It does **not** support streaming decoding yet.

k2/torch/csrc/nbest.cu
```diff
@@ -0,0 +1,118 @@
...
+
+  fsa.Scores().zero_();  // Just in case it has scores set
+
+  K2_CHECK(lattice->HasTensorAttr("aux_labels") ||
+           lattice->HasRaggedTensorAttr("aux_labels"));
+
+  // We don't need the aux labels for this->fsa,
+  // as we are going to use the one from lattice.
+  Fsa word_fsa_with_epsilon_self_loops;
+  RemoveEpsilonAndAddSelfLoops(fsa.fsa, fsa.Properties(),
```
pkufool, line 59: I think it is a little confusing to use both raw Fsa and FsaClass, we'd better wrap the algorithms to FsaClass for consistency. Also the IntersectDevice in line 71.

csukuangfj: Thanks. I've changed `IntersectDevice`.

But for `RemoveEpsilonAndAddSelfLoops`, it does not output 1-d array arc maps and I don't want to make things complicated. We don't need to handle attribute propagation for `RemoveEpsilonAndAddSelfLoops` in decoding. I think it is clear from the calling arguments whether it is a `k2::Fsa` or `k2::FsaClass`.

pkufool: Em.. either is ok for me, up to you.

PR #872, opened 2021-11-11
[Not for merge] Add full-chunk mode CTC decoding for models from WeNet

csukuangfj:
This PR shows that we can also use k2 for decoding with models from other frameworks.

For n-gram LM rescoring and attention decoder rescoring, it shares a lot of code with the following files and is very easy to implement.
- https://github.com/k2-fsa/k2/blob/v2.0-pre/k2/torch/bin/hlg_decode.cu
- https://github.com/k2-fsa/k2/blob/v2.0-pre/k2/torch/bin/ngram_lm_rescore.cu
- https://github.com/k2-fsa/k2/blob/v2.0-pre/k2/torch/bin/attention_rescore.cu

### DIFF
--- a/k2/torch/CMakeLists.txt
+++ b/k2/torch/CMakeLists.txt
@@ -1,3 +1,3 @@
 add_subdirectory(csrc)
-
 add_subdirectory(bin)
+add_subdirectory(sp)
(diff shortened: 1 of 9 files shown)


pzelasko:
Just for clarification: does "full-chunk decoding" mean "offline decoding"?

csukuangfj:
> Just for clarification: does "full-chunk decoding" mean "offline decoding"?

Yes, you are right.  The name [full-chunk decoding](https://github.com/wenet-e2e/wenet/blob/main/wenet/bin/recognize.py#L75) is from WeNet, which is misleading at the first sight.
I think it is equivalent to `full utterance`.

PR #882, opened 2021-11-30
[Not for merge] Out-of-range bug of FSA or RaggedTensor index representation

drawfish:
Just to point out that when constructing large FST graph or RaggedTensor, there will be numerical out-of-range by using int32_t for index or offset representation in k2 frameworks.

One of the impact of this bug is that the framework can not be applied to large-scale speech recognition decoding graph construction, (example can be found [here](https://github.com/k2-fsa/icefall/issues/132)).

The commits in this PR is only to fix a small bug in Tensor construction. However, there are still a lot of modifications related to index representation that need to be made in the future.

### DIFF
--- a/k2/csrc/dtype.h
+++ b/k2/csrc/dtype.h
@@ -277,6 +277,38 @@ struct DtypeOf {
     }                                                                    \
   } while (0)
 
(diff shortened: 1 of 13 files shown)


csukuangfj:
Thanks

pkufool:
Merged here. https://github.com/k2-fsa/k2/pull/892

PR #891, opened 2021-12-13
Add rnnt loss

pkufool:
- [X] Add mutual information
- [x] Add rnnt loss
- [x] Add unit tests
- [x] Update documents and fix code style

### DIFF
--- a/k2/python/csrc/torch/mutual_information.h
+++ b/k2/python/csrc/torch/mutual_information.h
@@ -22,7 +22,10 @@
 #define K2_PYTHON_CSRC_TORCH_MUTUAL_INFORMATION_H_
 
 #include <torch/extension.h>
+
 #include <vector>
+
+#include "k2/python/csrc/torch.h"
 
 namespace k2 {
 /*
@@ -70,10 +73,17 @@ namespace k2 {
    The block-dim and grid-dim must both be 1-dimensional, and the block-dim must
    be at least 128.
 */
-torch::Tensor mutual_information(torch::Tensor px,  // [B][S][T+1]
-                                      torch::Tensor py,  // [B][S+1][T]
-                                      torch::Tensor boundary,  // [B][4], int64_t.
-                                      torch::Tensor p);  //  [B][S+1][T+1]; an output
+torch::Tensor MutualInformationCpu(
+    torch::Tensor px,        // [B][S][T+1]
+    torch::Tensor py,        // [B][S+1][T]
+    torch::Tensor boundary,  // [B][4], int64_t.
+    torch::Tensor p);        //  [B][S+1][T+1]; an output
+
+torch::Tensor MutualInformationCuda(
+    torch::Tensor px,        // [B][S][T+1]
+    torch::Tensor py,        // [B][S+1][T]
+    torch::Tensor boundary,  // [B][4], int64_t.
+    torch::Tensor p);        //  [B][S+1][T+1]; an output
 
 /*
   backward of mutual_information; returns (grad_px, grad_py)
@@ -83,14 +93,18 @@ torch::Tensor mutual_information(torch::Tensor px,  // [B][S][T+1]
   very close to the value of ans_grad at entry.  This can be used
   to validate the correctness of this code.
 */
-std::vector<torch::Tensor> mutual_information_backward(
-    torch::Tensor px,
-    torch::Tensor py,
-    torch::Tensor boundary,
-    torch::Tensor p,
-    torch::Tensor ans_grad,
-    bool overwrite_ans_grad);
+std::vector<torch::Tensor> MutualInformationBackwardCpu(torch::Tensor px,
+                                                        torch::Tensor py,
+                                                        torch::Tensor boundary,
+                                                        torch::Tensor p,
+                                                        torch::Tensor ans_grad);
+
+std::vector<torch::Tensor> MutualInformationBackwardCuda(
+    torch::Tensor px, torch::Tensor py, torch::Tensor boundary, torch::Tensor p,
+    torch::Tensor ans_grad, bool overwrite_ans_grad);
 
 }  // namespace k2
+
+void PybindMutualInformation(py::module &m);
 
 #endif  // K2_PYTHON_CSRC_TORCH_MUTUAL_INFORMATION_H_
--- a/k2/python/csrc/torch/mutual_information.h
+++ b/k2/python/csrc/torch/mutual_information.h
@@ -73,16 +73,16 @@ namespace k2 {
    be at least 128.
 */
 torch::Tensor MutualInformationCpu(
-    torch::Tensor px,        // [B][S][T+1]
-    torch::Tensor py,        // [B][S+1][T]
-    torch::Tensor boundary,  // [B][4], int64_t.
-    torch::Tensor p);        //  [B][S+1][T+1]; an output
+    torch::Tensor px,                         // [B][S][T+1]
+    torch::Tensor py,                         // [B][S+1][T]
+    torch::optional<torch::Tensor> boundary,  // [B][4], int64_t.
+    torch::Tensor p);                         //  [B][S+1][T+1]; an output
 
 torch::Tensor MutualInformationCuda(
-    torch::Tensor px,        // [B][S][T+1]
-    torch::Tensor py,        // [B][S+1][T]
-    torch::Tensor boundary,  // [B][4], int64_t.
-    torch::Tensor p);        //  [B][S+1][T+1]; an output
+    torch::Tensor px,                         // [B][S][T+1]
+    torch::Tensor py,                         // [B][S+1][T]
+    torch::optional<torch::Tensor> boundary,  // [B][4], int64_t.
+    torch::Tensor p);                         //  [B][S+1][T+1]; an output
 
 /*
   backward of mutual_information; returns (grad_px, grad_py)
@@ -92,15 +92,13 @@ torch::Tensor MutualInformationCuda(
   very close to the value of ans_grad at entry.  This can be used
   to validate the correctness of this code.
 */
-std::vector<torch::Tensor> MutualInformationBackwardCpu(torch::Tensor px,
-                                                        torch::Tensor py,
-                                                        torch::Tensor boundary,
-                                                        torch::Tensor p,
-                                                        torch::Tensor ans_grad);
+std::vector<torch::Tensor> MutualInformationBackwardCpu(
+    torch::Tensor px, torch::Tensor py, torch::optional<torch::Tensor> boundary,
+    torch::Tensor p, torch::Tensor ans_grad);
 
 std::vector<torch::Tensor> MutualInformationBackwardCuda(
-    torch::Tensor px, torch::Tensor py, torch::Tensor boundary, torch::Tensor p,
-    torch::Tensor ans_grad, bool overwrite_ans_grad);
+    torch::Tensor px, torch::Tensor py, torch::optional<torch::Tensor> boundary,
+    torch::Tensor p, torch::Tensor ans_grad, bool overwrite_ans_grad);
 
 }  // namespace k2
 
--- a/k2/python/csrc/torch/mutual_information.h
+++ b/k2/python/csrc/torch/mutual_information.h
@@ -1,0 +1,96 @@
+/**
+ * @copyright
+ * Copyright      2021  Xiaomi Corporation (authors: Daniel Povey)
+ *
+ * @copyright
+ * See LICENSE for clarification regarding multiple authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef K2_PYTHON_CSRC_TORCH_MUTUAL_INFORMATION_H_
+#define K2_PYTHON_CSRC_TORCH_MUTUAL_INFORMATION_H_
+
+#include <torch/extension.h>
+#include <vector>
+
+namespace k2 {
+/*
+  Forward of mutual_information.  See also """... """ comment of
+  `mutual_information` in mutual_information.py.  This It is the core recursion
+  in the sequence-to-sequence mutual information computation.
+
+  Args:
+      px:     Tensor of shape [B][S][T + 1]; contains the log-odds ratio of
+              generating the next x in the sequence, i.e.
+              xy[b][s][t] is the log of
+                 p(x_s | x_0..x_{s-1}, y_0..y_{s-1}) / p(x_s),
+             i.e. the log-prob of generating x_s given subsequences of lengths
+             (s, t), divided by the prior probability of generating x_s.  (See
+              mutual_information.py for more info).
+      py:     The log-odds ratio of generating the next y in the sequence.
+              Shape [B][S + 1][T]
+      p:      This function writes to p[b][s][t] the mutual information between
+              sub-sequences of x and y of length s and t respectively, from the
+              b'th sequences in the batch.  Its shape is [B][S + 1][T + 1].
+              Concretely, this function implements the following recursion,
+              in the case where s_begin == t_begin == 0:
+
+               p[b,0,0] = 0.0
+               p[b,s,t] = log_add(p[b,s-1,t] + px[b,s-1,t],
+                                  p[b,s,t-1] + py[b,s,t-1])
+                       if s > 0 or t > 0,
+                       treating values with any -1 index as -infinity.
+              .. if `boundary` is set, we start fom p[b,s_begin,t_begin]=0.0.
+   boundary:  If set, a tensor of shape [B][4] of type int64_t, which
+              contains, where for each batch element b, boundary[b] equals
+              [s_begin, t_begin, s_end, t_end]
+              which are the beginning and end (i.e. one-past-the-last) of the
+              x and y sequences that we should process.  Alternatively, may be
+              a tensor of shape [0][0] and type int64_t; the elements will
+              default to (0, 0, S, T).
+     ans: a tensor `ans` of shape [B], where this function will set
+            ans[b] = p[b][s_end][t_end],
+            with s_end and t_end being (S, T) if `boundary` was specified,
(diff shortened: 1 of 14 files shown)


k2/python/csrc/torch/mutual_information_cuda.cu
```diff
@@ -0,0 +1,821 @@
...
+        //   p[b][s+sbb][t+tbb] = LogAdd(p[b][s+sbb-1][t+tbb] +
+        //   px[s+sbb-1][t+tbb],
+        //                               p[b][s+sbb][t+tbb-1] +
+        //                               py[s+sbb][t+tbb-1]
+        //
+        // where you can see that apart from the offsets of tbb and sbb, this is
+        // the same as the recursion defined for p in
+        // mutual_information.py:mutual_information_recursion(); and (eq. 0)
+        // above.
+#if 1
```
danpovey, line 303: Oh, this is something I forgot to test...
More ideally this would be #if 0, that path is slightly more optimized, but I have not tested it.

pkufool: Ok, I will test it.

pkufool:
I think this rnnt loss runs fine now, I will update the experiment results in Icefall (https://github.com/k2-fsa/icefall/pull/156).

Todos:
- [ ]  Add lm_only_scale & am_only_scale to pruned_rnnt_loss.

k2/python/k2/rnnt_loss.py
```diff
@@ -0,0 +1,995 @@
...
+            torch.full(
+                (B, S, 1), float("-inf"), device=px_am.device, dtype=px_am.dtype
+            ),
+        ),
+        dim=2,
+    )  # now: [B][S][T+1], index [:,:,T] has -inf..
+
+    if boundary is not None:
+        assert boundary.shape == (B, 4)
+        mask = torch.stack([torch.arange(0, T + 1, device=px_am.device)] * B)
```
danpovey, line 140: Here, instead of the torch.stack expression, it would be more efficient to do
`torch.arange(0, T + 1, device=px_am.device).reshape(1, T + 1).expand(B, T + 1)`
which is a zero-copy expression.
Also, below, instead of doing `mask = torch.where(mask < boundary[:, 3].reshape(B, 1), True, False)`
you can just use `mask = mask < boundary[:, 3].reshape(B, 1)` , which is already a boolean expression.

k2/python/k2/mutual_information.py
```diff
@@ -0,0 +1,298 @@
...
+            ctx.save_for_backward(px_grad, py_grad)
+        return ans, px_grad, py_grad
+
+    @staticmethod
+    def backward(
+            ctx, ans_grad: Tensor, dummy_px_grad: Tensor,
+            dummy_py_grad: Tensor) -> Tuple[torch.Tensor, torch.Tensor, None]:
+        (px_grad, py_grad) = ctx.saved_tensors
+        B, = ans_grad.shape
+        ans_grad = ans_grad.reshape((B, 1, 1))  # (B, 1, 1)
```
csukuangfj, line 71:
```suggestion
        ans_grad = ans_grad.reshape(B, 1, 1)  # (B, 1, 1)
```

k2/python/k2/mutual_information.py
```diff
@@ -0,0 +1,298 @@
...
+    assert px.dtype == py.dtype
+    (B, S, T) = px.shape
+    if boundary is not None:
+        assert boundary.dtype == torch.int64
+        assert boundary.shape == (B, 4)
+        for [s_begin, t_begin, s_end, t_end] in boundary.to('cpu').tolist():
+            assert 0 <= s_begin <= s_end <= S
+            assert 0 <= t_begin <= t_end <= T
+    # The following assertions are for efficiency
+    assert px.stride()[-1] == 1
```
csukuangfj, line 175:
```suggestion
    assert px.stride(-1) == 1
```

csukuangfj: I would recommend
```python3
assert px.is_contiguous()
assert py.is_contiguous()
```

k2/python/k2/mutual_information.py
```diff
@@ -0,0 +1,298 @@
...
+    assert py.stride()[-1] == 1
+    m, px_grad, py_grad = MutualInformationRecursionFunction.apply(
```
csukuangfj, line 177: If `return_grad` is False, is it possible to not return `px_grad` and `py_grad` to save memory?

k2/python/k2/mutual_information.py
```diff
@@ -0,0 +1,298 @@
...
+    return (m, (px_grad, py_grad)) if return_grad else m
+
+
+def _inner(a: Tensor, b: Tensor) -> Tensor:
+    """
+    Does inner product on the last dimension, with expected broadcasting,
+    i.e. equivalent to (a * b).sum(dim=-1)
+    without creating a large temporary.
+    """
+    assert a.shape[-1] == b.shape[-1]  # last last dim be K
```
csukuangfj, line 188:
```suggestion
    assert a.shape[-1] == b.shape[-1]  # The last dim must be equal
```

k2/python/k2/mutual_information.py
```diff
@@ -0,0 +1,298 @@
...
+            assert 0 <= t_begin <= t_end <= T
+    # The following assertions are for efficiency
+    assert px.stride()[-1] == 1
+    assert py.stride()[-1] == 1
+    m, px_grad, py_grad = MutualInformationRecursionFunction.apply(
+        px, py, boundary)
+    return (m, (px_grad, py_grad)) if return_grad else m
+
+
+def _inner(a: Tensor, b: Tensor) -> Tensor:
```
csukuangfj, line 182:
```suggestion
def _inner_product(a: Tensor, b: Tensor) -> Tensor:
```

k2/python/k2/mutual_information.py
```diff
@@ -0,0 +1,298 @@
...
+
+    px_cat = torch.stack(px, dim=0)  # (N, B, S, T+1)
+    py_cat = torch.stack(py, dim=0)  # (N, B, S+1, T)
+    px_tot = px_cat.sum(dim=0)  # (B, S, T+1)
+    py_tot = py_cat.sum(dim=0)  # (B, S+1, T)
+
+    if boundary is not None:
+        assert boundary.dtype == torch.int64
+        assert boundary.shape == (B, 4)
+        for [s_begin, t_begin, s_end, t_end] in boundary.to('cpu').tolist():
```
csukuangfj, line 257:
```suggestion
        for s_begin, t_begin, s_end, t_end in boundary.tolist():
```

k2/python/k2/mutual_information.py
```diff
@@ -0,0 +1,298 @@
...
+            assert 0 <= s_begin <= s_end <= S
+            assert 0 <= t_begin <= t_end <= T
+    else:
+        boundary = torch.zeros(0, 0, dtype=torch.int64, device=px_tot.device)
+
+    px_tot, py_tot = px_tot.contiguous(), py_tot.contiguous()
+    # The following assertions are for efficiency
+    assert px_tot.stride()[-1] == 1 and px_tot.ndim == 3
```
csukuangfj, line 265:
```suggestion
    assert px_tot.ndim == 3
```

`px_tot` is guaranteed to be contiguous as you invoked
```python3
px_tot = px_tot.contigous()
```

k2/python/csrc/torch/mutual_information.h
```diff
@@ -0,0 +1,109 @@
...
+                                   p[b,s,t-1] + py[b,s,t-1])
+                        if s > 0 or t > 0,
+                        treating values with any -1 index as -infinity.
+               .. if `boundary` is set, we start fom p[b,s_begin,t_begin]=0.0.
+    @param boundary  If set, a tensor of shape [B][4] of type int64_t, which
+                     contains, where for each batch element b, boundary[b]
+                     equals [s_begin, t_begin, s_end, t_end]
+                     which are the beginning and end (i.e. one-past-the-last)
+                     of the x and y sequences that we should process.
+                     Alternatively, may be a tensor of shape [0][0] and type
```
csukuangfj, line 62: I would recommend using `torch::optional<torch::Tensor>`.
You don't need to create a tensor of shape (0, 0). Just leave it unset.

Also, you can pass a `None` from Python.

k2/python/k2/mutual_information.py
```diff
@@ -0,0 +1,298 @@
...
+    # this is a kind of "fake gradient" that we use, in effect to compute
+    # occupation probabilities.  The backprop will work regardless of the
+    # actual derivative w.r.t. the total probs.
+    ans_grad = torch.ones(B, device=px_tot.device, dtype=px_tot.dtype)
+
+    (px_grad,
+     py_grad) = _k2.mutual_information_backward(px_tot, py_tot, boundary, p,
+                                                ans_grad)
+
+    px_grad, py_grad = px_grad.reshape(1, B, -1), py_grad.reshape(1, B, -1)
```
csukuangfj, line 282:
```suggestion
    px_grad = px_grad.reshape(1, B, -1)
    py_grad = py_grad.reshape(1, B, -1)
```
is more readable.

k2/python/k2/mutual_information.py
```diff
@@ -0,0 +1,298 @@
...
+    px_cat, py_cat = px_cat.reshape(N, B, -1), py_cat.reshape(N, B, -1)
+    # get rid of -inf, would generate nan on product with 0
+    px_cat, py_cat = px_cat.clamp(min=-1.0e+38), py_cat.clamp(min=-1.0e+38)
```
csukuangfj, line 285:
```suggestion
    px_cat, py_cat = px_cat.clamp(min=torch.finfo(torch.float32).min), py_cat.clamp(min=-1.0e+38)
```

Instead of doing
```
a, b = op(a), op(b)
```

I would recommend doing only one thing in a single line
```
a = op(a)
b = op(b)
```

k2/python/k2/rnnt_loss.py
```diff
@@ -0,0 +1,995 @@
...
+          represents the probability of extending the subsequences of length
+          (s,t) by one in the t direction,
+          i.e. of emitting the termination/next-frame symbol.
+
+          px[:,:,T] equals -infinity, meaning on the "one-past-the-last" frame
+          we cannot emit any symbols.  This is simply a way of incorporating
+          the probability of the termination symbol on the last frame.
+    """
+    assert (
+        lm.ndim == 3
```
csukuangfj, line 95: I would recommend separating these conjunctions out.
If one of the assertions fails, you don't know exactly which one failed.

k2/python/k2/rnnt_loss.py
```diff
@@ -0,0 +1,995 @@
...
+
+    # subtracting am_max and lm_max is to ensure the probs are in a good range
+    # to do exp() without causing underflow or overflow.
+    am_max, _ = torch.max(am, dim=2, keepdim=True)  # am_max: [B][T][1]
+    lm_max, _ = torch.max(lm, dim=2, keepdim=True)  # lm_max: [B][S+1][1]
+    am_probs = (am - am_max).exp()
+    lm_probs = (lm - lm_max).exp()
+    # normalizers: [B][S+1][T]
+    normalizers = (
+        torch.matmul(lm_probs, am_probs.transpose(1, 2)) + 1.0e-37
```
csukuangfj, line 112: What is the magic number `1.0e-37`?

danpovey: It's close to the lowest representable single precision float.  I was concerned making this too high, like 1e-20, might be the cause of instability.

csukuangfj: There is 
```
torch.finfo(torch.float32).tiny
```
which is `1.17549e-38`, Is this the value you want?

k2/python/k2/mutual_information.py
danpovey: ... that is a valid point.  In some circumstances we might not even need the backprop.  I'm pretty sure there was a time when it allowed to delay the backward part of the computation, it must have been simplified since then.
The memory requirements are not very huge because it's only one element per (s,t) position, not C elements, but it's still nonzero.

k2/python/k2/rnnt_loss.py
danpovey: Yes I suppose so.  Could even use the dtype of lm_probs instead of torch.float32.

pkufool:
Merging now, as it will be more convenient for me to integrate the modified version.

k2/python/csrc/torch/mutual_information.h
```diff
@@ -0,0 +1,107 @@
...
+namespace k2 {
+/*
+  Forward of mutual_information.  See also comment of `mutual_information`
+  in mutual_information.py.  This is the core recursion
+  in the sequence-to-sequence mutual information computation.
+
+    @param px  Tensor of shape [B][S][T + 1]; contains the log-odds ratio of
+               generating the next x in the sequence, i.e.
+               xy[b][s][t] is the log of
+                  p(x_s | x_0..x_{s-1}, y_0..y_{s-1}) / p(x_s),
```
csukuangfj, line 39: p(x_s | x_0..x_{s-1}, y_0..y_{s-1})

--> (the last one: s-1 --> t-1)

p(x_s | x_0..x_{s-1}, y_0..y_{t-1})

k2/python/csrc/torch/mutual_information.h
```diff
@@ -0,0 +1,107 @@
...
+                     of the x and y sequences that we should process.
+                     Alternatively, may be a tensor of shape [0][0] and type
+                     int64_t; the elements will default to (0, 0, S, T).
+    @return A tensor `ans` of shape [B], where this function will set
+               ans[b] = p[b][s_end][t_end],
+               with s_end and t_end being (S, T) if `boundary` was specified,
+               and (boundary[b][2], boundary[b][3]) otherwise.
+               `ans` represents the mutual information between each pair of
+               sequences (i.e. x[b] and y[b], although the sequences are not
+               supplied directy to this function).
```
csukuangfj, line 70: typo: directy -> directly

k2/python/k2/mutual_information.py
```diff
@@ -0,0 +1,302 @@
...
+    B, S, T1 = px.shape
+    T = T1 - 1
+    assert py.shape == (B, S + 1, T)
+    assert px.dtype == py.dtype
+    (B, S, T) = px.shape
+    if boundary is not None:
+        assert boundary.dtype == torch.int64
+        assert boundary.shape == (B, 4)
+        for s_begin, t_begin, s_end, t_end in boundary.tolist():
+            assert 0 <= s_begin <= s_end <= S
```
csukuangfj, line 172: https://github.com/pkufool/k2/blob/5dbf4081a99938cc9e3052de0bbb069346128725/k2/python/k2/mutual_information.py#L132
says:
>         with 0 <= s_begin <= s_end < S and 0 <= t_begin <= t_end < T

But this line is using
```python3
            assert 0 <= s_begin <= s_end <= S
```

Which one is correct?

k2/python/k2/mutual_information.py
```diff
@@ -0,0 +1,302 @@
...
+            assert 0 <= t_begin <= t_end <= T
+    # The following assertions are for efficiency
+    assert px.is_contiguous()
```
csukuangfj, line 175: https://github.com/pkufool/k2/blob/5dbf4081a99938cc9e3052de0bbb069346128725/k2/python/k2/mutual_information.py#L114
says

> 
        Note: we don't require px and py to be contiguous, but the
        code assumes for optimization purposes that the T axis has
        stride 1.

Buth here it is using
```python3
    assert px.is_contiguous()
    assert py.is_contiguous()
```

Which one is correct?

k2/python/k2/mutual_information.py
danpovey: should be <=.

k2/python/k2/mutual_information.py
danpovey: We could probably reduce the assertions to: `assert px.stride(-1) == 1 and py.stride(-1) == 1

csukuangfj: I find that the implementation is using tensor's accessor to access the underlying data, which handles the strides automagically. So I think it does not pose any constraints on the layout of `px` and `py`. 

We don't need to check the stride of `px` and `py` here, I think.

danpovey: It may not actually be required (although I would have to check the GPU implementation), but when the code was designed, I assumed it had stride 1, and it would likely be super slow if not.

k2/python/k2/rnnt_loss.py
```diff
@@ -0,0 +1,684 @@
...
+          the probability of the termination symbol on the last frame.
+    """
+    assert joint.ndim == 4
+    (B, T, S1, C) = joint.shape
+    S = S1 - 1
+    assert symbols.shape == (B, S)
+
+    max_value = torch.max(joint)
+    normalizers = (joint - max_value)
+    normalizers = torch.logsumexp(normalizers, dim=3)
```
csukuangfj: `torch.logsumexp()` is numerically stable, so there is no need to subtract the `max` here.

See https://pytorch.org/docs/stable/generated/torch.logsumexp.html
> Returns the log of summed exponentials of each row of the input tensor in the given dimension dim. The computation is numerically stabilized.

and see its internal implementation
https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/ReduceOps.cpp#L1260

```cpp
static Tensor& logsumexp_out_impl(Tensor& result, const Tensor& self, IntArrayRef dims, bool keepdim) {
  // can't take max of empty tensor
  if (self.numel() != 0) {
    auto maxes = at::amax(self, dims, true);
    auto maxes_squeezed = (keepdim ? maxes : squeeze_multiple(maxes, dims));
    maxes_squeezed.masked_fill_(maxes_squeezed.abs() == INFINITY, 0);
    at::sum_out(result, (self - maxes).exp_(), dims, keepdim);
    result.log_().add_(maxes_squeezed);
  } else {
```

pkufool: Yes, Dan had a comment on this line before, I forgot to delete it.  Will fix the comments you gave these days.  Thanks!

csukuangfj: I see. It has been fixed in the master.

k2/python/tests/mutual_information_test.py
```diff
@@ -0,0 +1,271 @@
...
+                            t_end = random.randint(t_begin, T)
+                            return [s_begin, t_begin, s_end, t_end]
+
+                        if device == torch.device('cpu'):
+                            boundary = torch.tensor(
+                                [get_boundary_row() for _ in range(B)],
+                                dtype=torch.int64,
+                                device=device)
+                        else:
+                            boundary = boundary.to(device)
```
csukuangfj, line 74: You are using the `boundary` that was defined when `device` is `cpu`.

The code should work if users disable the tests for CPU.

PR #892, opened 2021-12-21
Fix FSAVec TOPSORTED properity and RaggedTensor out-of-range bug.

drawfish:
Remove extra commas after 'TOPSORTED' properity.
Fix RaggedTensor constructer parameter 'byte_offset' out-of-range bug.


csukuangfj:
+2

PR #894, opened 2021-12-29
[Add] Support the detection of GCC version

TeaPoly:
Fix building error `invalid static_cast` when GCC is lower than 7.0.
Issue is mentioned here: https://github.com/k2-fsa/icefall/issues/161

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,13 @@ endif()
 set(languages CXX)
 set(_K2_WITH_CUDA ON)
 
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    # require at least gcc 7.0
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.0)
+        message(FATAL_ERROR "GCC version must be at least 7.0!")
+    endif()
+endif()
+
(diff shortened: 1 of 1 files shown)


danpovey:
Why is this closed, did we fix this another way?

csukuangfj:
I think the error is from PyTorch and exists only in PyTorch >= 1.10.

TeaPoly:
> Why is this closed, did we fix this another way?

The GCC version seems to be related to the `PyTorch` version.
I think this solution is not appropriate, so closed it.

danpovey:
So we don't have access to the PyTorch version in the CMakeLists.txt?
Surely there must be some reasonable way to enforce this.
Although it seems like this shoulds have been done inside PyTorch itself.

csukuangfj:
> So we don't have access to the PyTorch version in the CMakeLists.txt?
> 
> Surely there must be some reasonable way to enforce this.
> 
> Although it seems like this shoulds have been done inside PyTorch itself.

Yes, we do have. Please see k2/cmake/torch.cmake

csukuangfj:
If I remember correctly, that error occurs with torch 1.10 only when you enable CUDA support. If you build a CPU version of k2, you will not have such an error.

Issue #886, opened 2021-12-05
Hash64


danpovey:
This is something that we will likely need for the RNN-T decoding algorithm from #885, because our current
hash code may not support enough bits in the hash (since in addition to the FSA state, we now need the 2-symbol
history and the substep-index).

The task is to create a Hash64, which will be similar to the current Hash code, except the keys and values are both 64-bit
integers and you can use all bits.  There would, for now, be just one type of accessor (later on we may add support for
keys with slightly more than 64 bits in them, by using the bucket index similar to the current hash code).  The code would be substantially simpler; for now at least, there will be no keeping track of number of key bits or value bits, they would both
be 64.

The format/layout would be similar to the current hash code.  The layout would be: key value key value..., each 64-bit,
so a bucket is a key/value pair, and we fail over to the next bucket just like in the current hash code.  The only atomic
operations will be the operations on keys, so the Insert() operation would look like the current Insert() operations
except that we write the key, not the key/value, and once we successfully added a key, we then write the value using
a non-atomic operation as "fire and forget".  We will have to have a warning that you cannot use both Insert() and Find()
in the same kernel, because Find() may end up trying to read a key that has not yet been read.  [Actually this would
be fairly easy to fix; we could just just declare one particular key value, like -1, "invalid", and just wait in a loop until
you read a value different from -1].

pkufool:
Closed via #895

PR #895, opened 2021-12-30
Add Hash64

pkufool:
Implements #886.

- [x] Implements resize.

### DIFF
--- a/k2/csrc/hash.h
+++ b/k2/csrc/hash.h
@@ -1282,6 +1282,66 @@ class Hash64 {
     }
   }
 
+  /* Resize the hash to a new number of buckets.
+
+       @param [in] new_num_buckets   New number of buckets; must be a power of 2,
+                  and must be large enough to accommodate all values in the hash
+                  (we assume the caller is keeping track of the number of elements
+                  in the hash somehow).
+
+     CAUTION: Resizing will invalidate any accessor objects you have; you need
+     to re-get the accessors before accessing the hash again.
+  */
+  void Resize(int32_t new_num_buckets, bool copy_data = true) {
+    NVTX_RANGE(K2_FUNC);
+
+    K2_CHECK_GT(new_num_buckets, 0);
+    K2_CHECK_EQ(new_num_buckets & (new_num_buckets - 1), 0);  // power of 2.
+
+    ContextPtr c = data_.Context();
+    Hash64 new_hash(c, new_num_buckets);
+
+    if (copy_data) {
+      new_hash.CopyDataFromSimple(*this);
+    }
+
+    *this = new_hash;
+    new_hash.Destroy();  // avoid failed check in destructor (it would otherwise
+                       // expect the hash to be empty when destroyed).
+  }
+
+  /*
+    Copies all data elements from `src` to `*this`.
+   */
+  void CopyDataFromSimple(Hash64 &src) {
+    NVTX_RANGE(K2_FUNC);
+    int32_t num_buckets = data_.Dim() / 2,
+        src_num_buckets = src.data_.Dim() / 2;
+    const uint64_t *src_data = src.data_.Data();
+    uint64_t *data = data_.Data();
+    size_t new_num_buckets_mask = static_cast<size_t>(num_buckets) - 1,
+        new_buckets_num_bitsm1 = buckets_num_bitsm1_;
+    ContextPtr c = data_.Context();
+    K2_EVAL(c, src_num_buckets, lambda_copy_data, (int32_t i) -> void {
+        uint64_t key = src_data[2 * i];
+        uint64_t value = src_data[2 * i + 1];
+        if (~key == 0) return;  // equals -1.. nothing there.
+        uint64_t bucket_inc = 1 | ((key >> new_buckets_num_bitsm1) ^ key);
+        size_t cur_bucket = key & new_num_buckets_mask;
+        while (1) {
+          uint64_t assumed = ~((uint64_t)0),
+              old_elem = AtomicCAS((unsigned long long*)(data + 2 * cur_bucket),
+                                   assumed, key);
+          if (old_elem == assumed) {
+            *(data + 2 * cur_bucket + 1) = value;
+            return;
+          }
+          cur_bucket = (cur_bucket + bucket_inc) & new_num_buckets_mask;
+          // Keep iterating until we find a free spot in the new hash...
+        }
+      });
+  }
+
   // The destructor checks that the hash is empty, if we are in debug mode.
   // If you don't want this, call Destroy() before the destructor is called.
   ~Hash64() {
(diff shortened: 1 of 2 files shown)


k2/csrc/hash.h
```diff
@@ -1014,6 +1014,349 @@ class Hash {
...
+       @return  Returns true if this (key,value) pair was inserted, false
+     otherwise.
+
+       Note: the const is with respect to the metadata only; it is required, to
+       avoid compilation errors.
+    */
+    __forceinline__ __host__ __device__ bool Insert(
+        uint64_t key, uint64_t value, uint64_t *old_value = nullptr,
+        uint64_t **key_value_location = nullptr) const {
+      uint32_t cur_bucket = static_cast<uint32_t>(key) & num_buckets_mask_,
```
csukuangfj, line 1119: Any reason to discard the higher 32bits of the key?
There is no such a cast at line 1178.

k2/csrc/hash.h
```diff
@@ -1014,6 +1014,349 @@ class Hash {
...
+  // You should call this before the destructor is called if the hash will still
+  // contain values when it is destroyed, to bypass a check.
+  void Destroy() { data_ = Array1<uint64_t>(); }
+
+  void CheckEmpty() {
+    if (data_.Dim() == 0) return;
+    ContextPtr c = Context();
+    Array1<int32_t> error(c, 1, -1);
+    int32_t *error_data = error.Data();
+    uint64_t *hash_data = data_.Data();
```
csukuangfj, line 1268:
```suggestion
    const uint64_t *hash_data = data_.Data();
```

k2/csrc/hash.h
```diff
@@ -1014,6 +1014,349 @@ class Hash {
...
+    Return an Accessor object which can be used in kernel code (or on CPU if the
+    context is a CPU context).
+  */
+  Accessor GetAccessor() { return Accessor(*this); }
+
+  // You should call this before the destructor is called if the hash will still
+  // contain values when it is destroyed, to bypass a check.
+  void Destroy() { data_ = Array1<uint64_t>(); }
+
+  void CheckEmpty() {
```
csukuangfj, line 1263:
```suggestion
  void CheckEmpty() const {
```

k2/csrc/hash.h
```diff
@@ -1014,6 +1014,349 @@ class Hash {
...
+          if (~(hash_data[i]) != 0) error_data[0] = i;
+        });
+    int32_t i = error[0];
+    if (i >= 0) {  // there was an error; i is the index into the hash where
+      // there was an element.
+      int64_t elem = data_[i];
+      // We don't know the number of bits the user was using for the key vs.
+      // value, so print in hex, maybe they can figure it out.
+      K2_LOG(FATAL) << "Destroying hash: still contains values: position " << i
+                    << ", key,value = " << std::hex << elem;
```
csukuangfj, line 1281:
```suggestion
                    << ", content = " << std::hex << elem;
```
I think `elem` can be either a key or a value.

csukuangfj:
+1

k2/csrc/hash.h
```diff
@@ -1014,6 +1014,349 @@ class Hash {
 };
 
 
+/*
+  How class Hash64 works:
+
```
danpovey, line 1019: The documentation looks the same as Hash.. should be changed!

k2/csrc/hash.h
```diff
@@ -1014,6 +1014,349 @@ class Hash {
...
+          buckets_num_bitsm1_(hash.buckets_num_bitsm1_) {}
+
+    // Copy constructor
+    Accessor(const Accessor &src) = default;
+
+    /*
+     Try to insert pair (key,value) into hash.
+       @param [in] key  Key into hash, it is an error if ~key == 0, i.e. if all
+                        the allowed bits of `key` are set.
+       @param [in] value  Value to set, it is an error if ~value == 0, i.e. if
```
danpovey, line 1101: I think this "it is an error" stuff on "value" may not be necessary and the stipulation on 'key' only may be enough??
Is there a particular race condition that it solves?

k2/csrc/hash.h
csukuangfj: It does not document what a bucket contains and how a key-value pair is stored.

k2/csrc/hash.h
```diff
@@ -1014,6 +1014,349 @@ class Hash {
...
+          @param [in] key  Required to be the same key that was provided to
+                        Find(); it is an error otherwise.
+          @param [in] value  Value to write;
+
+      Note: the const is with respect to the metadata only; it is required, to
+      avoid compilation errors.
+     */
+    __forceinline__ __host__ __device__ void SetValue(
+        uint64_t *key_value_location, uint64_t key, uint64_t value) const {
+      *key_value_location = key;
```
danpovey, line 1210: Setting the key here is not necessary as it's required to be the same as the current key.

k2/csrc/hash.h
```diff
@@ -1014,6 +1014,349 @@ class Hash {
...
+      compilation errors.
+    */
+    __forceinline__ __host__ __device__ void Delete(uint64_t key) const {
+      uint32_t cur_bucket = key & num_buckets_mask_,
+               bucket_inc = 1 | ((key >> buckets_num_bitsm1_) ^ key);
+      while (1) {
+        uint64_t old_key = data_[2 * cur_bucket];
+        if (old_key == key) {
+          data_[2 * cur_bucket] = ~((uint64_t)0);
+          data_[2 * cur_bucket + 1] = ~((uint64_t)0);
```
danpovey, line 1233: It's possible that setting the value to ~0 is not necessary.

k2/csrc/hash.h
pkufool: `~(unint64_t)0` is used as `invalid`,  in `Find` it will wait in a loop until reading a value different from `~(unint64)0`.

https://github.com/k2-fsa/k2/blob/bb11575cf4fede0da96d877893a66347db3faacf/k2/csrc/hash.h#L1179-L1195

k2/csrc/hash.h
pkufool: Thanks, fixed.

k2/csrc/hash.h
pkufool: Maybe, it is better to change the bucket index to unint64_t.

danpovey: I think num_buckets_mask_ is uint32_t. 
I think we cannot index larger than int32_t into Array, so while it would be possible to make both those things uint64_t, I'm not sure that it would have a benefit.  But it might reduce the number of instructions slightly if we made them both uint64_t, so I  guess I'd be OK with that??  Or at least cast num_buckets_mask_ to uint64_t before using it...

PR #900, opened 2022-01-12
Extend interface of SubsampleRagged.

danpovey:
This is something else that will be needed for the RNN-T decoding.
It consists of extending the interface of SubsampleRagged to work on arbitrary/specified axes.
Hopefully this can be done in such a way that the performance of the current SubsampleRagged
invocations is not materially affected.
While doing this, please rename SubsampleRagged -> SubsetRagged.


danpovey:
I have added a new interface to this, that also needs to be implemented for the RNN-T decoding.
It's something like #899, but a bit simpler.
It's possible to treat this as a separate issue, but I included it in this PR as it was convenient for me.

pkufool:
See #905

PR #901, opened 2022-01-13
Fix k2.ragged.create_ragged_shape2

csukuangfj:
Before the fix, we have to specify both `row_splits` and `row_ids`
while calling `k2.create_ragged_shape2` even if one of them is `None`. 

The reason is that during the wrapping we did not specify the default value for `row_splits`
and `row_ids`.

After this fix, we only need to specify one of them.


danpovey:
cool

PR #902, opened 2022-01-17
Modified rnnt

pkufool:
Notes from Daniel:

I have something I want to try as a modification of RNN-T, basically to make it so there is max 1 symbol per frame in training time as well as test, and the way it would modify the recursion is, instead of a non-termination symbol taking you "up", it would take you diagonally up and to the right. 
It would also greatly simplify the decoding algorithm, as it would now be frame by frame, not sub-frame by sub-frame. For the decoder, the difference is we increment t every time we eat a symbol.


pkufool:
@danpovey  I think I have moved the code from you branch successfully, all the test cases passed (including the comparison with pytorch one).

danpovey:
cool!

pkufool:
@danpovey  Could we merge this now, so that other persons can use it.

danpovey:
Sure, merging is OK with me.

k2/python/k2/mutual_information.py
```diff
@@ -166,10 +170,10 @@ def mutual_information_recursion(
     """
     assert px.ndim == 3
     B, S, T1 = px.shape
-    T = T1 - 1
+    T = py.shape[-1]
+    assert px.shape[-1] in [T, T + 1]  # if T, then "modified".
```
csukuangfj, line 174: https://github.com/pkufool/k2/blob/16f3ac506a725e280016763784713827a993990a/k2/python/k2/mutual_information.py#L94

>       px:
>        A torch.Tensor of some floating point type, with shape ``[B][S][T+1]``,

The doc for `px` should also be updated for the modified case.

PR #911, opened 2022-01-28
Change the sign of the rnnt_loss and add reduction argument

pkufool:


k2/python/k2/rnnt_loss.py
```diff
@@ -201,25 +202,52 @@ def rnnt_loss_simple(
...
-    px, py = get_rnnt_logprobs(lm, am, symbols, termination_symbol, boundary)
-    return mutual_information_recursion(px, py, boundary, return_grad)
+    px, py = get_rnnt_logprobs(
+        lm=lm,
+        am=am,
+        symbols=symbols,
+        termination_symbol=termination_symbol,
+        boundary=boundary,
+    )
+    loss_and_grad = mutual_information_recursion(
```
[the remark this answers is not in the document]
pkufool: Yes, thanks! Fixed.

k2/python/k2/rnnt_loss.py
```diff
@@ -201,25 +202,51 @@ def rnnt_loss_simple(
...
         occupation probability is the output of the backward with a
         `fake gradient` input (all ones)  This is useful to implement the
         pruned version of rnnt loss.
     Returns:
-       If return_grad is False, returns a Tensor of shape (B,), containing the
-       NEGATED total RNN-T loss values for each element of the batch
-       (like log-probs of sequences).
+       If return_grad is False, returns a tensor of shape (B,), containing the
+       total RNN-T loss values for each element of the batch if reduction equals
+       to "none", otherwise a scalar with the reduction applied.
```
danpovey, line 219: Around line 214 we should clarify that the "fake gradient" is the same as the gradient you'd
get if you did `torch.grad( (-loss.sum()), [px, py])`.   
Now there is a - sign.
When you refer to the fake gradient below, you can add "(see above)" to clarify that it's explained above.

k2/python/k2/rnnt_loss.py
```diff
@@ -229,7 +256,7 @@ def get_rnnt_logprobs_joint(
     This function is called from rnnt_loss().
 
     Args:
-      joint:
+      logits:
         The output of joiner network, with shape (B, T, S + 1, C),
         i.e. batch, time_seq_len, symbol_seq_len+1, num_classes
```
danpovey, line 261: I think we should remove "possibly including EOS", since RNN-T does not use EOS.

pkufool:
Merging
Will release a new version after fixing the building docs.

PR #913, opened 2022-02-06
Implements the util functions needed by decoding

pkufool:
- [X] SubsampleRagged
- [X] PrunedRagged
- [ ] ArrayofRagged
- [ ] Unstack

PR #919, opened 2022-02-14
SubsetRagged & PruneRagged

pkufool:
Breaking https://github.com/k2-fsa/k2/pull/913 into two pull requests to make it easier to review, this pull request is about the `SubsetRagged` & `PruneRagged`.


k2/csrc/ragged_ops.h
```diff
@@ -804,18 +822,88 @@ RaggedShape RemoveEmptyListsAxis0(RaggedShape &src_shape,
...
+/*
+  This function creates a Renumbering object that can be used to obtain subsets
+  of ragged arrays via SubsetRaggedShape().  It implements beam pruning as
+  used in pruned Viterbi search and similar algorithms, where there is both a
+  beam and a max-active (`max_elems`) constraint.  T will probably be float or
+  double, interpreted as a "positive-is-better" sense, i.e. as scores.
+
+   @param [in] src  The ragged object to be subsampled.
+   @param [in] axis  The axis to be subsampled, must satisfy
+              0 <= axis < src.NumAxes().  The axis before `axis`, if axis < 0,
```
danpovey, line 869: I think `axis < 0` should be `axis > 0`.

k2/csrc/ragged_ops.h
```diff
@@ -804,18 +822,88 @@ RaggedShape RemoveEmptyListsAxis0(RaggedShape &src_shape,
...
+              will be interpreted as a "batch" axis.
+   @param [in] beam  The main pruning beam.  The sub-lists of elements on axis
+              `axis` will be removed if their maximum element (or the element
+              itself, if axis + 1 == src.NumAxes()) is less than
+              this_best_elem - beam, where this_best_elem
+              is the maximum element taken over axis `axis-1` (or over the
+              entire array, if axis == 0).   Think of axis `axis-1`, if
+              present, as the "batch" axis, and axis `axis` as the axis that we
```
danpovey, line 877: we can change "if present" to "if axis > 0", which is clearer.

k2/csrc/ragged_ops.h
```diff
@@ -804,18 +822,88 @@ RaggedShape RemoveEmptyListsAxis0(RaggedShape &src_shape,
...
+              actually remove elements or sub-lists on.  Empty sub-lists on axis
+              `axis` will always be pruned, as their score would be treated
+              as -infinity.
+   @param [in] max_elems  If max_elems > 0, it is the maximum number of sub-lists
+              or elements that are allowed within any sub-list on axis `axis-1`
+              (or the maximum number of top-level sub-lists after subsampling,
+              if axis == 0).  We keep the best ones, but behavior in case of ties is
+              undefined (TODO: check whether SortSublists() is a stable sort, and
```
danpovey, line 885: please resolve this TODO, we need to clarify if the behavior, if there are ties, is well-defined or could be random.

k2/csrc/ragged_ops_inl.h
```diff
@@ -759,6 +759,129 @@ Array2<T> PadRagged(Ragged<T> &src, const std::string &mode, T padding_value) {
   return res;
 }
 
+// Prune a two axes ragged tensor on axis0
+template <typename T>
```
danpovey, line 763: This needs more documentation!
OK to just say this is a special case of PruneRagged with axis == 0, and refer to
the docs for PruneRagged in ragged_ops.h.

k2/csrc/ragged_ops_inl.h
```diff
@@ -759,6 +759,129 @@ Array2<T> PadRagged(Ragged<T> &src, const std::string &mode, T padding_value) {
...
+    });
+  } else {
+    K2_EVAL(c, total_elements, lambda_set_keep, (int32_t i) {
+        keep_data[i] = sub_max_data[i] >= max_value - beam;
+    });
+  }
+  return renumbering;
+}
+
+// Prune a two axes ragged tensor on axis1
```
danpovey, line 806: same here

danpovey (commented):
Great, thanks!
Comments are small.

k2/csrc/algorithms.h
```diff
@@ -121,17 +121,40 @@ class Renumbering {
...
   }
 
+  /* Return a mapping from new index to old index, with one extra element
+     containing the total number of kept elements if extra_element == true.
+     If Keep() can be interpreted as a tails vector, i.e. with 1 at the end
+     of sub-lists of elements, then New2Old(true) would corresponds to a
+     row-splits array and Old2New(false) would correspond to a row-ids
+     array.
+  */
+  Array1<int32_t> New2Old(bool extra_element) {
```
csukuangfj, line 131: Should we make the interface the same as `Old2New(bool extra_element = false)`?
Just for consistency.

k2/csrc/algorithms.h
```diff
@@ -121,17 +121,40 @@ class Renumbering {
     return new2old_;
   }
 
+  /* Return a mapping from new index to old index, with one extra element
```
csukuangfj, line 124: The comment on line 115 needs also to be updated.

k2/csrc/ragged_ops.h
```diff
@@ -699,14 +699,32 @@ RaggedShape RandomRaggedShape(bool set_row_ids = false,
...
-  might give rise to empty lists.
+  Return ragged shape with only a subset of the elements or sub-lists
+  on the specified axis kept.  (This is not regular sampling, it is
+  irregular subsampling with specified elements kept).
+
+    @param [in] src  The ragged shape that we are subsampling
+    @param [in] renumbering  The renumbering object that dictates
+                    which elements of `src` we keep; we require
+                    renumbering.NumOldElems() == src.TotSize(axis2)
+                    where axis2 = (axis < 0 ? src.NumAxes() - axis : axis).
```
csukuangfj, line 710:
```suggestion
                    where axis2 = (axis < 0 ? src.NumAxes() + axis : axis).
```

k2/csrc/ragged_ops.h
```diff
@@ -699,14 +699,32 @@ RaggedShape RandomRaggedShape(bool set_row_ids = false,
...
+    @param [in] axis  The axis to subsample; if negative, will be
+                    interpreted as an offset from src.NumAxes().
+    @param [out] elems_new2old   If supplied, this function will
+                    output to this location a new2old vector that
+                    dictates how the elements of a ragged tensor
+                    with shape `src` would be renumbered.
+    @return  Returns the subsampled shape. All dimensions and tot-sizes
+       preceding the final axis will remain the same, which might give
```
csukuangfj, line 718: If `axis` is not -1, does this comment still valid?

k2/csrc/algorithms.h
danpovey: It is the same in effect; there is an overloaded version taking no arg, that doesn't return the extra element, but
returns the ref.  If we had default arg, it would cause ambiguity in function overloading.

k2/csrc/ragged_ops.h
danpovey: ah yes, perhaps should be "preceding axis `axis`".  But double-check this logic!

k2/csrc/algorithms.h
csukuangfj: I mean, removing the one with no argument.

danpovey: RE removing the one with no argument: the reason it's designed the way it is, is to avoid redundant copies of tensors, which could take time at runtime and also bulk up the compiled code with unnecessary instructions.  New2Old() is called quite a lot so I thought it was worth saving that time.

k2/csrc/ragged_ops.h
pkufool: I just checked the code, when running on CPU we use `std::sort` which is stable with elements number less than 16 (algorithm is insert sort) and is unstable with elements number equals to or greater than 16 (algorithm is quicksort).
when running on GPU we use the `segmented_sort` in modern_gpu lib which is modified from merge sort, and it is stable. So do we need to clarify the different behavior of different devices, especially the 16 elements thing?

pkufool: Maybe, we can change the cpu sorting algorithm from `std::sort` to `std::stable_sort`.

danpovey:
You could also test this on tedlium, if you have enough GPUs.
I am worried that your experiments and my experiments are in disagreement about whether the
max_frames_mask_fraction should be 0.3 or 0.15.  We need to explore that range, preferably
on other data.

k2/csrc/ragged_ops.h
danpovey: Sure, I guess that is OK.

pkufool:
I am merging it for further development. Will fix in other PRs if there are issues.

Issue #921, opened 2022-02-17
Bugs in Stack a shape itself


pkufool:
I found this bug when testing `Unstack` (i.e. stacking the Unstack result should give the original RaggedShape itself).

There is something wrong when stacking one RaggedShape (actually adding an axis), especially for the shape with nested empty lists.  Here is an example to reproduce this bug,

```cpp
ContextPtr c = GetCpuContext(); // or GetCudaContext();
auto shape = RaggedShape(c, "[ [ ] [ ] [ ] [ [ [ x x x ] [ ] [ x ] [ ] [ ] ] [ ] ] [ ] ]");
std::vector<RaggedShape *> ptr;
prt.emplace_back(&shape);
auto dest = Stack(2, 1, ptr.data());
// result should be: [ [ ] [ ] [ ] [ [ [ [ x x x ] [ ] [ x ] [ ] [ ] ] ] [ ] ] [ ] ]
```

The above is just one bad case,  you can find many cases by removing line 278 in the following test code.

https://github.com/k2-fsa/k2/blob/278a0becf7feca5a73872e4dfee516dda477c2de/k2/csrc/ragged_test.cu#L256-L285

pkufool:
I am busy doing other staff, I think it is not urgent, will fix it later, or maybe someone can help to fix it.

```
[F] /ceph-hw/kangwei/code/k2_rnnt_decode/k2/csrc/ragged.cu:416:k2::RaggedShape::Check() const::<lambda(int32_t)> Ragged shape has num_elems=2 for layers_[0],
vs. num_rows=5for the next layer.


[ Stack-Trace: ]
/ceph-hw/kangwei/code/k2_rnnt_decode/build/lib/libk2_log.so(k2::internal::GetStackTrace()+0x5b) [0x7f2189c3b82a]
./bin/cu_ragged_test(k2::internal::Logger::~Logger()+0x44) [0x55f058c3fbf2]
/ceph-hw/kangwei/code/k2_rnnt_decode/build/lib/libk2context.so(+0x3394dc) [0x7f218a3ed4dc]
/ceph-hw/kangwei/code/k2_rnnt_decode/build/lib/libk2context.so(k2::RaggedShape::Check() const+0xa51) [0x7f218a3eedd3]
[...]

PR #925, opened 2022-02-23
Fix Stack

wgb14:
closing #921

In my tests, the bug only appears when `num_src==1` and `axis>1`, here `s==*src[0]`, so `ans_layers[axis - 1]` and `ans_layers[axis]` should be buit by `s.Layers()[axis-1]` instead of `s.Layers()[0]`, i.e. https://github.com/k2-fsa/k2/blob/3cc74f19ccc045a7176a8a96a4b24c4af76e9653/k2/csrc/ragged_ops.cu#L1091-L1093
should be
```
  ans_layers[axis - 1] =
      RegularRaggedShape(c, s.TotSize(axis-1) / num_srcs, num_srcs).Layers()[0];
  ans_layers[axis] = s.Layers()[axis-1];
```
But to make the script consistent when `num_src>1`, I turned to modify `IntersperseRaggedLayer()` to let it return `*src.Layers()[axis-1]` when `num_src==1`.

Since I'm new to K2, I'm not sure whether this fix is elegant. Feel free to modify this PR.

### DIFF
--- a/k2/csrc/ragged_test.cu
+++ b/k2/csrc/ragged_test.cu
@@ -265,10 +265,7 @@ TEST(RaggedShapeOpsTest, UnstackRandom) {
       for (size_t i = 0; i < out.size(); ++i) {
         out_ptr.emplace_back(&(out[i]));
       }
-      // There is a bug in `Stack` for stacking a shape itself,
-      // not urgent, so skipping here.
-      // TODO: Remove this line when the bug fixed.
-      if (out.size() == 1) continue;
+
       auto dest = Stack(axis, out.size(), out_ptr.data());
       dest = RemoveEmptyLists(dest, axis);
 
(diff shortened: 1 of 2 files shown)


csukuangfj:
@wgb14 
Could you also add a test to cover

> In my tests, the bug only appears when num_src==1 and axis>1

?

wgb14:
> @wgb14 Could you also add a test to cover
> 
> > In my tests, the bug only appears when num_src==1 and axis>1
> 
> ?

This test has already been covered in `TEST(RaggedShapeOpsTest, UnstackRandom)`
https://github.com/k2-fsa/k2/blob/3cc74f19ccc045a7176a8a96a4b24c4af76e9653/k2/csrc/ragged_test.cu#L249-L278
I just unskipped it.

csukuangfj:
+1

pkufool:
Thanks！

PR #927, opened 2022-03-07
Implement ArrayOfRagged

LvHang:
Implement the ArrayOfRagged class.
The RowSplits()/RowIds() function returns the array of individual shapes.
The MetaRowSplits()/MetaRowIds() function reflects the relationship between the elements and the 'src-level' (each ragged) info.
I'm going to test it.


pkufool:
I only implemented `RowSplits()/RowIds()` that needed by rnnt decoding method.

k2/csrc/array_of_ragged.cu
```diff
@@ -0,0 +1,128 @@
...
+
+  Array2Accessor<int32_t *> row_splits_acc = row_splits_.Accessor(),
+                            row_ids_acc = row_ids_.Accessor();
+  // Bear in mind, when axis == 0, the TotSize() is row_splits.Dim() - 1.
+  // When 0 < axis < NumAxes(), the TotSize() is row_splits.Back().
+  int32_t tot_sizes_data = tot_sizes_.Data();
+
+  for (int32_t i = 1; i < num_axes_; ++i) {
+    for (int32_t j = 0; j < num_srcs_; ++j) {
+      row_splits_acc(i - 1, j) = srcs[j].RowSplits(i).Data();
```
pkufool, line 52: I don't think this line can run on GPUs.

LvHang:
Oh, I wasn't aware of this before. I skimmed Kang's PR just now, and as Kang said, this one tried to implement all the functions. I will try to compare the codes and make them unified.

One quick question: the class should be named as "Array1OfRaggedShape(RaggedShape *src, int32_t num_srcs)" or "ArrayOfRaggedShape(RaggedShape *srcs, int32_t num_srcs)"?

pkufool:
I think it is `Array1OfRaggedShape`.

pkufool:
I think it is `Array1OfRaggedShape`.

LvHang:
OK, I‘m confused as it seems not unified in Dan's comments and definitions.

danpovey:
> OK, I‘m confused as it seems not unified in Dan's comments and definitions.

likely just a typo, but I'm OK with keeping it as ArrayOfRaggedShape, since at the current time we only have one (and unlikely to create another with 2 axes).

k2/csrc/array_of_ragged.cu
danpovey: We actually need to run this on CPU because the pointers are only available from CPU, but that requires that
we initially have an Array1 on CPU, and then transfer it to GPU.

LvHang:
The 'bin/cu_array_of_ragged_test' has passed.
Thank Fangjun and Wei for answering tons of my questions!

In addition, I have a small question when I read the source code about "context", and how to use it in "ParallelRunner".
Assume, we have a GPU context with "StreamA" (i.e. a ```PytorchCudaContext c``` is initialized).
In the code, most of the time (except in "intersect_dense_pruned.cu"), it uses as ```ParallelRunner pr(c); With w(pr.NewStream());```, which doesn't pass the actual parameter to function--NewStream().
But, without actual parameter or "num_work_items < 10k", it will return the default stream (i.e. 0x0)，the "g_stream_override_" will not change. So, in "K2_EVAL" [inside K2_EVAL->EvalDevice(c->GetCudaStream())],  we still use "StreamA".
So, why do we try to use "ParallelRunner" with the default "NewStream()"?

k2/csrc/array_of_ragged.h
```diff
@@ -140,18 +179,28 @@ class Array1OfRaggedShape {
...
   int32_t num_axes_;
+
   Array2<const int32_t *> row_splits_;  // shape [num_axes_ - 1][num_srcs_]
   Array2<const int32_t *> row_ids_;     // shape [num_axes_ - 1][num_srcs_]
-  Array1<int32_t> tot_sizes_;           // dim num_axes_, this is on CPU
+  Array1<int32_t> tot_sizes_;           // dim num_axes_
+
+  Array2<int32_t> meta_row_splits_;     // shape [num_axes_][num_srcs_ + 1]
+  Array2<int32_t> offsets_;             // shape [num_axes_][num_srcs_ + 1]
+  std::vector<Array1<int32_t> > meta_row_ids_;  // dim num_axes_
```
csukuangfj, line 199: If you can use an Array2 to represent the meta_row_splits_, is it possible to also use Array2 for meta_row_ids_?

k2/csrc/array_of_ragged.cu
```diff
@@ -20,35 +22,130 @@
...
+    meta_row_ids_[axis] = meta_row_ids_[axis].To(c_);
+  }
+
+  meta_row_splits_ = meta_row_splits_.To(c_);
+  offsets_ = offsets_.To(c_);
+}
+
+
+template <typename T>
+Array1OfRagged<T>::Array1OfRagged(Ragged<T> *srcs, int32_t num_srcs) :
```
csukuangfj, line 129: Please don't move it from .h to .cu file and delete the explicit instantiations at the end of the file. You will not know what `T` could be in advance.

csukuangfj (changes requested):

k2/csrc/array_of_ragged.h
LvHang: No, in the comments, Dan has said: "This cannot be an Array2 because all the row-ids arrays are of different lengths".

LvHang:
Similar to the change of "MetaRowSplits()", I also change the following statements:
Array2<int32_t> Offsets()  --> const Array2<int32_t> &Offsets() const
Array1<int32_t*> MetaRowIds()  --> Array1<const int32_t *> MetaRowIds()
Array1<int32_t> MetaRowIds(int32_t axis)  --> const Array1<int32_t> &MetaRowIds(int32_t axis) const

Tests have passed.

danpovey:
> The 'bin/cu_array_of_ragged_test' has passed. Thank Fangjun and Wei for answering tons of my questions!
> 
> In addition, I have a small question when I read the source code about "context", and how to use it in "ParallelRunner". Assume, we have a GPU context with "StreamA" (i.e. a `PytorchCudaContext c` is initialized). In the code, most of the time (except in "intersect_dense_pruned.cu"), it uses as `ParallelRunner pr(c); With w(pr.NewStream());`, which doesn't pass the actual parameter to function--NewStream(). But, without actual parameter or "num_work_items < 10k", it will return the default stream (i.e. 0x0)，the "g_stream_override_" will not change. So, in "K2_EVAL" [inside K2_EVAL->EvalDevice(c->GetCudaStream())], we still use "StreamA". So, why do we try to use "ParallelRunner" with the default "NewStream()"?

The concept of ParallelRunner is that sometimes instead of dependencies a->b->c->d, we have things like a->b->d and a->c->d, so we can do b and c in parallel.  CUDA does have the concept of a "stream" that conceptually allows you to take advantage of this situation; and of course in principle we can use CPU streams as well to deal with this more efficiently (in addition to any other parallelization that we might do across streams within K2_EVAL.. which is possible in principle, even though we are not doing it now).

In practice, it turned out that the overhead of creating and deleting streams in CUDA is way too high for it to make sense to use this mechanism.  So in practice, the ParallelRunner thing is currently giving no benefit.  It's possible, though, that we might in future run on different hardware where it would make a difference; or that we might devise a way to have the mechanism be useful.

LvHang:
> The concept of ParallelRunner is that sometimes instead of dependencies a->b->c->d, we have things like a->b->d and a->c->d, so we can do b and c in parallel. CUDA does have the concept of a "stream" that conceptually allows you to take advantage of this situation; and of course in principle we can use CPU streams as well to deal with this more efficiently (in addition to any other parallelization that we might do across streams within K2_EVAL.. which is possible in principle, even though we are not doing it now).
> 
> In practice, it turned out that the overhead of creating and deleting streams in CUDA is way too high for it to make sense to use this mechanism. So in practice, the ParallelRunner thing is currently giving no benefit. It's possible, though, that we might in future run on different hardware where it would make a difference; or that we might devise a way to have the mechanism be useful.

Thanks, Dan, Got it. 
When I tried to know more about the concept of 'Context' of k2, I went through the source codes, so I encountered the above question. I think the ParallelRunner might be useful in the future.
The following words are left for other newbies like me who want to understand and know more about k2's 'Context'. I think the posters (https://medium.com/gpgpu/multi-gpu-programming-6768eeb42e2c; https://on-demand.gputechconf.com/gtc/2014/presentations/S4158-cuda-streams-best-practices-common-pitfalls.pdf, etc) will be helpful to know more about the parallelization of CUDA streams initially. When you read the source code, "default_context.cu" will show the basic conceptual implementation of k2's 'Context'. It's good to read. But, from the CMakeFile, you will know the 'pytorch_context' is actually used now.

pkufool:
Would you please add some tests for `Offsets`, I think this is good to merge after fixing this. Thanks!

LvHang:
The ```Offsets``` test code has been added and the bin/cu_array_of_ragged_test passed on my machine.

LvHang:
The "linear_fsa_with_self_loops_test.py" seems to cause the style check fails.

csukuangfj:
> The "linear_fsa_with_self_loops_test.py" seems to cause the style check fails.

It should be fixed with the latest master.

LvHang:
> > The "linear_fsa_with_self_loops_test.py" seems to cause the style check fails.
> 
> It should be fixed with the latest master.

It says the following line is too long:
https://github.com/k2-fsa/k2/blob/d9778653fab7a3a121463174c1816a270a3d14f2/k2/python/tests/linear_fsa_with_self_loops_test.py#L58

csukuangfj:
> > > The "linear_fsa_with_self_loops_test.py" seems to cause the style check fails.
> 
> > 
> 
> > It should be fixed with the latest master.
> 
> 
> 
> It says the following line is too long (I checked the latest code, It's true):
> 
> https://github.com/k2-fsa/k2/blob/d9778653fab7a3a121463174c1816a270a3d14f2/k2/python/tests/linear_fsa_with_self_loops_test.py#L58

Rebase on the latest master should solve your problem.

LvHang:
> > > > The "linear_fsa_with_self_loops_test.py" seems to cause the style check fails.
> > 
> > 
> > > 
> > 
> > 
> > > It should be fixed with the latest master.
> > 
> > 
> > It says the following line is too long (I checked the latest code, It's true):
> > https://github.com/k2-fsa/k2/blob/d9778653fab7a3a121463174c1816a270a3d14f2/k2/python/tests/linear_fsa_with_self_loops_test.py#L58
> 
> Rebase on the latest master should solve your problem.

Thanks, I missed the "#noqa" in the latest version. Done.

pkufool:
merging, thanks!

PR #938, opened 2022-03-18
Minor fixes for RNN-T decoding.

csukuangfj:


k2/csrc/nbest.cu
```diff
@@ -48,7 +48,7 @@ inline bool Leq(T a1, T a2, T a3, T b1, T b2, T b3) {
 */
 template <typename T>
 static void RadixPass(const T* a, T* b, const T* r, T n, T K) {
-  std::vector<T> c(K + 1, 0); // counter array
+  std::vector<T> c(K + 1, 0);  // counter array
```
pkufool, line 51: Do we need a 0 here? Your comment on line 72 said "entries are set to zero by default".

pkufool:
Thanks! Sorry, I should have checked them more carefully.

k2/csrc/nbest.cu
csukuangfj: I just added `0` to other usages of `vector` to make the fact explicit that entries default to be zero intialized.

PR #942, opened 2022-03-30
Update v2.0-pre

pkufool:


pkufool:
@danpovey @csukuangfj  I have a suggestion that we move v2.0-pre to a separate repository, say k2-runtime. In that repository, we need only feature extractor (i.e. kaldifeat or something)、C++ Fsa、torch script thing... and we can make k2 a submodule of it. And we can also put our device applications and server side code their.  What do you think?

pkufool:
> What're the disadvantages of putting everything in k2?

I don't think it bothers me to keep all these in the k2 repository.  Last time I suggested to merge v2.0-pre into master and you had different ideas, so I want to know your plans. What will be in v2.0 and where to put the code of our prototypes(i.e. the device apps, server side framework).

pkufool:
> But I don't see the connection of this PR to that. This contains RNN-T decoding which we would also want to use in Python. 

Yes, it has nothing to do with this PR, I just thought of this when I was merging mater into v2.0-pre,  we may discuss this in a separate issue.

> Also, I think for prototyping purposes we can just use k2 itself. Later when it comes time to actually build some specific application, there may be lots of extra constraints and they may want to just extrat some code from k2, as needed.

Ok, let's keep these two branches for now, we will decide it at the time we have to.

PR #943, opened 2022-03-31
support half precision for rnnt loss

pkufool:


pkufool:
See comments here https://github.com/k2-fsa/icefall/pull/283#issuecomment-1086833915.

pkufool:
See comments here https://github.com/k2-fsa/icefall/pull/283#issuecomment-1086833915.

PR #945, opened 2022-04-03
Rnnt allow different encoder/decoder dims

danpovey:


danpovey:
not tested yet, wait..

PR #949, opened 2022-04-08
Check the versions of PyTorch and CUDA at the import time.

csukuangfj:
If the version of PyTorch that is used to run k2 is different from the one used to compile k2, then it throws an `ImportError` exception.

Previously, it threw a mysterious error saying that some symbol cannot be found.

----

Assume k2 is compiled using torch 1.11.0 and  torch 1.10 is used to run it.

## Before this PR

It throws the following error while `import k2`:
```
ImportError: /ceph-fj/fangjun/open-source-2/k2-master-2/b3/lib/libk2context.so: undefined symbol: _ZN3c1015SmallVectorBaseIjE8grow_podEPvmm
```

## With this PR
```
  File "<string>", line 1, in <module>
  File "/ceph-fj/fangjun/open-source-2/k2-master-2/k2/python/k2/__init__.py", line 6, in <module>
    raise ImportError(
ImportError: k2 was built using PyTorch 1.11.0+cu102
But you are using PyTorch 1.10.0+cu102 to run it
```


desh2608:
If I have several versions of k2 built (for example with different PyTorch or CUDA versions), this causes an error because the last built version overwrites the torch_version.py file. A simple hack is to comment out the checks.

csukuangfj:
> If I have several versions of k2 built (for example with different PyTorch or CUDA versions), this causes an error because the last built version overwrites the torch_version.py file. A simple hack is to comment out the checks.

Yes, that is what I am doing.

An alternative is to provide an environment variable and skip the check if the environment variable is set.

PR #951, opened 2022-04-12
Fix precision

pkufool:
Something wrong with the precision if changing `pow` to `powf` (using `powf` to support windows build, see https://github.com/k2-fsa/k2/pull/946/files#diff-b7474255c771be7e6f33ddf04f0e5c9cb7c0588ccb7151713699d46841bc9e18) , the WER of test-clean increases from 2.72% -> 82.3%.

I have not figured out the root cause,  **do not merge**.

### DIFF
--- a/k2/csrc/rnnt_decode.cu
+++ b/k2/csrc/rnnt_decode.cu
@@ -159,8 +159,8 @@ void RnntDecodingStreams::GetContexts(RaggedShape *shape,
         int64_t state_value = states_values_data[state_idx01x],
                 context_state = state_value / num_graph_states,
                 exp = decoder_history_len - col,
-                state = context_state % (int64_t)K2_POW(vocab_size, exp);
-        state = state / (int64_t)K2_POW(vocab_size, exp - 1);
+                state = context_state % Pow(vocab_size, exp);
+        state = state / Pow(vocab_size, exp - 1);
         contexts_acc(row, col) = state;
       });
 }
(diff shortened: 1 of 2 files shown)


k2/csrc/rnnt_decode.cu
```diff
@@ -159,8 +159,8 @@ void RnntDecodingStreams::GetContexts(RaggedShape *shape,
         int64_t state_value = states_values_data[state_idx01x],
                 context_state = state_value / num_graph_states,
                 exp = decoder_history_len - col,
-                state = context_state % (int64_t)powf(vocab_size, exp);
-        state = state / (int64_t)powf(vocab_size, exp - 1);
+                state = context_state % (int64_t)pow(vocab_size, exp);
```
csukuangfj, line 162: You can use
```cpp
#ifndef _MSC_VER
#define K2_POW pow
#else
#define K2_POW powf
#endif

Comment (pkufool) [reply k2/csrc/rnnt_decode.cu:162]: Yes, I agree. But I am not sure whether this bug would happen on windows platform.

Comment (csukuangfj) [reply k2/csrc/rnnt_decode.cu:162]: One thing you can do is to add a CI test in icefall using fast beam search for windows.

- https://k2-fsa.org/nightly/whl/k2-1.14.dev20220409+cpu.torch1.10.0-cp38-cp38-win_amd64.whl
- https://github.com/k2-fsa/icefall/blob/master/.github/workflows/run-librispeech-2022-03-12.yml

Comment (danpovey) [reply k2/csrc/rnnt_decode.cu:162]: It needs that number to be exact, not approximate.  I think it would make more sense to code an int64 version of 'pow', rather than using floating-point versions.

**Comment by pkufool** [comment]
I checked that `powf` runs well on CPU but breaks CUDA decoding. `pow` breaks the CUDA building on windows platform.  

We don't have a GPU Windows machine, so not sure if `powf` will break CUDA decoding on Windows.  Currently, I will use the macros shows below to fix this issue.  We will fix CUDA decoding errors on Windows when someone reports it.
```
#ifndef _MSC_VER
#define K2_POW pow
#else
#define K2_POW powf
#endif
```

pkufool:
> @csukuangfj I think you missed my comment on the issue or PR... we shouldn't really be using pow or powf for this, we should compute it with a loop that multiplies in int64_t. That would be much safer.

Oh, yes. I didn't notice that. I  see.

PR #962, opened 2022-04-20
Fix invert

pkufool:


k2/python/k2/fsa.py
```diff
@@ -930,9 +930,16 @@ def invert_(self) -> 'Fsa':
             raise RuntimeError('current invert_ method only supports case '
                                'where aux_labels is a tensor')
 
+        self.__dict__['_properties'] = None
+
+        # Fix aux_labels in the final arcs to be -1, or the inverted fsa will
+        # be invalid.
+        self.aux_labels = torch.where(self.labels == -1, -1, self.aux_labels)
```
pkufool, line 937: @danpovey  Do you mean we can only change the 0 in aux_label to -1, like follows:
```
mask = torch.logical_and(self.labels == -1,  self.aux_labels == 0)
self.aux_labels = torch.where(mask, -1, self.aux_labels)
```

k2/python/k2/fsa.py
```diff
@@ -930,9 +930,16 @@ def invert_(self) -> 'Fsa':
...
+
         self.rename_tensor_attribute_('labels', '__temp')
         self.rename_tensor_attribute_('aux_labels', 'labels')
         self.rename_tensor_attribute_('__temp', 'aux_labels')
+        self.properties
```
csukuangfj, line 942: You can use
```python
_k2.fix_final_labels(self.arcs, None)
```

pkufool:
FixFinalLabel works fine, it will check the label can only be 0 or -1.

https://github.com/k2-fsa/k2/blob/b173c11ba379e2da0056281fe6b2d56f081419be/k2/csrc/fsa_utils.cu#L2760-L2766

PR #965, opened 2022-04-25
Clarify some comments in rm_epsilon; fix possible bug there.

danpovey:


danpovey:
RE that change in rm_epsilon.cu: that loop would have looped indefinitely if those 2 arrays were different after the 1st iter.
So it looks to me like maybe the epsilon removal algorithm was never tested for inputs with 2 successive epsilons???  My bad, I suppose, but it would be nice if someone e.g. @pkufool could fix this.

pkufool:
> but it would be nice if someone e.g. @pkufool could fix this.

Sure.

k2/csrc/rm_epsilon.cu
```diff
@@ -767,9 +767,9 @@ void ComputeEpsilonClosure(FsaVec &epsilon_fsa, FsaVec *closure_fsa,
   // We repeatedly call ComputeEpsilonClosureOneIter() until there is no further
   // change in the FsaVec (this can be by simple comparison on arcs vector,
   // since thanks to sorting the order is deterministic).
-  Array1<Arc> epsilon_fsa_arcs = epsilon_fsa.values;
+  Array1<Arc> &epsilon_fsa_arcs = epsilon_fsa.values;
```
danpovey, line 770: Actually, sorry, I realized this is not the source of the problem, I did not notice the statement below, `closure_fsa_arcs = cur_iter_closure_fsa.values;`
I will refactor a little for clarity though.

PR #954, opened 2022-04-15
Add max cpu mem allocation constraint

pkufool:
```
(rnnt2) kangwei@de-74279-k2-train-1-0307195509-54c966b95f-rtpfq:~/code/k2/build$ K2_MAX_CPU_MEM_ALLOCATE=1024 python
Python 3.8.0 (default, Nov  6 2019, 21:49:08)
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import k2
>>> a = k2.ctc_topo(5)
>>> a = k2.ctc_topo(20)
[F] /ceph-hw/kangwei/code/k2/k2/csrc/pytorch_context.cu:95:virtual void* k2::PytorchCpuContext::Allocate(std::size_t, void**) Check failed: static_cast<int64_
t>(bytes) <= max_bytes (1848 vs. 1024)


[ Stack-Trace: ]
/ceph-hw/kangwei/code/k2/build/lib/libk2_log.so(k2::internal::GetStackTrace()+0x5b) [0x7f743f40582a]
/ceph-hw/kangwei/code/k2/build/lib/libk2context.so(k2::internal::Logger::~Logger()+0x44) [0x7f743f8910ba]
/ceph-hw/kangwei/code/k2/build/lib/libk2context.so(k2::PytorchCpuContext::Allocate(unsigned long, void**)+0x14a) [0x7f743fb97ab2]
/ceph-hw/kangwei/code/k2/build/lib/libk2context.so(k2::NewRegion(std::shared_ptr<k2::Context>, unsigned long)+0xb0) [0x7f743f8d089f]
/ceph-hw/kangwei/code/k2/build/lib/libk2context.so(k2::Array1<int>::Init(std::shared_ptr<k2::Context>, int, k2::Dtype)+0x22a) [0x7f743f895612]
/ceph-hw/kangwei/code/k2/build/lib/libk2context.so(k2::Array1<int>::Array1(std::shared_ptr<k2::Context>, int, k2::Dtype)+0x5f) [0x7f743f893245]
/ceph-hw/kangwei/code/k2/build/lib/libk2context.so(k2::CtcTopo(std::shared_ptr<k2::Context> const&, int, bool, k2::Array1<int>*)+0x492) [0x7f743f8f426e]
/ceph-hw/kangwei/code/k2/build/lib/_k2.cpython-38-x86_64-linux-gnu.so(+0x124a7c) [0x7f74408eca7c]
[...]

PR #966, opened 2022-04-25
Add abort in log.h

yaozengwei:
Now users can set environment variable `K2_ABORT=1`, then it will call `abort()` if a K2_CHECK fails.

### DIFF
--- a/k2/csrc/log.h
+++ b/k2/csrc/log.h
@@ -192,9 +192,7 @@ class Logger {
       	abort();
 	// raise(SIGTERM);
       }
-      else {
-     	throw std::runtime_error(kErrMsg);
-      }
+      throw std::runtime_error(kErrMsg);
       // abort();
       //
       // NOTE: abort() will terminate the program immediately without
--- a/k2/csrc/log.h
+++ b/k2/csrc/log.h
@@ -32,7 +32,6 @@
 #define K2_CSRC_LOG_H_
 
 #include <cassert>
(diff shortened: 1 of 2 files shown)


k2/csrc/log.h
```diff
@@ -107,6 +107,15 @@ std::string GetStackTrace();
  */
 K2_CUDA_HOSTDEV LogLevel GetCurrentLogLevel();
 
+inline bool EnableAbort() {
```
csukuangfj, line 110: Please follow #954 to also modify the file `k2/python/k2/version/version.py`.

k2/csrc/log.h
```diff
@@ -179,11 +188,18 @@ class Logger {
...
+      }
+      else {
+     	throw std::runtime_error(kErrMsg);
+      }
       // abort();
       //
       // NOTE: abort() will terminate the program immediately without
       // printing the Python stack backtrace.
-      throw std::runtime_error(kErrMsg);
+      // throw std::runtime_error(kErrMsg);
```
csukuangfj, line 202:
```cpp
if (EnableAbort()) {
  abort();
}

throw std::runtime_error(kErrMsg);
```

is simpler.  

From https://en.cppreference.com/w/cpp/utility/program/abort, 
`abort` will never return, so you don't need to a use a separate `else` here.

yaozengwei: I see.

PR #967, opened 2022-04-25
Less pruning on last few frames for IntersectDensePruned

pkufool:
See https://github.com/k2-fsa/icefall/issues/292 and https://github.com/k2-fsa/icefall/issues/292#issuecomment-1093837846 for more details.

### DIFF
--- a/k2/csrc/intersect_dense_pruned.cu
+++ b/k2/csrc/intersect_dense_pruned.cu
@@ -529,6 +529,7 @@ class MultiGraphDenseIntersectPruned {
                     [arc] dim, so it's just [fsa_id][arc]
                     It is conceptually unchanged by this operation but non-const
                     because row-ids of its shape may need to be generated.
+       @param [in] t  Which frame we are processing (i.e. the frame index).
        @return      Returns a vector of log-likelihood cutoffs, one per FSA (the
                     cutoff will be -infinity for FSAs that don't have any active
                     states).  The cutoffs will be of the form: the best score
@@ -537,7 +538,7 @@ class MultiGraphDenseIntersectPruned {
                     'search_beam_' as long as the number of active states in
                     each FSA is between min_active and max_active.
   */
-  Array1<float> GetPruningCutoffs(Ragged<float> &arc_end_scores) {
+  Array1<float> GetPruningCutoffs(Ragged<float> &arc_end_scores, int32_t t) {
     NVTX_RANGE(K2_FUNC);
     int32_t num_fsas = arc_end_scores.shape.Dim0();
 
@@ -562,6 +563,8 @@ class MultiGraphDenseIntersectPruned {
           min_active = min_active_;
     K2_CHECK_LT(min_active, max_active);
 
+    const int32_t *b_fsas_row_splits1 = b_fsas_.shape.RowSplits(1).Data();
+
     Array1<float> cutoffs(c_, num_fsas);
     float *cutoffs_data = cutoffs.Data();
 
@@ -570,10 +573,17 @@ class MultiGraphDenseIntersectPruned {
           float best_loglike = max_per_fsa_data[i],
                 dynamic_beam = dynamic_beams_data[i];
           int32_t active_states = arc_end_scores_row_splits1_data[i + 1] -
-                                  arc_end_scores_row_splits1_data[i];
+                                  arc_end_scores_row_splits1_data[i],
+                  final_t = b_fsas_row_splits1[i+1] - b_fsas_row_splits1[i];
+          float current_min_active = min_active;
+          // Do less pruning on the final few frames, to ensure we don't prune
+          // away final states.
+          if (t + 5 < final_t) {
+              current_min_active = min(min_active, max_active / 2);
+          }
           if (active_states <= max_active) {
             // Not constrained by max_active...
-            if (active_states >= min_active || active_states == 0) {
+            if (active_states >= current_min_active || active_states == 0) {
               // Neither the max_active nor min_active constraints
               // apply.  Gradually approach 'beam'
               // (Also approach 'beam' if active_states == 0; we might as
@@ -593,6 +603,9 @@ class MultiGraphDenseIntersectPruned {
             // max_active active states.
             dynamic_beam *= 0.8;
           }
+          // no pruning on last frame; we want all final-arcs.
+          if (t == final_t) dynamic_beam = 1.0e+10;
+
           dynamic_beams_data[i] = dynamic_beam;
           cutoffs_data[i] = best_loglike - dynamic_beam;
         });
@@ -744,7 +757,7 @@ class MultiGraphDenseIntersectPruned {
     Ragged<float> ai_loglikes(arc_info.shape, ai_data_array1);
 
     // `cutoffs` is of dimension num_fsas.
-    Array1<float> cutoffs = GetPruningCutoffs(ai_loglikes);
+    Array1<float> cutoffs = GetPruningCutoffs(ai_loglikes, t);
     float *cutoffs_data = cutoffs.Data();
 
     // write certain indexes (into ai.values) to state_map_.Data().  Keeps
--- a/k2/csrc/intersect_dense_pruned.cu
+++ b/k2/csrc/intersect_dense_pruned.cu
@@ -598,13 +598,14 @@ class MultiGraphDenseIntersectPruned {
             }
           } else {
             // We violated the max_active constraint -> decrease beam
-            if (dynamic_beam > default_beam) dynamic_beam = default_beam;
+            // if (dynamic_beam > default_beam) dynamic_beam = default_beam;
             // Decrease the beam as long as we have more than
             // max_active active states.
             dynamic_beam *= 0.8;
           }
           // no pruning on last frame; we want all final-arcs.
-          if (t == final_t) dynamic_beam = 1.0e+10;
+          // -1 because t starts from 0.
+          if (t == final_t - 1) dynamic_beam = 1.0e+10;
 
           dynamic_beams_data[i] = dynamic_beam;
           cutoffs_data[i] = best_loglike - dynamic_beam;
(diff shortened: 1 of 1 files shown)


videodanchik:
Alright, after your last changes, then your modifications are the same as mine, but with it, I got doubled, the number of empty utterances in the decoding of multiple datasets. OK let me do decoding with exact your last changes in this PR and return with results tomorrow, for now I can provide some results for decoding with `master` branch and using my custom model for decoding `ami_dev` and `ted_dev` (each column is a separate model, last column is the strongest one in terms of WER, first is the weakest). I present 2 numbers separated by `/` first is empty utterances num., second - the total number of utterances. Beam is 12.
image
this is for beam 15
image

k2/csrc/intersect_dense_pruned.cu
```diff
@@ -570,10 +573,17 @@ class MultiGraphDenseIntersectPruned {
           float best_loglike = max_per_fsa_data[i],
                 dynamic_beam = dynamic_beams_data[i];
           int32_t active_states = arc_end_scores_row_splits1_data[i + 1] -
-                                  arc_end_scores_row_splits1_data[i];
+                                  arc_end_scores_row_splits1_data[i],
+                  final_t = b_fsas_row_splits1[i+1] - b_fsas_row_splits1[i];
```
danpovey, line 577: Is this really final_t.. wouldn't final_t be b_fsas_row_splits1[i+1] - b_fsas_row_splits1[i] - 1, if we intend final_t to be the index of the last frame (i.e. the frame on which final-arcs happen)?

videodanchik:
Ok, once again, with this PR I'm getting the following results (I considered utterance as empty if the reference transcript includes 3 or more words and hypothesis transcript is empty)

Master:
image

This PR:
image
I tried to decode with beam 15, but got CUDA OOM with my batch size, so can provide the table later if necessary, need to wait for finishing the decoding with smaller batch.

Additionally here are the WER I got with master vs this branch.

Master:
image

This PR:
image

k2/csrc/intersect_dense_pruned.cu
videodanchik: I did it the same during my experiments, similar to this line (https://github.com/k2-fsa/k2/blob/510cd726f91d5c6473a37f1714cc13e57b9c056b/k2/csrc/intersect_dense_pruned.cu#L410).

danpovey:
I'm a little confused how this PR could be making WERs worse.  Check if the results particular change toward the end of sentences.  And you could perhaps try printing out the beams, and decoding one minibatch with the vs. the original code (with print statements) to see how the beams change toward the end of the batch.

videodanchik:
I tried to print out the beams and didn't succeed, please provide some instructions on how to do it (may be @pkufool can make a temporal commit to this branch with an example). Instead I provide several examples of lattices where `master` branch code succeeds to produce transcript and this PR fails:

[lattice_1_master.pt.zip](https://github.com/k2-fsa/k2/files/8566906/lattice_1_master.pt.zip)
[lattice_1_this_pr.pt.zip](https://github.com/k2-fsa/k2/files/8566907/lattice_1_this_pr.pt.zip)
[lattice_connected_1_master.pdf](https://github.com/k2-fsa/k2/files/8566909/lattice_connected_1_master.pdf)
[lattice_whole_1_master.pdf](https://github.com/k2-fsa/k2/files/8566910/lattice_whole_1_master.pdf)
[lattice_whole_1_this_pr.pdf](https://github.com/k2-fsa/k2/files/8566911/lattice_whole_1_this_pr.pdf)

[lattice_2_master.pt.zip](https://github.com/k2-fsa/k2/files/8566913/lattice_2_master.pt.zip)
[lattice_2_this_pr.pt.zip](https://github.com/k2-fsa/k2/files/8566914/lattice_2_this_pr.pt.zip)
[lattice_connected_2_master.pdf](https://github.com/k2-fsa/k2/files/8566915/lattice_connected_2_master.pdf)
[lattice_whole_2_master.pdf](https://github.com/k2-fsa/k2/files/8566916/lattice_whole_2_master.pdf)
[lattice_whole_2_this_pr.pdf](https://github.com/k2-fsa/k2/files/8566917/lattice_whole_2_this_pr.pdf)

[lattice_3_master.pt.zip](https://github.com/k2-fsa/k2/files/8566919/lattice_3_master.pt.zip)
[lattice_3_this_pr.pt.zip](https://github.com/k2-fsa/k2/files/8566922/lattice_3_this_pr.pt.zip)
[lattice_connected_3_master.pdf](https://github.com/k2-fsa/k2/files/8566923/lattice_connected_3_master.pdf)
[lattice_whole_3_master.pdf](https://github.com/k2-fsa/k2/files/8566924/lattice_whole_3_master.pdf)
[lattice_whole_3_this_pr.pdf](https://github.com/k2-fsa/k2/files/8566932/lattice_whole_3_this_pr.pdf)

I have an impression that for some reason this PR cuts off the sub-lattice that has a final state. Please, try out yourself on some more or less challenging dataset where you have a `conformer_ctc` pretrained model like GigaSpeech for example. I bet you will find some empty hyps there.

videodanchik:
Ah, okey I have added something like 
```
for (int32_t k = 0; k != num_fsas; ++k) {
    std::cout << dynamic_beams_data[k] << ", ";
}
std::cout << std::endl;
```
and it didn't work for GPU (got segmentation fault), but it worked fine on CPU, I update the dynamic_beam history for a batch from my previous comment.

k2/csrc/intersect_dense_pruned.cu
pkufool: Ah, yes. final_t should be  `b_fsas_row_splits1[i+1] - b_fsas_row_splits1[i] - 1` . But I think this should not be the reason that making the results worse.

pkufool:
> Ah, okey I have added something like
> 
> ```
> for (int32_t k = 0; k != num_fsas; ++k) {
>     std::cout << dynamic_beams_data[k] << ", ";
> }
> std::cout << std::endl;
> ```
> 
> and it didn't work for GPU (got segmentation fault), but it worked fine on CPU, I update the dynamic_beam history for a batch from my previous comment.

You can use `K2_LOG` to print out the whole array.

```
K2_LOG(INFO) << "frame : " << t << "\t beams : " << dynamic_beam_;
```

Note, put this line out of the `K2_EVAL`,  for example, line 612 in this PR, just before we return cutoffs.

pkufool:
I think the reason of WER degradation might be it soon violated the `max-activate` after we increased the `min-activate`.
You can see in the code below, in line 591, the beam value will reset to initial beam then scale with 0.8 if `max-activate` violates. So the beam values might be 12 -> 15 -> 18.75 -> 9.6.

Just a analysis according to the code, let's see the beam values printing out by @videodanchik .

https://github.com/k2-fsa/k2/blob/510cd726f91d5c6473a37f1714cc13e57b9c056b/k2/csrc/intersect_dense_pruned.cu#L568-L598

videodanchik:
Thanks, folks for help! So here are the logs for master and this PR. Each file includes 3 utterances from my previous comment.
Master:
[master.txt](https://github.com/k2-fsa/k2/files/8567870/master.txt)
This PR:
[this_pr.txt](https://github.com/k2-fsa/k2/files/8567873/this_pr.txt)

videodanchik:
And with replacement `b_fsas_row_splits1[i+1] - b_fsas_row_splits1[i] - 1` I I still got first and last utterances empty.
[this_pr_update.txt](https://github.com/k2-fsa/k2/files/8567942/this_pr_update.txt)

pkufool:
That is it, it violates `max-states` at frame `final_t - 3`.
image

image

image

So, do we need to reset the beam  to  the initial value? @danpovey  I think we can just delete that line.

danpovey:
Yes, you could definitly try deletng that line and see what happens.  I hope it doesn't increase the chances of memory blowup in the normal case.

videodanchik:
Hi, I did a couple of experiments and decode several datasets. At first, I want to say that after complete commenting out `if (dynamic_beam > default_beam) dynamic_beam = default_beam;` I got some reduction of WER and the number of empty utterances, though I run into several CUDA OOM errors for beam `15.0`, unless I decreased `max_dur` for batch decoding. Anyway, to add some memory safety I wrapped [`else`](https://github.com/k2-fsa/k2/blob/master/k2/csrc/intersect_dense_pruned.cu#L589-L598) statement like following:

```
} else {
  // We don't want to modify dynamic_beam for the last few frames
  // in order to avoid final states pruning.
  if (t + 5 < final_t) {
    // We violated the max_active constraint -> decrease beam
    if (dynamic_beam > default_beam) dynamic_beam = default_beam;

    // Decrease the beam as long as we have more than
    // max_active active states.
    dynamic_beam *= 0.8;
  }
}
```
to not reset `dynamic_beam` for the last few frames (everything remaining code I keep like in this PR). That helped with OOM, so I tried decoding for `t+5` and `t+10` and got results for my model:
image
As you can see removing pruning the last 5 frames really helped when the beam was 12.0 and further increasing to 10 last frames didn't help. I tried to decode several utterances that still were empty after `one_best_decoding`, though `k2.connect(lattice)` returned not empty lattice this time. I am attaching 2 examples:
[lattice1.pt.zip](https://github.com/k2-fsa/k2/files/9059398/lattice1.pt.zip)
[lattice1.pdf](https://github.com/k2-fsa/k2/files/9059399/lattice1.pdf)
[lattice2.pt.zip](https://github.com/k2-fsa/k2/files/9059401/lattice2.pt.zip)
[lattice2.pdf](https://github.com/k2-fsa/k2/files/9059402/lattice2.pdf)
This is already a very rare cases, but I'm curious if there could be any corner cases bug or something...

danpovey:
@videodanchik  Great, thanks!    Perhaps you could create a branch with your change, modified from the branch in this PR, and we can merge that?

videodanchik:
Sure.

videodanchik:
I make a PR, though there's still some weird stuff is happening with `one_best_decoding` of 2 lattices from my previous comment.

pkufool:
Closed by #1004

danpovey:
I think we overloaded things like
 std::cout << [ragged array]

PR #974, opened 2022-05-15
Port kaldi::ParseOptions for parsing commandline options.

csukuangfj:
Note: gflags is not used here since k2 may be used as a dependency for other projects and other projects may use a different version of gflags. If gflags was used, there would be some conflicts for linking with different versions of gflags.


danpovey:
Thanks!  Yes, makes sense I suppose.

csukuangfj:
Ok, merging.

PR #975, opened 2022-05-18
Greedy search and modified beam search for pruned stateless RNN-T.

csukuangfj:
Also, here is a proposal to support both streaming decoding and non-streaming decoding.

Untitled drawing

## Clients

The client sends audio samples to the server for recognition and receives results from the server.
It can either send a whole sound file or send samples in a streaming fashion.

There can be multiple clients sending at the same time.

We can provide APIs for various languages, e.g., C++, C, java, python, etc. (See https://github.com/alphacep/vosk-api for possible languages) and we can provide more than one protocol for communication, e.g., gRPC and WebSockets.

## Server

The server accepts connections from the clients. For each connection, it assigns a producer to handle it.

The server is implemented in C++.

## Producers

There will be a pool of producers. A producer contains 
- a socket, for communicating with the client
- a streaming feature extractor, for converting audio samples to features
- partial decoded results
- states of the encoder neural network in case of streaming recognition
- a handle to the queue. It sends feature frames, model states, and partial decoded results to the queue, which will be handled by the consumer.

## Consumers

There will also be a pool of consumers. A consumer is responsible for extracting data from the queue, **batching** them up, sending them to the Nnet computer and decoder, fetching updated model states and partial results from the nnet computer and decoder and sending them to the producer.

## Nnet computers

There will be a pool of Nnet computers. Each computer accepts batched data from the consumer and feeds them into the neural network model

## Decoders

There will be a pool of decoders. It takes input from the Nnet computer, fetches partial decoded results from the producer,
runs decoding (i.e., greedy search, modified beam search, fast beam search). The decoded results are returned to the consumer, which are sent to the producer.

### DIFF
--- a/k2/csrc/ragged_ops.h
+++ b/k2/csrc/ragged_ops.h
@@ -1202,7 +1202,7 @@ RaggedShape ComposeRaggedShapes3(const RaggedShape &a, const RaggedShape &b,
   If cached_tot_sizeN is not -1, it must equal the total size on
   that axis which will equal the last element of row_splitsN (if
   provided) and must equal the row_idsN.Dim(), if provided. See
-  documentation above for RagggedShape2 for details.
+  documentation above for RaggedShape2 for details.
 
   We also require that (supposing both row_splitsN and row_idsN are non-NULL):
   row_splits1[row_splits1.Dim() - 1] == row_ids1.Dim()
--- a/k2/torch/bin/CMakeLists.txt
+++ b/k2/torch/bin/CMakeLists.txt
@@ -64,7 +64,6 @@ add_executable(online_decode ${online_decode_srcs})
 set_property(TARGET online_decode PROPERTY CXX_STANDARD 14)
 target_link_libraries(online_decode ${bin_dep_libs})
 
-
 #-------------------------------------------
 #      rnnt demo
 #-------------------------------------------
@@ -77,3 +76,14 @@ add_executable(rnnt_demo ${rnnt_demo_srcs})
 set_property(TARGET rnnt_demo PROPERTY CXX_STANDARD 14)
 target_link_libraries(rnnt_demo ${bin_dep_libs})
 
+#-------------------------------------------
+#      pruned stateless transducer
+#-------------------------------------------
+set(pruned_stateless_transducer_srcs pruned_stateless_transducer.cu)
+if(NOT K2_WITH_CUDA)
+  transform(OUTPUT_VARIABLE pruned_stateless_transducer_srcs SRCS ${pruned_stateless_transducer_srcs})
+endif()
+
+add_executable(pruned_stateless_transducer ${pruned_stateless_transducer_srcs})
+set_property(TARGET pruned_stateless_transducer PROPERTY CXX_STANDARD 14)
+target_link_libraries(pruned_stateless_transducer ${bin_dep_libs})
--- a/k2/torch/bin/pruned_stateless_transducer.cu
+++ b/k2/torch/bin/pruned_stateless_transducer.cu
@@ -1,0 +1,171 @@
+/**
+ * Copyright      2022  Xiaomi Corporation (authors: Fangjun Kuang)
+ *
+ * See LICENSE for clarification regarding multiple authors
+ *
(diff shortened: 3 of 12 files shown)


pzelasko:
Nice! 

Did you consider existing solutions such as Triton? I know they support at least some of your requirements (e.g. dynamic batching), and additionally if you’re able to convert the model to some optimized runtime like ONNX/TensorRT, it can be pretty fast. https://github.com/triton-inference-server/server

danpovey:
I am open minded about the Triton thing but I don't know much about such things.
One option is to have some kind of demo of a run-time in a separate repo from k2, so that we don't pollute k2's core library with too much non-core stuff.

csukuangfj:
> One option is to have some kind of demo of a run-time in a separate repo from k2, so that we don't pollute k2's core library with too much non-core stuff.

Yes, will do that.

I am trying to implement the above framework in Python first.  I am using asyncio to handle communications and the thread pools are managed in  Python. 

Only the CPU-bound tasks such as neural network computation and decoding are implemented in C++, which is called by Python threads with the global interpreter lock being released.  All other parts are in Python. I will create a PR soon.

@pzelasko Thanks for the information. I will look into it.

yuekaizhang:
There is a triton solution for offline ASR in Wenet. https://github.com/wenet-e2e/wenet/tree/main/runtime/server/x86_gpu
Also, it is possible for streaming ASR with triton stateful model. I would love to implement a similar triton solution for icefall if you're interested to triton. @csukuangfj

danpovey:
Sounds cool to me!  As far as I am concerned it would be great if you would implement a Triton solution, but for the time being I'm not ready to commit the team to a Triton-based runtime because we don't know how much work it would be and whether it would work with some of the C++ extensions we are doing.  (Also I don't want to shift all of our current effort over to creating the runtime, because we still have work to do on the core modeling side).  So basically, I think it would be great if you would do that, but for the time being it would be a kind of demo that we'd put somewhere in the k2-fsa organization; we wouldn't be committing the team to the Triton framework at this stage.

Let's see what @csukuangfj and others think too, though.

csukuangfj:
> https://github.com/wenet-e2e/wenet/tree/main/runtime/server/x86_gpu

Yes, you are very welcome to add triton.  

Could you add it to https://github.com/k2-fsa/sherpa

yuekaizhang:
Sounds good. Let me first try if k2 works with triton python backend. Would updates under sherpa.

csukuangfj:
> Sounds good. Let me first try if k2 works with triton python backend. Would updates under sherpa.

Thanks a lot!

csukuangfj:
Usage:

```bash

nn_model=/ceph-fj/fangjun/open-source-2/icefall-master-2/egs/librispeech/ASR/pruned_transducer_stateless3/exp/cpu_jit.pt
bpe_model=/ceph-fj/fangjun/open-source-2/icefall-master-2/egs/librispeech/ASR/data/lang_bpe_500/bpe.model
wav1=/ceph-fj/fangjun/open-source-2/icefall-master-2/egs/librispeech/ASR/pruned_transducer_stateless3/exp/test_wavs/1089-134686-0001.wav
wav2=/ceph-fj/fangjun/open-source-2/icefall-master-2/egs/librispeech/ASR/pruned_transducer_stateless3/exp/test_wavs/1221-135766-0001.wav
wav3=/ceph-fj/fangjun/open-source-2/icefall-master-2/egs/librispeech/ASR/pruned_transducer_stateless3/exp/test_wavs/1221-135766-0002.wav

for m in greedy_search modified_beam_search; do
./bin/pruned_stateless_transducer \
  --nn-model=$nn_model \
  --bpe-model=$bpe_model \
  --use-gpu=true \
  --decoding-method=$m \
  $wav1 \
  $wav2 \
  $wav3
done
```

Note: You can download `nn_model`, `bpe_model` and test waves from
https://huggingface.co/csukuangfj/icefall-asr-librispeech-pruned-transducer-stateless3-2022-05-13

---

Note: Our server framework is moved to a separate repo:
https://github.com/k2-fsa/sherpa

Issue #955, opened 2022-04-16
Penalizing and measuring symbol delay


danpovey:
I'd like to have a way to penalize symbol delay in RNN-T computations, and to measure it during training.
(This is relevant for systems that use time masking, to avoid the time-masking encouraging the network to
delay symbols).

My proposal is this: we introduce a user-specifiable constant \alpha >= 0that says how much we'll penalize symbol delay.
Suppose we have a sequence of lengths S_n, T_n, (n is the batch index), i.e. there are S_n symbols in the sequence
and T_n frames; px[n,s,t] will contain the log-probability of emitting a non-blank symbol starting from position s,t of sequence n.

To penalize symbol delay, we'll do: px[n,s,t] +=  offset_n - alpha t.  [We could equivalently do this on the py's].
the "- alpha t" term will encourage emitting symbols earlier (because we reduce the log-probs of later time-steps).
offset_n is a constant per sequence that is not going to affect where symbols are emitted but which
keeps the loss function at about the old value; we will set offset_n = alpha (T_n-1) / 2, which is the average of
 [0,1,..T_n-1] which are all the possible positions a symbol could be emitted.  We'll have to do this in at least 2 different 
places: once for the "simple" loss and once for the "pruned" loss.

To measure symbol delay (for diagnostic purposes), I propose to measure it only from the "simple" model, 
because we compute those grads anyway
during the forward pass.  I would like to define "normalized symbol delay" as the average (over symbols) of: 
  (position symbol is emitted, in [0,1,..T_n-1]) - (T_n-1),
i.e. it's positive if symbols tend to be emitted later than the middle of the audio, on average; negative if before.
This should be computable from the sum or average of px_grad * arange(T+1), since px_grad[n,s,t] is the probability of
the symbol at position s being emitted on frame t.  We can average and report this "sym_delay" with our other stats, if alpha is
specified.

This "sym_delay" does not directly reflect the real delay because it could be that our corpus has more trailing than preceding
silence; but it's valid for comparisons, and we can always compute this delay from a reference system that was trained with
alpha=0 and no time masking, which should reflect the actual time symbols were spoken (since there is a time-symmetry which
should prevent symbols from being emitted, on average, earlier or later than the true time).

pkufool:
I will try this idea with this pull request(https://github.com/k2-fsa/icefall/pull/242).

desh2608:
[...]

PR #976, opened 2022-05-19
add delay_penalty in rnnt loss

pkufool:
Implements https://github.com/k2-fsa/k2/issues/955

@danpovey Would you please have a look at the code, in case I did something wrong, thanks!

### DIFF
--- a/k2/python/k2/rnnt_loss.py
+++ b/k2/python/k2/rnnt_loss.py
@@ -266,17 +266,17 @@ def rnnt_loss_simple(
         T = T0 if modified else T0 - 1
         if boundary is None:
             offset = torch.tensor(
-                [(T - 1) / 2 * delay_penalty] * B,
+                (T - 1) / 2,
                 dtype=px.dtype,
                 device=px.device,
-            )
+            ).expand(B, 1, 1)
         else:
-            offset = (boundary[:, 3] - 1) / 2 * delay_penalty
-        penalty = offset.reshape(B, 1, 1) - delay_penalty * torch.arange(
+            offset = (boundary[:, 3] - 1) / 2
+        penalty = offset.reshape(B, 1, 1) - torch.arange(
             T0, device=px.device
         ).reshape(1, 1, T0)
-        penalty = penalty.to(px.dtype)
-        px += penalty
+        penalty = penalty * delay_penalty
+        px += penalty.to(px.dtype)
 
     scores_and_grads = mutual_information_recursion(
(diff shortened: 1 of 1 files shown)


k2/python/k2/rnnt_loss.py
```diff
@@ -255,6 +260,24 @@ def rnnt_loss_simple(
         boundary=boundary,
         modified=modified,
     )
+
+    if delay_penalty > 0.0:
+        B, S, T0 = px.shape
+        T = T0 if modified else T0 - 1
+        if boundary is None:
+            offset = torch.tensor(
+                [(T - 1) / 2 * delay_penalty] * B,
```
danpovey, line 269: If I were you I would multiply by delay_penalty after combining with the offset, to avoid an extra op.

k2/python/k2/rnnt_loss.py
```diff
@@ -1159,6 +1235,24 @@ def rnnt_loss_smoothed(
...
+        else:
+            offset = (boundary[:, 3] - 1) / 2 * delay_penalty
+        penalty = offset.reshape(B, 1, 1) - delay_penalty * torch.arange(
+            T0, device=px.device
+        ).reshape(1, 1, T0)
+        penalty = penalty.to(px.dtype)
+        px += penalty
+
     scores_and_grads = mutual_information_recursion(
         px=px, py=py, boundary=boundary, return_grad=return_grad
```
danpovey, line 1257: I think it might be useful to keep track of the average delay during training, which we could get
by computing the dot-product of the (arange - offset) with the returned grads.  
But I'm not sure that it would be super convenient to add an extra return value to these
functions because it would require the calling code to be changed at the exact same time.
It would definitely be possible to also compute it outside this function, in icefall, by duplicating the
code that computes the (arange - offset) and dotting that with the returned grad.  I'd say that
would be an acceptable solution.
That would make it obvious if we were penalizing the delay too much or too little; it would also
provide an easy way to measure what the delay "should" be as a baseline, if we ran non-chunked
systems.  (I.e. I'm saying we could perhaps keep track of this avg-delay even in cases when we were not
doing the delay-penalty).

pkufool: Yes, I think so. I kept track of the average delay in Icefall, in the model.py and return it along loss value. I have not created a PR yet.

k2/python/k2/rnnt_loss.py
pkufool: Will fix, thanks!

PR #981, opened 2022-06-15
Support using k2 as a third party library.

csukuangfj:
The motivation is to let users do
```
pip install k2
```
and then he/she can use k2 in its own projects, which can use CMake, makefile, or other build systems.

Neither need the users to re-compile k2 nor do they need NVCC to compile their projects.

They can use k2 in their projects like what we are using PyTorch in k2.

---

Will make another PR to support using k2 in a project, e.g., [sherpa](https://github.com/k2-fsa/sherpa), using CMake as the build system.


k2/csrc/torch_api_test.cc
```diff
@@ -0,0 +1,56 @@
+/**
```
csukuangfj, line 1: Note this is a `.cc` file, not a `.cu` file.

k2/csrc/torch_api.h
```diff
@@ -0,0 +1,75 @@
+/**
```
csukuangfj, line 1: Only this header file is visible to users.

We can wrap other classes of k2 when needed.

csukuangfj: @danpovey 

If you find this approach is feasible, I will continue with it.

PR #990, opened 2022-06-23
test K2_TRANS_EXCSUM

glynpu:
1. fix comment error of "K2_TRANS_EXCSUM" in k2/csrc/macros.h
2. add a test for it in k2/csrc/macros_test.cu
3. include "kernel_scan.hxx" into k2/csrc/moderngpu.h  to fix following compilation  error
 ```
[ 46%] Building CUDA object k2/csrc/CMakeFiles/cu_macros_test.dir/macros_test.cu.o
/ceph-ly/open-source/k2/k2/csrc/macros_test.cu(74): error: namespace "mgpu" has no member "transform_scan"
```

   It seems that macro "K2_TRANS_EXCSUM" is not being used except this new added unit test.
   So this compilation error is not met before.

 4. BTW, the comment related to "EvalWithRedirect" seems obsolete since it  is already removed by https://github.com/k2-fsa/k2/pull/727/files#diff-dc2a6f83d656522a88ebe371fdb539a836fd1f9b2655f924a5b70e248f26caecL445

### DIFF
--- a/k2/csrc/eval.h
+++ b/k2/csrc/eval.h
@@ -35,8 +35,6 @@
 #include "k2/csrc/context.h"
 #include "k2/csrc/log.h"
 
-// Caution: see also utils.h where there is EvalWithRedirect that's very related
-// to this.
 namespace k2 {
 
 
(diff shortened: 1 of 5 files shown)


csukuangfj:
> It seems that macro "K2_TRANS_EXCSUM" is not being used except this new added unit test.
So this compilation error is not met before.

I would recommend removing that macro if it is never used.

Also, the test about it can also be removed.

glynpu:
just close it. See detail in https://github.com/k2-fsa/k2/pull/1005

PR #998, opened 2022-07-04
Fix for CMake 3.10

csukuangfj:
From https://cmake.org/cmake/help/v3.10/command/target_compile_definitions.html

> The named `<target>` must have been created by a command such as [add_executable()](https://cmake.org/cmake/help/v3.10/command/add_executable.html#command:add_executable) or [add_library()](https://cmake.org/cmake/help/v3.10/command/add_library.html#command:add_library) and must not be an [Imported Target](https://cmake.org/cmake/help/v3.10/manual/cmake-buildsystem.7.html#imported-targets).

This PR fixes that by removing `target_compile_definitions()`.

### DIFF
--- a/cmake/k2Config.cmake.in
+++ b/cmake/k2Config.cmake.in
@@ -62,8 +62,7 @@ foreach(lib IN LISTS K2_LIBRARIES)
       CXX_STANDARD 14
   )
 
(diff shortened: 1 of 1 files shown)


csukuangfj:
Tested by @luomingshuang  and it works.

PR #1002, opened 2022-07-04
Fix parsing FSA in openfst format.

csukuangfj:
```
!line_is.eof() || line_is.fail()
```

I think checking whether it is `EOF` is enough.

Fixes https://github.com/k2-fsa/icefall/issues/459

@danpovey @pkufool 

Do you think it is safe to remove the condition `line_is.fail()`?

References:
- https://en.cppreference.com/w/cpp/io/basic_ios/fail
- https://en.cppreference.com/w/cpp/io/ios_base/iostate

### DIFF
--- a/k2/csrc/fsa_utils.cu
+++ b/k2/csrc/fsa_utils.cu
@@ -200,14 +200,12 @@ static Fsa K2FsaFromStream(std::istringstream &is,
     return ans;
 }
 
-// `expected` must not be whitespace.  expect (read) `expected, set failbit if
-// not.
+// `expected` must not be whitespace.  expect (read) `expected
 void ExpectChar(std::istream &is, char expected) {
   is >> std::ws;
   char ch;
   is >> ch;
-  if (ch != expected)
-    is.setstate(std::ios::failbit);
+  K2_CHECK_EQ(ch, expected);
 }
 
 
@@ -291,7 +289,7 @@ class OpenFstStreamReader {
       line_is >> cost;
       if (!line_is.eof()) line_is >> std::ws;
     }
-    if (!line_is.eof() || line_is.fail() || src_state < 0 || dest_state < 0) {
+    if (!line_is.eof() || src_state < 0 || dest_state < 0) {
       K2_LOG(FATAL) << "Invalid line: " << line << ", eof=" << line_is.eof()
                     << ", fail=" << line_is.fail()
                     << ", src_state=" << src_state
--- a/k2/csrc/fsa_utils.cu
+++ b/k2/csrc/fsa_utils.cu
@@ -263,7 +263,7 @@ class OpenFstStreamReader {
     int32_t src_state, dest_state, label;
     line_is >> src_state >> dest_state >> label;
     if (!line_is.good()) {
-      // Our only hope of parsing othis is that `line` was empty, or a
+      // Our only hope of parsing this is that `line` was empty, or a
       // final-state; try that.
       if (!ProcessFinalProbOrEmpty(line)) {
         K2_LOG(FATAL) << "Invalid line: " << line;
@@ -291,9 +291,8 @@ class OpenFstStreamReader {
       line_is >> cost;
       if (!line_is.eof()) line_is >> std::ws;
     }
-    if (!line_is.eof() || line_is.fail() || src_state < 0 || dest_state < 0) {
+    if (!line_is.eof() || src_state < 0 || dest_state < 0) {
       K2_LOG(FATAL) << "Invalid line: " << line << ", eof=" << line_is.eof()
-                    << ", fail=" << line_is.fail()
                     << ", src_state=" << src_state
                     << ", dest_state=" << dest_state;
     }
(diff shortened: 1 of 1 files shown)


pkufool:
> Do you think it is safe to remove the condition `line_is.fail()`?

I think so.

k2/csrc/fsa_utils.cu
```diff
@@ -291,9 +291,8 @@ class OpenFstStreamReader {
       line_is >> cost;
       if (!line_is.eof()) line_is >> std::ws;
     }
-    if (!line_is.eof() || line_is.fail() || src_state < 0 || dest_state < 0) {
+    if (!line_is.eof() || src_state < 0 || dest_state < 0) {
```
danpovey, line 294: I don't think it's safe to ignore line_is.fail() here.  E.g. above, it does:
`ExpectChar(line_is, ']');  // sets failbit if not..`
That check would be ineffective if we ignore failbit.   If this is failing we need to figure out exactly why, I think.

k2/csrc/fsa_utils.cu
```diff
@@ -263,7 +263,7 @@ class OpenFstStreamReader {
     int32_t src_state, dest_state, label;
     line_is >> src_state >> dest_state >> label;
     if (!line_is.good()) {
```
danpovey, line 265: I think the actually problem may be here, in "if (!line_is.good()) {".
According to the table here, https://en.cppreference.com/w/cpp/io/ios_base/iostate, if eofbit is set but neither failbit nor badbit is set, good() will return false.  This is not what I intended.  The if-statement should probably read: "if (line_is.fail()) { .. }"

k2/csrc/fsa_utils.cu
csukuangfj: > I don't think it's safe to ignore line_is.fail() here. E.g. above, it does:
ExpectChar(line_is, ']'); // sets failbit if not..

In that case, I recommend throwing as soon as we detect that `line_is.fail()` is `true`.
I just reverted the changes and change `if(!line_is.good())` to `if(line.fail())` and 
someone just tested it and the problem still exists.

Here is the error message:
[error.txt](https://github.com/k2-fsa/k2/files/9052985/error.txt)

csukuangfj: @danpovey 
> I don't think it's safe to ignore line_is.fail() here. E.g. above, it does:
ExpectChar(line_is, ']'); // sets failbit if not..

I have changed the semantics of `ExpectChar`: If the char is unexpected, we invoke
`K2_LOG(FATAL)` so that we don't need to use `line.is_fail()` later.

csukuangfj:
Someone encounters the same problem in https://github.com/k2-fsa/k2/issues/1022#issuecomment-1194887000

@danpovey 

What do you think about the fix in this PR?

k2/csrc/fsa_utils.cu
danpovey: Sorry, I don't think the change in ExpectChar is good, because the current error-checking code prints out the line that failed, and other information like src_state.  Inside ExpectChar, we can't print out that info.  If we have a failure in future, this change would make it hard to debug.
I had a look at the error message, which was:
```
k2::OpenFstStreamReader::ProcessLine(std::string&) Invalid line: 3	0	137054	0, eof=true, fail=true, src_state=3, dest_state=0
```
... this is very hard to understand.  I cannot find a code path that should cause these results.  What we could perhaps do is if there is a way to reproduce the error, add other commands to detect at what point failbit is set.  (or find someone who is willing to get this in gdb and step through it).  

Changing:
```
    if (line_is.fail())  {
```
to
```
    if (line_is.good())  {
```
should be safe, even if it is not necessary or does not help here.
I would rather not mask this error.  If something here is not behaving as it should, that indicates a deeper problem, and we should find the problem before we attempt to mask it.

danpovey:
Sorry, I don't agree with this change, I think it is masking some deeper bug.   The real cause could be a bug in the standard library or a compiler problem or a synchronization issue, for instance, which we should find.  Or we misunderstood something.  It's annoying that no-one seems to be willing or able to help us by getting this in gdb and finding at what point the failbit is being set.
Once we find a way to reproduce this, or find someone who is able to reproduce it who is willing to debug, we can find the true source of the issue.  For now I would rather leave the code as it is.

danpovey:
... we could prepare a branch that prints out the fail bit at each stage where it could possibly be being set, and have people use this to print out more info.
We could also print out num_aux_labels_ and num_ragged_labels_.  Maybe they are getting overwritten somehow.
Now we know that the issue is failbit being set when it should not be set, so we made some progress at least.

csukuangfj:
ok, I posted a comment at https://github.com/k2-fsa/k2/issues/1022#issuecomment-1195138561

If it still does not help, let's wait for the next one who can reproduce it.

csukuangfj:
> ... we could prepare a branch that prints out the fail bit at each stage where it could possibly be being set, and have people use this to print out more info. We could also print out num_aux_labels_ and num_ragged_labels_. Maybe they are getting overwritten somehow. Now we know that the issue is failbit being set when it should not be set, so we made some progress at least.

I just created https://github.com/k2-fsa/k2/pull/1024 to check `line_is.fail()` after every read operation.

PR #1004, opened 2022-07-07
less pruning last 5 frames

videodanchik:
See https://github.com/k2-fsa/icefall/issues/292 and https://github.com/k2-fsa/icefall/issues/292#issuecomment-1093837846 for more details. This is a memory-safe modification of https://github.com/k2-fsa/k2/pull/967.


danpovey:
Thanks!   @pkufool, perhaps you could merge it if you think it's OK?
I think this strategy is pretty safe, there is something similar in Kaldi I think.

pkufool:
> Thanks! @pkufool, perhaps you could merge it if you think it's OK? I think this strategy is pretty safe, there is something similar in Kaldi I think.

OK, I will merge it once all the unit test passed, I mean the github CI.

PR #1005, opened 2022-07-07
remove unused typename

glynpu:
1. remove the unused typename in eval_lambda
2. other fixes are from https://github.com/k2-fsa/k2/pull/990, only fix the comment error and remove the new added test.
As Dan suggested last time: leave the unused macro "TRANS_EXCSUM" in code base in case of we may use it in the future.

So I will also close   https://github.com/k2-fsa/k2/pull/990.


csukuangfj:
Thanks!

PR #1009, opened 2022-07-11
Fix pruning bounds

pkufool:
This PR fixes the issue of returning nan or inf loss (posted at https://github.com/danpovey/fast_rnnt/issues/10).
It also updates the method to generate pruning bounds (i.e. the method we wrote in our pruned rnn-t paper). I did some experiments on April, but fogot to submit the code, sorry! About the new pruning bounds, I think we can not 100% gurantee it would be better than the old one, as both of them are local optimal bounds.

### DIFF
--- a/k2/python/k2/rnnt_loss.py
+++ b/k2/python/k2/rnnt_loss.py
@@ -557,8 +557,11 @@ def get_rnnt_prune_ranges(
             s_range >= 2
         ), "Pruning range for standard RNN-T should be equal to or greater than 2, or no valid paths could survive pruning."
 
+    (B_stride, S_stride, T_stride) = py_grad.stride()
     blk_grad = torch.as_strided(
-        py_grad, (B, S1 - s_range + 1, s_range, T), (S1 * T, T, T, 1)
+        py_grad,
+        (B, S1 - s_range + 1, s_range, T),
+        (B_stride, S_stride, S_stride, T_stride),
     )
     # (B, S1 - s_range + 1, T)
     blk_sum_grad = torch.sum(blk_grad, axis=2)
@@ -572,13 +575,17 @@ def get_rnnt_prune_ranges(
 
     # (B, T)
     s_begin = torch.argmax(final_grad, axis=1)
-    s_begin = s_begin[:, :T]
 
     # Handle the values of s_begin in padding positions.
-    # -1 here means we fill the position of the last frame of real data with
+    # -1 here means we fill the position of the last frame (before padding) with
     # padding value which is `len(symbols) - s_range + 1`.
-    # This is to guarantee that we reach the last symbol at last frame of real
-    # data.
+    # This is to guarantee that we reach the last symbol at last frame (before
+    # padding).
+    # The shape of the mask is (B, T), for example, we have a batch containing
+    # 3 sequences, their lengths are 3, 5, 6 (i.e. B = 3, T = 6), so the mask is
+    # [[True, True, False, False, False, False],
+    #  [True, True, True,  True,  False, False],
+    #  [True, True, True,  True,  True,  False]]
     mask = torch.arange(0, T, device=px_grad.device).reshape(1, T).expand(B, T)
     mask = mask < boundary[:, 3].reshape(B, 1) - 1
 
@@ -589,7 +596,7 @@ def get_rnnt_prune_ranges(
     s_begin = torch.where(mask, s_begin, s_begin_padding)
 
     # adjusting lower bound to make it satisfied some constrains, see docs in
-    # `adjust_pruning_lower_bound` for more details of these constrains.
+    # `_adjust_pruning_lower_bound` for more details of these constrains.
     # T1 == T here means we are using the modified version of transducer,
     # the third constrain becomes `s_begin[i + 1] - s_begin[i] < 2`, because
     # it only emits one symbol per frame.
@@ -699,10 +706,15 @@ def get_rnnt_prune_ranges_deprecated(
     s_begin = s_begin[:, :T]
 
     # Handle the values of s_begin in padding positions.
-    # -1 here means we fill the position of the last frame of real data with
+    # -1 here means we fill the position of the last frame (before padding) with
     # padding value which is `len(symbols) - s_range + 1`.
-    # This is to guarantee that we reach the last symbol at last frame of real
-    # data.
+    # This is to guarantee that we reach the last symbol at last frame (before
+    # padding).
+    # The shape of the mask is (B, T), for example, we have a batch containing
+    # 3 sequences, their lengths are 3, 5, 6 (i.e. B = 3, T = 6), so the mask is
+    # [[True, True, False, False, False, False],
+    #  [True, True, True,  True,  False, False],
+    #  [True, True, True,  True,  True,  False]]
     mask = torch.arange(0, T, device=px_grad.device).reshape(1, T).expand(B, T)
     mask = mask < boundary[:, 3].reshape(B, 1) - 1
 
@@ -713,7 +725,7 @@ def get_rnnt_prune_ranges_deprecated(
     s_begin = torch.where(mask, s_begin, s_begin_padding)
 
     # adjusting lower bound to make it satisfied some constrains, see docs in
-    # `adjust_pruning_lower_bound` for more details of these constrains.
+    # `_adjust_pruning_lower_bound` for more details of these constrains.
     # T1 == T here means we are using the modified version of transducer,
     # the third constrain becomes `s_begin[i + 1] - s_begin[i] < 2`, because
     # it only emits one symbol per frame.
--- a/k2/python/k2/rnnt_loss.py
+++ b/k2/python/k2/rnnt_loss.py
@@ -505,8 +505,118 @@ def get_rnnt_prune_ranges(
     of symbols given a particular frame.
 
     Note:
-      For the generated tensor ranges, ranges[:, 0] is a monotonic increasing
-      tensor from 0 to `len(symbols)` and it satisfies
+      For the generated tensor ranges (assuming batch size is 1), ranges[:, 0]
+      is a monotonic increasing tensor from 0 to `len(symbols)` and it satisfies
+      `ranges[t+1, 0] - ranges[t, 0] < s_range` which means we won't skip any
+      symbols.
+
+    Args:
+      px_grad:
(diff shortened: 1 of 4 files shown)


k2/python/tests/rnnt_loss_test.py
```diff
@@ -488,23 +546,211 @@ def test_rnnt_loss_pruned(self):
...
+
+    # Test more exact pruning bounds.
+    # In our previous versions we use a less exact method to generate
+    # pruning bounds which is different from the method we
+    # publish in our paper(https://arxiv.org/pdf/2206.13236.pdf).
+    # This tests the difference between these two methods.
+    # We won't do any assertion in this test, just printing out the losses,
+    # because we can not 100% sure that the new method is better than the old
+    # one all the time, both of them are local optimal bounds.
+    def test_prune_ranges(self):
```
pkufool, line 672: I tested the two pruning bounds here, first, I generated a mask according to the ranges, and filled the logits  with 0.0 at the positions outof ranges.  then, I computed the loss of the modifed logits.  If the new pruning bounds is better than the old one, the loss should be smaller.  From my experiments (both on real data and random generated data), most of the time, the loss of the logit pruned with new bounds is smaller than the one pruned with old bounds. But it can not 100% gurantee it. If I did something wrong, please LMK @danpovey, thanks!

k2/python/k2/rnnt_loss.py
```diff
@@ -531,9 +657,23 @@ def get_rnnt_prune_ranges(
...
     assert boundary.shape == (B, 4)
-    assert s_range >= 1
+    assert S >= 1
+    assert T >= S
+
+    # s_range > S means we won't prune out any symbols. To make indexing with
+    # ranges runs normally, s_range should be equal to or less than ``S + 1``.
     if s_range > S:
-        s_range = S
+        s_range = S + 1
```
pkufool, line 666: The buggy code that raises nan or inf loss is here.  It only occurs when `S==1` and `modified==False`, because when `modified==False`  the transitions can only go up and right, so `s_range` MUST be equal to or greater than 2, or no path can survive pruning.

pkufool:
I tested the new bounds and old bounds on real data,  in the following figure, pink one is the curve of loss pruned by old bounds, blun one is the curve of loss pruned by new bounds (Note, I used the trivial joiner to calculate the losses). You can see in the figure, only in the early batches, the loss of the new one is a little larger than the old one, after 3k iterations, the old one is identical to the new one.

I also compared the pruning bounds, in the early 3k iterations, `torch.mean(new_bounds - old_bounds) == 0.01535`, for the iterations from 5k -> 10k, `torch.mean(new_bounds - old_bounds) == 0.00558`.

image

pkufool:
It seems that the old / new bounds are almost identical, I will use the new bounds to reproduce one of our recipes.

pkufool:
I tried this new pruning bounds on our two baseline recipes, here are the results:

From the following results, training with new pruning bounds makes some improvements on pruned_transducer_stateless4 recipe, especially on test-other. While on pruned_transducer_stateless5 (exp-B, 88MB parameters), I got slightly worse results.

### [pruned transducer stateless4](https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/RESULTS.md#librispeech-bpe-training-results-pruned-stateless-transducer-4)

  | greedy_search | modified_beam_search | fast_beam_search | comments
-- | -- | -- | -- | --
baseline | 2.69 & 6.64 | 2.62 & 6.57 | 2.66 & 6.6 | --epoch 30 --avg 6
New bounds | 2.62 & 6.28 | 2.58 & 6.18 | 2.57 & 6.13 | --epoch 25 --avg 13

| tensorboard for baseline | tensorboard for new bounds |
|---|---|
|image|image|

### [pruned transducer stateless5 exp-B](https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/RESULTS.md#baseline-2)
  | greedy_search | modified_beam_search | fast_beam_search | comments
-- | -- | -- | -- | --
baseline | 2.54 & 5.72 | 2.47 & 5.71 | 2.5 & 5.72 | --epoch 30 --avg 10
New bounds | 2.53 & 5.77 | 2.53 & 5.7 | 2.52 & 5.71 | --epoch 25 --avg 11

| tensorboard for baseline | tensorboard for new bounds |
|---|---|
|image|image|

pkufool:
I conducted two experiments to compare the pruned loss value of old bounds and new bounds.

It seems that whether we use "old bounds" or "new bounds" in train, the loss values of "new bounds" are always larger than that of "old bounds" at the early iterations.  BTW from the picture, we can see "new bounds" helps to converge faster,  the only difference of these two experiments are the bounds used for pruning.  

### training with new boulds
The blue curve is for new bounds
The pink curve is for old bounds
image

code to generate tensorboard above
```python
   ranges = k2.get_rnnt_prune_ranges(
      px_grad=px_grad,
      py_grad=py_grad,
      boundary=boundary,
      s_range=prune_range,
    )

    # this loss will be used to train the model
    pruned_loss = k2.rnnt_loss_pruned(
        logits=logits.float(),   # this logits is from joiner output.
        symbols=y_padded,
        ranges=ranges,
        termination_symbol=blank_id,
        boundary=boundary,
        reduction="sum",
    )

      new_am_pruned, new_lm_pruned = k2.do_rnnt_pruning(am=am, lm=lm, ranges=ranges)
      new_logits = new_am_pruned + new_lm_pruned

     # this loss is used to draw the blue curve
      new_pruned_loss = k2.rnnt_loss_pruned(
          logits = new_logits.float(),
          symbols=y_padded,
          ranges=ranges,
          termination_symbol=blank_id,
          boundary=boundary,
          reduction="sum",
      )

      old_ranges = k2.get_rnnt_prune_ranges_deprecated(
          px_grad=px_grad,
          py_grad=py_grad,
          boundary=boundary,
          s_range=prune_range,
      )

      old_am_pruned, old_lm_pruned = k2.do_rnnt_pruning(am=am, lm=lm, ranges=old_ranges)
      old_logits = old_am_pruned + old_lm_pruned

      # this loss is used to draw the pink curve
      old_pruned_loss = k2.rnnt_loss_pruned(
          logits = old_logits.float(),
          symbols=y_padded,
          ranges=old_ranges,
          termination_symbol=blank_id,
          boundary=boundary,
          reduction="sum",
      )
      return (simple_loss, pruned_loss, new_pruned_loss, old_pruned_loss)
```

### training with old boulds
The blue curve is for new bounds
The pink curve is for old bounds
image

k2/python/k2/rnnt_loss.py
```diff
@@ -505,8 +512,127 @@ def get_rnnt_prune_ranges(
...
+        assert (
+            s_range >= 1
+        ), "Pruning range for modified RNN-T should be equal to or greater than 1, or no valid paths could survive pruning."
+
+    else:
+        assert (
+            s_range >= 2
+        ), "Pruning range for standard RNN-T should be equal to or greater than 2, or no valid paths could survive pruning."
+
+    blk_grad = torch.as_strided(
```
danpovey, line 560: Here,  I think it would be better not to assume that py is contiguous, in place of S1 you  should probably 
be using py_grad.stride(2), and so on.

k2/python/k2/rnnt_loss.py
```diff
@@ -505,8 +512,127 @@ def get_rnnt_prune_ranges(
...
+    px_pad = torch.zeros((B, 1, T1), dtype=px_grad.dtype, device=px_grad.device)
+    # (B, S1, T)
+    px_grad_pad = torch.cat((px_pad, px_grad), dim=1)
+
+    # (B, S1 - s_range + 1, T)
+    final_grad = blk_sum_grad - px_grad_pad[:, : S1 - s_range + 1, :T]
+
+    # (B, T)
+    s_begin = torch.argmax(final_grad, axis=1)
+    s_begin = s_begin[:, :T]
```
danpovey, line 575: this statement may be redundant?

k2/python/k2/rnnt_loss.py
```diff
@@ -505,8 +512,127 @@ def get_rnnt_prune_ranges(
...
+
+    # Handle the values of s_begin in padding positions.
+    # -1 here means we fill the position of the last frame of real data with
+    # padding value which is `len(symbols) - s_range + 1`.
+    # This is to guarantee that we reach the last symbol at last frame of real
+    # data.
+    mask = torch.arange(0, T, device=px_grad.device).reshape(1, T).expand(B, T)
```
danpovey, line 582: can we comment to say what mask will be, e.g. its shape and when it is true?

danpovey:
After fixing those small issues and making sure the tests run, I incline towards merging this.
Generally if there are inconsistent WER results that are not clearly significant, I prefer to just average them to get the overall story, so I tend to think this is better than the previous method; also it's encouraging that it converges faster.
Regarding the difference between the loss functions: firstly, double-check that you didn't swap the two of them at some point in your code, but if the difference persists, I suppose it might have to do with the continuity constraints somehow, e.g. the new method might be more locally optimal but less globally optimal somehow.

pkufool:
> After fixing those small issues and making sure the tests run, I incline towards merging this. Generally if there are inconsistent WER results that are not clearly significant, I prefer to just average them to get the overall story, so I tend to think this is better than the previous method; also it's encouraging that it converges faster. Regarding the difference between the loss functions: firstly, double-check that you didn't swap the two of them at some point in your code, or when you interpreted the graph; but if the difference persists, I suppose it might have to do with the continuity constraints somehow, e.g. the new method might be more locally optimal but less globally optimal somehow.

Ok, will fix the issue and check the code again. Thanks!

danpovey:
Thanks... looks OK to me.

RuABraun:
On real-world datasets the assert `T>=S` fails a non-insignificant number of times for me. I think a subsampling factor of 4 might be too aggressive.

danpovey:
We should make sure there's a way for calling code to ignore this error.
@RuABraun  Are you sure that the transcripts in these cases are correct?

RuABraun:
Yup (it's German if that matters). Doesn't happen that often honestly but just wanted to raise awareness.

danpovey:
Ah, OK.  For German it may be different than English (or might need larger BPE vocab).
It should be possible to train with modified==False (on all minibatches), so we shouldn't have this limitation (IIRC).
To take advantage of that in test time, we can only use certain decoding methods.
In any case we should make it possible to skip over batches with this problem.

PR #1018, opened 2022-07-14
Expose more k2 interface

pkufool:
Some fixes to support fast_beam_search in sherpa.


k2/csrc/CMakeLists.txt
```diff
@@ -265,10 +269,19 @@ file(COPY
   DESTINATION
     ${PROJECT_BINARY_DIR}/include/k2
 )
+file(GLOB HEADER_LIST "*.h")
```
pkufool, line 272: @csukuangfj  I am not sure is it a good way to copy all the header files to install directory, what's your oponions?

k2/csrc/torch_api.h
```diff
@@ -21,6 +21,13 @@
...
 #include <memory>
 
+// TODO(WeiKang): Move the following lines to k2/csrc/all.h
+#include "k2/csrc/fsa_algo.h"
+#include "k2/csrc/fsa_utils.h"
+#include "k2/csrc/ragged.h"
+#include "k2/csrc/ragged_ops.h"
+#include "k2/csrc/rnnt_decode.h"
+#include "k2/csrc/torch_util.h"
 #include "torch/script.h"
```
pkufool, line 31: @csukuangfj  Why we need this extra torch wrapper, I think we can do it just like pytorch by putting the header files in one header file (like all.h), no need to wrap them again.

csukuangfj: Do you need NVCC to compile sherpa if you do it this way?

The original design tries to achieve the goal that whether you use a CPU version of k2 or a CUDA version as a pre-compiled library, you can always use GCC to compile sherpa.

pkufool: I try compile sherpa with CPU version Pytorch and CUDA version k2, it compiles successfully and runs normally.

csukuangfj: > I tried compiling sherpa with CPU version Pytorch and CUDA version k2, it compiles successfully and runs normally.

Do you have CI tests to cover that?

It looks surprising to me that k2 headers can be compiled with GCC when CUDA is enabled as there are stuffs that can only be processed by NVCC, e.g., kernels.

csukuangfj: Can you show the output of

make VERBOSE=1


You have to run 

make clean

Before you run "make VERBOSE=1".

pkufool:
[log.txt](https://github.com/k2-fsa/k2/files/9121824/log.txt)

It indeed compiles by gcc and links the cuda version so.

k2/csrc/torch_api.h
csukuangfj: > I tried compiling sherpa with CPU version Pytorch and CUDA version k2, it compiles successfully and runs normally.

Have you tried to compile sherpa and k2 with CUDA support?

---

There are header files in k2 that use different macros depending on whether NVCC is present.
For instance,

https://github.com/k2-fsa/k2/blob/6a899805cacf82e69953e591e8683951d167fa26/k2/csrc/context.h#L44-L46


In your case, even if it does not throw compile-time errors, it does not mean it can run correctly.

The headers of k2 that you include in sherpa are compiled with different flags from those used by pre-compiled k2 libraries.

In short, the header files of k2 that you are using in sherpa are always for CPU, not for CUDA.
That is, the same template may be instantiated differently in sherpa and k2 if k2 is using CUDA, which I think
can cause undefined behavior during link time.

pkufool: > There are header files in k2 that use different macros depending on whether NVCC is present.

I know what you said.

> In your case, even if it does not throw compile-time errors, it does not mean it can run correctly.

At least, it runs normally when decoding with fast_beam_search.

Will do more experiments.

pkufool: If I use cuda version Pytorch, how could I make sherpa compile with gcc?

danpovey: It may depend what functions we use.  If we use certain functions from a .cc file compiled with gcc that would normally directly instantiate a CUDA kernel, it may compile something that will just generate an error instead of really using CUDA.

k2/python/csrc/torch/array_ops.cu
```diff
@@ -65,7 +65,8 @@ static void PybindMonotonicLowerBound(py::module &m) {
         // Unreachable code, to make compiler happy
         return torch::Tensor();
       },
-      py::arg("src"), py::arg("inplace") = false);
+      py::arg("src"), py::arg("inplace") = false,
+      py::call_guard<py::gil_scoped_release>());
```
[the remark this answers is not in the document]
pkufool: OK, will only add to functions used by fast_beam_search.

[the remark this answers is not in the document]
pkufool: I checked the functions wrapped in k2, I think we need most of them for `fast_beam_search` (including `fast_beam_search_nbest`, `fast_beam_search_LG`), If we want to do it this way, we should figure out what are the potential problems.

BTW, what do you mean "check that functions using this argument are thread-safe",  I think most of our functions return a new fsa object and won't modify fsas in place, that would be safe for multithreaded application. As far as I am concerned, it is the user's responsibility to handle the multithreaded issues(i.e. protecting fsa objects).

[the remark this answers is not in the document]
danpovey: The main potential problem I can think of is if the torch allocator was not thread safe, but it seems to be thread safe.  
We don't have much interaction with Python objects ,in k2 code.

pkufool:
@csukuangfj  Do you have any ideas of figuring out what's going on among these CI tests, the tests are all passed in my local env, I can not reproduce they. Thanks!

csukuangfj:
> @csukuangfj Do you have any ideas of figuring out what's going on among these CI tests, the tests are all passed in my local env, I can not reproduce they. Thanks!

The CI is using torch 1.12.0, python 3.10. Are you using the same version as the CI locally?

yonatankahana:
hi @csukuangfj,
any progress with that?

PR #1030, opened 2022-08-09
[WIP] Implements RNNT+MMI

pkufool:
It runs normally in my self constructed test case, not fully tested yet, though.


k2/csrc/fsa_algo.cu
```diff
@@ -1991,4 +1991,323 @@ FsaOrVec RemoveEpsilonSelfLoops(FsaOrVec &src,
...
+
+          arc_map_value = states_idx012;
+          arc.label = sampled_paths_data[states_idx012];
+          float sampling_prob = sampling_probs_data[states_idx012];
+
+          int32_t us_idx0 = us_row_ids1_data[ss_idx01],
+                  repeat_num = us_row_splits1_data[us_idx0 + 1] -
+                    us_row_splits1_data[us_idx0];
+
+          arc.score = -logf(1 - powf(1 - sampling_prob, repeat_num));
```
pkufool, line 2234: I only include the "predictor" head output in C++ part, the other two scores (i.e. hybrid output and lm_output) will add on python part, it would be easier to enable autograd for hybrid output.

k2/python/k2/fsa_algo.py
```diff
@@ -1381,3 +1381,57 @@ def union(fsas: Fsa) -> Fsa:
...
+        left_symbols=k2.RaggedTensor(left_symbols),
+        sampling_probs=k2.RaggedTensor(sampling_probs),
+        vocab_size=vocab_size,
+        context_size=context_size,
+    )
+    lattice = Fsa(ragged_arc)
+    a_value = getattr(lattice, "scores")
+    # Enable autograd for path_scores
+    b_value = index_select(path_scores.flatten(), arc_map)
+    value = a_value + b_value
```
pkufool, line 1435: `path_scores` here will contain hybrid_output and detached lm_output.  I include the `path_scores` here and enable antograd to `path_scores`.

k2/python/tests/generate_denominator_lattice_test.py
```diff
@@ -0,0 +1,216 @@
...
+
+        frame_ids_list.append(t_index)
+
+        left_symbols_list.append(left_symbols)
+
+        # update (t, s) for each path
+        # index == 0 means the sampled symbol is blank
+        t_mask = index == 0
+        #         t_index = torch.where(t_mask, t_index + 1, t_index)
+        t_index = t_index + 1
```
pkufool, line 146: If we use regular RNN-T, it is possible to generate too many symbols for a specific frame, and that might be chances to generate a lattice containing cycles, which is not expected.  I am not sure whether we will encounter such a issue at the very beginning of training.

pkufool (commented):
@danpovey  Do you have any good idea to test this function, I can only think of constructing simple test cases.

k2/python/tests/generate_denominator_lattice_test.py
danpovey: > 

Hm, a valid point.  Yes, computing forward backward scores would not work correctly if there are cycles.  One possibility would be to augment the state with a sub-frame, i.e. instead of (ctx, t) it becomes (ctx, t, sub_t) with sub_t = (0, 1, 2, ..).  That would prevent cycles, although it might prevent a small number of paths from recombining that might otherwise be able to recombine.

k2/python/k2/fsa_algo.py
danpovey: Yes, OK.  Right, we treat those as differentiable, but the negated sampling_prob is treated as just a constant.

PR #1032, opened 2022-08-10
Link python lib to fix mkl not found

pkufool:
The Pytorch distribution on CentOS does not include MKL library, so there will be linking error when compiling k2.
This PR can fix the issue because Python library contains MKL library, not a good way though. We might need other method to fix it, as we don't want k2core to depend on Python library.


pkufool:
@csukuangfj  Do you have any other ideas to fix this issue? I think we discussed this issue before, you are in your holiday at that time.  @glynpu  found this issue.

csukuangfj:
What are the error logs?

I think the core library of k2 does not need Python. It is PyTorch that needs MKL.
Also, this happens only for some versions of PyTorch, I think.

I suggest that we update the documentation to mention how to fix it rather than adding PYTHON as a dependency.

PR #1039, opened 2022-08-15
optimize shortestpath

glynpu:
Implement Dan's suggestion:
```
// ... incidentally, it's possible to further optimize this so the run
// time is less than linear, by using methods similar to what I use
// in GetStateBatches(); imagine computing a table that instead of
// the best traceback, is the best 2-step traceback; and then the 4-step
// traceback, and so on. There's no need for this right now, since the
// forward-pass algorithm is already at least linear-time in the length
// of this path. But we can consider it for the future.
```

### DIFF
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -866,15 +866,6 @@ void ArcSort(Fsa &src, Fsa *dest, Array1<int32_t> *arc_map /*= nullptr*/) {
   *dest = tmp;
 }
 
-// TODO(fangjun): use the following method suggested by Dan
-//
-// ... incidentally, it's possible to further optimize this so the run
-// time is less than linear, by using methods similar to what I use
-// in GetStateBatches(); imagine computing a table that instead of
-// the best traceback, is the best 2-step traceback; and then the 4-step
-// traceback, and so on. There's no need for this right now, since the
-// forward-pass algorithm is already at least linear-time in the length
-// of this path. But we can consider it for the future.
 Ragged<int32_t> ShortestPath(FsaVec &fsas,
                              const Array1<int32_t> &entering_arcs) {
   NVTX_RANGE(K2_FUNC);
@@ -895,6 +886,9 @@ Ragged<int32_t> ShortestPath(FsaVec &fsas,
   Array1<int32_t> state_best_arc_index_array(context, num_states, -1);
   int32_t *state_best_arc_index_array_data = state_best_arc_index_array.Data();
 
+#if 0
+  // This is a simple version of the kernel that demonstrates what we're trying
+  // to do with the more complex code.
   K2_EVAL(
       context, num_fsas, lambda_set_num_best_arcs, (int32_t fsas_idx0) {
         int32_t state_idx01 = row_splits1_data[fsas_idx0];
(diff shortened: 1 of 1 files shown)


csukuangfj:
Is it ready for review?

glynpu:
> Is it ready for review?

Ready now. All unit tests are passed last night.

danpovey:
Cool!  Liyong, did you ever manage to test whether this gives us a speedup in practice?
I'm not sure that the unit test would do that adequately.

glynpu:
> did you ever manage to test whether this gives us a speedup in practice?

Not yet. 
Is there some suggestions about what/how to measure?

> I'm not sure that the unit test would do that adequately.
Fsas in the unit test is too small to compare the speed difference. 

Shall I create  a new unitest with larger fsas to  test the speed differences?

danpovey:
You could do that, or just test it yourself on a GPU by writing some kind of loop.
I suppose you could just create linear FSAs for testing purposes, should probably be fine.  Up to you though.

glynpu:
> "some kind of loop" and "linear FSAs"
Good idea. Thanks, will do this.

glynpu:
With a rough glance, the optimized is 4 times faster than original one
![image](https://user-images.githubusercontent.com/14951566/184804254-cd29dc27-4c4b-4f04-857b-a74df2f80f31.png)

test code:

```
import os                                                                                                                                                                   
import sys                                                                                                                                                                  
from time import time                                                                                                                                                       
import torch                                                                                                                                                                
                                                                                                                                                                            
NUM_LOOPS = 10                                                                                                                                                              
NUM_ARCS = 5000                                                                                                                                                             
def measure_time():                                                                                                                                                         
    # GPU 1 is free when testing this.                                                                                                                                      
    device = torch.device('cuda', 1)                                                                                                                                        
    labels = list(range(NUM_ARCS))                                                                                                                                          
    fsa = k2.linear_fst(labels, labels)                                                                                                                                     
    fsav = k2.create_fsa_vec([fsa] * 10).to(device)                                                                                                                         
    durations = []                                                                                                                                                          
    for _ in range(NUM_LOOPS):                                                                                                                                              
        torch.cuda.synchronize()                                                                                                                                            
        start=time()                                                                                                                                                        
        _ = k2.shortest_path(fsav, use_double_scores=False)                                                                                                                 
        torch.cuda.synchronize()                                                                                                                                            
        end = time()                                                                                                                                                        
                                                                                                                                                                            
        durations.append(end-start)                                                                                                                                         
    return durations                                                                                                                                                        
                                                                                                                                                                            
import k2                                                                                                                                                                   
import _k2                                                                                                                                                                  
print(_k2.__file__)                                                                                                                                                         
durations = measure_time()                                                                                                                                                  
print(durations)                                                                                                                                                            
print(sum(durations))  
```

related envs:

```
k2_root=/ceph-ly/open-source/linear_fst_k2/k2/                                                                                                                              
export PYTHONPATH=${k2_root}/k2/python:$PYTHONPATH                                                                                                                          
export PYTHONPATH=${k2_root}/build_release/lib/:$PYTHONPATH                                                                                                                 
                                                                                                                                                                            
# k2_root=/ceph-hw/kangwei/code/k2_release/k2/                                                                                                                              
# export PYTHONPATH=${k2_root}/k2/python:$PYTHONPATH                                                                                                                        
# export PYTHONPATH=${k2_root}/build/lib/:$PYTHONPATH
```

danpovey:
Great-- thanks!!

PR #1043, opened 2022-08-18
[WIP] Publish k2 as android arr

pkufool:


csukuangfj:
I suggest that we use
https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-android

That is, we only need to change `CMakeLists.txt` and provide a toolchain file for cross compiling.

pkufool:
> I suggest that we use https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-android
> 
> That is, we only need to change `CMakeLists.txt` and provide a toolchain file for cross compiling.

I think the only thing we want to do is to distribute the header files and so libs, by doing this way, we don't have to do any extra building task, we just release our version on github then the users can get k2 from gradle.

pkufool:
> That is, we only need to change `CMakeLists.txt` and provide a toolchain file for cross compiling.

Actually, I did only change the `CMakeLists.txt`, the others files added is to make the build system to gennerate so files, the same as the scripts you added to the github CI. As far as I know, most people if not all use gradle build system to build the android application. I think current method is the easiest way for users.

pkufool:
You can read the demo code (I will submit to android-demo repo soon),  you can make decision then.

PR #1056, opened 2022-09-14
Update array_ops_inl.h

armusc:
replace
int32_t elem_size = src[0]->ElementSize();
by
int64_t elem_size = src[0]->ElementSize();
in
template <typename T>
Array1<T> Cat(ContextPtr c, int32_t num_arrays, const Array1<T> **src) 

to prevent overflow with big fst (> 4GB)


csukuangfj:
Thanks! Will merge after the CI passes.

PR #1059, opened 2022-09-18
Add allow_partial option in rnnt_decode.

pkufool:
In our current implementation all the states in the graph can be final state in the generated lattice.  In this PR, I add an option `allow_partial`, if `allow_partial` is True(default), any states can be final state. if `allow_partial` is False, only the final state in the graph can be final state.  I think we will need this option when implementing `context biasing` feature.


danpovey:
Why do you think we might need this in the context-biasing thing?
I would have thought that we could just make any states in the context biasing graph that we want final, to have an arc to the super-final state.

pkufool:
> Why do you think we might need this in the context-biasing thing? I would have thought that we could just make any states in the context biasing graph that we want final, to have an arc to the super-final state.

So we will augment not only the context paths but also the prefixes of the context paths? I think if we augment only the full context paths we need to set `allow_partial == False`, or we can not penalty the boost score we added when going into the grammar.

danpovey:
> > Why do you think we might need this in the context-biasing thing? I would have thought that we could just make any states in the context biasing graph that we want final, to have an arc to the super-final state.
> 
> So we will augment not only the context paths but also the prefixes of the context paths? I think if we augment only the full context paths we need to set `allow_partial == False`, or we can not penalty the boost score we added when going into the grammar.

Oh sorry, I assumed allow_partial=False was the default (like in Kaldi).

I'm surpised that allow_partial=True was the behavior before.  I would have thought if we wanted a state to be final it would be easy to add an arc to the superfinal state.
But also IDK how that would interact with pruning, and whether there might be a danger of getting no successful paths if final state was pruned.

pkufool:
> I'm surpised that allow_partial=True was the behavior before.  But also IDK how that would interact with pruning, and whether there might be a danger of getting no successful paths if final state was pruned.

Yes, the default behavior is allow_partial=True in master code,  this code was wrote months ago, I did some testing, it seemed to make no difference for trivial graph whether allow_partial is True or not. But I did remember that this option will affect the WERs if decoding graph is general LG. I can't find the exact numbers, will re-test it later.

danpovey:
OK, no problem.  OK to merge this as far as I am concerned.

pkufool:
I re-test how allow_partial will affect the WERs, here is the resutls:
| decoding method | decoding graph | allow_partial | WERs (clean & other) |
|---------------------|-------------------|---------------|-------------------------|
| fast_beam_search | trivial graph         | True             |  2.67 & 6.26                |
| fast_beam_search | trivial graph       |  False            | 2.66 & 6.25             |
| fast_beam_search_nbest | LG graph | True  | 2.82 & 6.24 |
| fast_beam_search_nbest | LG graph | False | 2.81 & 6.23|
| fast_beam_search |  LG graph | True | 4.63 & 8.09 |
| fast_beam_search | LG graph | False | 2.81 & 6.32 |

I will set the default value of `allow_partial` to False (i.e. the same as Kaldi).

danpovey:
I'm very surprised that you got such big differences with that option, in the 2nd-to-last line.  What's going on there?  See if you can see what's different in the transcripts.

pkufool:
> I'm very surprised that you got such big differences with that option, in the 2nd-to-last line. What's going on there? See if you can see what's different in the transcripts.

The 2nd-to-last line has a lot of tailing deletions, I think it is because the `allow_partial` option makes the shorter paths have higher scores. 

I realize I might mistaken the semantic of `allow_partial`,  `allow_parital` should mean we allow partial results if not reaching final state, now I implement it as allowing partial results for all the sequences if `allow_partial=True`.

pkufool:
I fix the issue, I think `allow_partial` works correctly now.

| decoding method | decoding graph | allow_partial | WERs (clean & other) |
|---------------------|-------------------|---------------|-------------------------|
| fast_beam_search | trivial graph         | True             |  2.67 & 6.26                |
| fast_beam_search | trivial graph       |  False            | 2.66 & 6.25             |
| fast_beam_search_nbest | LG graph | True  | 2.82 & 6.24 |
| fast_beam_search_nbest | LG graph | False | 2.81 & 6.23|
| fast_beam_search |  LG graph | True | 2.81 & 6.32 |
| fast_beam_search | LG graph | False | 2.81 & 6.32 |

I am going to merge it.

PR #1076, opened 2022-10-18
to_str_openfst

glynpu:
Before: k2.Fsa.from_str(openfsat=True) failed to accept the output of k2.to_str(openfst=True)
image

Previous problem is fixed by this pr and "openfst=False" is not affected:

image


csukuangfj:
Could you also add a test to import the exported string with k2?

glynpu:
> Could you also add a test to import the exported string with k2?

Sure. Now the original existing tests of to_str(openfst=True) are made compatible with new code.

danpovey:
Looks OK to me.

Issue #1078, opened 2022-10-20
segmentation fault error whiling calling k2.shortest_path


jiaji-huang:
I am referring to this [repo](https://github.com/luomingshuang/k2-speechbrain/blob/python-version/utils/decode.py#L426) to run decoding. My code differs a bit, but here are the gists:

1) I obtain lattice calling
`k2.intersect_dense_pruned(decoding_graph, dense_FSA, beam_size, beam_size, min_active_states, max_active_states)`
Here decoding_graph is a composition of the (default) CTC topology and lexicon. dense_FSA is the denseFSA built from neural net's outputs (softmax probabilities). beam_size is 5, and min_active_states=30, max_active_states=10000.

2. Then I run 
`k2.shortest_path(lattice, use_double_scores=False)`

On a few utterances, this decoding pipeline returns me segmentation error without any further error msg. But sometimes, it does print out more stack trace, e.g.,

`[ Stack-Trace: ]
/tools/venv/lib64/python3.7/site-packages/k2-1.21.dev20221015+cuda11.3.torch1.11.0-py3.7-linux-x86_64.egg/k2/lib64/libk2_log.so(k2::internal::GetStackTrace()+0x34) [0x7f4217009e74]
/tools/venv/lib64/python3.7/site-packages/k2-1.21.dev20221015+cuda11.3.torch1.11.0-py3.7-linux-x86_64.egg/k2/lib64/libk2context.so(k2::Array1<int>::Init(std::shared_ptr<k2::Context>, int, k2::Dtype)+0x75d) [0x7f42175c767d]
/tools/venv/lib64/python3.7/site-packages/k2-1.21.dev20221015+cuda11.3.torch1.11.0-py3.7-linux-x86_64.egg/k2/lib64/libk2context.so(k2::RaggedShape::RowIds(int)+0xd8) [0x7f42177488a8]
/tools/venv/lib64/python3.7/site-packages/k2-1.21.dev20221015+cuda11.3.torch1.11.0-py3.7-linux-x86_64.egg/k2/lib64/libk2context.so(k2::ShortestPath(k2::Ragged<k2::Arc>&, k2::Array1<int> const&)+0x92a) [0x7f421762a1ba]
/tools/venv/lib64/python3.7/site-packages/k2-1.21.dev20221015+cuda11.3.torch1.11.0-py3.7-linux-x86_64.egg/_k2.cpython-37m-x86_64-linux-gnu.so(+0x6dc64) [0x7f4218822c64]
[...]

Issue #1077, opened 2022-10-19
Array size is coming as negative


kbramhendra:
Hi,
    While I am decoding with k2, i am facing this issue `Check failed: size >= 0 (-8388608 vs. 0) Array size MUST be greater than or equal to 0, given :-8388608` in k2.shortest_path, can you please help in this.

csukuangfj:
Which decoding method are you using? How large is your graph used for decoding?

kbramhendra:
Thanks for replying. The graph size is 1.1G, vocab size is 231670. I am using 1best decoding in k2.

csukuangfj:
What's your max duration?

kbramhendra:
To give you full context, I am using espnet + k2 , K2 i am using for decoding ( i took from icefall ), max duration of an audio file is 30 seconds.

kbramhendra:
Further to add...this problem is happening when one short segments (1 sec) is added in the batch with a bit long sentences (avg 9 sec), but it is getting recognised when processed alone. And if are increasing max_states then also it processing in the batch. I am not able to figure out the exact reason why its failing. please help in this.

ngoel17:
[F] /home/ngoel/k2/k2/csrc/array.h:501:void k2::Array1<T>::Init(k2::ContextPtr, int32_t, k2::Dtype) [with T = int; k2::ContextPtr = std::shared_ptr<k2::Context>; int32_t = int] Check failed: size >= 0 (-402279408 vs. 0) Array size MUST be greater than or equal to 0, given :-402279408

I am getting a similar error while decoding. No graph, no espnet, just plain k2 fast-beam-search decoding from a pruned-transducer-stateless2 model. The longest audio file is 186.4 seconds.

Python stack is 
    File "/mnt/dsk1/icefall/egs/dh/pruned_transducer_stateless2/beam_search.py", line 77, in fast_beam_search_one_best
    best_path = one_best_decoding(lattice)
  File "/home/ngoel/icefall/icefall/decode.py", line 474, in one_best_decoding
    best_path = k2.shortest_path(lattice, use_double_scores=use_double_scores)
  File "/home/ngoel/k2/k2/python/k2/fsa_algo.py", line 594, in shortest_path
    ragged_arc, ragged_int = _k2.shortest_path(fsa.arcs, entering_arcs)

desh2608:
[...]

PR #1080, opened 2022-10-24
shortest_path for non connected fst

glynpu:
Fix issue https://github.com/k2-fsa/k2/issues/1078
May also fix issue  https://github.com/k2-fsa/k2/issues/1077
Current k2.shortest_path in master branch "requires" connected input fsa. While in issue https://github.com/k2-fsa/k2/issues/1078, the input fst is not(Thanks to @jiaji-huang for providing the test input).

BTW, in icefall inputs of k2.shortest_path are always connected, so this "bug" is not triggered before:

https://github.com/k2-fsa/icefall/blob/6af5a82d8f9c9c38e3c87eaaab706b292c15bd9b/icefall/decode.py#L371-L375

### DIFF
--- a/k2/csrc/fsa_algo.cu
+++ b/k2/csrc/fsa_algo.cu
@@ -877,7 +877,7 @@ Ragged<int32_t> ShortestPath(FsaVec &fsas,
   ContextPtr &context = fsas.Context();
 
   // allocate an extra element for ExclusiveSum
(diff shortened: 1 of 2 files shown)


danpovey:
This looks OK to merge.  In hindsight, this algorithm was perhaps over-optimized and we could maybe have done OK with the simple for-loop like the `#if 0` version of it.

PR #1085, opened 2022-10-31
Merge the branch v2.0-pre to master

csukuangfj:
So that we can use HLG decoding in sherpa using C++.


csukuangfj:
Note: To keep the git commit history, this PR won't use `squash and merge`.

PR #1087, opened 2022-11-03
allow_partial for intersect_dense_pruned

glynpu:
Following results are tested with a deliberately selected input and decoding configuration.

## with allow_partial=False (the original code):
There is no final arc(i.e. arc.label == -1).
```
0 1 0 0 -8.41582e-05                                                                                                                                                        
1 2 0 0 -4.69674e-05                                                                                                                                                        
2 3 0 0 -9.17907e-06                                                                                                                                                        
3 4 0 0 -1.10864e-05                                                                                                                                                        
4 5 0 0 -1.4305e-05                                                                                                                                                         
5 6 0 0 -2.63449e-05                                                                                                                                                        
6 7 0 0 -6.55649e-06                                                                                                                                                        
7 8 0 0 -1.19209e-05                                                                                                                                                        
8 9 0 0 -3.12323e-05                                                                                                                                                        
9 10 0 0 -2.36032e-05                                                                                                                                                       
11
```

## with allow_partial=True (current pr):
image

### DIFF
--- a/k2/csrc/fsa_algo.h
+++ b/k2/csrc/fsa_algo.h
@@ -161,10 +161,10 @@ void AddEpsilonSelfLoops(FsaOrVec &src, FsaOrVec *dest,
          @param[in] b_fsas   Input FSAs that correspond to neural network
                          outputs (see documentation in fsa.h).
          @param[in] search_beam   Beam for frame-synchronous beam pruning,
-                    e.g. 20. Smaller is faster, larger is more exact
-                    (less pruning). This is the default value; it may be
-                    modified by {min,max}_active which dictate the minimum
-                    or maximum allowed number of active states per frame.
+                         e.g. 20. Smaller is faster, larger is more exact
+                         (less pruning). This is the default value; it may be
+                         modified by {min,max}_active which dictate the minimum
+                         or maximum allowed number of active states per frame.
          @param[in] output_beam   Beam with which we prune the output (analogous
                          to lattice-beam in Kaldi), e.g. 8.  We discard arcs in
                          the output that are not on a path that's within
@@ -178,6 +178,10 @@ void AddEpsilonSelfLoops(FsaOrVec &src, FsaOrVec *dest,
                          of states are active.  The hash size used per FSA is 4
                          times (this rounded up to a power of 2), so this
                          affects memory consumption.
+       @param [in] allow_partial If true, we will treat all the states on the
+                         last frame to be final state. If false, we only
+                         care about the real final state in the decoding
+                         graph on the last frame when generating lattice.
          @param[out] out Output vector of composed, pruned FSAs, with same
                          Dim0() as b_fsas.  Elements of it may be empty if the
                          composition was empty, either intrinsically or due to
@@ -196,6 +200,7 @@ void AddEpsilonSelfLoops(FsaOrVec &src, FsaOrVec *dest,
 void IntersectDensePruned(FsaVec &a_fsas, DenseFsaVec &b_fsas,
                           float search_beam, float output_beam,
                           int32_t min_active_states, int32_t max_active_states,
+                          bool allow_partial,
                           FsaVec *out, Array1<int32_t> *arc_map_a,
                           Array1<int32_t> *arc_map_b);
 
--- a/k2/csrc/fsa_algo.h
+++ b/k2/csrc/fsa_algo.h
@@ -178,8 +178,9 @@ void AddEpsilonSelfLoops(FsaOrVec &src, FsaOrVec *dest,
                          of states are active.  The hash size used per FSA is 4
                          times (this rounded up to a power of 2), so this
                          affects memory consumption.
-       @param [in] allow_partial If true, we will treat all the states on the
-                         last frame to be final state. If false, we only
+       @param [in] allow_partial If true and there was no final state active,
+                         we will treat all the states on the last frame
+                         to be final state. If false, we only
                          care about the real final state in the decoding
                          graph on the last frame when generating lattice.
          @param[out] out Output vector of composed, pruned FSAs, with same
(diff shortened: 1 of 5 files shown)


k2/csrc/fsa_algo.h
```diff
@@ -178,6 +178,10 @@ void AddEpsilonSelfLoops(FsaOrVec &src, FsaOrVec *dest,
                          of states are active.  The hash size used per FSA is 4
                          times (this rounded up to a power of 2), so this
                          affects memory consumption.
+       @param [in] allow_partial If true, we will treat all the states on the
+                         last frame to be final state. If false, we only
+                         care about the real final state in the decoding
+                         graph on the last frame when generating lattice.
          @param[out] out Output vector of composed, pruned FSAs, with same
```
danpovey, line 185: the way allow_patrial works in Kaldi is, it is conditional: we only make the all final if there was no final state active.

glynpu: The code is implemented as "only make the all final if there was no final state active".
It's controlled by variable "has_valid_final_arc".
I will change this comment.

k2/python/k2/autograd.py
```diff
@@ -650,6 +657,7 @@ def intersect_dense_pruned(a_fsas: Fsa,
                            b_fsas: DenseFsaVec,
                            search_beam: float,
                            output_beam: float,
+                           allow_partial: bool,
```
csukuangfj, line 660: Please give it a default value so that it won't break existing programs.

csukuangfj: Also, please add it as the last argument to make it compatible with previous code.

glynpu: fixed.

k2/csrc/intersect_dense_pruned.cu
```diff
@@ -698,6 +747,17 @@ class MultiGraphDenseIntersectPruned {
...
+          auto final_t = b_fsas_row_splits1[ai_fsa_idx0+1] - b_fsas_row_splits1[ai_fsa_idx0];
+          if (final_t - 1 == t && !has_valid_final_arc_data[ai_fsa_idx0] &&
+              allow_partial_) {
+              int32_t a_fsas_idx0 = a_fsas_row_ids1[sinfo.a_fsas_state_idx01];
+              // state_idx1 is 0-based.
+              // So "-1" is used when calculating a_fsas_final_state_idx1.
+              int32_t a_fsas_final_state_idx1 = a_fsas_row_splits1[a_fsas_idx0 + 1] - 1 - a_fsas_row_splits1[a_fsas_idx0];
+              dest_state = a_fsas_final_state_idx1;
+              acoustic_score = 0.0;
+          }
```
pkufool, line 760: I think this block and the above block are not necessary,  we can know which sequence has no final state by the shape of  ArcInfo at the last frame (see first K2_EVAL in FormatOutput),  at the last frame, there will be only one state (the final state) or no state at all. 
Another thing, if we modify `ai.u.dest_a_fsas_state_idx01` here, it will mess the arc-info and might raise an error for chunk by chunk decoding.  Actually, if we know which sequcence has no arc, we can set the dest-state of the arcs at the last frame to the extra state we added (see first K2_EVAL in FormatOutput) without knowing `ai.u.dest_info_state_idx1`.

pkufool (changes requested):
I recommend we implement `allow_partial` only within the `FormatOutput` function without modifying code in other functions, I think it is possible, I did not test it by myself, though.

k2/csrc/intersect_dense_pruned.cu
```diff
@@ -498,12 +506,27 @@ class MultiGraphDenseIntersectPruned {
...
           arc.dest_state = dest_state_idx012 - oarc_idx0xx;
-          arc.label = a_fsas_arcs[arc_info.a_fsas_arc_idx012].label;
+          int32_t arc_label = a_fsas_arcs[arc_info.a_fsas_arc_idx012].label;
+          arc.label = arc_label;
+          int32_t final_t = b_fsas_row_splits1[oarc_idx0+1] - b_fsas_row_splits1[oarc_idx0];
+          if (t == final_t - 1 && arc_label != -1) {
+            if (allow_partial_) {
+              arc.label = -1;
+            } else {
+              // Unreachable code.
```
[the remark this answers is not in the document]
glynpu: For arcs pointing to super-final state, their labels must be -1 if allow_partial==false.
Just add this "else" branch to trigger some un-realized bug in the future.

k2/csrc/intersect_dense_pruned.cu
glynpu: The last frame log_probs is manually added [0.0, -inf, -inf, -inf, ..., -inf, -inf].
The main purpose of this block is setting those -inf to 0.0.
Or all active arcs will be pruned by function `PruneTimeRange `

This is also the reason when the input num_frames == 20, while the generated lattice length is only 10!
Not only the final arc is missing, but also the last "10" frames are pruned by `PruneTimeRange `.

Before fix, lattice length is 10.
```
0 1 0 0 -8.41582e-05                                                                                                                                                        
1 2 0 0 -4.69674e-05                                                                                                                                                        
2 3 0 0 -9.17907e-06                                                                                                                                                        
3 4 0 0 -1.10864e-05                                                                                                                                                        
4 5 0 0 -1.4305e-05                                                                                                                                                         
5 6 0 0 -2.63449e-05                                                                                                                                                        
6 7 0 0 -6.55649e-06                                                                                                                                                        
7 8 0 0 -1.19209e-05                                                                                                                                                        
8 9 0 0 -3.12323e-05                                                                                                                                                        
9 10 0 0 -2.36032e-05                                                                                                                                                       
11
```

After fix the length is 20.
image

pkufool:
Merged in #1218

PR #923, opened 2022-02-22
Asynchronous support for ctc/hlg decoding

pkufool:
Asynchronous decoding means that the sequences of consecutive chunks can come from different sources, for the production scenarios, not all the waves would come at the same time. With asynchronous decoding we can randomly combine the received chunks into batches.

PR #1091, opened 2022-11-04
Asynchronous support for ctc/hlg decoding

pkufool:
Merge https://github.com/k2-fsa/k2/pull/923 into master branch.

### DIFF
--- a/k2/torch/bin/online_decode.cu
+++ b/k2/torch/bin/online_decode.cu
@@ -1,5 +1,5 @@
 /**
- * Copyright      2021  Xiaomi Corporation (authors: Fangjun Kuang)
+ * Copyright      2021  Xiaomi Corporation (authors: Fangjun Kuang, Wei Kang)
  *
  * See LICENSE for clarification regarding multiple authors
  *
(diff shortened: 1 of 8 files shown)


k2/torch/bin/online_decode.cu
```diff
@@ -110,7 +114,7 @@ int main(int argc, char *argv[]) {
     ./bin/online_decode \
       --use_ctc_decoding true \
       --jit_pt <path to exported torch script pt file> \
-      --tokens <path to tokens.txt> \
+      --bpe_model <path to pretrained BPE model> \
```
csukuangfj, line 117: We don't support passing `bpe.model` anymore as we have removed `sentencepiece`.

pkufool: Yes, I saw it. thanks!

pkufool:
v2.0-pre and this PR modify `intersect_dense_pruned.cu`, I will train a ctc model based on it to check the correctness.

pkufool:
I trained a ctc model based on this PR, the training ran normally and the decoding results are almost the same as the model trained with the code before mergeing v2.0-pre. I think it is safe to merge the code, will do it after the CI passed.

pkufool:
The failed checks relate to env issue, have nothing to do with this PR, will fix them in seperate PR. Merging now.

PR #1093, opened 2022-11-06
Export CTC decoding algorithm to sherpa

pkufool:
See https://github.com/k2-fsa/sherpa/pull/177 for our plans.

- [ ] Move https://github.com/k2-fsa/k2/blob/master/k2/csrc/torch_api.h to https://github.com/k2-fsa/k2/tree/master/k2/torch/csrc 
- [ ] Remove the dependency `kaldifeat_core` from [`k2_torch`](https://github.com/k2-fsa/k2/blob/v2.0-pre/k2/torch/csrc/CMakeLists.txt#L26). We can make the binaries depend on `kaldifeat_core` directly.
- [ ] Export the library `k2_torch` in https://github.com/k2-fsa/k2/blob/c3a7404e0d9da4d8d99f3f15c54d80b321c1b960/cmake/k2Config.cmake.in#L48
- [ ] Add the following functions to `torch_api.h`
  - [ ] A function to load `HLG.pt`. The returned type can be `std::shared_ptr<k2::FsaClass>`. We can define an alias `FsaClassPtr` for it, like  [RaggedShapePtr](https://github.com/k2-fsa/k2/blob/master/k2/csrc/torch_api.h#L29)
  - [ ] A function to wrap `k2::CtcTopo()`. It should also return a value of type `FsaClassPtr`.
  - [ ] A function for CTC/HLG decoding. It takes the following inputs:
         - log_softmax_out: a 3-D tensor of shape (N, T, C)
         - log_softmax_out_lens, a 1-D tensor of shape (N,)
         - FsaClassPtr, can be either a CtcTopo or an HLG
         and it returns `std::vector<std::vector<int32_t>>`


k2/torch/csrc/CMakeLists.txt
```diff
@@ -65,3 +87,21 @@ if(K2_ENABLE_TESTS)
     k2_add_torch_test(${source})
   endforeach()
 endif()
+
+file(MAKE_DIRECTORY
+  DESTINATION
```
csukuangfj, line 92:
```suggestion
```

k2/torch/csrc/CMakeLists.txt
```diff
@@ -65,3 +87,21 @@ if(K2_ENABLE_TESTS)
...
+    ${PROJECT_BINARY_DIR}/include/k2
+)
+
+file(COPY
+  torch_api.h
+  DESTINATION
+    ${PROJECT_BINARY_DIR}/include/k2
+)
+
+install(TARGETS k2_torch_api
```
csukuangfj, line 102:
```suggestion
install(TARGETS k2_torch_api k2_torch
```

k2/torch/bin/CMakeLists.txt
```diff
@@ -4,13 +4,15 @@ include(transform)
...
+  kaldifeat_core
 )
 
+set(feature_srcs ../csrc/features.cc)
+
 #----------------------------------------
 #       CTC decoding
 #----------------------------------------
-set(ctc_decode_srcs ctc_decode.cu)
+set(ctc_decode_srcs ctc_decode.cu ${feature_srcs})
```
csukuangfj, line 15: It will recompile feature_srcs for each binary. Shall we make it a library that can be shared?

csukuangfj:
I suggest that we create a new PR to bind the exposed APIs to Python and provide python APIs and examples to decode models trained using CTC loss from various frameworks, such as icefall, nemo, espnet, speechbrain, wenet, etc.

pkufool:
> I suggest that we create a new PR to bind the exposed APIs to Python and provide python APIs and examples to decode models trained using CTC loss from various frameworks, such as icefall, nemo, espnet, speechbrain, wenet, etc.

I think we **MUST** create a new PR, because the binding and demo code will in sherpa, not k2.

k2/torch/bin/CMakeLists.txt
pkufool: OK, I thinks so. so, let's change back to its original

csukuangfj:
k2 only requires log_softmax_out + TLG for decoding, so it is easier to use.
People can just import k2 and provide the required inputs for decoding and they may don't want to install sherpa.

pkufool:
> k2 only requires log_softmax_out + TLG for decoding, so it is easier to use. People can just import k2 and provide the required inputs for decoding and they may not want to install sherpa.

OK, I got your idea, I think to do that the functions to be wrapped is in `k2/torch/*` not in `torch_api.h`.  It does not relate to this PR, will do it in another change, thanks!

k2/torch/csrc/torch_api.cu
```diff
@@ -74,13 +74,13 @@ FsaClassPtr LoadFsaClass(const std::string &filename,
...
 }
 
-std::vector<std::vector<int32_t>> Decode(torch::Tensor log_softmax_out,
-                                         torch::Tensor log_softmax_out_lens,
-                                         FsaClassPtr decoding_graph,
-                                         float search_beam, float output_beam,
-                                         int32_t min_activate_states,
-                                         int32_t max_activate_states,
-                                         int32_t subsampling_factor) {
+FsaClassPtr GetLattice(torch::Tensor log_softmax_out,
```
csukuangfj, line 77: I suggest that we provide only a single method `Decode()` to give the results directly.

In the current approach, users have to call `BestPath` on the returned lattice and that is the only function that users can call for the returned lattice. 

That is an implementation detail and we can hide it from the users.

pkufool: > I suggest that we provide only a single method `Decode()` to give the results directly.
> 
> In the current approach, users have to call `BestPath` on the returned lattice and that is the only function that users can call for the returned lattice.
> 
> That is an implementation detail and we can hide it from the users.

I thought we can use this lattice to do rescoring, we can implement `one_best_decoding` and `ngram_rescoring` in sherpa.

pkufool: I think it is no need to hide lattice.

csukuangfj (approved):

Issue #1061, opened 2022-09-18
Gradients through nbest lattice?


un-baised:
Hi,

I'm trying to implement MWER loss in k2. Is there an easy way to get gradients through nbest decoding? Or have you implemented something similar in the past?

danpovey:
The gradients should propagate from the original lattice to the sampled lattice you get from nbest, so you could compute an expected loss over a weighted sum of the nbest paths (weighted in a way involving the scores) and gradients would be propagated correctly.

un-baised:
Thanks for your note, now got gradients back to the initial lattice scores. Now looking at getting gradients through k2.RnntDecodingStreams .advance back to the joiner output log probabilities but not having much luck yet. Do you have any recommendations for setting this up? I can post a code snippet if that would help.

francisr:
I'd also like to know if it's possible to propagate lattice gradients to the frame log probs through the decoder.

danpovey:
I missed this conversation.  @pkufool does the RnntDecodingStreams object support gradients?

francisr:
I had a quick look at the cuda code. I'm not very familiar with it, but from what I understand it's not supported.  
Would it be complicated to implement?

csukuangfj:
I think it will return the arc indexes , so it should be possible to support gradient back propagation.

csukuangfj:
> I think it will return the arc indexes , so it should be possible to support gradient back propagation.

https://github.com/k2-fsa/k2/blob/1b151e5280046757a118fd1e9a5ed9dde8d18345/k2/python/k2/rnnt_decode.py#L182-L184

https://github.com/k2-fsa/k2/blob/1b151e5280046757a118fd1e9a5ed9dde8d18345/k2/csrc/rnnt_decode.h#L191-L199

You can see that it returns the arc_map.

danpovey:
Do one of you guys have time to implement this, or shoud I try to figure out how?

danpovey:
Oh.  I thought perhaps it would be outputting the scores somehow, e.g. into some kind of lattice-like object.  Yes, unless it is for the scores it doesn't make sense.

francisr:
We do get the lattice out from the decoder, whose scores will come from the log probs we feed in `.advance()`.

francisr:
The k2 code is quite cryptic to me right now (especially the CUDA stuff), so it would take me a bit too long.

csukuangfj:
[...]

PR #1094, opened 2022-11-08
[Ready] track arc_map_token during rnnt decoding

glynpu:
## Updated on Nov. 30, 2022:
k2.mwer is already merged by https://github.com/k2-fsa/k2/pull/1103

This pr will focus on tracking arc_map_token during decoding.

## Original:

An implementation of MWER, equation 2 of https://arxiv.org/pdf/2106.02302.pdf

Also solving issue https://github.com/k2-fsa/k2/issues/1061

## An example usage is:
1. slightly modification of [fast_beam_search](https://github.com/k2-fsa/icefall/blob/c30b8d3a1c095671d89321323390f8858722bcd1/egs/librispeech/ASR/transducer_stateless/beam_search.py#L226) to return log_probs of each time step.
```

log_probs_list = []
for t in range(T):
    *****(code of decoding)
    log_probs = (logits / temperature).log_softmax(dim=-1) 
    log_probs_list.append(log_probs)
    decoding_streams.advance(log_probs)
decoding_streams.terminate_and_flush_to_streams()
log_probs = torch.cat(log_probs_list) 
lattice, arc_map_token = decoding_streams.format_output(encoder_out_lens.tolist(), allow_partial=False)
return lattice, log_probs, arc_map_token 
```
2. Compute MWER based on the `lattice`, `log_probs`, `arc_map_token` 
```
lattice, log_probs, arc_map_token = fast_beam_search(                                                                                                                       
    model=model,                                                                                                                                                            
    decoding_graph=decoding_graph,                                                                                                                                          
    encoder_out=encoder_out,                                                                                                                                                
    encoder_out_lens=encoder_out_lens,                                                                                                                                      
    beam=beam,                                                                                                                                                              
    max_states=max_states,                                                                                                                                                  
    max_contexts=max_contexts,                                                                                                                                              
    temperature=temperature,                                                                                                                                                
)  
mwer_loss = k2.mwer_loss(nbest_scale, num_paths, log_probs, lattice, arc_map_token, ref_texts)
mwer_loss.backward()
```

## Unittest
Now I only manually check whether the gradient backprogates to the correct tokens, and it seems no obvious bugs.
Need more checks and unittests about this.
```
        # After loss.backward(),
        # users could manually check if the grad go to the right place,
        # An example:
        # By comapring hyps.labels and torch.where(log_probs.grad != 0)[1],
        # at least we could say that gradients seems to backpropate to the correct token,
        # i.e. 0, 15, 33, 10, 101 in following examples.
        # But why does token 95 in hyps.lables fail to get a gradient?
        # hyps.labels
        # >>>  tensor([  0,   0,  15,   0,  -1,   0,   0,  33,   0,  -1,   0,   0,  95,   0,
        #               -1,   0,   0, 101,   0,  -1,   0,   0,  10,   0,   0,  10,   0,  -1,
        #                0,   0,  10,   0,   0,  10,   0,   0,  10,   0,  -1],
        #             device='cuda:0', dtype=torch.int32)
        #
        # torch.where(log_probs.grad != 0)[0]  # corresopnds to context_index
        # >>>  tensor([ 0,  1,  1,  5,  6,  9, 10, 13, 14, 17, 18, 21, 22, 25, 26, 29, 30, 33,
        #               34, 38, 39, 39, 40, 44, 45, 46, 48, 49, 51, 52], device='cuda:0')
        #
        # torch.where(log_probs.grad != 0)[1]  # corresponds to token
        # >>>  tensor([  0,   0,  15,   0,   0,  33,   0,   0,   0,   0,   0,   0,   0,   0,
        #                0,   0,   0,   0,   0,  10,  10, 101,  10,  10,   0,  10,   0,   0,
        #                0,   0], device='cuda:0')
```

### DIFF
--- a/k2/csrc/rnnt_decode.cu
+++ b/k2/csrc/rnnt_decode.cu
@@ -936,6 +936,7 @@ void RnntDecodingStreams::FormatOutput(const std::vector<int32_t> &num_frames,
           arc.dest_state = oarc_idx01xx_next - oarc_idx0xxx;
           arc.label = -1;
           arc.score = 0;
+          out_map_data[oarc_idx01234] = -1;
         } else {
           const Arc *graph_arcs_data = graphs_arcs_data[oarc_idx0];
           arc.src_state = oarc_idx0123 - oarc_idx0xxx;
@@ -951,13 +952,12 @@ void RnntDecodingStreams::FormatOutput(const std::vector<int32_t> &num_frames,
           // make the generated lattice a valid k2 fsa.
           if (arc_info.graph_arc_idx01 == -1 || arc_info.label == -1) {
             arc.label = 0;
-            arc_info.graph_arc_idx01 = -1;
           } else {
             arc.label = graph_arcs_data[arc_info.graph_arc_idx01].label;
           }
           arc.score = arc_info.score;
+          out_map_data[oarc_idx01234] = arc_info.graph_arc_idx01;
         }
-        out_map_data[oarc_idx01234] = arc_info.graph_arc_idx01;
         arcs_out_data[oarc_idx01234] = arc;
         if (arc_map_b != nullptr) {
           t = t - num_padded_frames_data[oarc_idx0];
--- a/k2/csrc/rnnt_decode.cu
+++ b/k2/csrc/rnnt_decode.cu
@@ -91,6 +91,7 @@ void RnntDecodingStreams::TerminateAndFlushToStreams() {
   // detach prev_frames_
   std::vector<Ragged<ArcInfo> *> frames_ptr;
   for (size_t i = 0; i < prev_frames_.size(); ++i) {
+    // i is time index
     frames_ptr.emplace_back(prev_frames_[i].get());
   }
   // stack_frames has a shape of [t][stream][state][arc]
@@ -669,8 +670,9 @@ void RnntDecodingStreams::Advance(const Array2<float> &logprobs) {
         pruned_arcs_data[pruned_arc_idx0123] = info;
       });
 
+  // pruned_arcs is indexed [stream][context][src_state][arc].
   prev_frames_.emplace_back(
-      std::make_shared<Ragged<ArcInfo>>(pruned_arcs.RemoveAxis(1)));
+      std::make_shared<Ragged<ArcInfo>>(pruned_arcs));
 }
 
 void RnntDecodingStreams::GatherPrevFrames(
@@ -682,15 +684,17 @@ void RnntDecodingStreams::GatherPrevFrames(
   Array1<int32_t> stream2t_row_splits(GetCpuContext(), num_frames.size() + 1);
 
   for (size_t i = 0; i < num_frames.size(); ++i) {
+    // i is stream index
     stream2t_row_splits.Data()[i] = num_frames[i];
     K2_CHECK_LE(num_frames[i],
                 static_cast<int32_t>(srcs_[i]->prev_frames.size()));
     for (int32_t j = 0; j < num_frames[i]; ++j) {
+      // j is time index
       frames_ptr.push_back(srcs_[i]->prev_frames[j].get());
     }
   }
 
-  // frames has a shape of [t][state][arc],
+  // frames has a shape of [t][context][state][arc],
   // its Dim0() equals std::sum(num_frames)
   auto frames = Stack(0, frames_ptr.size(), frames_ptr.data());
 
@@ -715,6 +719,16 @@ void RnntDecodingStreams::FormatOutput(const std::vector<int32_t> &num_frames,
                                        bool allow_partial, FsaVec *ofsa,
                                        Array1<int32_t> *out_map) {
   NVTX_RANGE(K2_FUNC);
+  FormatOutput(num_frames, allow_partial, ofsa, out_map,
+               nullptr /* arc_map_token */, RaggedShape());
+}
+
+void RnntDecodingStreams::FormatOutput(const std::vector<int32_t> &num_frames,
+                                       bool allow_partial, FsaVec *ofsa,
+                                       Array1<int32_t> *out_map,
+                                       Array1<int32_t> *arc_map_b,
+                                       const RaggedShape &t2s2c_shape) {
+  NVTX_RANGE(K2_FUNC);
   K2_CHECK(!attached_)
       << "You can only get outputs after calling TerminateAndFlushToStreams()";
   K2_CHECK(ofsa);
@@ -728,15 +742,17 @@ void RnntDecodingStreams::FormatOutput(const std::vector<int32_t> &num_frames,
 
   auto has_final = Array1<bool>(c_, num_streams_, false);
   const ArcInfo *last_frame_arc_data = prev_frames_[frames - 1]->values.Data();
-  const int32_t *lfs_row_ids2_data = last_frame_shape.RowIds(2).Data(),
+  const int32_t *lfs_row_ids3_data = last_frame_shape.RowIds(3).Data(),
+                *lfs_row_ids2_data = last_frame_shape.RowIds(2).Data(),
                 *lfs_row_ids1_data = last_frame_shape.RowIds(1).Data();
   bool *has_final_data = has_final.Data();
 
   K2_EVAL(
       c_, last_frame_shape.NumElements(), lambda_set_has_final,
-      (int32_t idx012) {
-        ArcInfo ai = last_frame_arc_data[idx012];
-        int32_t idx01 = lfs_row_ids2_data[idx012],
+      (int32_t idx0123) {
+        ArcInfo ai = last_frame_arc_data[idx0123];
+        int32_t idx012 = lfs_row_ids3_data[idx0123],
+                idx01 = lfs_row_ids2_data[idx012],
                 idx0 = lfs_row_ids1_data[idx01];
         if (ai.label == -1) has_final_data[idx0] = true;
       });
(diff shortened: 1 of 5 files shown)


danpovey:
In my opinion this is not really in the correct spirit of implementing backprop.  What we should be doing, I think, is replacing the scores in the lattice with scores looked-up, via arc_map_token, from the log_probs tensor, and from the decoding graph.  [I'm not sure of the details of the backprop; possibly just assigning the tensor would work... see what we do in functions for existing FSA operations]  These scores should have the same numerical values, but should have the correct backprop information.
Then all kinds of lattice/FSA operations-- not just this specific one-- would work correctly.

glynpu:
> replacing the scores in the lattice with scores looked-up, via arc_map_token, from the log_probs tensor, and from the decoding graph

It's a good idea and helps me figure out a way to do "unittest".

## An example usage now: 
1. slightly modification of [fast_beam_search](https://github.com/k2-fsa/icefall/blob/c30b8d3a1c095671d89321323390f8858722bcd1/egs/librispeech/ASR/transducer_stateless/beam_search.py#L226) to get a lattice whose scores are  already **tracked by auto-grad machine.**

```

log_probs_list = []
t2stream_row_splits = [0]
stream2context_row_splits = [0]
num_log_probs = 0 
for t in range(T):
    *****(code of decoding)
    log_probs = (logits / temperature).log_softmax(dim=-1) 
    log_probs_list.append(log_probs)
    t2stream_row_splits += [shape.tot_size(0) + t2stream_row_splits[-1]]                                                                                                
    stream2context_row_splits += (shape.row_splits(1) + num_log_probs)[1:].tolist()                                                                                     
    num_log_probs = stream2context_row_splits[-1] 
    decoding_streams.advance(log_probs)
decoding_streams.terminate_and_flush_to_streams()
log_probs = torch.cat(log_probs_list) 
# NOTE: now lattice.scores is already tracked by auto-grad machine.
lattice = decoding_streams.format_output(encoder_out_lens.tolist(), log_probs=log_probs, t2s2c_shape=t2stream2context_shape3)
return lattice
```

2.  `lattice` and `ref_texts` are enough to compute MWER.
     i.e. current k2.mwer loss doesn't care how the lattice generated.
     It could be generated from a rnnt model by `fast_beam_search`, 
     or from a ctc model by `intersect_dense_pruned`.   
```
mwer_loss = k2.mwer_loss(nbest_scale, num_paths, lattice, ref_texts)
mwer_loss.backward()
```

## Unittest
Now a simple test is added for scores looking up by `arc_map_token` from log_probs.
With line-277, at least we could say `arc_map_token` is generated correctly.

https://github.com/k2-fsa/k2/blob/e5a8e4e2546e65c73c3a72353c044a52fb4ef707/k2/python/k2/rnnt_decode.py#L273-L279

k2/python/k2/mwer_loss.py
```diff
@@ -0,0 +1,436 @@
...
+
+
+        # Group each log_prob into [path][arc]
+        ragged_nbest_logp = k2.RaggedTensor(nbest.kept_path.shape, nbest_logp)
+        # Get the probalitity of each path, in log format, with shape [path].
+        path_logp = ragged_nbest_logp.sum()
+
+        # Group path_logp into [stream][path] to compute denominator of each stream.
+        # A stream here means an input wav.
+        ragged_path_prob = k2.RaggedTensor(nbest.shape, path_logp.exp())
```
danpovey, line 397: You could perhaps write a softmax function accepting RaggedTensor to replace these few lines, and  call that instead, it will be clearer.

k2/python/k2/mwer_loss.py
```diff
@@ -0,0 +1,436 @@
...
+    assert aux_labels.num_axes == 2
+    if return_ragged:
+        return aux_labels
+    else:
+        return aux_labels.tolist()
+
+
+# Copied from icefall.
+# TODO(liyong) simplify this.
+class Nbest(object):
```
danpovey, line 53: If this is going to be used it should probably be in a different file.   But are you sure it's not already in k2?

k2/python/k2/mwer_loss.py
```diff
@@ -0,0 +1,436 @@
...
+        return k2.RaggedTensor(self.shape, tot_scores)
+
+    def build_levenshtein_graphs(self) -> k2.Fsa:
+        """Return an FsaVec with axes [utt][state][arc]."""
+        word_ids = get_texts(self.fsa, return_ragged=True)
+        return k2.levenshtein_graph(word_ids)
+
+
+class MWERLoss(nn.Module):
+    # Minimus Word Error Rate.
```
danpovey, line 354: I am concerned that, especially for long utterances, the dynamic range of scores will be too large and the posteriors will be mostly 0 or 1.  To prevent this it might be a good idea to have an extra argument that functions like a temperature-- something we can scale the logprobs by before doing the softmax.

danpovey (commented):
some comments..

k2/python/k2/mwer_loss.py
glynpu: A function `k2.ragged_softmax` is added to k2/ops.py.

k2/python/k2/mwer_loss.py
glynpu: Yes, class `Nbest` is also exists in k2. 
But it has fewer functions than that in icefall, i.e. `from_lattice` and `build_levenshtein_graphs`. So I just copied the icefall one here. 
Now in the latest pr, this two needed functions  are added into the k2.Nbest.

glynpu:
The most important unit test for this pr is following assertion:

https://github.com/k2-fsa/k2/blob/d83d79ae1d43b2c6e9fd486953cb5c4eb62cf1ed/k2/python/k2/rnnt_decode.py#L284-L286

`fsa.scores` is generated during decoding.
`scores_tracked_by_autograd` is selected by `arc_map_token`

If `fsa.scores` is identical to `scores_tracked_by_autograd`,  it means `arc_map_token` is tracked correctly.

k2/csrc/rnnt_decode.cu
```diff
@@ -885,12 +952,36 @@ void RnntDecodingStreams::FormatOutput(const std::vector<int32_t> &num_frames,
...
+                    t2s2c_shape_row_splits2[t2s2c_stream_idx01];
+          // oarc_idx2 is for context.
+          int32_t oarc_idx2 = oarc_idx012 - oarc_idx01x;
+          int32_t t2s2c_context_idx012 = t2s2c_context_idx01x + oarc_idx2;
+          // Manually map arc to super_final state(i.e. arc.label == -1)
+          // to blank_id == 0.
+          int32_t arc_label = 0;
+          if (arc.label != -1) {
+            arc_label = arc.label;
+          }
```
pkufool, line 972: I think you can set the `arc_map_token` to `-1` for final arcs,  and `k2.index_select` can map `-1` to default value  automatically.

k2/python/k2/rnnt_decode.py
```diff
@@ -173,16 +175,27 @@ def format_output(
             If false, we only care about the real final state in the
             decoding graph on the last frame when generating lattice.
             Default False.
```
pkufool, line 177: Please add documents for `log_probs`.

k2/python/k2/rnnt_decode.py
```diff
@@ -264,4 +277,14 @@ def format_output(
             for name, value in src.named_non_tensor_attr():
                 setattr(fsa, name, value)
 
+        if log_probs is not None:
+            # Make fsa.scores tracked by autograd
+            # to make the whole decoding process differentiable.
+            scores_tracked_by_autograd = torch.index_select(
+                log_probs.reshape(-1), 0, arc_map_token)
+            final_arc_index = torch.where(fsa.arcs.values()[:, 2] == -1)
+            scores_tracked_by_autograd[final_arc_index] *= 0
```
pkufool, line 286: See the comment above, if you add `-1` to `arc_map_token` and use `k2.index_select` you can avoid line 285-286.

k2/csrc/rnnt_decode.cu
```diff
@@ -823,55 +850,95 @@ void RnntDecodingStreams::FormatOutput(const std::vector<int32_t> &num_frames,
...
+                *t2s2c_shape_row_splits1 = nullptr;
+
+  // In `GatherPrevFrames`, we use
+  // Unstack(frames, 1, false /*pad_right*/, &prev_frames);
+  // to generated prev_frames.
+  // Since pad_right=false, shorter utterances are padded to the left side,
+  // i.e. padded beginning frames.
+  // We need to subtract number of padded frames when calculating
+  // the real time index.
+  auto num_padded_frames = Array1<int32_t>(c_, num_frames);
```
pkufool, line 875: You initialize `num_padded_frames` with real frames here, it takes me a while to get your idea till I read the following lines. Please add comments here.

pkufool (commented):
LGTM

Leave some comments here. It would be great if you can run the `fast_beam_search` and `fast_beam_search_LG` in icefall with this change to double check its correctness.

k2/python/k2/rnnt_decode.py
```diff
@@ -264,4 +277,14 @@ def format_output(
...
+            # This assertion statement is kind of unit test.
+            assert torch.all(fsa.scores == scores_tracked_by_autograd)
```
pkufool, line 288: I think it is not right if we are using an LG graph.

pkufool (approved):
+2

desh2608:
> t2stream2context_shape3

@glynpu Could you tell me how you obtain `t2stream2context_shape3`?

desh2608:
> > t2stream2context_shape3
> 
> @glynpu Could you tell me how you obtain `t2stream2context_shape3`?

Found the answer here: https://github.com/k2-fsa/k2/blob/42e92fdd4097adcfe9937b4d2df7736d227b8e85/k2/python/tests/rnnt_decode_test.py#L138

PR #1099, opened 2022-11-10
fix error "UnicodeDecodeError: 'ascii' codec can't decode byte"

yuekaizhang:
When I tried to load Chinese tokens.txt, sometimes I would get this error "UnicodeDecodeError: 'ascii' codec can't decode byte".


csukuangfj:
Thanks!

PR #1103, opened 2022-11-13
[Ready]minimum word error rate

glynpu:
MWER computation part of https://github.com/k2-fsa/k2/pull/1094.

https://github.com/k2-fsa/k2/pull/1094 mixed two parts together:
Part 1. make lattice generated by [fast_beam_search](https://github.com/k2-fsa/icefall/blob/c30b8d3a1c095671d89321323390f8858722bcd1/egs/librispeech/ASR/transducer_stateless/beam_search.py#L226) tracked by autograd.
Part 2.  Compute MWER based on a lattice and corresponding ref_texts. (What this pr does)

For part 1, modifications of [fast_beam_search](https://github.com/k2-fsa/icefall/blob/c30b8d3a1c095671d89321323390f8858722bcd1/egs/librispeech/ASR/transducer_stateless/beam_search.py#L226) still needs some time to be compatible with others use cases that do not require gradient, i.e. pure decoding processes during evaluation.   

For part 2, It is an independent work and not necessarily used together with part 1, i.e. an RNN-T decoding method. 
It's moved here so users working with ctc trained pipeline could use it immediately.
Because k2.intersect_dense_pruned is already based on `torch.autograd.Function`.

An example usage pipeline could be:
``` 
lattice = k2.intersect_dense_pruned(*correct arguments*)
mwer_loss = k2.mwer_loss(lattice, ref_texts, nbest_scale, num_paths, temperature) 
loss = other_losses + mwer_loss
loss.backward()
```

### DIFF
--- a/k2/python/k2/mwer_loss.py
+++ b/k2/python/k2/mwer_loss.py
@@ -1,0 +1,140 @@
+# Copyright (c)  2022  Xiaomi Corporation (authors: Liyong Guo)
+
+from typing import List, Union
+
+import torch
+import k2
+
+
+class MWERLoss(torch.nn.Module):
+    '''Minimum Word Error Rate Loss compuration in k2.
+
+    See equation 2 of https://arxiv.org/pdf/2106.02302.pdf about its definition.
+    '''
+
+    def __init__(self,
+                 temperature: float = 1.0,
+                 use_double_scores: bool = True):
+        '''
+        Args:
+          temperature:
+            Softmax temperature similar to (logits / temperature).log_softmax().
+            For long utterances, the dynamic range of scores will be too large
+            and the posteriors will be mostly 0 or 1.
+            To prevent this it might be a good idea to have an extra argument
+            that functions like a temperature.
+            Something we can scale the logprobs by before doing the softmax.
+          use_double_scores:
+            True to use double precision floating point.
+            False to use single precision.
+        '''
+
+        super().__init__()
+        self.temperature = temperature
+        self.use_double_scores = use_double_scores
+
+    def forward(self,
+                lattice: k2.Fsa,
+                ref_texts: Union[k2.RaggedTensor, List[List[int]]],
+                nbest_scale: float,
+                num_paths: int) -> torch.Tensor:
+        '''Compute the Minimum Word Error loss given
+        a lattice and corresponding ref_texts.
+
+        Args:
+          lattice:
+            An FsaVec with axes [utt][state][arc].
+          ref_texts:
+            It can be one of the following types:
+              - A list of list-of-integers, e..g, `[ [1, 2], [1, 2, 3] ]`
+              - An instance of :class:`k2.RaggedTensor`.
+                Must have `num_axes == 2` and with dtype `torch.int32`.
+          nbest_scale:
+            Scale `lattice.score` before passing it to :func:`k2.random_paths`.
+            A smaller value leads to more unique paths at the risk of being not
+            to sample the path with the best score.''
+          num_paths:
+            Number of paths to **sample** from the lattice
+            using :func:`k2.random_paths`.
+        Returns:
+            Minimum Word Error Rate loss.
+        '''
+
+        nbest = k2.Nbest.from_lattice(
+            lattice=lattice,
+            num_paths=num_paths,
+            use_double_scores=self.use_double_scores,
+            nbest_scale=nbest_scale,
+        )
+
+        hyps = nbest.build_levenshtein_graphs()
+        refs = k2.levenshtein_graph(ref_texts, device=hyps.device)
+        levenshtein_alignment = k2.levenshtein_alignment(
+            refs=refs,
+            hyps=hyps,
+            hyp_to_ref_map=nbest.shape.row_ids(1),
+            sorted_match_ref=True,
+        )
+        tot_scores = levenshtein_alignment.get_tot_scores(
+            use_double_scores=self.use_double_scores, log_semiring=False
+        )
+        # Each path has a corresponding wer.
+        wers = -tot_scores
+
+        # Group each log_prob into [path][arc]
+        ragged_nbest_logp = k2.RaggedTensor(nbest.kept_path.shape,
+                                            nbest.fsa.scores)
+        # Get the probalitity of each path, in log format,
+        # with shape [stream][path].
+        ragged_path_logp = k2.RaggedTensor(
+            nbest.shape, ragged_nbest_logp.sum() / self.temperature)
+
+        # Function `normalize(use_log=True)` equals to apply
+        # `torch.log_softmax(dim=-1)` over each sublist.
+        # TODO(liyong): Implement RaggedTensor.softmax
+        # to replace following two lines?
+        ragged_path_logp_normalized = ragged_path_logp.normalize(use_log=True)
+        prob_normalized = ragged_path_logp_normalized.values.exp()
+        loss = (prob_normalized * wers).sum()
+        return loss
+
+
+def mwer_loss(
+        lattice,
+        ref_texts,
+        nbest_scale=0.5,
+        num_paths=200,
+        temperature=1.0,
+        use_double_scores=True) -> torch.Tensor:
+    '''Compute the Minimum loss given a lattice and corresponding ref_texts.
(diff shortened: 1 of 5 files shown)


csukuangfj:
All Python APIs have some tests in k2.
I suggest that we also add some tests for this PR.

danpovey:
From a brief look, this seems OK to me.

k2/python/k2/mwer_loss.py
```diff
@@ -0,0 +1,140 @@
...
+            An FsaVec with axes [utt][state][arc].
+          ref_texts:
+            It can be one of the following types:
+              - A list of list-of-integers, e..g, `[ [1, 2], [1, 2, 3] ]`
+              - An instance of :class:`k2.RaggedTensor`.
+                Must have `num_axes == 2` and with dtype `torch.int32`.
+          nbest_scale:
+            Scale `lattice.score` before passing it to :func:`k2.random_paths`.
+            A smaller value leads to more unique paths at the risk of being not
+            to sample the path with the best score.''
```
csukuangfj, line 55:
```suggestion
            to sample the path with the best score.
```

k2/python/k2/mwer_loss.py
```diff
@@ -0,0 +1,140 @@
...
+        )
+        # Each path has a corresponding wer.
+        wers = -tot_scores
+
+        # Group each log_prob into [path][arc]
+        ragged_nbest_logp = k2.RaggedTensor(nbest.kept_path.shape,
+                                            nbest.fsa.scores)
+        # Get the probalitity of each path, in log format,
+        # with shape [stream][path].
+        path_logp = ragged_nbest_logp.sum() / self.temperature
```
csukuangfj, line 89: Is this the intended use of `self.temperature`?
Looks like it does not match the documentation.

k2/python/k2/mwer_loss.py
```diff
@@ -0,0 +1,140 @@
...
+        ragged_path_prob = k2.RaggedTensor(stream_path_shape, path_logp.exp())
+
+        # Normalize prob of each path.
+        den_prob = ragged_path_prob.sum()
+        den_logp = torch.index_select(
+            den_prob.log(), 0, ragged_path_prob.shape.row_ids(1))
+        prob_normalized = (path_logp - den_logp).exp()
+
+        loss = (prob_normalized * wers).sum()
```
GNroy, line 99: Hi k2 team,
Great PR!

I'm not sure what size `prob_normalized` and `wers` have here.
If their length is `num_utterances` (batch size), then wouldn't it be better to return the loss as is (without summation)? In particular, so that the user can apply `.sum()` or `.mean()` to the final loss (e.g. after adding mwer loss to the `.get_tot_scores()` result of the main loss).
If their length is `num_utterances * num_paths` then does it make sense to do summation with respect to the utterance number, to, again, get the loss size as `num_utterances`?

glynpu: The length is not num_utterance. While it's  not num_utterances * num_paths either.  Since num_paths of each utterance could be unequal.  
Its size is `num_paths_of_utt1 + num_paths_of_utt2 + ... + num_paths_of_uttN`.

Maybe there should be a `reduction` option. 
If `none` then ` return the loss as is (without summation)` as you suggested.
If `sum` then return a summed value as what it does now.
If `mean` return a mean value.

Do you have suggestions?  @GNroy

GNroy: @glynpu thanks for answering!

I believe that there should be an internal reduction option of loss values belonging to the same utterance. In particular, get one number as the sum or mean over all paths of `utt1`, then the same for `utt2`, and so on.
I find it quite intuitive to get `n` loss values when feeding `n` utterances (lattices). Also, having loss values per utterance (instead of one per batch) enables tricks like post scaling (e.g. focal loss).

Regular `reduction` is also an option for those who just want a single loss value.

danpovey: I agree about the reducction option.

glynpu: What do you think of current design? @GNroy @danpovey 

> I find it quite intuitive to get n loss values when feeding n utterances (lattices). Also, having loss values per utterance (instead of one per batch) enables tricks like post scaling (e.g. focal loss).

This requirement is quite easy, you just need one extra step to get loss values per utterance. 
An example usage could be
```
# loss is a k2.RaggedTensor as explained in the comment.
loss = k2.mwer_loss(lattice, refs_texts, reduction='none')
loss_per_utt = loss.sum()
```

https://github.com/k2-fsa/k2/blob/cc67ec0f4622dadc1467f7e9cd040e73015ef478/k2/python/k2/mwer_loss.py#L160-L172

GNroy: LGTM except for a small typo on lines 43 and 172: tot**o**l -> tot**a**l
Thanks!

k2/python/k2/mwer_loss.py
```diff
@@ -0,0 +1,177 @@
...
+        )
+        tot_scores = levenshtein_alignment.get_tot_scores(
+            use_double_scores=self.use_double_scores, log_semiring=False
+        )
+        # Each path has a corresponding wer.
+        wers = -tot_scores.to(device)
+
+        # Group each log_prob into [path][arc]
+        ragged_nbest_logp = k2.RaggedTensor(path_arc_shape, nbest.fsa.scores)
+        # Get the probalitity of each path, in log format,
```
csukuangfj, line 104:
```suggestion
        # Get the probability of each path, in log format,
```

k2/python/k2/mwer_loss.py
```diff
@@ -0,0 +1,177 @@
...
+
+        hyps = nbest.build_levenshtein_graphs()
+        refs = k2.levenshtein_graph(ref_texts, device=hyps.device)
+        levenshtein_alignment = k2.levenshtein_alignment(
+            refs=refs,
+            hyps=hyps,
+            hyp_to_ref_map=nbest.shape.row_ids(1),
+            sorted_match_ref=True,
+        )
+        tot_scores = levenshtein_alignment.get_tot_scores(
```
csukuangfj, line 96: Please also describe the shape of `tot_scores` for readability.

k2/python/k2/mwer_loss.py
```diff
@@ -0,0 +1,177 @@
...
+        # Group each log_prob into [path][arc]
+        ragged_nbest_logp = k2.RaggedTensor(path_arc_shape, nbest.fsa.scores)
+        # Get the probalitity of each path, in log format,
+        # with shape [stream][path].
+        path_logp = ragged_nbest_logp.sum() / self.temperature
+        ragged_path_prob = k2.RaggedTensor(stream_path_shape, path_logp.exp())
+
+        # Normalize prob of each path.
+        den_prob = ragged_path_prob.sum()
+        den_logp = torch.index_select(
```
csukuangfj, line 111: For some versions of PyTorch, e.g., torch 1.7.1, the index type has to be `torch.int64`, I think.

k2/python/k2/mwer_loss.py
```diff
@@ -0,0 +1,177 @@
...
+        )
+        # Each path has a corresponding wer.
+        wers = -tot_scores.to(device)
+
+        # Group each log_prob into [path][arc]
+        ragged_nbest_logp = k2.RaggedTensor(path_arc_shape, nbest.fsa.scores)
+        # Get the probalitity of each path, in log format,
+        # with shape [stream][path].
+        path_logp = ragged_nbest_logp.sum() / self.temperature
+        ragged_path_prob = k2.RaggedTensor(stream_path_shape, path_logp.exp())
```
csukuangfj, line 107: Is it possible to use `normalize` of the RaggedTensor to simplify the code?

glynpu: It's possible. I am implementing backward for normalize(use_log=False).

csukuangfj: Can we use
```python
ragged_path_prob = k2.RaggedTensor(stream_path_shape, path_logp)
prob_normalized = ragged_path_log_prob.normalize(use_log=True).exp()
```

csukuangfj:
+2

Issue #1120, opened 2022-11-28
fsa attributes may be on different devices


glynpu:
See code below:
```
import torch                                                                                                                    
import k2                                                                                                                       
import _k2                                                                                                                      
f1 = k2.linear_fsa([1,2])                                                                               
scores = torch.tensor([0.1, 0.2, 0.3], device=torch.device('cuda'))                                                             
f1.scores = scores                                                                                                              
print(f"device f1.device: {f1.device}")                                                                                         
print(f"device f1.scores: {f1.scores.device}")                                                                                  
print(f"device of arcs: {f1.arcs.values().device}") 
```
Results:
```
device f1.device: cuda:0
device f1.scores: cuda:0
device of arcs: cpu
```
This could trigger some fatal errors. Since `arc_map` is always got from `arcs`, so it will be also on `cpu`. 
While `scores` is on GPU.
So following code would crash(It may be used in some functions):
```
_k2.index_select(f1.scores,arc_map, 0.0)
```
with log:
```
k2/k2/python/csrc/torch/index_select.cu:70:at::Tensor k2::IndexSelect1D(at::Tensor,
 at::Tensor, T) [with T = float] Check failed: src.device() == index.device() (cuda:0 vs. cpu) Expected in the same device Given
 : cuda:0, cpu
```

glynpu:
So I suggest following changes:

https://github.com/k2-fsa/k2/blob/0a6cf0b7b5e4f58a3444cac85800f397f7f317f0/k2/python/k2/fsa.py#L407

to:
```
self._tensor_attr[name] = value.to(self.device)
```

PR #1121, opened 2022-11-28
[Ready] avoid attibutes on different device

glynpu:
fix issue https://github.com/k2-fsa/k2/issues/1120

### DIFF
--- a/k2/python/k2/fsa.py
+++ b/k2/python/k2/fsa.py
@@ -418,7 +418,7 @@ class Fsa(object):
             assert value.dim0 == self.arcs.values().shape[0], \
                     f'value.dim0: {value.dim0}, shape[0]: {self.arcs.values().shape[0]}'  # noqa
             # See https://github.com/k2-fsa/k2/issues/1120
-            assert value.device == self.device
+            assert value.device == self.device, (value.device, self.device)
             self._tensor_attr[name] = value
         else:
             self._non_tensor_attr[name] = value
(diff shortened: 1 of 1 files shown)


k2/python/k2/fsa.py
```diff
@@ -391,6 +391,8 @@ def __setattr__(self, name: str, value: Any) -> None:
 
         if isinstance(value, torch.Tensor):
             assert value.shape[0] == self.arcs.values().shape[0]
+            # See https://github.com/k2-fsa/k2/issues/1120
+            assert value.device == self.device
```
csukuangfj, line 395: Please also change line 417 and add tests for your changes.

k2/python/k2/fsa.py
```diff
@@ -417,6 +417,8 @@ def __setattr__(self, name: str, value: Any) -> None:
         elif isinstance(value, k2.RaggedTensor):
             assert value.dim0 == self.arcs.values().shape[0], \
                     f'value.dim0: {value.dim0}, shape[0]: {self.arcs.values().shape[0]}'  # noqa
+            # See https://github.com/k2-fsa/k2/issues/1120
+            assert value.device == self.device
```
csukuangfj, line 421:
```suggestion
            assert value.device == self.device, (value.device, self.device)
```

csukuangfj (changes requested):

k2/python/k2/fsa.py
csukuangfj: See https://github.com/k2-fsa/k2/pull/1123

Please always add tests for your changes in the future if possible.

PR #1125, opened 2022-12-04
Remove unused file

pkufool:


k2/csrc/intersect_dense_pruned.cu
```diff
@@ -289,12 +288,8 @@ class MultiGraphDenseIntersectPruned {
         K2_CHECK_EQ(state_map_.NumKeyBits(), 40);
         frames_.push_back(PropagateForward<40>(t, frames_.back().get()));
       }
-      if (t != 0 && (T_ + t) % prune_shift == 0 ||
-          t == b_fsas_->shape.MaxSize(1)) {
-        int32_t prune_t_begin =
-            (T_ + t - prune_num_frames) > 0 ? (T_ + t - prune_num_frames) : 0;
-        int32_t prune_t_end = T_ + t;
-        PruneTimeRange(prune_t_begin, prune_t_end);
+      if (t == b_fsas_->shape.MaxSize(1)) {
```
danpovey, line 291: @pkufool  I'm just curious about this change here, was it to fix some kind of bug?
I think the goal here with the periodic pruning was to reduce memory consumption, but it's not impossible that the implementation could have been buggy (or unnecessary in practice).

pkufool: For streaming decoding, we will invoke this function every chunk, so I think we only need to prune once for each chunk, this fix is for efficiency.

PR #1127, opened 2022-12-07
[WIP] Fix default allocator

pkufool:
The purpose of this PR is to fix the `default_context.cu` so that we can use k2 without Pytorch. There are still some issues to be fixed.

- [ ] `NvToolExt.h` Not found error (if K2_ENABLE_NVTX=ON). I found that the cuda env was handle by Pytorch before,  I am reading `FindCUDA.cmake` in Pytorch and try to extract some script there.
- [ ] About the cudaStream, I am still a little confued about how k2 or pytorch handle cuda streams, now I always allocate memory on the default stream (i.e. stream 0).
- [ ] The caching strategy (using default now, might need some tuning and benchmark).
- [ ] A global Caching allocator or each `Context` has its own allocator instance? Which one is better?

Even though this is a small fixes, allocator is the fundamental part, I may miss someting important, hope for your suggestions.

### DIFF
--- a/k2/csrc/default_context.cu
+++ b/k2/csrc/default_context.cu
@@ -90,9 +90,7 @@ class CudaContext : public Context {
       K2_CHECK_CUDA_ERROR(ret);
       gpu_id_ = current_gpu_id;
     }
-
-    auto ret = cudaStreamCreate(&stream_);
-    K2_CHECK_CUDA_ERROR(ret);
+    allocator_ = new cub::CachingDeviceAllocator();
 #else
     K2_LOG(FATAL) << "Unreachable code.";
 #endif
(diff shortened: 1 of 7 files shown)


k2/csrc/default_context.cu
```diff
@@ -52,65 +53,126 @@ class CpuContext : public Context {
...
+    DeviceGuard guard(gpu_id_);
+    auto ret = allocator_->DeviceFree(data);
     K2_CHECK_CUDA_ERROR(ret);
+#endif
   }
 
   cudaStream_t GetCudaStream() const override {
-    return g_stream_override.OverrideStream(stream_);
+#ifdef K2_WITH_CUDA
+    return g_stream_override.OverrideStream(0);
```
pkufool, line 151: Here, in `Pytorch_Context.cu`, we get the stream by `c10::cuda::getCurrentCudaStream`, I read the source code of Pytorch, if we don't `setCurrentCudaStream` the return stream is always the default stream (i.e. stream 0).

So, can we use the default stream 0 here? (0 might not be good here, we can add a macro some like `kDefaultStream`)

csukuangfj:
> The purpose of this PR is to fix the default_context.cu so that we can use k2 without Pytorch

What is the use case?  That is, how will we use k2 with cuda but without PyTorch?

pkufool:
> > The purpose of this PR is to fix the default_context.cu so that we can use k2 without Pytorch
> 
> What is the use case? That is, how will we use k2 with cuda but without PyTorch?

For decoding purpose, if we use onnxruntime or tensorRT as the inference engine, I don't think we should depend on Pytorch at that time.  The inference engine might have allocator, though, I am not sure.

csukuangfj:
>  The inference engine might have allocator, though, I am not sure.

In that case, I would suggest creating another context that wraps the allocator from the given framework.

PR #1137, opened 2023-01-07
Fix installing Python 3.11 in GitHub CI.

csukuangfj:


csukuangfj:
It turns out there are no packages about torch 1.13.1 with Python 3.11 at
https://anaconda.org/pytorch/pytorch/files

See https://github.com/pytorch/pytorch/issues/91839

Issue #1104, opened 2022-11-14
Wrong forward and backward scores when an arc weight is changed


jctian98:
Hi! I'm trying to modify the weight on a certain arc. But after that, the forward-backward scores seem not correct.

The code is:
```
print('idx: ', zero_idx)
lats.scores[zero_idx] = 0.0
print('scores: ', lats.scores)
lats[0].draw('lat.png')
bw_scores = lats.get_backward_scores(True, True)
fw_scores = lats.get_forward_scores(True, True)
print("backward_scores: " , bw_scores)
print("forward_scores:", fw_scores)
```

The `lats` has only one FSA and looks like below. The arc with the index of 6 is to modify: `idx:  tensor([6], device='cuda:0')`.
![rnnt_lat](https://user-images.githubusercontent.com/92212251/201662081-ccb0a93e-f7c1-4a50-b595-ce2e9a060d09.png)

The scores have been changed:
`scores:  tensor([-1.441691, -1.249144, -1.893759, -1.694620, -1.370999, -1.370999,
         0.000000], device='cuda:0')`

But the forward and backward scores suggest the weight on the last arc has NOT been changed. (the last-second backward score is not 0).
```
backward_scores:  tensor([-5.169402, -4.635757, -4.436618, -2.741998, -2.741998, -1.370999,
         0.000000], device='cuda:0', dtype=torch.float64)
forward_scores: tensor([ 0.000000, -1.441691, -1.249144, -3.335450, -2.943764, -3.798403,
        -5.169402], device='cuda:0', dtype=torch.float64)
```

The K2 version is 1.16.

Thanks a lot!

csukuangfj:
```
lats[0].draw('lat.png')
```
So your `lats` is an FsaVec, which means `lats.scores` and `lats.arcs.values()[:, 3]` are two different things.

In general, if you assign a value to `lats.scores`, it will also update `lats.arcs.values()[:, 3]`.
That is, if you do
```
lats.scores = some_tensor
```
then `lats.arcs.values()[:, 3]` is also updated.

`lats.scores = some_tensor` invokes `__setattr__` of an `Fsa`, where we also update `lats.arcs.values()[:, 3]`.

But this is not the case when you use `lats.scores[some_index] = some_value`.

`lats.scores` invokes `__getattr__` and it is not possible to know tell when it should update `lats.arcs.values()[:, 3]`.

---
```
lats[0].draw('lat.png')
```

[...]

Issue #978, opened 2022-05-26
Wrong intersection result after renaming labels


GNroy:
Hi,

I use the latest (1.5.1) k2 release.

I've found severe mismatch between intersection (with `treat_epsilons_specially=False`) results if some of the labels are changed.

Say, if we intersect FSAs `a` and `b`, then the result will be correct.
But if we do `a.labels[a.labels == x] = y` and `b.labels[b.labels == x] = y` before the intersection, then the result will be different and therefore wrong.

The attached notebook contains steps to reproduce the issue (and a workaround I came up with):

[intersection_mismatch.ipynb.txt](https://github.com/k2-fsa/k2/files/8779032/intersection_mismatch.ipynb.txt) (remove the .txt)

csukuangfj:
> But if we do a.labels[a.labels == x] = y and b.labels[b.labels == x] = y before the intersection, then the result will be different and therefore wrong.

Could you use
```
a.labels[a.labels == x] = y
a.__dict__["_properties"] = None

b.labels[b.labels == x] = y
b.__dict__["_properties"] = None
```
and see it it helps?

csukuangfj:
Note: The following link to the Colab notebook is to simplify the process to view the provided `intersection_mismatch.ipynb.txt`

https://colab.research.google.com/drive/1mlNo6soSBxmY_GCNVjudIiITZjLqxU_g?usp=sharing

csukuangfj:
> Though we still need to do k2.arc_sort for this to work on CPU.

Yes, you still need to arc sort them if you are doing intersection on CPU, as described in the doc:
https://k2-fsa.github.io/k2/python_api/api.html#k2.intersect

> The two input FSAs MUST be arc sorted if treat_epsilons_specially is True.

https://github.com/k2-fsa/k2/blob/ecfe7bd6d9189964bf3ff043038918d889a43185/k2/python/k2/fsa_algo.py#L279-L281

---

> Setting {a,b}.__dict__["_properties"] = None helped!

Please see https://github.com/k2-fsa/k2/issues/874#issuecomment-969974902
We have had such an issue before. I think we don't have much to do in k2 to detect this problem automatically.

GNroy:
Ok, thanks for the explanation!

One last question: is it safe to set  `.__dict__["_properties"] = None`?
I mean, could it potentially break something?

danpovey:
It's safe, just resets the properties to say they are unknown.
The difficulty with setting them automatically is that when you do:
[...]

Issue #874, opened 2021-11-15
Score a text by word-level N-gram model


jctian98:
Hi. I'm trying to score a (Chinese) text by a word-level N-gram model.  To achieve this, I build a lattice to enumerate all possible word sequences. After that, I call `k2.intersect` to the lattice and `G.fst`. Finally, `get_tot_scores` is called to compute the score.  
An error is encountered when I call `get_tot_scores`. I suppose there is no loop in the output lattice (see the graph later). So I don't know how to fix this problem. 
Code and details are below :) thanks for the help!
```
import k2
import re
from pathlib import Path

def is_disambig_symbol(symbol: str, pattern: re.Pattern = re.compile(r'^#\d+$')) -> bool:
    return pattern.match(symbol) is not None

def find_first_disambig_symbol(symbols: k2.SymbolTable) -> int:
    return min(v for k, v in symbols._sym2id.items() if is_disambig_symbol(k))


lang=Path("data/eg_ngram")

with open(lang / 'G.fst.txt') as f:
    G = k2.Fsa.from_openfst(f.read(), acceptor=False)
symbol_table = k2.SymbolTable.from_file(lang / 'words.txt')
first_word_disambig_id = find_first_disambig_symbol(symbol_table)

G.labels[G.labels >= first_word_disambig_id] = 0
G = k2.arc_sort(G)
del G.aux_labels
G.draw("G.svg", title="G")

def text2lat(text, gram_len, symbol_table):
    arcs = []
    aux_labels = []
    for s in range(len(text) + 1):
        for r in range(1, gram_len + 1):
            if len(text) - s - r < 0:
                continue
            w = "".join(text[s: s+r])
            if not w in symbol_table:
                continue
            wid = symbol_table[w]
            arc = [s, s+r, wid, 0]
            arcs.append(arc)
            aux_labels.append(wid)
    arcs.append([len(text), len(text) + 1, -1, 0.0]) 
    aux_labels.append(-1)

    arcs = torch.Tensor(arcs).int()
    lat = k2.Fsa.from_dict({"arcs": arcs})
    lat = k2.arc_sort(lat)

    aux_labels = torch.Tensor(aux_labels)
    lat.aux_labels = aux_labels
    lat.labels_sym = symbol_table
    return lat

lat = text2lat("向人们展示", 2, symbol_table)
lat = k2.create_fsa_vec([lat])
lat = k2.add_epsilon_self_loops(lat)
[...]

PR #1140, opened 2023-01-12
Fix fsa labels

yaozengwei:
This PR fixes the issue when we change the fsa attribute `labels`:
- https://github.com/k2-fsa/k2/issues/1104
- https://github.com/k2-fsa/k2/issues/874
- https://github.com/k2-fsa/k2/issues/978

### DIFF
--- a/k2/python/k2/fsa.py
+++ b/k2/python/k2/fsa.py
@@ -444,7 +444,8 @@ class Fsa(object):
         if properties is not None:
             if self.labels._version != self.labels_version:
                 raise RuntimeError(
-                    "The fsa attribute (labels) has been inappropriately modified like:\n"
+                    "The fsa attribute (labels) has been inappropriately "
+                    "modified like:\n"
                     "    fsa.labels[xxx] = yyy\n"
                     "The correct way should be like:\n"
                     "    labels = fsa.labels\n"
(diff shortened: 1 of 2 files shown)


k2/python/k2/fsa.py
```diff
@@ -222,6 +222,8 @@ def __init__(
             self.__dict__[name] = dict()
 
         self._tensor_attr['scores'] = _k2.as_float(self.arcs.values()[:, -1])
+        self._tensor_attr["labels"] = self.arcs.values()[:, 2]
+        self.labels_version = self._tensor_attr["labels"]._version
```
csukuangfj, line 226: Could you add a comment here explaining why
we need `labels_version`?

You can place the link to this PR https://github.com/k2-fsa/k2/pull/1140
to the comment.

yaozengwei:
By adding an attribute `labels_version` for class `Fsa`, we can check whether the labels has been inappropriately modified. For example:
22061673527903_ pic

k2/python/k2/fsa.py
yaozengwei: ok

csukuangfj (approved):

csukuangfj:
Thanks! Looks good to me.

csukuangfj:
Could you tell us which version of k2 you are using?

(Please include the date info in the version number.)

csukuangfj:
Could you replace
```
with torch.inference_mode():
```
with
```
with torch.no_grad():
```
and retry?

csukuangfj:
> self.labels_version = self._tensor_attr["labels"]._version

As a workaround, you can use
```
self.labels_version = 0
```
if you are sure you won't modify `fsa.labels` once `fsa` is constructed.

---

You can change the file  `fsa.py` in-place.

PR #1149, opened 2023-01-27
Allow zero length references for rnnt loss

desh2608:
Based on some discussions with @danpovey, it seems the RNNT loss computation itself does not have any inherent requirement of references being non-zero. Allowing this relaxation would be useful for training with possibly empty references, e.g. cases where utterance is just noise or silence, or for multi-talker ASR (e.g. see https://github.com/k2-fsa/icefall/issues/845).

### DIFF
--- a/k2/python/k2/rnnt_loss.py
+++ b/k2/python/k2/rnnt_loss.py
@@ -146,7 +146,9 @@ def get_rnnt_logprobs(
     S = lm.shape[1] - 1
     assert symbols.shape == (B, S), (symbols.shape, B, S)
     assert S >= 0, S
-    assert T >= S, (T, S)
+    assert (
+        rnnt_type != "modified" or T >= S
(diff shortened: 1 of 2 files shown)


csukuangfj:
I think you can also remove the constraint 

assert T >= S

That constraint is required only for modified transducer, where it can only move horizontally or diagonally.

desh2608:
> I think you can also remove the constraint
> 
> assert T >= S
> 
> That constraint is required only for modified transducer, where it can only move horizontally or diagonally.

Okay, I will make that constraint conditional on if the `rnnt_type` is modified.

csukuangfj:
Thanks. Would you mind also creating a PR in the repo danpovey/fast_rnnt ?

csukuangfj:
Will merge it tomorrow. Thanks!

csukuangfj:
Thanks for your contribution!

PR #1151, opened 2023-01-30
fix swoosh.cu for building [issue 1143]

treya-lin:
This fix is relevant to the error log reported in https://github.com/k2-fsa/k2/issues/1143#issuecomment-1408131855


csukuangfj:
Thanks for your contribution!

PR #1154, opened 2023-02-05
fix the tar cudnn cmd

Honei:
I can download the package `cudnn-11.3-linux-x64-v8.2.0.53.tgz`, and cannot download the `cudnn-11.3-linux-x64-v8.2.1.32.tgz`.
And the tar cmd use the `cudnn-11.3-linux-x64-v8.2.1.32.tgz`.


csukuangfj:
Thanks for your contribution!

Issue #1161, opened 2023-02-16
Typo in documentation


treya-lin:
Not a big finding but just a minor typo in documentation: 

https://k2-fsa.github.io/k2/core_concepts/index.html#example-1-autograd-in-tropical-semiring

```
fsa_vec = k2.create_fsa_vec([fsa])
total_scores = fsa.get_tot_scores(log_semiring=False, use_double_scores=False)
```
should be 
```
fsa_vec = k2.create_fsa_vec([fsa])
total_scores = fsa_vec.get_tot_scores(log_semiring=False, use_double_scores=False)
```
right : )

csukuangfj:
Yes, you are absolutely right.

Would you mind creating a pull-request to fix it?

treya-lin:
> Yes, you are absolutely right.
> 
> Would you mind creating a pull-request to fix it?

Sure. Just created it https://github.com/k2-fsa/k2/pull/1162 . Cheers.

PR #1162, opened 2023-02-17
Fix minor typo in doc of k2 core concepts (#1161)

treya-lin:
Fix a minor typo in doc of k2 core concepts


csukuangfj:
Thanks for your contribution!

PR #1170, opened 2023-03-16
Add doc about installing cuda 11.8 for torch 2.0

csukuangfj:


wangtiance:
Wow, it's amazing how fast torch 2 support is added!

PR #1198, opened 2023-05-17
fix and remove the assumption that symbols in the decoding graph must…

DongjiGao:
… exist in nnet_output (dense fsa)


csukuangfj:
@pkufool Could you have a review?

danpovey:
I think this is OK, will merge.

Issue #1199, opened 2023-05-18
Build with CUDA 12.x is broken in Docker


artbataev:
CUDA 12.x build is broken in Docker (when compiling without GPU / for all architectures).

## How to reproduce

`Dockerfile`:

```dockerfile
# nvidia-pytorch official docker image
ARG BASE_IMAGE=nvcr.io/nvidia/pytorch:23.04-py3

FROM ${BASE_IMAGE}

RUN K2_MAKE_ARGS="-j" pip install -v "git+https://github.com/k2-fsa/k2#egg=k2" || { echo "k2 installation failed"; exit 1; }
```

Execute:
```shell
docker build --tag pytorch_k2_cuda12 .
```

Error:
```
...
#6 70.68     [  8%] Building CXX object _deps/kaldifeat-build/kaldifeat/csrc/CMakeFiles/kaldifeat_core.dir/online-feature.cc.o
#6 70.69     nvcc fatal   : Unsupported gpu architecture 'compute_35'
#6 70.69     make[2]: *** [k2/csrc/CMakeFiles/k2_log.dir/build.make:76: k2/csrc/CMakeFiles/k2_log.dir/log.cu.o] Error 1
#6 70.69     make[1]: *** [CMakeFiles/Makefile2:364: k2/csrc/CMakeFiles/k2_log.dir/all] Error 2
...
```

## Potential solution

Remove `35` from `K2_COMPUTE_ARCH_CANDIDATES` for CUDA 12.x

csukuangfj:
Closing via #1200

PR #1200, opened 2023-05-18
Fix architectures for CUDA 12 and 11.6

artbataev:
- add support for Hopper (90) architecture for CUDA 11.6+
- remove compute architecture `35` for CUDA 12.0+ (no longer supported)

Fixes issue #1199

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,6 +240,19 @@ if(K2_WITH_CUDA)
 
       string(APPEND CMAKE_CUDA_FLAGS " -Wno-deprecated-gpu-targets ")
   endif()
+
+  if(CUDA_VERSION VERSION_GREATER_EQUAL "11.6")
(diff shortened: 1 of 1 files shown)


csukuangfj:
Thanks for your contribution!

MartinKocour:
We have an issue with this update in our cluster... is Hopper really supported with CUDA 11.6? Shouldn't it be from 11.8?
https://docs.nvidia.com/cuda/hopper-compatibility-guide/index.html

artbataev:
> We have an issue with this update in our cluster... is Hopper really supported with CUDA 11.6? Shouldn't it be from 11.8?
> https://docs.nvidia.com/cuda/hopper-compatibility-guide/index.html

I'm sorry, it seems to be my mistake, CUDA should be "11.8" for Hopper architecture. I will provide a fix. 
@MartinKocour, Did you try to modify the version? Does it fully fix your problem?

PR #1211, opened 2023-06-15
Remove c++14 overrides in torch/bin/CMakeLists

wrongbad:
They inherit c++14 by default from top-level anyway

But override needs to be removed to link with
recent torch builds which require c++17.

Now you can specify -DCMAKE_CXX_STANDARD=17 from CLI


csukuangfj:
The current code also works for the latest PyTorch version, e.g., torch 2.0.1.

Could you describe what this PR tries to fix?

csukuangfj:
Thanks! Fixed in #1249

PR #1215, opened 2023-06-27
Disable find mkl

pkufool:
Prevent cmake from trying to link with system installed mkl since we not directly use it. mkl libraries should be linked with pytorch already.


csukuangfj:
See also https://github.com/csukuangfj/kaldifeat/pull/64

csukuangfj (approved):

PR #1218, opened 2023-07-07
Allow partial for intersect_dense_pruned

pkufool:


geoph9:
Is there any plan to add the `allow_partial` option on `intersect` and `intersect_device`? I am doing whole lattice rescoring with a 4-gram LM, and for some reason the lattice for some of the utterances does not have any final states and hence `k2.connect(lattice)` outputs an empty Fsa.

I have tried increasing the beam size to no avail. I think a workaround would be to allow partial paths, i.e. if not final states are encountered in a lattice then we assume that all states on the last frames are final.

pkufool:
> Is there any plan to add the `allow_partial` option on `intersect` and `intersect_device`? I am doing whole lattice rescoring with a 4-gram LM, and for some reason the lattice for some of the utterances does not have any final states and hence `k2.connect(lattice)` outputs an empty Fsa.
> 
> I have tried increasing the beam size to no avail. I think a workaround would be to allow partial paths, i.e. if not final states are encountered in a lattice then we assume that all states on the last frames are final.

Yes,  I think it is doable, @danpovey  what do you think?

k2/csrc/intersect_dense_pruned.cu
```diff
@@ -722,6 +828,30 @@ class MultiGraphDenseIntersectPruned {
     Ragged<ArcInfo> ai(ai_shape);
     ArcInfo *ai_data = ai.values.Data();  // uninitialized
 
+    // A valid final arc means its label == -1.
+    auto has_valid_final_arc = Array1<bool>(c_, NumFsas(), false);
+    bool *has_valid_final_arc_data = has_valid_final_arc.Data();
+    bool allow_partial = allow_partial_;
+
+    if (allow_partial_) {
```
danpovey, line 836: maybe meant to use allow_partial not allow_partial_ (or the variable was not needed?)

danpovey (commented):
Thanks!
I am surprised this didn't already have an allow_partial option.

pkufool:
Merging now, I found the failed tests (mainly windows build) existed for several months, we should fix it in a separate PR.

PR #1236, opened 2023-07-26
Add scripts for building manylinux wheels

csukuangfj:
Also, remove auto-generated files from doc.


k2/csrc/dtype.h
```diff
@@ -68,9 +68,6 @@ class DtypeTraits {
   const char *name_;  // name, e.g. "float", "int8", "int32"
 };
 
-// We initialize this in dtype.cu
-extern const DtypeTraits g_dtype_traits_array[];
```
csukuangfj, line 72: When building shared libs for windows, it complains that the array
`g_dtype_traits_array` is undefined.

PR #1244, opened 2023-09-26
Hybrid autoregressive transducer (HAT)

desh2608:
This is an implementation of the HAT loss proposed in https://arxiv.org/abs/2003.07705.

The test produces reasonable looking losses. I am working on a LibriSpeech zipformer recipe using this loss. In general, it is not expected to improve upon the RNNT loss by itself, but may be useful for things like using external LMs. I am planning to use it in multi-talker ASR for speaker attribution (e.g. https://arxiv.org/abs/2309.08489)


danpovey:
Great!

desh2608:
@csukuangfj could you also check this when you have some time? Thanks!

csukuangfj:
> @csukuangfj could you also check this when you have some time? Thanks!

Thanks! Left a minor comment. Otherwise, it looks good to me.

desh2608:
> > @csukuangfj could you also check this when you have some time? Thanks!
> 
> Thanks! Left a minor comment. Otherwise, it looks good to me.

Sorry it took a while since I was on vacation last 2 weeks. I have made the change.

csukuangfj:
Thanks!

PR #1273, opened 2024-03-21
update CMakeLists.txt

KarelVesely84:
- compute capability 90 appeared first in CUDA 11.8, causing error for CUDA 11.7


csukuangfj:
Thank you for your first-time contribution!

PR #1294, opened 2024-07-02
Wrap unsqueeze

pkufool:


k2/python/csrc/torch/v2/ragged_any.cu
```diff
@@ -575,6 +575,14 @@ RaggedAny RaggedAny::Cat(const std::vector<RaggedAny> &srcs, int32_t axis) {
   return {};
 }
 
+RaggedAny RaggedAny::Unsqueeze(int32_t axis) {
+  DeviceGuard guard(any.Context());
+  Dtype t = any.GetDtype();
+  FOR_REAL_AND_INT32_TYPES(t, T, {
+    return RaggedAny(k2::Unsqueeze(any.Specialize<T>(), axis).Generic());
```
csukuangfj, line 582: k2:: can be removed since it is inside the k2 namespace.

csukuangfj (changes requested):

PR #1303, opened 2024-09-06
Publish pre-built wheels for torch 2.4.1

csukuangfj:


csukuangfj:
please see

- https://k2-fsa.github.io/k2/cuda.html
- https://k2-fsa.github.io/k2/cpu.html

PR #1349, opened 2026-03-14
Add Apple MPS (Metal) backend: native Metal kernels, autograd CPU bridges

dredozubov:
## Motivation

k2 is used for FSA-based speech recognition (CTC, RNN-T, MMI), and Apple Silicon Macs have become the primary development machine for a lot of researchers. The problem: k2 had no MPS support at all. Every operation fell through PyTorch's CPU fallback, so GPU-accelerated training on macOS simply didn't work.

## Why this was hard

k2's C++ core uses the \`K2_EVAL\` macro extensively — it reads raw device pointers from the CPU side to walk arc data. On CUDA this works because pinned memory and unified addressing let the CPU see GPU memory. On MPS, \`data_ptr()\` on a tensor doesn't return a valid CPU pointer, so any path through \`K2_EVAL\` crashes immediately.

Two things fix this. For hot paths, native Metal kernels that receive \`id<MTLBuffer>\` handles and never touch device memory from the CPU. For topology-dependent operations where native Metal isn't worth it yet, CPU bridges in the Python autograd layer: copy the FSA to CPU, run the existing C++ algorithm, return a gradient-connected result to MPS via a lightweight \`_MpsScoresBridge\` autograd function.

### \`mutual_information_mps.mm\` — native Metal MI kernel

Replaces the CPU round-trip in \`mutual_information_recursion\` with a Metal compute kernel using the blocked antidiagonal wavefront pattern from the CUDA version. Forward (\`mi_forward\`) and backward (\`mi_backward\`) kernels are compiled at runtime from an embedded MSL string, cached via \`dispatch_once\`. Expected speedup: 5–8× over CPU bridging for typical (T=300, S=50) sequences.

CUDA→Metal primitive mapping:
- \`__shared__ float arr[N]\` → \`threadgroup float arr[N]\`
- \`__syncthreads()\` → \`threadgroup_barrier(mem_flags::mem_threadgroup)\`
- \`__syncwarp()\` → \`simdgroup_barrier(mem_flags::mem_threadgroup)\`

### \`mps_utils.h\` — ATen int32 ops

PyTorch MPS supports int32 \`cumsum\`, \`diff\`, \`repeat_interleave\`, and \`searchsorted\` natively since PyTorch 2.2. \`ExclusiveSumMps\`, \`RowSplitsToRowIdsMps\`, and \`MaxSizeMps\` are rewritten to use int32 throughout, removing redundant int64 casts and intermediate allocations.

### \`mps_fsa_scores.mm\` — native Metal \`GetForwardScores\`

Level-by-level BFS dispatch: each BFS layer's arcs are independent (antecedent states are already scored), so one Metal command encoder per layer is dispatched onto PyTorch's existing MPS command buffer. Supports log semiring (CAS loop with \`log_add\`) and tropical semiring (CAS loop with \`max\`). NaN guards (\`if (isnan(candidate)) return;\`) prevent the CAS loops from spinning forever on invalid inputs — \`NaN >= x\` is always false in IEEE 754, so without the guard the loop never exits.

### \`fsa.py\` — zero-copy arc-batch computation

A full \`self.to('cpu')\` copies 16 bytes × num_arcs plus all tensor attributes. The BFS-level sort only needs the \`dest_state\` column (4 bytes × num_arcs). \`_compute_bfs_arc_batches_mps\` extracts only that column, sorts arc indices on CPU via \`torch.argsort\`, and returns \`sorted_arc_ids\` as an MPS tensor — 4× less data crossing the bus.

### \`autograd.py\` — MPS bridges for \`IntersectDense\` and \`IntersectDensePruned\`

\`_IntersectDensePrunedFunction\` and \`_IntersectDenseFunction\` run the forward pass on CPU (the intersection algorithm is topology-dependent and relies on \`K2_EVAL\`). Gradients are scattered via \`_k2.index_add\` on CPU and moved back to MPS. \`_MpsScoresBridge\` keeps the gradient graph intact through the device transfer.

### \`mps_fsa_scores.mm\` — Hillis-Steele O(log N) associative scan

For single-FSA, tropical-semiring graphs with 4 ≤ N ≤ 128 states, the sequential N-step BFS dispatch is replaced by a Hillis-Steele prefix scan over N per-state transition matrices:

- Each state \`d\` has an N×N transfer matrix \`M[d]\`: diagonal entries are 0, \`M[d][d][src]\` = max arc weight from \`src\` to \`d\`, all else −∞.
- \`M[0]\` = identity (start state, no entering arcs).
- Prefix product \`P[s] = M[s] ⊗ M[s-1] ⊗ … ⊗ M[0]\` (tropical matrix-multiply).
- Forward score of state \`s\` = \`P[s][s][0]\` (best path from state 0 to state s).
- Hillis-Steele doubling reduces Metal encoder calls from N to ⌈log₂N⌉.
- Power-of-2 padding uses identity matrices (neutral element for tropical matmul).

### Additional autograd bridges

Auditing \`K2_EVAL\` call paths turned up three more autograd functions that would crash on MPS:

- \`_GetForwardScoresFunction.backward\` — bridges to CPU; also fixes a stride-0 tensor issue in the \`.expand_as().contiguous()\` chain
- \`_GetBackwardScoresFunction\` (forward + backward) — makes \`get_backward_scores\` and \`get_arc_post\` work on MPS
- \`_GetArcPostFunction\` (forward + backward) — full MPS support for \`get_arc_post\`

### Build system

- \`CMakeLists.txt\`: detect MPS via Python/PyTorch probe; add \`K2_WITH_MPS\` option; link \`Metal.framework\` + \`Foundation.framework\`
- \`k2/python/csrc/CMakeLists.txt\`: separate \`.mm\` files from \`.cu\` before \`transform()\` — the transform renames source extensions, and \`.mm\` → \`.cc\` breaks Objective-C++ compilation
- \`k2/python/csrc/torch/CMakeLists.txt\`: conditionally add \`.mm\` sources under \`K2_WITH_MPS\`

### CI (\`.github/workflows/build-mps-macos.yml\`)

Runs on \`macos-14\` (Apple Silicon, MPS hardware available). Triggers on push to master (path-filtered), PR labeled \`ready\`, or \`workflow_dispatch\`. Steps:
1. Build wheel with \`K2_WITH_MPS=ON\`
2. Verify \`k2.with_mps == True\`
3. Run \`test_mps.py\` (65 MPS-specific tests)
4. Run all other tests as CPU sanity check
5. Build + ctest C++ tests

## Test coverage (65 tests, all passing)

| Class | Tests | Coverage |
|---|---|---|
| \`TestMpsTensor\` | 6 | FSA/tensor transfer, flag, basic ops |
| \`TestMpsMutualInformation\` | 4 | Forward parity, backward, boundary tensor |
| \`TestMpsTotScores\` | 5 | Log/tropical backward, double, non-unit grad |
| \`TestMpsForwardScores\` | 4 | Log/tropical parity, differentiable, multi-FSA |
| \`TestMpsIntersectDense\` | 5 | Forward/backward parity, seqframe attr (pruned + non-pruned) |
| \`TestMpsAssocScan\` | 5 | Linear/branching parity, tot_scores, large fallback |
| \`TestMpsEdgeCases\` | 6 | Empty FSA, unreachable state, NotImplementedError guards, NaN-free grad |
| \`TestMpsAssocScanBoundaries\` | 8 | N=4/5/7/16/128/129, diamond topology, multi-arc dest |
| \`TestMpsArcPost\` | 4 | Tropical/log parity + gradient |
| \`TestMpsGetForwardScoresDifferentiable\` | 3 | Tropical/log/non-unit gradient |
| \`TestMpsIntersectDenseExtended\` | 5 | 2-utterance batch, seqframe parity (fixed seed) |
| \`TestMpsMutualInformationExtended\` | 4 | Varied sizes, gradient, batch, boundary |
| \`TestMpsForwardScoresNumericalStress\` | 3 | 50-state random DAG, NaN-free gradient |

## Known limitations

- \`_get_forward_scores(use_double_scores=True)\` on MPS raises \`NotImplementedError\` — MPS has no float64. Use \`get_forward_scores\` (differentiable, float32) instead.
- \`_get_backward_scores\` and \`_get_tot_scores\` (non-differentiable) on MPS raise \`NotImplementedError\`. Use the differentiable \`get_backward_scores\` / \`get_tot_scores\` instead.
- The O(log N) assoc-scan path only kicks in for single-FSA, tropical semiring, N in [4, 128]. Anything outside that falls back silently to the sequential Metal path.
- \`IntersectDense\` / \`IntersectDensePruned\` forward+backward run on CPU. Only the score tensors are MPS-resident.

## Test plan

- [x] 65/65 MPS tests pass locally on M-series Mac
- [x] CPU tests unmodified and unaffected
- [ ] CI run on \`macos-14\` via \`build-mps-macos.yml\` (add \`ready\` label to trigger)

### DIFF
--- a/.github/workflows/build-mps-macos.yml
+++ b/.github/workflows/build-mps-macos.yml
@@ -1,0 +1,130 @@
+# Copyright      2024  k2 contributors
+
+# See ../../LICENSE for clarification regarding multiple authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: build-mps-macos
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - '.github/workflows/build-mps-macos.yml'
+      - 'CMakeLists.txt'
+      - 'cmake/**'
+      - 'k2/csrc/**'
+      - 'k2/python/**'
+  pull_request:
+    types: [labeled]
+    paths:
+      - '.github/workflows/build-mps-macos.yml'
+      - 'CMakeLists.txt'
+      - 'cmake/**'
+      - 'k2/csrc/**'
+      - 'k2/python/**'
+  workflow_dispatch:
+
+concurrency:
+  group: build-mps-macos-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  BUILD_TYPE: Release
+
+jobs:
+  build-mps-macos:
+    if: github.event.label.name == 'ready' || github.event_name == 'push' || github.event_name == 'workflow_dispatch'
+    name: Build (macOS, CPU+MPS)
+    runs-on: macos-14  # Apple Silicon M1/M2 — MPS is available
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Display clang version
+        run: clang --version
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Display Python version
+        run: python -c "import sys; print(sys.version)"
+
+      - name: Install PyTorch (macOS arm64 with MPS)
+        shell: bash
+        run: |
+          python3 -m pip install -qq --upgrade pip
+          python3 -m pip install -qq wheel twine
+          python3 -m pip install -qq torch --index-url https://download.pytorch.org/whl/cpu
+          python3 -c "import torch; print('torch version:', torch.__version__); print('MPS available:', torch.backends.mps.is_available())"
+
+      - name: Build wheel
+        shell: bash
+        run: |
+          export K2_CMAKE_ARGS="-DCMAKE_BUILD_TYPE=$BUILD_TYPE -DK2_WITH_CUDA=OFF -DK2_WITH_MPS=ON"
+          export K2_MAKE_ARGS="-j3"
+          python3 setup.py bdist_wheel
+          ls -lh dist/
+          ls -lh build/*
+
+      - name: Upload wheel
+        uses: actions/upload-artifact@v4
+        with:
+          name: k2-macos14-mps-python3.11
+          path: dist/*.whl
+
+      - name: Install wheel and verify MPS support
+        shell: bash
+        run: |
+          pip install dist/*.whl
+          python3 -m k2.version
+          python3 -c "
+          import k2
+          assert k2.with_mps, 'k2 was not built with MPS support'
+          print('k2.with_cuda:', k2.with_cuda)
+          print('k2.with_mps:', k2.with_mps)
+          print('MPS build verified.')
+          "
+
+      - name: Run MPS tests
+        shell: bash
+        env:
+          PYTORCH_ENABLE_MPS_FALLBACK: "1"
+        run: |
+          pip install pytest
+          python3 -m pytest k2/python/tests/test_mps.py -v --tb=short
+
+      - name: Run CPU tests (sanity check)
+        shell: bash
+        run: |
+          python3 -m pytest k2/python/tests/ -v --ignore=k2/python/tests/test_mps.py -x --tb=short -q
+
+      - name: Build k2 (cmake + C++ tests)
+        shell: bash
+        run: |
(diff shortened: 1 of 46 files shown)


dredozubov:
Re the two Gemini review comments:

**`array_ops_inl.h` sync (HIGH)** — agreed, fixing. Adding `torch::mps::synchronize()` before each of the four MPS→CPU read paths in `array_ops_inl.h` (lines 346, 743, 782, 1041), matching the pattern already in `array_ops.cu`.

**`array.h` comment (MEDIUM)** — partial disagreement on the framing. PyTorch's MPS allocator uses `MTLStorageModeShared` (not `MTLStorageModePrivate`) for standard tensors, so `data_ptr()` does return a valid CPU-dereferenceable pointer — the "unified memory" statement is accurate. The real risk is stale data from uncommitted Metal commands, which is a synchronization issue, not a pointer-validity issue. Updating the comment to clarify both points.

dredozubov:
@csukuangfj @pkufool Can you guys take a look, please? I'd love to see it merged anytime soon.

PR #1356, opened 2026-06-26
Build wheels for AMD GPU with ROCm

csukuangfj:
See also #1353

Installation doc: https://k2-fsa.github.io/k2/installation/from_wheels.html#linux-rocm-example

See
https://k2-fsa.github.io/k2/rocm.html

Screenshot 2026-06-26 at 22 26 30

cc @jeffdaily


jeffdaily:
Tested on an AMD Instinct MI250X (gfx90a), ROCm 7.2.1, Python 3.10, torch 2.12.1+rocm7.2. The wheel installs and runs correctly: `arc_sort`, `get_forward_scores`, and `intersect_dense` + `shortest_path` (CTC) all execute on-device with correct results. `k2.with_cuda` is True.

Two notes:

**1. The documented pip command fails.** This, from the installation page:

```
pip install k2==1.24.4.dev20260626+rocm7.2.torch2.12.1 -f https://k2-fsa.github.io/k2/rocm.html
```

errors with `inconsistent version: expected '1.24.4.dev20260626+rocm7.2.torch2.12.1', but metadata has '1.24.4.dev20260626+rocm7.2.53211.torch2.12.1'`. The wheel's internal metadata version embeds the full ROCm patch (`rocm7.2.53211`) while the filename truncates to `rocm7.2`, and pip rejects the mismatch. Downloading the wheel file and `pip install ./<wheel>` (the direct-download path in the docs) works. Worth aligning the filename and the metadata version so the documented one-liner works.

**2. A small FYI on the arch list (not a problem, just an option).** The arch selection in `scripts/github_actions/build-ubuntu-rocm.sh` is clearly deliberate -- enumerating the gfx targets from the ROCm support docs is a sound, well-reasoned way to do it, and the wheel runs correctly on every arch it ships. So this is more of a "did you know" than a suggested change.

If it's ever useful to have the list stay in lockstep with the torch build the wheel is paired against, `torch.cuda.get_arch_list()` returns exactly the gfx targets the installed ROCm torch wheel was compiled for:

```bash
HIP_ARCH=$(python -c "import torch; print(';'.join(a for a in torch.cuda.get_arch_list() if a.startswith('gfx')))")
cmake -DCMAKE_HIP_ARCHITECTURES="$HIP_ARCH" ...
```

For reference, on this build the two lists differ only by `gfx900`/`gfx906` (old Vega) -- the torch wheel still includes them, while the k2 list (reasonably) tracks the current ROCm support matrix instead. Either policy is defensible; deriving from torch is just one way to avoid maintaining the list by hand as new arches land.

csukuangfj:
@jeffdaily 

Thanks for your detailed information.

> 1. The documented pip command fails. This, from the installation page:

This issue should be fixed now. Please retry.

----

> If it's ever useful to have the list stay in lockstep with the torch build the wheel is paired against, torch.cuda.get_arch_list()

Since GitHub Actions provides only CPU machines, `torch.cuda.get_arch_list()` returns an empty list, thus, we keep the current approach.

